PR #18, opened 2015-07-30
Reporting: use fixed precision for AvrTime and StdDev.

ppanyukov:
- Makes report look much better with everything aligned;
- Four decimals is a good comporomise;
- Implemented using compile-time feature switch.

Sample:

```
 Method |    AvrTime |    StdDev |           op/s |
------- |----------- |---------- |--------------- |
 Sqrt13 | 60.7802 ns | 0.2933 ns |  16,452,713.42 |
 Sqrt14 |  1.6146 ns | 0.0162 ns | 619,364,025.60 |
```

### DIFF
--- a/BenchmarkDotNet/Reports/BenchmarkTimeSpan.cs
+++ b/BenchmarkDotNet/Reports/BenchmarkTimeSpan.cs
@@ -14,21 +14,16 @@
 
         public override string ToString()
         {
-            if (Nanoseconds < 0.001)
-                return string.Format(EnvironmentHelper.MainCultureInfo, "{0:0.000000} ns", Nanoseconds);
-            if (Nanoseconds < 0.01)
-                return string.Format(EnvironmentHelper.MainCultureInfo, "{0:0.00000} ns", Nanoseconds);
-            if (Nanoseconds < 0.1)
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
LGTM. But you can remove FEATURE_REPORT_FIXED_PRECISION and set you formatting rules by default.

ppanyukov:
OK, if you are sure.

ppanyukov:
How's this looking now?

BTW, what's the best way to make amendments to pull requests?

- revert back; then make the correct change; or
- just make the adjustments on top?

I did the revert -> correct change. Probably not the right way :)

AndreyAkinshin:
> How's this looking now?

Looks good.

> BTW, what's the best way to make amendments to pull requests?

You should [squash your commits](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) and execute `git push -f`.

ppanyukov:
So, should I do it for this? Even if it's already pushed to github?

AndreyAkinshin:
> So, should I do it for this?

Yes. In this case, history will look better.

> Even if it's already pushed to github?

Yes. Just don't forget the `-f`.

ppanyukov:
All done. Neat technique, even if a somewhat dangerous, thanks for that.

PR #21, opened 2015-08-04
Reporting: uniform time units across all benchmarks.

ppanyukov:
As per https://github.com/PerfDotNet/BenchmarkDotNet/issues/16

For usability use the same time unit when reporting `AvrTime` and `StdDev`.

Sample output:

```
  Method |        AvrTime |      StdDev |       op/s |
-------- |--------------- |------------ |----------- |
 Slower1 |      1.8112 us |   0.2015 us | 552,120.14 |
 Slower2 |      2.4149 us |   0.2425 us | 414,095.82 |
 Slower3 |      9.0558 us |   0.3052 us | 110,426.47 |
 Slower4 |     81.2006 us |   1.2585 us |  12,315.18 |
 Slower5 |  4,148.1727 us | 740.3897 us |     241.07 |
 Slower6 | 15,047.0119 us |  99.5290 us |      66.46 |
```

### DIFF
--- a/BenchmarkDotNet.Samples/Intro_04_UniformReportingTest.cs
+++ b/BenchmarkDotNet.Samples/Intro_04_UniformReportingTest.cs
@@ -1,0 +1,80 @@
+﻿using BenchmarkDotNet.Tasks;
+
+namespace BenchmarkDotNet.Samples
+{
+    using System;
+    using System.Threading;
+
+    /// <summary>
+    /// A special benchmark to see that operation times are reported in uniform time units
+    /// and align nicely. These benchmarks produce widely different results but still we
+    /// should be able to compare them side-by-side without doing any mental translations
+    /// between nanoseconds to microseconds to milliseconds.
+    /// 
+    /// Expected output along these lines:
+    /// 
+    ///   Method |        AvrTime |      StdDev |       op/s |
+    /// -------- |--------------- |------------ |----------- |
+    ///  Slower1 |      1.8112 us |   0.2015 us | 552,120.14 |
+    ///  Slower2 |      2.4149 us |   0.2425 us | 414,095.82 |
+    ///  Slower3 |      9.0558 us |   0.3052 us | 110,426.47 |
+    ///  Slower4 |     81.2006 us |   1.2585 us |  12,315.18 |
+    ///  Slower5 |  4,148.1727 us | 740.3897 us |     241.07 |
+    ///  Slower6 | 15,047.0119 us |  99.5290 us |      66.46 |
+    /// 
+    /// As can be seen, all time units are microseconds, even though the two slowest of them
+    /// is in the milliseconds territory.
+    /// </summary>
+    [BenchmarkTask(mode: BenchmarkMode.SingleRun)]
+    public class Intro_04_UniformReportingTest
+    {
+        [Benchmark]
+        public void Slower1()
+        {
+            Iterate(5);
+        }
+
+        [Benchmark]
+        public void Slower2()
+        {
+            Iterate(100);
+        }
+
+        [Benchmark]
+        public void Slower3()
+        {
+            Iterate(1000);
+        }
+
+        [Benchmark]
+        public void Slower4()
+        {
+            Iterate(10000);
+        }
+
+        [Benchmark]
+        public void Slower5()
+        {
+            Iterate(500000);
+        }
+
+        [Benchmark]
+        public void Slower6()
+        {
+            Iterate(500000);
+            Thread.Sleep(10);
+        }
+
+        private static int[] Iterate(int arraySize)
+        {
+            var rand = new Random();
+            var array = new int[arraySize];
+            for (var i = 0; i < arraySize; i++)
+            {
+                array[i] = rand.Next();
+            }
+
+            return array;
+        }
+    }
+}
--- a/BenchmarkDotNet/BenchmarkRunner.cs
+++ b/BenchmarkDotNet/BenchmarkRunner.cs
@@ -73,21 +73,16 @@ namespace BenchmarkDotNet
                     r.Benchmark,
                     Stat = new BenchmarkRunReportsStatistic("Target", r.Runs)
                 }).ToList();
+
+            // Ensure uniform number formats and use of time units via these helpers.
+            var averageTimeStats = reportStats.Select(reportStat => reportStat.Stat);
+            var timeToStringFunc = GetTimeMeasurementFormattingFunc(averageTimeStats);
+            var opsPerSecToStringFunc = GetOpsPerSecFormattingFunc();
+
             var table = new List<string[]> { new[] { "Type", "Method", "Mode", "Platform", "Jit", ".NET", "AvrTime", "StdDev", "op/s" } };
             foreach (var reportStat in reportStats)
             {
                 var b = reportStat.Benchmark;
-
-                // Ops/sec number formatting:
-                //      - Thousand separators: we generally expect large numbers so these 
-                //        would make it easier to view.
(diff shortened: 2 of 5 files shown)


BenchmarkDotNet/BenchmarkRunner.cs
```diff
@@ -295,5 +290,76 @@ private static void AssertBenchmarkMethodIsNotGeneric(MethodInfo methodInfo)
...
+        public static Func<double, string> GetTimeMeasurementFormattingFunc(IEnumerable<BenchmarkRunReportsStatistic> statistics)
+        {
+            // Find the smallest measurement in the primary statistics, which is the Median.
+            // This will determine the time unit we will use for all measurements.
+            var minRecordedMedian = statistics.Min(stat => stat.AverageTime.Median);
+
+            // Use the largest unit to display the smallest recorded measurement without loss of precision.
+            // TODO: This is duplicated in BenchmarkMeasurementStatistic.ToString(), figure out how to refactor this later?
+            Func<double, string> measurementToString;
+            if (minRecordedMedian < 1000)
```
AndreyAkinshin: Maybe we can declare an array `string[] { "ns", "us", "ms", " s"}` and write a `for` loop?

BenchmarkDotNet.Samples/Intro_04_UniformReportingTest.cs
```diff
@@ -0,0 +1,85 @@
...
+    ///  Slower5 |  4,148.1727 us | 740.3897 us |     241.07 |
+    ///  Slower6 | 15,047.0119 us |  99.5290 us |      66.46 |
+    /// 
+    /// As can be seen, all time units are microseconds, even though the two slowest of them
+    /// is in the milliseconds territory.
+    /// </summary>
+    public class Intro_04_UniformReportingTest
+    {
+        [Benchmark]
+        [BenchmarkTask(mode: BenchmarkMode.SingleRun)]
```
AndreyAkinshin: You can use the `BenchmarkTask` attribute once for the class.

BenchmarkDotNet/BenchmarkRunner.cs
ppanyukov: Good idea

ppanyukov: OK, we could do it in several ways, but I am not convinced any of it makes things any clearer or better.

Using `for` loop:
```
const int baseScale = 1000;
var timeUnits = new[] { "ns", "us", "ms", " s" };

int selectedTimeUnitIndex;
for (selectedTimeUnitIndex = 0;
    selectedTimeUnitIndex < timeUnits.Length; 
    selectedTimeUnitIndex++, minRecordedMedian /= baseScale)
{
    if (minRecordedMedian < baseScale) break;
}

var adjustedScale = (int)Math.Pow(baseScale, selectedTimeUnitIndex);
var adjustedTimeUnit = timeUnits[selectedTimeUnitIndex];
return (value) => string.Format(EnvironmentHelper.MainCultureInfo, "{0:N4} {1}", value / adjustedScale, adjustedTimeUnit);
```

A more fun(c) way:
```
const int baseScale = 1000;
var timeUnits = 
    new[] { "ns", "us", "ms", " s" }
    .Select((timeUnit, index) =>
    {
        var adjustedScale = Math.Pow(baseScale, index);
        var isPicked = minRecordedMedian / adjustedScale < baseScale;
        return new { timeUnit, adjustedScale, picked = isPicked, };
    });

var selected = timeUnits.FirstOrDefault(t => t.picked) ?? timeUnits.Last();
return (value) => string.Format(EnvironmentHelper.MainCultureInfo, "{0:N4} {1}", value / selected.adjustedScale, selected.timeUnit);
```

Unless I'm doing something completely stupid here?

BenchmarkDotNet.Samples/Intro_04_UniformReportingTest.cs
ppanyukov: The benefit of this approach you keep the benchmark task setup very close to the benchmark code so you don't have to scroll up to see it.  Besides, they didn't start with all the same setup.

But yeah, I can pull that to the class level if that's the preference. No strong feelings on this from my side.

AndreyAkinshin: My opinion is the following: if all of benchmark method should has the same `BenchmarkTask` attribute, the attribute should be applied for the class, it reduces code duplication.

BenchmarkDotNet/BenchmarkRunner.cs
AndreyAkinshin: Hmm... I think, we can create an additional class like the following:
```cs
public class MeasurementTimeUnit
{
    public string FullName { get; }
    public string ShortName { get; }
    public double SecondsFactor { get; }

    public MeasurementTimeUnit(string fullName, string shortName, double secondsFactor)
    {
        FullName = fullName;
        ShortName = shortName;
        SecondsFactor = secondsFactor;
    }

    public static readonly MeasurementTimeUnit Nanoseconds = new MeasurementTimeUnit("Nanoseconds", "ns", 1e-9);
    // ...

    public static readonly MeasurementTimeUnit[] All =
    {
        Nanoseconds,
        // ...
    };

    public static MeasurementTimeUnit GetBestTimeUnitFromNanoseconds(double ns)
    {
        // check all time units from All
    }

    public string FormatNanoseconds(double ns)
    {
        return string.Format(EnvironmentHelper.MainCultureInfo, "{0:N4} {1}", ns / 1e9 * SecondsFactor, ShortName);
    }
}
```
It allows to remove the `BenchmarkTimeSpan` class and make the reporting logic more readable.

ppanyukov: Well surely the original 4 if statements is the simplest possible way and pretty much in line with the existing codebase, isn't it just good enough for now? Aren't we overthinking :)

Besides, you'd still have to implement similar kind of logic in `GetBestTimeUnitFromNanoseconds` right?

Also having this class wouldn't automatically remove the need for `BenchmarkTimeSpan` because it's used in several places. The `BenchmarkTimeSpan.ToString()` probably can be removed, I don't see any direct usages of it.

And to get rid of `BenchmarkTimeSpan` entirely. It would be nice. It's possible to replace it with with `TimeSpan` and a couple of extension methods to get `us` and `ns`. But then everywhere else we use `double` so the use of `TimeSpan` just in one place seems odd. Maybe we could get rid of `BenchmarkTimeSpan` and replace it with raw `double` and just do conversion to seconds (or whatever) directly in one place where it's needed. Or have properties in `BenchmarkRunReportsStatistic` to do `OpsPerSecond` and `AverageNanoseconds` instead of calculating them somewhere else.

You see, there are lots of options and other things we could do, and we also want the formatting done.

How about we do this:

- proceed with the `if` statements as they are for now, if only to make progress with these additions:

    - get rid of `ToString()` in `BenchmarkTimeSpan`
    - move `[BenchmarkTask]` to class level (you seem to care, and I don't have any strong opinion on that).

After this, as a separate PR:

- get rid of `BenchmarkTimeSpan` entirely, or replace it with something else. And we can discuss these changes separately. (And I'm not saying that I'm entirely disagreeing with your idea to have this `MeasurementTimeUnit` class or something similar)

What do you think about this plan.

PS. Do you know what will happen to this discussion if I squash commits and do `push -f`?

AndreyAkinshin: Ok, the plan looks good. The discussion should be available by the link.

ppanyukov:
New commit as per the plan: squashed pushed etc.

PR #26, opened 2015-08-18
message for Obsolete warning + updated README

adamsitnik:
Just a minor fix: message for Obsolete warning which tells what to use now + updated README (I was confused after update so fixed this)

PS. This tool is just awesome!!


AndreyAkinshin:
Thanks!

PR #32, opened 2015-09-10
Update BenchmarkProgram.txt

mattwarren:
So that any errors in the [Benchmark] method are shown in the console, at the moment you just get the cryptic message:

> Something bad happened during the execution of {exeName}. Try to run the benchmark again using an AnyCPU application

@AndreyAkinshin are there any perf issues with wrapping the measurement code like this? i.e. is it a problem putting in an exception block at the top-level?

**Before**:

exception thrown by the benchmark - default

**After**:

exception thrown by the benchmark - proposed behaviour


AndreyAkinshin:
>  are there any perf issues with wrapping the measurement code like this? i.e. is it a problem putting in an exception block at the top-level?

I don't think so, it looks ok.

mattwarren:
thanks

PR #43, opened 2015-10-27
More examples also couldnt make CoreCLR to work on my machine. Check the example. 

redknightlois:
Added RotateBits will be interesting when https://github.com/dotnet/coreclr/issues/1619 is included in the Desktop RyuJIT and the CoreCLR runtime benchmarking works properly.

Added alternative implementations of PopCount for reference
 * Naive If based implementation
 * Parallel PopCount (two operations simultaneously avoiding data hazards).


AndreyAkinshin:
Thanks!
About https://github.com/dotnet/coreclr/issues/1619: it is awesome! I want to add the CoreCLR support in the nearest future (but it is not a simple issue because we should build benchmarks special way).

PR #47, opened 2015-11-19
Change Jit_GenericsMethod benchmark to reproduce on x64 both Legacy & RuyJit

alexandrnikitin:
That one reproduces on all platforms. Let's think about more concise repro and figure out what optimizations x64 performs comparing to x86 that the previous one didn't reproduce.


mattwarren:
@alexandrnikitin BTW thanks for your article/presentation on [.NET Generics under the hood](https://alexandrnikitin.github.io/blog/dotnet-generics-under-the-hood/), it was a great read!

alexandrnikitin:
@mattwarren Thank you! I'm glad you liked it.

PR #58, opened 2015-12-09
build fix: removing post build event that tried to copy non-existing dll

adamsitnik:
I could not compile the solution due to lack of file that was copied in post build event.

I have checked and "CLRMD\x86\msdia120.dll" was never a part of the repo. But from what I see is that @mattwarren has added CLRMD folder with this post build event [recently](https://github.com/PerfDotNet/BenchmarkDotNet/commit/37f468db46c997d67f9ef3efe399bfe9a8e31b49). So Matt my question to you is whether you should add the missing file or is my fix ok?


mattwarren:
@adamsitnik thanks for spotting and fixing this. Yeah it's my bad, I left this in after I tidied a few things up!!

mattwarren:
Sorry about this, I'm not sure what happened here.

> I have checked and "CLRMD\x86\msdia120.dll" was never a part of the repo. 

It should be, see https://github.com/PerfDotNet/BenchmarkDotNet/tree/master/CLRMD, maybe I did something wrong when I added them

mattwarren:
Uggh, sorry I mis-read your post, CLRMD\\**x86**\\msdia120.dll is indeed missing. I'll fix this. 

Also see #62

adamsitnik:
I can see that you have added the files, so I close my PR now.

mattwarren:
Thanks

PR #106, opened 2016-02-28
Changed GCDiagnoser to inspect live ETW event stream instead of recording to a file

### LINKED ISSUE #102: GCDiagnoser should create a real-time ETW session and not use logman to record to a file
Continuing the discussion from https://github.com/PerfDotNet/BenchmarkDotNet/commit/ae71f296843b2d9bd05892477241c4193573b42e. Need to make the diagnosers project refer to the TraceEvent NuGet package (which supports .NET 3.5), and then the `TraceEventSession` class can be used for real-time event processing.

@mattwarren

goldshtn:
If merged this closes #102.


mattwarren:
@goldshtn thanks again for doing this, I ran it locally and it all looks good.

PR #115, opened 2016-03-11
Executing single Benchmark for multiple Runtimes

adamsitnik:
We can use the advantages of xprojs and project.json files that allows targeting multiple frameworks and run single benchmark for multiple runtimes.

The mandatory thing is to set Framework for Runtime.Clr in explicit way. The app that contains the benchmarks must also be targeting all the frameworks. Just like BenchmarkDotNet.Samples does.

Sample config:
```
internal class AllRuntimesConfig : ManualConfig
{
	public AllRuntimesConfig()
	{
		Add(Job.Default.With(Runtime.Dnx).With(Jit.Host));
		Add(Job.Default.With(Runtime.Core).With(Jit.Host));
		Add(Job.Default.With(Runtime.Clr).With(Jit.Host).With(Jobs.Framework.V40));
	}
}
```

Sample results:
image

### DIFF
--- a/BenchmarkDotNet.IntegrationTests/MultipleRuntimesTest.cs
+++ b/BenchmarkDotNet.IntegrationTests/MultipleRuntimesTest.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Linq;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains;
+using Xunit;
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+    public class MultipleRuntimesTest
+    {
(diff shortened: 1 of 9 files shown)


AndreyAkinshin:
Hi @adamsitnik, I have several points:
* Don't use master branch for PR, use `develop`. By desing, `master` is only for stable releases.
* Here is the plan: 
  1. publish v0.9.3 with CoreCLR support (I still play with different settings in `project.json` and prerelease versions, try it on several OS)
  2. merge master into `develop` (probably, an issue for you)
  3. continue to write new stuff in the `develop` branch and merge it into master per release.

adamsitnik:
Sorry for trouble, I am new to git workflow. Thanks for the explanation, it makes more sense to me now. I should probably also start using the chat :)
Btw. I hope you like this feature.

adamsitnik:
I am closing, will reopen as PR to develop branch

Issue #53, opened 2015-11-28
Diagnostic improvements


AndreyAkinshin:
@mattwarren, I want the following list of features:
- [X] 1. I want be able to add a diagnostic class with the `AddDiagnoster` method in manual mode (automatic loading should work in audo mode).
- [X] 2. I want a special diagnostic `BenchmarkIterationMode` for diagnostic. It should be run before the pre-warmup stage.
- [ ] 3. Now I have troubles with diagnostic of x86 benchmark from the x64 host application. Can we fix it?
- [x] 4. It would be cool to split the current diagnostic class to source diagnoster (IL and ASM) and runtime diagnoster (GC, Segments, and so on).
- [x] 5. We should cover the diagnostics by unit tests.

What do you think?

AndreyAkinshin:
I have tried to implement tasks 1 and 4. Please, review (https://github.com/PerfDotNet/BenchmarkDotNet/commit/28083486f9423229545969617cab4b05909ba346).
Now you can add specific diagnostic with a command like arguments like `-d=source,runtime` (`d` means diagnostic, target diagosers will be matched by name).

mattwarren:
> I have tried to implement tasks 1 and 4. Please, review (2808348).

Thanks for doing that, at first glance it looks good, but I'll have a more of a look later on today.

> 2 - I want a special diagnostic BenchmarkIterationMode for diagnostic. It should be run before the pre-warmup stage.

Just curious, what's the reason for this? Is it so the diagnostic doesn't interfere with calculating the `invokeCount`

> 3 - Now I have troubles with diagnostic of x86 benchmark from the x64 host application. Can we fix it?

Yeah that's a problem with the CLRMD library, it doesn't want to analyse a 32-bit process from a 64-bit host app. I'll take a look, I *think* I can fix it by having a small shim 32-bit process that can then launch the 32-bit benchmark.

AndreyAkinshin:
> Thanks for doing that, at first glance it looks good, but I'll have a more of a look later on today.

Ok. I just implemented basic concepts. Please, refactor the diagnostic classes at its discretion.

> Just curious, what's the reason for this? 

Theoretically, it can affects the warmup process. We should try to avoid any additional logic between our runs. So, I suggest to create a special stage: diagnostic. We can run it directly without hacks like `if (line.StartWith("// Warmup") ...`. The source code will be more clear.

[...]

PR #117, opened 2016-03-13
Executing single Benchmark for multiple Runtimes [with Diagnoser attached]

adamsitnik:
project.json files allows us to target multiple frameworks. The master process of BenchmarkDotNet always generates, builds and executes new program that executes single benchmark. Master process monitors and parses the output of the the child process. When any diagnoser is used, it also attaches to the child process.

What I have noticed is that the runtime and framework of the child and master process do NOT have to be the same. The only limitation is for some Diagnosers the processor architecture must be the same for both processes.

So with project.json files and dotnet cli we can build and execute single program for different runtimes/frameworks configuration. Currently diagnosers do not support Core, but if our master process is not Core, and the child is, then we can simply attach to it (only on Windows) and use existing diagnosers.

I am not a native english speaker, but I hope that my explanation is good enough to understand it. Please ask questions if something is unclear.

The mandatory thing is to set Framework for Runtime.Clr in explicit way. The app that contains the benchmarks must also be targeting all the frameworks. Just like BenchmarkDotNet.Samples does.

Sample config:
```
internal class MultipleRuntimesConfig : ManualConfig
{
	public MultipleRuntimesConfig()
	{
		Add(Job.Default.With(Runtime.Dnx));
		Add(Job.Default.With(Runtime.Core));
		Add(Job.Default.With(Runtime.Clr).With(Jobs.Framework.V40));
	}
}
```

Sample results:
image

Sample results for GCDiagnoser:
image

### DIFF
--- a/BenchmarkDotNet.IntegrationTests/MultipleRuntimesTest.cs
+++ b/BenchmarkDotNet.IntegrationTests/MultipleRuntimesTest.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Linq;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains;
+using Xunit;
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+    public class MultipleRuntimesTest
+    {
+        private class MultipleRuntimesConfig : ManualConfig
+        {
+            public MultipleRuntimesConfig()
+            {
+                Add(Job.Dry.With(Runtime.Dnx).With(Jit.Host));
+                Add(Job.Dry.With(Runtime.Core).With(Jit.Host));
+                Add(Job.Dry.With(Runtime.Clr).With(Jit.Host).With(Framework.V45));
+            }
+        }
+
+        [Fact]
+        public void SingleBenchmarkCanBeExecutedForMultpleRuntimes()
+        {
+            var summary = BenchmarkRunner.Run<MultipleRuntimesTest>(new MultipleRuntimesConfig());
+
+            Assert.True(summary.Reports.Values
+                .All(report => report.ExecuteResults
+                .All(executeResult => executeResult.FoundExecutable)));
+
+            Assert.True(summary.Reports.Values.All(report => report.AllMeasurements.Any()));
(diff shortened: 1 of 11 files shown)


AndreyAkinshin:
@mattwarren, can you take a look?

BenchmarkDotNet.IntegrationTests/MultipleRuntimesTest.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public class MultipleRuntimesTest
+    {
+        private class MultipleRuntimesConfig : ManualConfig
+        {
+            public MultipleRuntimesConfig()
+            {
+                Add(Job.Dry.With(Runtime.Dnx).With(Jit.Host));
+                Add(Job.Dry.With(Runtime.Core).With(Jit.Host));
+                Add(Job.Dry.With(Runtime.Clr).With(Jit.Host).With(Framework.V45));
```
AndreyAkinshin: You shouldn't write `With(Jit.Host)`, `Jit.Host` is the default value.

mattwarren:
> The only limitation is for some Diagnosers the processor architecture must be the same for both processes.

@adamsitnik Yeah I need to remove this restriction, so we allow 32-bit/64-bit `Source` diagnostics regardless of the bit-ness of the parent process. It's being tracked as part of #53

mattwarren:
@AndreyAkinshin @adamsitnik I realise I'm assigned to this issue, but I've lost track of what I was supposed to be doing ;-)

Do I just need to look through the code or am I included because of the 32-bit/64-bit issue with `Source` diagnostics? If it's the latter then I don't think the PR shouldn't be blocked by this issue.

adamsitnik:
@mattwarren I guess that @AndreyAkinshin assigned you because this issue covers Diagnosers. I think that I will just close it (because of the conflicts) and push the changes directly to develop. As far as you are ok with that.

mattwarren:
@mattwarren I guess that @AndreyAkinshin assigned you because this issue covers Diagnosers

Ah that makes sense. I've looked through the PR and all the Diagnoser stuff looks fine.

>  I think that I will just close it (because of the conflicts) and push the changes directly to develop. As far as you are ok with that.

That's absolutely fine with me.

adamsitnik:
I have pushed the changes directly to develop, so now I am closing this.

Issue #41, opened 2015-10-15
Seems, not supported "sub-folder"


constructor-igor:
My binaries folder has next structure:
```
folder        Compilation
folder       Compilation\Common\
tesproject TestProject.dll
```

I added test in TestProject 
```c#
    [TestFixture]
    public class BenchmarkTests
    {
        public BenchmarkTests()
        {
        }
        [Benchmark]
        public void ByDoubleReader()
        {
            <code>
        }

        [Test]
        public void BenchMark()
        {
            new BenchmarkRunner().RunCompetition(new BenchmarkTests());
        }
    }
```
and found next problem:
folder <root>\Compilation\BenchmarkTests_ByDoubleReader_Throughput_HostPlatform_HostJit_NET-HostFramework
doesn't contain subfolder Common and "benchmark" cannot be loaded

AndreyAkinshin:
@constructor-igor, I'm not completely understand your situation. Can you share a minimal reproducible example?

constructor-igor:
@AndreyAkinshin I'll try to prepare the sample.
Main idea: binary folder contains subfolder and test project has reference to other binaries from the subfolder.

constructor-igor:
https://github.com/constructor-igor/TechSugar/tree/master/BenchmarkingLab/BenchmarkDotNet
working sample: BenchmarkWorkingSample
problem sample: BenchmarkProblemSample

mattwarren:
I'll take a look at this, I think it's similar to #42

mattwarren:
Okay, I've looked at this a bit more and I've been able to reproduce it outside of your test.

However I'm not entirely sure that we want to support this.

Let me explain, to make this happen you have to do **all** of the following steps:
  1. Add a reference to a 3rd party dll **directly** (**not** as a [project reference](https://msdn.microsoft.com/en-us/library/f3st0d45.aspx#Anchor_1)), i.e. in VS you do "Add Reference" -> "Browse..".
  1. You have to build your 3rd party dll into a separate folder "..\Binaries\Common", compared to you main dll "..\Binaries"
[...]

Issue #66, opened 2016-01-06
Friendliness to LinqPad


omariom:
Would be nice to have ability to run benchmarks in LinqPad and get nicely formattet results in LinqPad's Results output.

AndreyAkinshin:
Good idea, added to roadmap: https://github.com/PerfDotNet/BenchmarkDotNet/wiki

mattwarren:
We actually have LinqPad support (or we did, I haven't tested it since), I made it work as part of #42.

So maybe it's enough to just do something like this in the `.linq` file:

``` csharp
public static void Main()
{
	var report = new BenchmarkRunner().Run<Intro_00_Basic>();
	report.Dump();
}
```

mattwarren:
@AndreyAkinshin sorry, I've only just read the note on the wiki:

> Please, use the develop branch for developing. The master branch should correspond the latest NuGet package of the library.

I'll make sure I use the develop branch in the future.

AndreyAkinshin:
@mattwarren, yes, we should implement the `Dump` method.

mattwarren:
Sorry, I wasn't clear. `Dump()` is built-in to LinqPad and it (usually) does a pretty good job of displaying whatever object you give it.

When I get a chance I'll run that LinqPad code and post the output here.

mattwarren:
> We actually have LinqPad support (or we did, I haven't tested it since), I made it work as part of #42.

Whoops, seems like our LinqPad support broke somewhere along the way, I'll take a look at fix it.

pjmagee:
Yep, it says it cannot find UserQuery namespace. :(

mattwarren:
@adamsitnik looks like your recent changes fixed this also. When running against your [references branch](https://github.com/PerfDotNet/BenchmarkDotNet/tree/references) (i.e. PR #125) in LINQPad I get the following output (i.e. it works!):

```
// **************************
// Benchmark: Intro_00_Basic_Sleep_Dry
// *** Generate *** 
// File doesn't exist: C:\Program Files (x86)\LINQPad4\benchmarkdotnet.dll
//   Actually at: C:\Users\warma11\AppData\Local\Temp\LINQPad\_abcwrien\shadow_bljeos\benchmarkdotnet.dll
//   Copying benchmarkdotnet.dll
//   from: C:\Users\warma11\AppData\Local\Temp\LINQPad\_abcwrien\shadow_bljeos
[...]

PR #125, opened 2016-03-21
supporting all kinds of references for generated project

### LINKED ISSUE #49: Dependent assemblies are not copied or added to the project file.
When benchmarking complex operations you may end up with situations like the following:

On assembly ```assembly1.dll```
```csharp
public class A
{}
```
On assembly ```benchmark.dll```
```csharp
public class Program : A
{
   [Benchmark]
   public void Method ()
   {
       // calling some method of A for initialization.  
   }
}
```
In this case ```benchmark.dll``` will be referenced while  ```assembly1.dll``` is not causing a build failure like the following:

```
// **************************
// Benchmark: Program_Method (Throughput_HostPlatform_HostJit_NET-HostFramework_ArraySize) [-w=5 -t=10]
// Generated project: G:\Src\ravendb-3.5-git\Raven.Voron\Voron.Tryout\bin\Release\Program_InsertInTable_Throughput_HostPlatform_HostJit_NET-HostFramework_ArraySize
// Build:
// Program.cs(14,18): error CS0012: The type 'Voron.Tests.Tables.TableStorageTest' is defined in an assembly that is not  referenced. You must add a reference to assembly 'Voron.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=37f41c7f99471593'.
// OverallResult = Failure
// ***** Competition: Finish  *****
```

### LINKED ISSUE #72: Referenced assembly dll-file (directly via file) not referenced in generated Program.csproj
Referencing a file directly (e.g. Browse... for reference) does not work, the generated Program.csproj does not reference it.

### LINKED ISSUE #123: Can not run benchmark that references custom framework library (like WindowsBase)
using System.Windows;

    public class Repro
    {
        [Benchmark]
        public Point? Foo()
        {
            return new Point();
        }
    }

adamsitnik:
I take all assemblies referenced by assembly that defines target benchmark and for each of it that is not on our predefined assembly list (System, System.Core etc) I load it and check whether it was loaded from GAC or not.

If it was loaded from GAC then in csproj I don't specify path at all  (it is most probably framework assembly like in #123), otherwise I specify absolute path. Why absolute? Because it supports custom paths as well (#72). Perhaps it could also help with LINQPad problems?

By taking all referenced assemblies we can fix a problem when benchmark method was using, but not returning some type from custom dll.

This hoepfully closes #49, closes #72, and closes  #123

@mattwarren @AndreyAkinshin In order to test it you can run runClassicTests.cmd or modify Sample's project and add some custom references to project.json file and debug #CLASSIC profile.

It is quite fresh idea, please be suspicious ;)

**EDIT**
Another change: if the target framework version has not been set in explicit way I take the highest one used to avoid any framework mismatches.
I have also made 3.5 Obsolete and added 4.6.1 support.


mattwarren:
@adamsitnik thanks for doing this!

BTW It might also solve #41?

adamsitnik:
@mattwarren You welcome. It would be great if we could fix all the bugs, then we could focus only on the new features. And yes, it should fix #41 .

mattwarren:
> It would be great if we could fix all the bugs, then we could focus only on the new features.

Yeah that would be nice!

mattwarren:
@adamsitnik I think you might need to check this fix against [the code sample](https://github.com/PerfDotNet/BenchmarkDotNet/issues/72#issuecomment-172197979) that @nietras provided in #72. For me it's not working, although I could be doing something wrong!?

Here's the [Program.csproj.zip](https://github.com/PerfDotNet/BenchmarkDotNet/files/183874/Program.csproj.zip) and [Program.cs.zip](https://github.com/PerfDotNet/BenchmarkDotNet/files/183880/Program.cs.zip) files that were generated, it has your extra references (with HintPath), but it still fails to compile first time:

image

I think that it would be good to have an integration test that mimics the scenario in #72, i.e. a helper/utility project/dll that's referenced from the main Benchmark, do you agree?

adamsitnik:
@mattwarren I agree, I will try to reproduce this problem and write test for it. Do you have any idea what could be the cause here? Could it be a missing 'using' statement in 'Program.cs'?

mattwarren:
I don't think it's that, because when we shell out and run `BuildBenchmark.bat` it compiles fine and Program.cs/Program.csproj are not changed. I'll keep digging and see if I can see the problem

mattwarren:
I've enabled logging of the MSBuild task that we run from inside BenchmarkDotNet, here's some interesting lines from the output, it may be my machine setup that's causing the problem:

```
Build FAILED.

C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.Common.targets(1605,5): warning MSB3268: The primary reference "DotNetCross.Memory.Benchmark" could not be resolved because it has an indirect dependency on the framework assembly "System.Runtime, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" which could not be resolved in the currently targeted framework. ".NETFramework,Version=v4.5.2". To resolve this problem, either remove the reference "DotNetCross.Memory.Benchmark" or retarget your application to a framework version which contains "System.Runtime, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.Common.targets(1605,5): warning MSB3268: The primary reference "DotNetCross.Memory.Unsafe" could not be resolved because it has an indirect dependency on the framework assembly "System.Runtime, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" which could not be resolved in the currently targeted framework. ".NETFramework,Version=v4.5.2". To resolve this problem, either remove the reference "DotNetCross.Memory.Unsafe" or retarget your application to a framework version which contains "System.Runtime, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
Program.cs(10,7): error CS0246: The type or namespace name 'DotNetCross' could not be found (are you missing a using directive or an assembly reference?)
Program.cs(17,36): error CS0400: The type or namespace name 'DotNetCross' could not be found in the global namespace (are you missing an assembly reference?)
    2 Warning(s)
    2 Error(s)

Time Elapsed 00:00:00.26
```

And here's the full log [BenchmarkDotNet-MSBuild.zip](https://github.com/PerfDotNet/BenchmarkDotNet/files/183901/BenchmarkDotNet-MSBuild.zip)

nietras:
Hi all, perhaps I can give some extra information here. I'm assuming you use the "old" code from   `DotNetCross.Memory.Unsafe` repo, since the newest version actually doesn't work with regard to the benchmark, a project reference vs package reference issue.

Anyway,  `DotNetCross.Memory.Unsafe.dll` was not properly packaged before (`<= 0.2.2`), which should not mean anything with regard to it working as such, but there was an issue here since it uses the reference dll "trick" introduced .NET 4.5 so it didn't support 4.0 for example even though, the package claimed this. Perhaps, the warnings are related to that, and depends on how the project made by `BenchmarkDotNot` is configured.

Additionally, the `Unsafe` package included `.dll` in it's name, another mistake on my behalf. Had to go through hoops to get this changed on nuget, they deleted the old packages, anyway since `0.2.3` packaging and naming of the package should work. However, during this the benchmarks and tests have stopped working in the repo, so don't use the newest one for now.

To summarize, there are some packaging issues with old `Unsafe` however I don't think this should cause the problem with the benchmark not working. The benchmark does not have a `using DotNetCross.Memory.Unsafe;` in [Program.cs](https://github.com/DotNetCross/Memory.Unsafe/blob/master/src/DotNetCross.Memory.Unsafe.Benchmark/Program.cs) since the benchmark is isolated in a separate class [BasicReadWriteBenchmark](https://github.com/DotNetCross/Memory.Unsafe/blob/master/src/DotNetCross.Memory.Unsafe.Benchmark/BasicReadWriteBenchmark.cs). Which I hope will be supported :)

mattwarren:
@nietras thanks for the extra info

adamsitnik:
> or retarget your application to a framework version which contains "System.Runtime, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".

@mattwarren This information is very helpful! I suspect that we set wrong version of .NET in .csproj. Will investigate it.

adamsitnik:
@mattwarren I have cloned the code from #72 sample and it works for me. The difference is that the generated csproj file has `<TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>` You can find the csproj [here](https://gist.github.com/adamsitnik/d2f9389f5ee0585e4252)

mattwarren:
@adamsitnik glad you found this issue, thanks for working on this.

BTW do you plan to add an integration test that mimics the project that @nietras provided, or do you think the test you've already added is enough? (i.e. `ValuesReturnedByBenchmarkTest.cs`)

adamsitnik:
> BTW do you plan to add an integration test that mimics the project that @nietras provided, or do you think the test you've already added is enough? (i.e. ValuesReturnedByBenchmarkTest.cs)

My current plan is to at least try. The problem is that it might be cumbersome. I will try the following:
1. Create another project that defines a benchmark that is using few different libs (custom paths, nugets, explicit references)
2. Create new Testing project (csproj based) that targets only "classic runtimes" and reuses 1 project

adamsitnik:
Ok I have everything covered with the integration tests and the 'conflict' bug fixed.

I added 3 new projects:
* BenchmarkDotNet.IntegrationTests.DifferentRuntime that is 4.6 and using nuget library that has indirect reference to System.Runtime v 4.0.20.0 (to mimic @nietras case)
* BenchmarkDotNet.IntegrationTests.CustomPaths that is 4.5.1 and has dependency to dll in custom path (I used Microsoft.Diagnostics.Runtime from our repo for that)
* BenchmarkDotNet.IntegrationTests.Classic that is Test project for Classics only, currently it has only tests for the comiplation of these complex scenarios

@mattwarren the cure for conflicts was to use assembly.FullName when referencing dll that has multiple versions. To avoid extra bat call I used newer version of Msbuild dlls (4.0 -> 12.0). Msbuild change affects only our 4.5 target, I did not remove backup call to msbuild

@AndreyAkinshin it would be great if you could test it and include in the upcoming release, this closes 4 issues ;)

mattwarren:
@adamsitnik I think your fixed LINQPad also (i.e. #66), see https://github.com/PerfDotNet/BenchmarkDotNet/issues/66#issuecomment-200440575

PR #138, opened 2016-04-04
Percentiles added into Statistics, StatisticColumn, BaselineDiffColumn

ig-sinicyn:
Redone as #164. This one will be closed

As I wrote [here](https://github.com/PerfDotNet/BenchmarkDotNet/issues/126#issuecomment-200302063) I begin to do small PR to make it easier to use BenchmarkDotNet as an competition tests runner.

This PR adds Nth percentile and Nth percenile scaled columns. These should be used to rank and to compare competitors if the timings are varying from run to run.

More info on these: 
http://www.goland.org/average_percentile_services/
http://apmblog.dynatrace.com/2012/11/14/why-averages-suck-and-percentiles-are-great/

P.S. I'm not sure that proposed API design is good enough so it's definitely subject to discuss:)

### DIFF
--- a/BenchmarkDotNet.Tests/StatisticsTests.cs
+++ b/BenchmarkDotNet.Tests/StatisticsTests.cs
@@ -128,5 +128,36 @@ namespace BenchmarkDotNet.Tests
             Assert.Equal(12.34974, summary.ConfidenceInterval.Lower, 4);
             Assert.Equal(18.65026, summary.ConfidenceInterval.Upper, 4);
         }
+
+        [Fact]
+        public void PercentileValuesTest()
+        {
+            var summary = new Statistics(Enumerable.Range(1, 30));
+            Print(summary);
+            Assert.Equal(1, summary.Percentiles.P0);
+            Assert.Equal(8.25, summary.Percentiles.P25);
+            Assert.Equal(15.5, summary.Percentiles.P50);
+            Assert.Equal(20.3333, summary.Percentiles.P67, 4);
+            Assert.Equal(24.2, summary.Percentiles.P80, 4);
+            Assert.Equal(25.65, summary.Percentiles.P85);
+            Assert.Equal(27.1, summary.Percentiles.P90);
+            Assert.Equal(28.55, summary.Percentiles.P95, 4);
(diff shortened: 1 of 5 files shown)


AndreyAkinshin:
@ig-sinicyn, thanks for the PR.

Can you use an integer value for percentile instead of `PercentileLevel`?

ig-sinicyn:
@AndreyAkinshin, yes, of course :)

Just to clarify. There're two overloads of the PercentileValues.Percentile method, first takes enum and second - double value in range 0..1.

The first should be removed and the second should be changed to accept int in range 0..100, right?

AndreyAkinshin:
> The first should be removed and the second should be changed to accept int in range 0..100, right?

Yep, LGTM.

ig-sinicyn:
Ok, will do.

mattwarren:
Firstly thanks for doing this, having percentiles is great!

Secondly, I'm a bit late to this PR, but I think it would be nice if there was an [integration test](https://github.com/PerfDotNet/BenchmarkDotNet/tree/master/BenchmarkDotNet.IntegrationTests) that covered this new feature.

You could create a brand new test, or extend the relevant existing ones, i.e. BaselineDiffColumnTest.cs, StatResultExtenderTests.cs, etc)

ig-sinicyn:
@mattwarren Not a problem, will do :)

It looks that the PR is slightly outdated. Should I close it and create a new one from a current develop branch?

mattwarren:
> It looks that the PR is slightly outdated. Should I close it and create a new one from a current develop branch?

Sorry, I'm not a GitHub expert. If that's easiest go ahead and create a new one

ig-sinicyn:
Well, me too :)

Ok, I'll push the commits into the current pull request and if there will be some troubles with the merging I'll redo the PR.

goldshtn:
What you're looking for is squash :-)

ig-sinicyn:
Well, it turns out I'll have to redo the PR. The BaselineDiffColumnTest in the month-old version is failing.

I'll close this after submitting a new one.

ig-sinicyn:
Redone as #164

PR #173, opened 2016-05-11
The benchmark is improperly testing ILP

redknightlois:
The benchmark is improperly testing the Max operation because as all the numbers are 0 only a single path will be exercised.

The side effect is that the microarchitecture is able to branch predict everything and also there are no load-store dependencies (which is not true in the general case).
I modified the test to account for a random array and also added other ways to calculate Math.Max which would showcase the impact on instruction level parallelism in the face of store-load dependencies.


AndreyAkinshin:
Thanks!

PR #178, opened 2016-05-13
Prerequisites for machines running VS2015+ - Update README.md

voltagex:
Add a note about needing MSBuild12/Microsoft Build Tools 2013 installed on machines that are only running 2015+

I posted in Gitter about this issue a little while ago but forgot to test this before moving on to other tasks - https://www.microsoft.com/en-au/download/details.aspx?id=40760 allows BenchmarkDotNet to run correctly. The wording of this and creating a whole new section is a little clunky but I thought I better get the PR in before I completely forgot about it

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@
 
 ## Content
 
+* [Prerequisites](#prerequisites)
 * [Getting started](#getting-started)
 * [Configs](#configs)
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
We want to migrate to Roslyn in the next version. So, BenchmarkDotNet will work without installed MSBuild.

PR #182, opened 2016-05-19
Use same configured counts for IdleWarmup/Target as Main

leppie:
The values should match the config.
Alternatively, it should be exposed for configuration.

### DIFF
--- a/BenchmarkDotNet/Running/MethodInvoker.cs
+++ b/BenchmarkDotNet/Running/MethodInvoker.cs
@@ -94,8 +94,8 @@ namespace BenchmarkDotNet.Running
             if (job.Mode == Mode.Throughput)
             {
                 invokeCount = RunPilot(multiInvoke, job.IterationTime);
-                RunWarmup(multiInvoke, invokeCount, IterationMode.IdleWarmup, Count.Auto);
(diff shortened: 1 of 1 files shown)


mattwarren:
LGTM

@adamsitnik or @AndreyAkinshin is there any reason that you can think of why we were using `Count.Auto` in this case and/or why we can't use `job.WarmupCount` instead?

AndreyAkinshin:
Yep, it's by design. We don't want to specify iteration count for the idle iteration. And it shouldn't match to the main iteration. Maybe we should write it in documentation or xml docs.
Also I'm not sure that we should be able to configure idle settings. `Count.Auto` should work fine. Any use-cases where we need this feature?

mattwarren:
> Any use-cases where we need this feature?

Good question, I'm not sure what the issue that @leppie had was, I'll let him answer that.

leppie:
It is good for 99% of the cases, but when you start running very tight benchmarks (and many of them), the idle measurement takes forever (~30-40% of the actual bench due to huge op count).

Perhaps a better idea would be to only do idle measurement once per param combo (do you really need it on every benchmark?).

AndreyAkinshin:
@leppie, ok, I will upgrade idle measurements.

PR #207, opened 2016-06-12
Fixed tests not using OutputLogger

gigi81:
Wired up OutputLogger to tests that were not using it so that output can
be seen in AppVeyor under the tests tab


AndreyAkinshin:
@gigi81, can you rename `_output` to `output` (without the underscore)?

gigi81:
@AndreyAkinshin done.

Issue #184, opened 2016-05-22
[Proposal] remove jitting calls from BenchmarkProgram.txt


ig-sinicyn:
The `Templates\BenchmarkProgram.txt` file contains the following lines:
```cs
				$ParamsContent$
				instance.setupAction();
				instance.targetAction();
```
The last two should be removed as jitting is expected to be performed by  `new MethodInvoker().Invoke()` call:
```cs
        public void Invoke(IJob job, long operationsPerInvoke, ...)
        {
            // Jitting
            setupAction();
            targetAction();
            idleAction();

            // Run
            ...
        }
```

AndreyAkinshin:
@ig-sinicyn, we should run jitting as soon as possible. We can't remove these two lines from `BenchmarkProgram.txt` because it can spoil some benchmarks. E.g., it is important to run jitting for the target method _before_ the first use of the Stopwatch (See also: http://aakinshin.net/en/blog/dotnet/legacyjitx86-and-first-method-call/). Perhaps, we can remove the jitting section from the `Invoke` method.

ig-sinicyn:
@AndreyAkinshin
 wouldn't it be better to create explicit `Init()` step and to move all init logic (setting properties, performing the jitting calls etc) into it?

 I've naive-but-working implementation in `RunnableBenchmark` [here](https://github.com/rsdn/CodeJam/tree/master/BenchmarkDotNet/src/%5BL1_InProcess%5D/Toolchains).
Of course the API is just a mockup. For example, it should provide a way to notify diagnosers, but that's subject for another issue, #177.

AndreyAkinshin:
I will think about it.

AndreyAkinshin:
I have decided not to introduce an additional `Init()` method, jitting is too tricky, we could have some troubles with additional invocation. In the generated program, it is very important to jit our method as soon as possible.

Issue #215, opened 2016-06-27
Allow a [Cleanup] method to be specified


mattwarren:
To mirror the invocations of the `[Setup]` method, i.e.

- `[Setup]`
- Run and measure Benchmarks iterations
- `[Cleanup]`

This is useful when there is a resource that needs cleaning up after each set of iterations

adamsitnik:
Good idea! I would also Dispose the object if it is IDisposable (at the end)

mattwarren:
@adamsitnik I'm not following, which object do you mean?

adamsitnik:
@mattwarren  the one that is implementing Benchmark method(s)

AndreyAkinshin:
@mattwarren, can you write an example of benchmark where we need a cleanup method?

mattwarren:
This idea came from [this tweet](https://twitter.com/Aaronontheweb/status/746085821359362048), but I'm not really sure of the exact scenario, I'll try and find out.

DenisIstomin:
I think that Aaronontheweb (as a contributor to NBench) at [his tweet](https://twitter.com/Aaronontheweb/status/746085821359362048) said about "Cleanup" logic after **each** iteration. But BenchmarkDotNet uses "Setup" attribute for set of iterations called "Launch" (I hope so). That's why I think that "Cleanup" attribute could be used at benchmarks that do some work with "out of the process" resources - filesystem, for example.

AndreyAkinshin:
It makes sense, but I still want to have a real use-case for cleanup.

nblumhardt:
Hi all - thanks for the great project!

We're using BenchmarkDotNet to profile a [new Serilog "async" sink](https://github.com/serilog/serilog-sinks-async), and we need to:

 * Gracefully shut down worker threads
 * Delete temporary files (in a PR I'm working on)

...after running a benchmark. Without `[Cleanup]` it's hard to do this properly.

Does this cover the scenario information you need to consider including this feature?

Thanks!

nblumhardt:
Ah, now I see the commit above - great! Sorry about the noise :-)

adamsitnik:
@nblumhardt If you don't want to wait for 0.9.9 release you can download the pre-release package from our development feed `<add key="appveyor-bdn" value="https://ci.appveyor.com/nuget/benchmarkdotnet" />`

If that's not an option you can add a finalizer to your benchmark class. But personally I don't like this workaround.

PR #233, opened 2016-07-18
Add support of Cleanup attribute #215

DenisIstomin:


AndreyAkinshin:
@DenisIstomin, thanks for the PR. Code looks good, but I don't like your example: `List_LinkedListAllocation`. Benchmarks in this class don't have a steady state, so, we can't benchmark it with default settings (cleanup doesn't help in this case). Can you write another example?

DenisIstomin:
Wrote another example. I think that there are some redundant [operations](https://github.com/DenisIstomin/BenchmarkDotNet/blob/f728f4a60a104c966b477a383a961e530df8d07a/src/BenchmarkDotNet.Core/Templates/BenchmarkProgram.txt#L28-L30), because [that code](https://github.com/DenisIstomin/BenchmarkDotNet/blob/f728f4a60a104c966b477a383a961e530df8d07a/src/BenchmarkDotNet.Core/Templates/BenchmarkProgram.txt#L39) does [JITting](https://github.com/PerfDotNet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Running/MethodInvoker.cs#L40-L42) itself. But I'm decided not to remove them.

adamsitnik:
@DenisIstomin Thanks for the PR! Code looks good to me, I am merging ;)

AndreyAkinshin:
@DenisIstomin,
> I think that there are some redundant operations
There are no any redundant operations.
1. It is very important to perform JITting of the `setup` and `target` methods as soon as possible (before any BenchmarkDotNet-specific code). E.g. `BenchmarkEnvironmentInfo.GetCurrent().ToFormattedString()` can significantly affect resulted asm code of the `target` method.
2. We can't skip the JITting phase in the `Invoke` method because we can't be sure that it was already done (it can be false for a custom toolchain).

adamsitnik:
@AndreyAkinshin thanks for the explanation! Does it mean that you can close #184?

AndreyAkinshin:
I think so. There is a suggestion in #184 to introduce an `Init` method, but I'm afraid of unexpected side-effects. The source code of `MethodInvoker` will be refactored anyway (I'm trying to do it right now).

adamsitnik:
>The source code of MethodInvoker will be refactored anyway (I'm trying to do it right now). 

Great, I will implement #236 when you are done with this refactoring.

Issue #219, opened 2016-07-03
Online documentation and API reference


AndreyAkinshin:
Now we have almost all our documentation in a single README file. The documentation is great, but the file is too big, it's hard to maintain it, it doesn't include the API reference. Possibly, it makes sense to use a tool for automatic documentation generating. For examples, [DocFX](https://dotnet.github.io/docfx/index.html) looks good. We can integrate it to our CI build and generate docs fully automatically.

@mattwarren, @adamsitnik, what do you think?

## Current status

- [X] Guide
- [X] API Reference

adamsitnik:
>  it makes sense to use a tool for automatic documentation generating

Would you like to generate it from the comments from source code?

Maybe we could create a github.io web page with tutorial that tell how BDN works and how to use it?

AndreyAkinshin:
> Would you like to generate it from the comments from source code?

API reference from xml-docs in the source code, Guide from a set of markdown files.

> Maybe we could create a github.io web page with tutorial that tell how BDN works and how to use it?

Yes!

FransBouma:
Besides DocFX, you can also use [DocNet](http://fransbouma.github.io/DocNet/) (disclaimer: I wrote it). The advantage of DocNet is that it can autogenerate your ToC from headers and can deal with deeper nesting than DocFX. Of course it's suitable for automatic generation of a documentation website (we generate all our docs using it automatically). 

It's not suitable for pure API docs, it only consumes md files which form a documentation site. I'd go for sandcastle helpfile builder for that (https://github.com/EWSoftware/SHFB) as it can output to multiple formats (e.g. offline documentation sets in vs.net format, for the people who want those).

FransBouma:
Is this still up for grabs? Should I go ahead and create a documentation site from the readme using DocNet?

adamsitnik:
> Is this still up for grabs? Should I go ahead and create a documentation site from the readme using DocNet? 

@FransBouma Yes. That would be great!

AndreyAkinshin:
@FransBouma, sounds great! I have only two requests:
1. We still need the API reference. Can you implement it?
2. It would be cool two have at least two version of the documentation: for the latest stable version and for the latest commit from the master branch (for example, see http://nodatime.org/). Is it possible to do it with DocNet?

FransBouma:
[...]

PR #239, opened 2016-08-06
Added documentation source and output based on readme.md

FransBouma:
The documentation is created from the readme.md file and uses docnet
v0.13 or higher to generate its output. Documentation\UserGuideSource is
the folder which contains the markdown. Run generatehtml.cmd which will
generate html files from the md files into documentation\userguidehtml.
I committed the html output as well so there's an initial batch.
Configure where the files are generated in
documentation\userguidesource\docnet.json as well as other things
related to the html generation process.

For #219


AndreyAkinshin:
@FransBouma, we shouldn't keep the generated documentation in the master branch. Here we need only documentation sources and building scripts/infrastructure. Also I suggest to do a few folder renaming: `documentation` -> `docs`, `UserGuideSource` -> `guide`.

FransBouma:
> we shouldn't keep the generated documentation in the master branch

Ok, will remove that. 

> Here we need only documentation sources and building scripts/infrastructure. Also I suggest to do a few folder renaming: documentation -> docs, UserGuideSource -> guide.

Ok. The 'documentation' folder is linked to in the current develop.md file btw, so I'll adjust that too.

AndreyAkinshin:
Awesome! I will review it tonight.

FransBouma:
:) The 'documentation' folder wasn't referenced by the original develop.md so that's fine. The images folder in the documentation folder was unused. 

I just noticed I left an unused include in the docnet.json file: the Images folder isn't there anymore and therefore the 
```
	"SourceFoldersToCopy" : ["Images"],
```
line isn't needed anymore. It won't break or anything, but it's not needed anymore. Should I remove it or leave it? Removing it will trigger another appveyor build for no real reason. 

The docnet.json file dictates now to build the html to docs\guidehtml. This folder is excluded in gitignore, but you can change it to any other folder you want.

AndreyAkinshin:
> Should I remove it or leave it? Removing it will trigger another appveyor build for no real reason.

Don't hesitate to do any improvements, nothing wrong with additional appveyor builds.

FransBouma:
Agreed. :) Done.

FransBouma:
Any problems with the PR?

AndreyAkinshin:
@FransBouma, sorry for the late merge, I was too busy on this week. Thanks again, it looks awesome! Waiting for a PR with API reference.

FransBouma:
@AndreyAkinshin I haven't created a setup for the API ref docs as there are no xml doc comments in the source, so creating that would be a bit odd. If you really really want to, I could create one, but the end results are barely useable. (you can see what properties there are etc. but nothing tells you what they mean).

AndreyAkinshin:
I really really want to have it. =) In fact, I want to have a way to build the API reference. I will not publish it online for now, but I just want to have a possibility to do it as soon as xml-docs will be ready.

PR #251, opened 2016-08-24
Better wording for error message

arthrp:
If you have a code like this:
```
class A 
{
    [Benchmark]
    public void Test()
    {
        var b = 42;
    }
}
```
you will get a message
> Benchmark method Test defined within type MyProject.A has incorrect access modifiers.

>Declaring type must be public.
IMO it's confusing as it implies that method, not type has incorrect access modifier.
I think
>MyProject.A containing Benchmark method Test has incorrect access modifiers.
is less confusing.


AndreyAkinshin:
@arthrp, thanks!

PR #253, opened 2016-08-27
Mark [Benchmark] as implying implicit use

roji:
Added Resharper code annotations to the project, and marked BenchmarkAttribute as implying implicit use. For users of BenchmarkDotNet which have R#, this will remove the "unused method" warning on benchmark methods.

Note that this PR includes the entire R# CodeAnnotations.cs, which makes sense if you intend to use the annotations in BenchmarkDotNet itself (to improve code quality). If not I can amend the PR, removing all annotations except the relevant [MeansImplicitUse].


AndreyAkinshin:
@roji, that's a great idea, thanks!

PR #267, opened 2016-09-28
Make shipped assemblies have CLSCompliant(true) applied

lahma:
By applying CLSCompliant(true) attribute to assemblies library consumers no longer need to mark their public code (methods, constructors) referencing BenchmarkDotNet assets as CLSCompliant(false).


AndreyAkinshin:
LGTM. @adamsitnik, what do you think?

adamsitnik:
LGTM

@AndreyAkinshin It looks that BenchmarkDotNet reached the moment when people build other software on top of it ;)

Issue #273, opened 2016-10-03
Create a tail call diagnoser


neoeinstein:
The implementation should mirror the `InlineDiagnoser` pretty closely, but use the facilities for hooking into JIT Tail Call optimization events. These events are described on MSDN and referenced below.
- [Description of Inlining and Tail Call ETW JIT events](https://blogs.msdn.microsoft.com/clrcodegeneration/2009/05/11/jit-etw-tracing-in-net-framework-4/)
- [Event DTOs](https://msdn.microsoft.com/en-us/library/ff356158%28v=vs.110%29.aspx)

This diagnoser would be extremely helpful for developers of F# code to verify that returns that are function calls are properly in tail position or that recursive functions are properly tailed as well.

GeorgePlotnikov:
Hi guys, is it case still actual? If yes, I'd like to grab it.

adamsitnik:
@GeorgePlotnikov Great! Please let me know if you need any help.

GeorgePlotnikov:
Hi, I've pulled #573

adamsitnik:
@neoeinstein do you know all scenarios which should be covered by the Tail Call Diagnoser?

When F# compiler can use goto? When tail call opt?

adamsitnik:
@GeorgePlotnikov I asked on twitter for some help and @tpetricek provided [this](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1011-warn-on-recursive-without-tail-call.md) link

GeorgePlotnikov:
@adamsitnik it's awesome, sincerely appreciate you

adamsitnik:
@GeorgePlotnikov big thanks for doing this! I checked your new diagnoser and it work very well!

Fixed by #573 and #598

GeorgePlotnikov:
@adamsitnik great, many thanks!

Issue #186, opened 2016-05-23
GC Diagnoser should not include allocations done by Setup method


Bobris:
See subj.

mattwarren:
Hmm, yeah that's a good point, I'll take a look (also see #133)

Bobris:
I read whole #133 before submitting this one, but because I don't have problem with unstable results and it didn't mentioned Setup method, I though should be good to separate these.

mattwarren:
> I read whole #133 before submitting this one, but because I don't have problem with unstable results and it didn't mentioned Setup method, I though should be good to separate these.

Yeah that's absolutely fine, I wasn't saying it was a dupe, it was more a note to myself to fix that #133 at the same time.

Thanks for reporting this BTW, I didn't even consider the effect of `[Setup]` along with the `MemoryDiagnoser`.

PR #277, opened 2016-10-05
Changed diagnosers flow, reduced heap allocations in Engine to 0

adamsitnik:
I wanted to implement a tail call diagnoser with my friends for #273 but we got stucked. We signed up for a jit event but we were never getting it, most probably because the method was already jitted before we even started the ETW session. But I am not sure, it's a guess.

The diagnosers flow was quite complicated and not very flexible. So I have changed it a little.

How it works now (Auto-generated program.cs):
1. The first thing to do is to notify master process that diagnoser can attach.
2. The JIT based diagnosers can attach at this moment, and we are sure that no jitting was done.
3. When Engine is created and prepared, we run Setup **once** and notify the diagnosers that this can happen. The GC diagnoser can attach at this moment and avoid the problem of including `[Setup]` allocations #186 .
4. Engine runs, but it does not call Setup and/or Cleanup anymore
5. We notify diagnosers that we are going to call Cleanup, so for example GC Diagnoser can detach at this moment, and don't include Cleanup allocations.
6. Call Cleanup

@AndreyAkinshin @mattwarren could  you guys take a look at the code? I created a PR to start the discussion. 

If you like this solution then I would like to continue the work: 

- reduce the number of heap allocations in Engine to 0 to give better accuracy for Memory Diagnoser.

- And improve allocated bytes calculation per operation to get better accuracy, for example 100 bytes for following benchmark:

```cs
[Benchmark]
public void New() => new byte[100];
```

@AndreyAkinshin what do you think about setup/cleanup changes I made?


AndreyAkinshin:
@adamsitnik, I like your changes, we are definitely on the right way. Will do a code review on the weekend.

adamsitnik:
>  I like your changes, we are definitely on the right way. Will do a code review on the weekend.

@AndreyAkinshin cool!

Now I am profiling the Engine to reduce the number of allocations. Currently when I run a benchmark with  empty method (does nothing, 0 allocations) with TargetCount = 5000 the results are : 7x GC 0,  2.302,08 Bytes Allocated/Op. Let's see how far I can get ;)

adamsitnik:
Ok, little update. I was able to get 0 GCs for empty method, the overhead went from 2000 bytes to 20 bytes per target invocation. My current problem is that Visual Studio Profiler can not tell me where the 20 bytes are ;) So I will have to try another tool.

@AndreyAkinshin Most of the changes so far were not ugly. The only ugly thing was to pre-allocate `List<Measurement>` that contains the results. 

The breaking change is to not display results during run, but display them all at the end of benchmark run. But I hope it's not a problem.

AndreyAkinshin:
> The breaking change is to not display results during run, but display them all at the end of benchmark run. But I hope it's not a problem.

Could we keep the old behavior? It's really nice to look at results one by one during benchmarking. Moreover, it allows to understand what's going on right now and check how long single iteration takes (so, it's possible to detect mistakes in the benchmark design on early stages).

P.S. Probably, we have some allocations because of the `Statistics` instances. In this case, we could implement incremental logic.

adamsitnik:
I have one idea: there is always a separate run when any diagnoser is attached, so I could print results immediately by default and in the diagnosers run delay the prining.

AndreyAkinshin:
LGTM.

adamsitnik:
Ok, I updated the code.

To do:

- find the missing 18-22 bytes for most simple scenario: no warmup, no target, fixed target count
- analyse and improve for scenarion with warmup and pilot

adamsitnik:
@AndreyAkinshin Ok, I updated the code once again. I am very close to getting it done ;) 

I also tried to set InvocationCount to some value, but then realized it was not passed to engine. Is [this commit](https://github.com/PerfDotNet/BenchmarkDotNet/pull/277/commits/2f0df729998788051cfc00475f479ab0d3f7b7ad) correct fix?

The results are getting more stable accurate, but 100KB limitation of ETW allocation event is still a problem.

Default Config + MemoryDiagnoser:
image

mattwarren:
> The results are getting more stable and accurate, but 100KB limitation of ETW allocation event is still a problem.

Yeah that's always going to be a fundamental problem with ETW events, maybe it's time to look at using [AppDomain.MonitoringTotalAllocatedMemorySize](https://msdn.microsoft.com/en-us/library/system.appdomain.monitoringtotalallocatedmemorysize(v=vs.110).aspx) instead?

adamsitnik:
> Yeah that's always going to be a fundamental problem with ETW events, maybe it's time to look at using AppDomain.MonitoringTotalAllocatedMemorySize instead?

I like this idea, I will do some PoC.

adamsitnik:
Ok, I am done. All possible heap allocations were eliminated from `Engine.Run()`

Memory profiling results of auto-genearated benchmark runner process for benchmark:

```cs
[Benchmark(Description = "new byte[10]")]
public byte[] TenBytes() => new byte[10];
```

image


Next steps:

1. @AndreyAkinshin could you do the code review?
2. When the code passes code review Itry to make the Memory Diagnostics a build in part of BDN. For classic toolchain I want to use `AppDomain.MonitoringTotalAllocatedMemorySize` as @mattwarren  suggested, for Core I want to use `GC.GetAllocatedBytesForCurrentThread` which got exposed [2 days ago](https://github.com/dotnet/corefx/pull/12489)

AndreyAkinshin:
@adamsitnik, great! Could you also fix failed test? I will do the full review in a few days.

src/BenchmarkDotNet.Core/Templates/BenchmarkProgram.txt
```diff
@@ -17,50 +19,87 @@ $AdditionalLogic$
 
 namespace BenchmarkDotNet.Autogenerated
 {
-    public class Program : global::$TargetTypeName$
+    public class Program
```
AndreyAkinshin: I like new approach!

src/BenchmarkDotNet.Diagnostics.Windows/Configs/InliningDiagnoserConfigAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using BenchmarkDotNet.Configs;
+
+namespace BenchmarkDotNet.Diagnostics.Windows.Configs
+{
+    public class InliningDiagnoserConfigAttribute : Attribute, IConfigSource
```
AndreyAkinshin: Rename to `InliningDiagnoser`

src/BenchmarkDotNet.Diagnostics.Windows/Configs/MemoryDiagnoserConfigAttribute.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using BenchmarkDotNet.Configs;
+
+namespace BenchmarkDotNet.Diagnostics.Windows.Configs
+{
+    public class MemoryDiagnoserConfigAttribute : Attribute, IConfigSource
```
AndreyAkinshin: Rename to `MemoryDiagnoser`

src/BenchmarkDotNet.Diagnostics.Windows/EtwDiagnoser.cs
```diff
@@ -1,18 +1,103 @@
...
+            WaitUntilStarted(task); 
+        }
+
+        protected void Stop()
+        {
+            // ETW real-time sessions receive events with a slight delay. Typically it
+            // shouldn't be more than a few seconds. This increases the likelihood that
+            // all relevant events are processed by the collection thread by the time we
+            // are done with the benchmark.
+            Thread.Sleep(TimeSpan.FromSeconds(3));
```
AndreyAkinshin: Could we wrap this line in a separated method with nice name (like `WaitFor...`)?

src/BenchmarkDotNet.Diagnostics.Windows/InliningDiagnoser.cs
```diff
@@ -1,128 +1,56 @@
...
-            var expected = benchmark.Target.Method.DeclaringType.Namespace  + "." +
+            var expected = benchmark.Target.Method.DeclaringType.Namespace + "." +
                            benchmark.Target.Method.DeclaringType.Name;
 
-            logger.WriteLine();
-            logger.WriteLineHeader(new string('-', 20));
-            logger.WriteLineInfo($"{benchmark.DisplayInfo}");
-            logger.WriteLineHeader(new string('-', 20));
+            Logger.WriteLine();
+            Logger.WriteLineHeader(new string('-', 20));
```
AndreyAkinshin: Could we move `new string('-', 20)` to a static readonly field (e.g. `LogSeparator`)?

tests/BenchmarkDotNet.IntegrationTests/MemoryDiagnoserTests.cs
```diff
@@ -12,44 +16,61 @@
...
-            list = Enumerable.Range(0, 100).ToList();
+            var bytes = stackalloc byte[100];
+            Blackhole(bytes);
         }
 
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void Blackhole<T>(T input) { }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private unsafe void Blackhole(byte* input) { }
```
AndreyAkinshin: Could we rename it to `Consume`?

tests/BenchmarkDotNet.IntegrationTests/MemoryDiagnoserTests.cs
```diff
@@ -63,47 +84,89 @@ public MemoryDiagnoserTests(ITestOutputHelper outputHelper)
...
+                var allocations = allocationColumn.GetValue(summary, benchmark);
+
+                AssertParsed(allocations, allocatedBytes => allocatedBytes == 0);
+            }
+        }
+
+        private IConfig CreateConfig(IDiagnoser diagnoser, int targetCount)
+        {
+            return ManualConfig.CreateEmpty()
+                               .With(Job.Dry.WithLaunchCount(1).WithWarmupCount(1).WithTargetCount(targetCount).With(GcMode.Default.WithForce(false)))
```
AndreyAkinshin: This line is too long. Could we introduce a variable for job here?
Also, you could use `WithGcForce`.

samples/BenchmarkDotNet.Samples/JIT/Jit_Inlining.cs
```diff
@@ -5,6 +5,9 @@ namespace BenchmarkDotNet.Samples.JIT
 {
     // See http://en.wikipedia.org/wiki/Inline_expansion
     // See http://aakinshin.net/en/blog/dotnet/inlining-and-starg/
+#if !CORE
+    [Diagnostics.Windows.Configs.InliningDiagnoserConfig]
```
adamsitnik: Yes and no ;)

In general these attributes have to have dependency to our Diagnosers, and Diagnosers are in the `BenchmarkDotNet.Diagnostics.Windows` project, which supports only NET 4.5+, no .NET Core support.

But we could workaround that by adding the class with same name, but without implementation to our main project. Then all classic projects that have dependency to our Diagnostics package would use the real one, and the .NET Core ones would get empty class.  By using the Obsolete attribute we can give them warning.

```cs
#if CORE
using System;

namespace BenchmarkDotNet.Diagnostics.Windows.Configs
{
    [Obsolete("InliningDiagnoser is not supported for .NET Core yet")]
    public class InliningDiagnoserAttribute : Attribute, IConfigSource
    {
        public IConfig Config { get; }

        public InliningDiagnoserAttribute() { }
    }
}
#endif
```


What do you think? Personally I am afraid that this could cause some problems in the future and we would gain only simplier sample.

adamsitnik:
@AndreyAkinshin Thanks! I updated the code.

samples/BenchmarkDotNet.Samples/JIT/Jit_Inlining.cs
AndreyAkinshin: Both options looks bad. =( Ok, let keep the current behavior without obsolete attributes.
By the way, is it possible to implement diagnosers for .NET Core?

src/BenchmarkDotNet.Core/Mathematics/MeasurementsStatistics.cs
```diff
@@ -0,0 +1,122 @@
...
+            else
+            {
+                Q1 = GetQuartile(measurements, measurements.Count / 2);
+                Median = GetQuartile(measurements, measurements.Count);
+                Q3 = GetQuartile(measurements, measurements.Count * 3 / 2);
+            }
+
+            var InterquartileRange = Q3 - Q1;
+            var LowerFence = Q1 - 1.5 * InterquartileRange;
+            var UpperFence = Q3 + 1.5 * InterquartileRange;
```
AndreyAkinshin: Could you use lower first letter (because here we are using local variables instead of properties)?

mattwarren:
@adamsitnik Great job, you've given the Diagnosers a long overdue and sorely needed fix/tidy-up thanks for doing that and the new `IDiagnoser` API looks really nice. 

Plus the work on reducing the BenchmarkDotNet memory allocations is really cool, I like this approach of [forcing allocations to happen up-front](https://github.com/PerfDotNet/BenchmarkDotNet/pull/277/commits/376bd86bf2a06b77b4ba4150cd8720d530ecbba0)

Just one small thing I noticed, we now have [quite a few warnings messages](https://ci.appveyor.com/project/PerfDotNet/benchmarkdotnet/build/0.9.9.178/messages) when you build the code (which don't seem to be [there on master](https://ci.appveyor.com/project/PerfDotNet/benchmarkdotnet/build/0.9.9.176/messages)), can some/all of these be removed or suppressed?

mattwarren:
Slight tongue-in-cheek but also slightly serious, should we ban `using System.Linq` from certain parts of BenchmarkDotNet? Like the Roslyn [code-guidelines do](https://github.com/dotnet/roslyn/wiki/Contributing-Code#coding-conventions)

> - Avoid allocations in compiler hot paths:
>   - **Avoid LINQ.**
>   - Avoid using foreach over collections that do not have a struct enumerator.
>   - Consider using an object pool. There are many usages of object pools in the compiler to see an example.

That always makes me laugh as the people who designed/wrote LINQ are banned from using it (admittedly only in certain parts of their code)

adamsitnik:
@mattwarren Thanks! It was a lot of fun to me to track even the smallest allocations. Btw. first call to list.Sort was allocating 1740 bytes!! That's insane

As for the LINQ I was wondering if we could somehow ban it on unit-test level. 

Pseudocode:   `Assert.False(typeof(Engine).References.Any(reference => reference.Name == "System.Linq"))`

Anyway I will most probably check for heap allocations in engine before each BDN release.

samples/BenchmarkDotNet.Samples/JIT/Jit_Inlining.cs
adamsitnik: > By the way, is it possible to implement diagnosers for .NET Core?

We are limited to what API Microsoft/other software vendors are exposing. If MS releases .NET Core ETW lib then there is no problem. I am planning to implement .NET Core Memory Diagnoser next week.

mattwarren:
> Anyway I will most probably check for heap allocations in engine before each BDN release.

Maybe we should [dog-food our own product](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) and use BenchmarkDotNet to automate this for us? I.e. a unit test that runs on each build and fails if we allocate too much or certain code paths?

adamsitnik:
> Maybe we should dog-food our own product and use BenchmarkDotNet to automate this for us? I.e. a unit test that runs on each build and fails if we allocate too much or certain code paths?

Good idea! I could cover that with integration tests

src/BenchmarkDotNet.Core/Mathematics/MeasurementsStatistics.cs
adamsitnik: yes, sure! I will do this today

adamsitnik:
@AndreyAkinshin  I have pushed the last commit with rename + I added and verified possibility to use custom Engine.

Next week I should be able to implement universal Memory Diagnoser, so if you could wait with 0.10.0 until then it would be great!

AndreyAkinshin:
@adamsitnik,
1. Could we merge this PR into master right now? Or should we wait for the universal Memory Diagnoser?
2. I'm going to release v0.10.0 at the end of October.

adamsitnik:
@AndreyAkinshin I will do separate PR with universal Memory Diagnoser within next week.

Issue #279, opened 2016-10-10
[Suggestion]: Simplify the configuration API


ig-sinicyn:
All of following is not a strong opinion, just fresh impressions on trying to use the current codebase.
1. How much settings do we need to alter N of iterations?
   Here they are:
   
   ```
   RunMode.Default.InvocationCount / IterationTime
   AccuracyMode.Default.MinInvokeCount / AccuracyMode.Default.MinIterationTime
   RunMode.Default.LaunchCount
   RunMode.Default.TargetCount
   RunMode.Default.WarmupCount
   RunMode.Default.UnrollFactor
   OperationsPerInvoke attribute
   ```
   
   can these be simplified down to
   
   ```
   RunMode.Default.TargetCount / TargetTime
   RunMode.Default.WarmupCount / WarmupTime
   RunMode.Default.UnrollFactor => OperationsPerInvoke
   ```
   
   :scrathes_head:?
2. Please, consider to use human-oriented API. The creation of a new job should not look like this:
   
   ```
             var job = Job.Default
                 .WithLaunchCount(1)
                 .WithWarmupCount(100)
                 .WithTargetCount(300)
                 .WithUnrollFactor(1)
                 .Mutate(
                     new JobMutator("have no idea what")
                         .Add(RunMode.Default.RunStrategy.Mutate(RunStrategy.Throughput))
                         .Add(RunMode.Default.InvocationCount.Mutate(1))
                         .Add(GcMode.Default.Force.Mutate(false))
                         .Add(InfrastructureMode.Default.Toolchain.Mutate(InProcessToolchain.Instance)))
   ```
   
   Can freezable objects be used instead? [Fast prototype](https://gist.github.com/ig-sinicyn/b4cc36175a002f6c08f2eeb9f35faea2) to show what I'm talking about.

AndreyAkinshin:
1. We can't do this simplification, all the listed characteristics are very important
a) When you are working with `RunMode`, you specify exact values. For example, if you set `RunMode.IterationTime` = `100ms`, you say: "Single iteration should take 100ms". When you set `AccuracyMode.MinIterationTime` = `100ms`, you say: "Magic algorithm should choose ideal iteration time automatically, however it should be at least 100ms".
[...]

PR #286, opened 2016-10-19
Mutable Job implementation

ig-sinicyn:
As proposed in #279.

Before:
```cs
    var job = Job.Dry
       .WithLaunchCount(1)
       .WithWarmupCount(1)
       .WithTargetCount(targetCount)
       .WithGcForce(false));       
```
After:
```cs
    var job = new Job(Job.Dry)
    {
        Run = { LaunchCount = 1, WarmupCount = 1, TargetCount = targetCount },
        Env = { Gc = { Force = false } }
    }.Freeze();
```

The property system is inspired by WPF dependency properties + freezable objects approach. 
However, there are restrictions made to maintain compatibility with the current codebase: 
* Each property (`Characteristic`) can be used only once per object instance (no way to assign different values for `Job.Id` and `Job.EnvMode.Id`).
* The value of the properties with type derived from JobMode (e.g. `job.EnvMode`) cannot be null.

The expected use-case pattern is
- create
- set properties (via obj initializer syntax or using the `.Apply()` method)
- freeze and reuse.

Frozen objects are immutable and therefore can be shared across threads or stored in static fields.
Jobs are frozen automatically when added into the config, so the `did we forget to freeze the job?' check is unnecessary.

**P.S.** Tests [are green](https://ci.appveyor.com/project/ig-sinicyn/benchmarkdotnet), suggestions are welcome:)


adamsitnik:
@ig-sinicyn  I really appreciate the amount of hard work you have done to implement it but personally I don't like the new syntax. Some of my arguments:

1. Some of our users are used to fluent api, now it's gone and the backward compatibility is broken. If they want to upgrade to the new version for their existing benchmarks they have to learn new syntax and change existing code.
2. With the old, "fluent" api it was an implementation detail covered by us how properties of the Job were named.  For example `Job.With(Runtime.Clr)` was leaving it to us how we save this state, with `new Job() { Env = { Runtime = Runtime.Core } }` we are not allowed to rename `Env` or any other public properties if we want to keep backward compatibility for future versions. 
3. The goal was to simplify the code, but the size of it has doubled (731 lines removed, 1618 added)

@AndreyAkinshin I think that we should keep the backward compatibility with the fluent api and only extend it by new, optional "object initializer" syntax. We should follow [OCP](https://en.wikipedia.org/wiki/Open/closed_principle)

ig-sinicyn:
@adamsitnik , see your point.

At first, there's a Job .ctor that allows you to pass multiple job mutators, e.g.
```cs
var job = new Job("MyJob", EnvMode.Clr, RunMode.Dry, GcMode.Server);
```
and it's very easy to create your own ones.
The obj initializer syntax is fallback for case there's no predefined mutator or you do want to set all properties explicitly.

However, I do agree that breaking changes are bad and will do another PR with WithXyz() fluent extensions coming back. Or we can just revert the PR,  either way will do : )

P.S. 
> The goal was to simplify the code, but the size of it has doubled (731 lines removed, 1618 added)

Yep, looks like a fail ; ) 
OK, seriously, the goal was to simplify the public API, not the implementation. Obviously there's a room for making it better, so any ideas / suggestions / fixes are welcome!

AndreyAkinshin:
Support of the `WithXXX()` methods in `v0.10.0` looks like a good idea. @ig-sinicyn, I'm waiting for another PR.
In general, I will care a lot about backward compatibility since `v1.0.0`. Right now it's ok to do breaking changes, *if it improves the API*. I like new approach to specify jobs and I don't know which approach (old or new) is better. Let's keep them both and try to use it.
What about the internal implementation, this PR is a step forward. A lot of ugly methods were rewritten, the architecture has become more flexible.

PR #289, opened 2016-10-21
Jobs: .With() extension methods.

ig-sinicyn:
As [requested](https://github.com/PerfDotNet/BenchmarkDotNet/pull/286#issuecomment-255080585) by @adamsitnik 

Changelist:
- Jobs: WithXxx() extension methods added back, docs updated.
- Jobs, minor bug fixed: Id was cleared on explicit JobMode assignment + test for it (`Test03IdDoesNotFlow`).
- Jobs.Accuracy, typo: AnaylyzeLaunchVariance => AnalyzeLaunchVariance, docs updated.

### DIFF
--- a/src/BenchmarkDotNet.Core/Jobs/JobExtensions.cs
+++ b/src/BenchmarkDotNet.Core/Jobs/JobExtensions.cs
@@ -9,171 +9,48 @@ namespace BenchmarkDotNet.Jobs
 {
     public static class JobExtensions
     {
+        private static Job WithCore(this Job job, Action<Job> updateCallback)
+        {
+            job = new Job(job);
+            updateCallback(job);
+            return job;
+        }
+        public static Job With(this Job job, Platform platform) => job.WithCore(j => j.Env.Platform = platform);
+
         public static Job WithId(this Job job, string id) => new Job(id, job);
 
         // Env
-        public static Job With(this Job job, Platform platform)
-        {
-            job = new Job(job);
-            job.Env.Platform = platform;
-            return job;
-        }
-        public static Job With(this Job job, Jit jit)
-        {
-            job = new Job(job);
-            job.Env.Jit = jit;
-            return job;
-        }
-        public static Job With(this Job job, Runtime runtime)
-        {
-            job = new Job(job);
-            job.Env.Runtime = runtime;
-            return job;
-        }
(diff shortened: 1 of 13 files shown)


AndreyAkinshin:
In the original design, the `WithXXX()` methods don't change the original job, it always creates a copy of the original job with requested change. Could we keep this approach?

ig-sinicyn:
@AndreyAkinshin  yep, I've think about that, but it feels very counterintuitive.

There're mutable properties, there's .Apply() method that updates the instance of the job and there're .With() methods that creates a new copy of the Job on each call. Kinda Js/PHP style : )

I'd prefer to keep standard Create-Modify-Freeze-Reuse approach until there're strong reasons against it.

AndreyAkinshin:
Ok, let's consider the following situation: I want to have a dry job for Mono. Before the refactoring, I could take the predefined `Dry` job and create a DryMono job in the following way: `Job.Dry.With(Runtime.Mono)`. It was the main use-case of the `WithXXX()` methods.

ig-sinicyn:
@AndreyAkinshin 

Ok, I've fixed the `.With()` method logic. Should not break compatibility now.

ig-sinicyn:
Oops, last minute fix had broken smth.

src/BenchmarkDotNet.Core/Jobs/JobExtensions.cs
```diff
@@ -0,0 +1,184 @@
+﻿using System;
+using BenchmarkDotNet.Toolchains;
+using BenchmarkDotNet.Characteristics;
+using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Environments;
+using BenchmarkDotNet.Horology;
+
+namespace BenchmarkDotNet.Jobs
+{
+    public static class JobExtensions
```
ig-sinicyn: @AndreyAkinshin 
yep, but each call will incur additional allocation due to closure on WithXxx() arg.
If that's ok - will do.

AndreyAkinshin: It's ok. When we are working with Jobs, we usually don't care about allocations. =)

ig-sinicyn: Ok, fixed 'n' pushed

PR #314, opened 2016-12-03
Improved information about job environments in summary

AndreyAkinshin:
Many people share benchmark results via simple screenshots which include information about the environment. So, it would be cool to keep the environment info as small as possible but still print all the necessary information. I made some changes; there is the result:

Before (12 lines):
```ini
BenchmarkDotNet=v0.10.0-develop
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-4702MQ CPU 2.20GHz, ProcessorCount=8
Frequency=2143476 Hz, Resolution=466.5319 ns, Timer=TSC
Host Runtime=Clr 4.0.30319.42000, Arch=64-bit RELEASE [RyuJIT]
GC=Concurrent Workstation
JitModules=clrjit-v4.6.1586.0
Job Runtime(s):
        .NET Core 4.6.24628.01, Arch=64-bit  [RyuJIT]
        Clr 4.0.30319.42000, Arch=64-bit RELEASE
        Mono 4.6.2 (Visual Studio built mono), Arch=64-bit RELEASE
        Clr 4.0.30319.42000, Arch=64-bit RELEASE [RyuJIT]
```

After (9 lines):
```ini
BenchmarkDotNet=v0.10.0-develop
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-4702MQ CPU 2.20GHz, ProcessorCount=8
Frequency=2143476 Hz, Resolution=466.5319 ns, Timer=TSC
  [Host]       : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1586.0
  Core         : .NET Core 4.6.24628.01, 64bit RyuJIT
  LegacyJitX64 : Clr 4.0.30319.42000, 64bit LegacyJIT-v4.6.1586.0
  Mono         : Mono 4.6.2 (Visual Studio built mono), 64bit
  RyuJitX64    : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1586.0
```
Changes:
* JitModules now is a part of the runtime info. Also, we have smart logic which combines knowledge about JitModules and Jit kind into a short form
* We don't print the `Job Runtime(s):` line anymore. I think, our users will understand what is it without additional line with a hint.
* The list of runtimes includes the host runtime (always on the first place)
* Each runtime has a prefix with the JobId (sometimes it's hard to associate runtimes with jobs without an explicit hint).
* An indent of the runtime lines is two spaces instead of a tab (it allows to save two more symbols in these lines
* Now we print `32bit` and `64bit` instead of `Arch=32-bit` and `Arch=64-bit`. The reason is the same: we want to throw away all the unimportant information and print a small amount of characters which fit into a terminal with a small width.
* Now we also don't print information about GC in the summary because I don't think that it's the first importance information. I also added information about GC (and about runtime) for *each* job in the details section (which is before the summary).
* We don't print `RELEASE` anymore, only the `DEBUG` label will be printed.

### DIFF
--- a/src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
@@ -1,9 +1,11 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Text.RegularExpressions;
 using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Extensions;
@@ -21,15 +23,15 @@ namespace BenchmarkDotNet.Portability
     {
         private static readonly bool isMono = Type.GetType("Mono.Runtime") != null; // it allocates a lot of memory, we need to check it once in order to keep Enging non-allocating!
 
-        private const string Debug = "DEBUG";
-        private const string Release = "RELEASE";
+        private const string DebugConfigurationName = "DEBUG";
+        internal const string ReleaseConfigurationName = "RELEASE";
         internal const string Unknown = "?";
 
         internal static string ExecutableExtension => IsWindows() ? ".exe" : string.Empty;
 
         internal static string ScriptFileExtension => IsWindows() ? ".bat" : ".sh";
 
-        internal static string GetArchitecture() => IntPtr.Size == 4 ? "32-bit" : "64-bit";
+        internal static string GetArchitecture() => IntPtr.Size == 4 ? "32bit" : "64bit";
 
         internal static bool IsWindows()
         {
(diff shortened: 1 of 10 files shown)


AndreyAkinshin:
@mattwarren, @adamsitnik, what do you think?

adamsitnik:
@AndreyAkinshin I like the new form!

Maybe we could go one step further and display the OS info together at the same line with BDN version?

```ini
BenchmarkDotNet=v0.10.0-develop, OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-4702MQ CPU 2.20GHz, ProcessorCount=8
Frequency=2143476 Hz, Resolution=466.5319 ns, Timer=TSC
  [Host]       : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1586.0
  Core         : .NET Core 4.6.24628.01, 64bit RyuJIT
  LegacyJitX64 : Clr 4.0.30319.42000, 64bit LegacyJIT-v4.6.1586.0
  Mono         : Mono 4.6.2 (Visual Studio built mono), 64bit
  RyuJitX64    : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1586.0
```

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -141,6 +152,33 @@ internal static Jit GetCurrentJit()
             return HasRyuJit() ? Jit.RyuJit : Jit.LegacyJit;
         }
 
+        internal static string GetJitInfo()
+        {
+            if (IsMono())
+                return ""; // There is no helpful information about JIT on Mono
+#if CORE
+            return "RyuJIT"; // CoreCLR supports only RyuJIT
```
adamsitnik: "CoreCLR supports only RyuJIT" - currently we allow our users to run only x64 benchmark for Core, so this is true today. 

However if we enable  32bit support for .NET Core 1.1 it won't be true, because right now .NET Core is using Legacy Jit for 32bit. And 32bit .NET Core has support for Windows now only.

.NET Core 1.2 will move from leagacy Jitr for 32bits to RyuJIT which will be used by default. Most probably then also other OSes will get 32bit support.

The goal of my comment is just to share the knowledge so we don't forget it about in the future

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -212,5 +250,17 @@ public bool IsMsX64(int step = 1)
                 return value == 20 + step;
             }
         }
+
+        public class JitModule
```
adamsitnik: when we move to C# 7.0 we will be able to use ValueTuples for such cases ;)

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
AndreyAkinshin: Thanks for the info!

AndreyAkinshin:
> Maybe we could go one step further and display the OS info together at the same line with BDN version?

Good idea!

mattwarren:
This is nice idea, LGTM

PR #331, opened 2016-12-19
[NO MERGE] Uap

zabulus:
PR is just for review and sharing considerations about further implementation


global.json
```diff
@@ -1,5 +1,10 @@
 {
   "projects": [
-    "src", "samples", "tests"
-  ]
+    "src",
+    "samples",
+    "tests"
+  ],
+  "sdk": {
+    "version": "1.0.0-preview2-1-003177"
```
adamsitnik: do you specify the sdk version explicity in order to get it running on VS 2017? If so, is the VS 2017 mandatory to compile the solution now? If yes, it's not a problem for me, I just want to know

samples/BenchmarkDotNet.Samples.Runner/project.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+  "version": "1.0.0-*",
+  "buildOptions": {
+    "emitEntryPoint": true
+  },
+
+  "dependencies": {
+    "BenchmarkDotNet": "0.10.1-develop",
```
adamsitnik: you can make it a reference to the project, then there is no need to specify the version (which would require change with every update)

```json
"BenchmarkDotNet": {
          "target": "project"
        }
```

samples/BenchmarkDotNet.Samples/Program.cs
```diff
@@ -1,9 +1,22 @@
 ﻿using System.Reflection;
 using BenchmarkDotNet.Running;
 
+#if UAP
+namespace System.Threading
+{
+    internal static class Thread
+    {
+        public static void Sleep(int milliseconds)
```
adamsitnik: I am not sure about the purpose of this class, but in general Thread.Sleep != Task.Delay so the name is wrong.

samples/BenchmarkDotNet.Samples/project.json
```diff
@@ -33,22 +33,35 @@
...
+        "System.Threading.Tasks": "4.3.0"
       }
     }
   },
   "dependencies": {
     "BenchmarkDotNet": {
       "target": "project"
     }
+  },
+  "runtimes": {
```
adamsitnik: we can not support only hardcoded list of runtimes, but you probably know that

Is it possible to set the runtimes per target framework?

src/BenchmarkDotNet.Core/Attributes/Exporters/RPlotExporterAttribute.cs
```diff
@@ -2,10 +2,12 @@
 
 namespace BenchmarkDotNet.Attributes.Exporters
 {
+#if !UAP
```
adamsitnik: please do that per whole file if possible (#if in the first line of the file)

src/BenchmarkDotNet.Core/Characteristics/CharacteristicHelper.cs
```diff
@@ -56,12 +56,18 @@ public static IReadOnlyList<Characteristic> GetThisTypeCharacteristics(Type jobM
 
         private static IReadOnlyList<Characteristic> GetThisTypeCharacteristicsCore(Type jobModeType)
         {
-            var fieldValues = jobModeType.GetTypeInfo()
+            var fieldValues = jobModeType
+#if !UAP
+                .GetTypeInfo()
```
adamsitnik: I think that you could create an extension method for Type called GetTypeInfo() and get rid of all the ifdefines around this call

sth like:
```cs
#if UAP
internal static Type GetTypeInfo(this Type type) => type;
#endiff
```

src/BenchmarkDotNet.Core/Code/CodeGenerator.cs
```diff
@@ -19,9 +19,14 @@ internal static class CodeGenerator
     {
         internal static string Generate(Benchmark benchmark)
         {
+            return Generate(benchmark, "BenchmarkProgram.txt");
+        }
+
+        internal static string Generate(Benchmark benchmark, string templateName)
```
adamsitnik: can this method be private now?

src/BenchmarkDotNet.Core/Diagnosers/CompositeDiagnoser.cs
```diff
@@ -19,13 +19,13 @@ public CompositeDiagnoser(params IDiagnoser[] diagnosers)
 
         public IColumnProvider GetColumnProvider() 
             => new CompositeColumnProvider(diagnosers.Select(d => d.GetColumnProvider()).ToArray());
-
+#if !UAP
         public void BeforeAnythingElse(Process process, Benchmark benchmark)
```
adamsitnik: does UAP has no definition of `Process` class?

src/BenchmarkDotNet.Core/Engines/RunResults.cs
```diff
@@ -62,16 +62,18 @@ public IEnumerable<Measurement> GetMeasurements()
...
+        public void Print(Action<string> printAction)
+        {
             foreach (var measurement in GetMeasurements())
-                WriteLine(measurement.ToOutputLine());
+                printAction(measurement.ToOutputLine());
 
-            WriteLine(GCStats.WithTotalOperations(totalOperationsCount).ToOutputLine());
-            WriteLine();
+            printAction(GCStats.WithTotalOperations(totalOperationsCount).ToOutputLine());
+            printAction("\r\n");
```
adamsitnik: can we use `Environment.NewLine` here?

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -101,6 +101,8 @@ internal static string GetRuntimeVersion()
             }
 #if CLASSIC
             return $"Clr {System.Environment.Version}";
+#elif UAP
+            return "Uap 10.0";
```
adamsitnik: is it possible to read the current version from somewhere?

src/BenchmarkDotNet.Core/Templates/UapBenchmarkProgram.txt
```diff
@@ -0,0 +1,188 @@
...
+                    finally
+                    {
+						lc.Dispose();
+                        Application.Current.Exit();
+                    }
+                };
+
+                Task.Factory.StartNew(action);
+            }
+            catch (Exception ex)
```
adamsitnik: you started the task asynchronously, without awaiting it the catch will not work

btw why is the task needed? why not to run in in synchronous way?

src/BenchmarkDotNet.Core/Templates/UapBenchmarkProject.json
```diff
@@ -0,0 +1,18 @@
+﻿{
+  "dependencies": {
+    "Microsoft.NETCore.UniversalWindowsPlatform": "5.2.2",
+    "system.console": "4.3.0",
```
adamsitnik: is dependency to system.console needed?

src/BenchmarkDotNet.Core/Toolchains/ConsoleHandler.cs
```diff
@@ -50,7 +51,11 @@ private void HandlerCallback(object sender, ConsoleCancelEventArgs e)
                     return;
 
                 // Give it a bit of time to exit!
+#if UAP
+                System.Threading.Tasks.Task.Delay(500);
```
adamsitnik: you are missing .Wait()

src/BenchmarkDotNet.Core/Toolchains/DotNetCli/DotNetCliGenerator.cs
```diff
@@ -91,7 +96,11 @@ protected override void Cleanup(ArtifactsPaths artifactsPaths)
                 }
                 catch (Exception) when (attempt++ < 5)
                 {
+#if UAP
```
adamsitnik: does the `#if !UAP` in the top of the file excludes a chance for this `#if UAP`?

src/BenchmarkDotNet.Core/Toolchains/Uap/DevicePortalApiWrapper.cs
```diff
@@ -0,0 +1,323 @@
...
+        public PackageStruct DeployApplication(string fullPath)
+        {
+            var installRequest = new RestRequest("api/app/packagemanager/package", Method.POST);
+            installRequest.AddCookie(csrfToken.Name, csrfToken.Value);
+            installRequest.AddCookie(wmid.Name, wmid.Value);
+            installRequest.AddFile(Path.GetFileName(fullPath), fullPath);
+            installRequest.AddQueryParameter("package", Path.GetFileName(fullPath));
+            installRequest.AddHeader("X-CSRF-Token", csrfToken.Value);
+            int i = 10;
+            for (;;)
```
adamsitnik: I believe that `while(true)` would be easier to understand, or maybe we can simplify it further?

```cs
while (client.Execute(installRequest).StatusCode != HttpStatusCode.Accepted)
{
    if (--i == 0)
    {
         throw new InvalidOperationException("Install failed");
    }
}
```

src/BenchmarkDotNet.Core/Toolchains/Uap/UapToolchain.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            var currentRuntime = "win10-x64";
+            if (!string.IsNullOrEmpty(currentRuntime))
+            {
+                return $"\"runtimes\": {{ \"{currentRuntime}\": {{ }} }},";
+            }
+
+            return string.Empty;
+        }
+    }
```
adamsitnik: the private methods seems to not be used anywhere, can they be removed?

src/BenchmarkDotNet.Core/project.json
```diff
@@ -48,7 +48,8 @@
         "System.Xml": "4.0.0.0"
       },
       "dependencies": {
-        "System.Threading.Tasks.Extensions": "4.0.0"
+        "System.Threading.Tasks.Extensions": "4.0.0",
+        "RestSharp": "105.2.3"
```
adamsitnik: "105.2.3" => this is a crazy version number ;)

global.json
zabulus: I have some mess with cli tooling on my local PC. preview3 that I have installed, catched if no version specified, and it doesn't work.
I planned to revert this.

src/BenchmarkDotNet.Core/Characteristics/CharacteristicHelper.cs
zabulus: Nice, will fix this

global.json
adamsitnik: @zabulus it's ok, I also needed this when installed VS 2017 on the same box.

src/BenchmarkDotNet.Core/Diagnosers/CompositeDiagnoser.cs
zabulus: There is System.Diagnostics.Process nuget package. But when I add ```runtimes``` section to a project where the package is added, tooling starts complaining about it.

src/BenchmarkDotNet.Core/Templates/UapBenchmarkProgram.txt
zabulus: It is UI application but without UI. 
I tried to run benchmark in constructor of Program, but no results appears. I assume that there is some timeout or restriction of operation that can be called in ctor.
Problem is that I can't reproduce that during device debug, only while benchmark running. So the fix is quite blind guess.

src/BenchmarkDotNet.Core/Templates/UapBenchmarkProject.json
zabulus: Will delete

src/BenchmarkDotNet.Core/Toolchains/DotNetCli/DotNetCliGenerator.cs
zabulus: I think whole Toolchains namespace should be excluded for UAP. There will be no support for running BDN as host.

src/BenchmarkDotNet.Core/Toolchains/Uap/DevicePortalApiWrapper.cs
zabulus: Yep, that's much nicer

src/BenchmarkDotNet.Core/Toolchains/ConsoleHandler.cs
zabulus: Whole file should excluded by UAP

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
zabulus: Not sure what side is this code running at.

src/BenchmarkDotNet.Core/Templates/UapBenchmarkProgram.txt
adamsitnik: @zabulus this requires further investigation, we need to be sure that it does not affect results

zabulus: Problem I've mentioned I've fixed with additional page (MainPage.xaml). The benchmark logic runs as background task, and application doesn't crash in release config.

What investigation should I perform to be sure that nothing affects a results?

adamsitnik: @zabulus I am not 100% sure, but my guess is to check the difference between main app thread and the thread that task scheduler schedules for the task

sample problem could be: background thread vs foreground thread, different priorities 

@AndreyAkinshin @mattwarren @goldshtn what is your opinion on that?

PR #339, opened 2017-01-05
Better detection of Rscript in RPlotExporter

### LINKED ISSUE #337: Problematic mechanism/docs for locating Rscript.exe
The [documentation](http://benchmarkdotnet.org/Configs/Exporters.htm#plots) instructs users to define %R_HOME%. However, the RPlotExporter expects to find Rscript.exe directly in that directory, where it seems more standard for a variable called R_HOME to point to the installation directory (as opposed to something like R_BIN). In addition, this triggers a Win32Exception with the error "The system cannot find the file specified", which isn't very explicit.

I'll be happy to submit a PR for this. IMHO RPlotExporter should be changed to look for RScript.exe under a bin directory in %R_HOME%, but that would be a somewhat breaking change.

roji:
Fixes #337.

RPlotExporter now logs an error if Rscript.exe isn't found and aborts. It may not be visible enough (is throwing an exception better?)

### DIFF
--- a/docs/guide/Configs/Exporters.md
+++ b/docs/guide/Configs/Exporters.md
@@ -207,8 +207,7 @@ Example of `IntroJsonExport-report-brief.json`:
 
 ## Plots
 
-If you have installed [R](https://www.r-project.org/), defined `%R_HOME%` variable and used `RPlotExporter.Default` and `CsvMeasurementsExporter.Default` 
-in your config, you will also get nice plots with help of the `BuildPlots.R` script in your bin directory. 
+You can install [R](https://www.r-project.org/) to automatically get nice plots of your benchmark results. First, make sure `Rscript.exe` or `Rscript` is in your path, or define an R_HOME environment variable pointing to the R installation directory (containing the `bin` directory). Use `RPlotExporter.Default` and `CsvMeasurementsExporter.Default` in your config, and the `BuildPlots.R` script in your bin directory will take care of the rest.
 
 ### Examples
 
--- a/src/BenchmarkDotNet.Core/Exporters/RPlotExporter.cs
+++ b/src/BenchmarkDotNet.Core/Exporters/RPlotExporter.cs
@@ -2,9 +2,11 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using BenchmarkDotNet.Exporters.Csv;
 using BenchmarkDotNet.Helpers;
 using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Portability;
 using BenchmarkDotNet.Properties;
 using BenchmarkDotNet.Reports;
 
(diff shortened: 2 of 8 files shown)


src/BenchmarkDotNet.Core/Exporters/RPlotExporter.cs
```diff
@@ -36,28 +36,33 @@ public IEnumerable<string> ExportToFiles(Summary summary)
...
+                consoleLogger.WriteLineError("RPlotExporter requires the R_HOME environment variable to be defined!");
+                yield break;
             }
-            else
+
+            var rscriptPath = Path.Combine(rHome, "Rscript.exe");
+            if (!File.Exists(rscriptPath))
             {
-                // TODO: print warning, if the folder is not found
+                consoleLogger.WriteLineError("RPlotExporter requires R_HOME to point to the directory containing Rscript.exe");
```
roji: Amended to do this

docs/guide/Configs/Exporters.md
```diff
@@ -207,8 +207,7 @@ Example of `IntroJsonExport-report-brief.json`:
 
 ## Plots
 
-If you have installed [R](https://www.r-project.org/), defined `%R_HOME%` variable and used `RPlotExporter.Default` and `CsvMeasurementsExporter.Default` 
-in your config, you will also get nice plots with help of the `BuildPlots.R` script in your bin directory. 
+If you have installed [R](https://www.r-project.org/), defined a `%R_HOME%` variable pointing to the directory containing `Rscript.exe`, and used `RPlotExporter.Default` and `CsvMeasurementsExporter.Default` in your config, you will also get nice plots with help of the `BuildPlots.R` script in your bin directory.
```
AndreyAkinshin: Could you also add examples of the typical `R_HOME` values for Windows, MacOS, and Linux?

AndreyAkinshin:
> It may not be visible enough (is throwing an exception better?)

`WriteLineError` is ok for now. In the future, I will add an ability to produce warnings from exporters via a nice API.

docs/guide/Configs/Exporters.md
AndreyAkinshin: We shouldn't cover all the distributions, but it will be nice to have several examples.

roji:
OK, I redid it... We know look for both Rscript.exe and Rscript (for non-Windows), and if R_HOME isn't defined we look in the PATH.

It seems that rather than an R_HOME pointing to a directory, it may be simpler to just have RSCRIPT_PATH pointing at the executable (yeah, I know I said this before).

roji:
>> it may be simpler to just have RSCRIPT_PATH pointing at the executable
> @roji, could you also add this feature?

Sure, but do you want me to also remove (R_HOME)? It doesn't make much sense to have both options, except for backwards compatibility.

AndreyAkinshin:
Hmm, I changed my mind: let's keep the `R_HOME` variable because it's a common way to specify the installation directory. It could makes sense even on Linux (in case if a user has several installation of R). So, I suggest to do the following:
* Step 1: If we have the `R_HOME` variable, we are looking for Rscript (`Rscript.exe` for WIndows, `Rscript` for other operation systems; we can get the correct name of executable in advance) in `R_HOME/bin`
* Step 2: If we failed to find Rscript in `R_HOME` (there is no such environment variable or there is no Rscript in `R_HOME/bin`), we are trying to find it in PATH.

src/BenchmarkDotNet.Core/Exporters/RPlotExporter.cs
```diff
@@ -36,33 +37,54 @@ public IEnumerable<string> ExportToFiles(Summary summary)
...
-                    FileName = Path.Combine(rHome, "Rscript.exe"),
-                    WorkingDirectory = summary.ResultsDirectoryPath,
-                    Arguments = $"\"{scriptFullPath}\" \"{fileNamePrefix}-measurements.csv\""
-                };
-                using (var process = Process.Start(start))
-                    process?.WaitForExit();
-                yield return fileNamePrefix + "-boxplot.png";
-                yield return fileNamePrefix + "-barplot.png";
+                    rscriptPath = Path.Combine(rHome, "Rscript.exe");
+                    if (!File.Exists(rscriptPath))
```
AndreyAkinshin: I believe, you did a copy-paste bug here: you check existence of `Path.Combine(rHome, "Rscript.exe")` twice and you don't use the bin directory.

src/BenchmarkDotNet.Core/Exporters/RPlotExporter.cs
```diff
@@ -36,33 +37,54 @@ public IEnumerable<string> ExportToFiles(Summary summary)
...
+                        consoleLogger.WriteLineError($"RPlotExporter requires R_HOME to point to the directory containing Rscript.exe (currently points to {rHome})");
+                        yield break;
+                    }
+                }
             }
-            else
+            else // No R_HOME, try the path
             {
-                // TODO: print warning, if the folder is not found
+                rscriptPath = FindInPath("Rscript.exe") ?? FindInPath("Rscript");
```
AndreyAkinshin: You can get the correct name of executable in advance, use `RuntimeInformation.IsWindows()`

src/BenchmarkDotNet.Core/Exporters/RPlotExporter.cs
```diff
@@ -36,33 +37,54 @@ public IEnumerable<string> ExportToFiles(Summary summary)
...
+                    rscriptPath = Path.Combine(rHome, "Rscript.exe");
+                    if (!File.Exists(rscriptPath))
+                    {
+                        consoleLogger.WriteLineError($"RPlotExporter requires R_HOME to point to the directory containing Rscript.exe (currently points to {rHome})");
+                        yield break;
+                    }
+                }
             }
-            else
+            else // No R_HOME, try the path
```
AndreyAkinshin: You could wrap each branch of the `if` in a separated method with a nice name (something like `TryFindRScriptInRHome` and `TryFindRScriptInPath`), it will make code simpler.

roji:
Have done the changes.

Have left the code in the main function (rather than creating `TryFindRScriptInRHome` and `TryFindRScriptInPath`) since we don't want to check the path if R_HOME exists but is wrong - in that case we want to issue the warning and abort (handling that with two functions seems needlessly convoluted).

Issue #320, opened 2016-12-06
Results table should be a GitHub Flavored Markdown table


svick:
Currently, the results table printed to the console looks something like this:

```
    Method |      Mean |    StdDev |    Median |
---------- |---------- |---------- |---------- |
 OldSchool | 2.6605 ms | 0.0601 ms | 2.6675 ms |
    Modern | 2.5060 ms | 0.1050 ms | 2.5105 ms |
```

This is *almost*, but not quite, the format used for tables in [GitHub Flavored Markdown](https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown) and if copy-pasted directly into GitHub renders broken:

    Method |      Mean |    StdDev |    Median |
---------- |---------- |---------- |---------- |
 OldSchool | 2.6605 ms | 0.0601 ms | 2.6675 ms |
    Modern | 2.5060 ms | 0.1050 ms | 2.5105 ms |

To make it a GFM table, prepending a pipe to each line seems to work:

```
|    Method |      Mean |    StdDev |    Median |
|---------- |---------- |---------- |---------- |
| OldSchool | 2.6605 ms | 0.0601 ms | 2.6675 ms |
|    Modern | 2.5060 ms | 0.1050 ms | 2.5105 ms |
```

Which renders as:

|    Method |      Mean |    StdDev |    Median |
|---------- |---------- |---------- |---------- |
| OldSchool | 2.6605 ms | 0.0601 ms | 2.6675 ms |
|    Modern | 2.5060 ms | 0.1050 ms | 2.5105 ms |

I think the format of the table should be changed to this, so that it can be easily copy-pasted into GitHub (and any other site that supports the same table format).

adamsitnik:
@AndreyAkinshin @alinasmirnova sorry for stealing the issue, but it was annoying me since quite a long time ;)

mattwarren:
Has this issue regressed? I'm seeing quite a few people posting tables that don't have the correct format?

For example see this comment, that is using `BenchmarkDotNet=v0.10.14` https://github.com/dotnet/coreclr/issues/17932#issuecomment-388008331 (anther example is here https://github.com/dotnet/coreclr/pull/17928#issuecomment-387810897)

The first link has tables like this (raw view):

```
  Method |     Mean |     Error |    StdDev |  Gen 0 | Allocated |
-------- |---------:|----------:|----------:|-------:|----------:|
[...]

PR #348, opened 2017-01-15
Approval tests for exporters

alinasmirnova:


AndreyAkinshin:
@alinasmirnova, great tests, thanks!

I have only one concern: we should mock also the environment info, the summary should include CPU invariant data (BDN version, OS, Processor, and so on).

mattwarren:
BTW I *think* we have a problem with the GitHub table markdown. If you take the [markdown from here](https://github.com/dotnet/BenchmarkDotNet/pull/348/files#diff-e3181db4ccf5ab8ea5896db85cddb1bd), the *raw* version looks like this:
```
 Method |      Mean |    StdDev |       P67 |
------- |---------- |---------- |---------- |
    Foo | 1.0000 ns | 0.0000 ns | 1.0000 ns |
    Bar | 1.0000 ns | 0.0000 ns | 1.0000 ns |
```

and it's rendered (incorrectly) like so:

 Method |      Mean |    StdDev |       P67 |
------- |---------- |---------- |---------- |
    Foo | 1.0000 ns | 0.0000 ns | 1.0000 ns |
    Bar | 1.0000 ns | 0.0000 ns | 1.0000 ns |

----

To make it work you need to use the following *raw* version (note the '|' at the beginning of each line)

```
| Method |      Mean |    StdDev |       P67 |
|------- |---------- |---------- |---------- |
|    Foo | 1.0000 ns | 0.0000 ns | 1.0000 ns |
|    Bar | 1.0000 ns | 0.0000 ns | 1.0000 ns |
```

which renders like so:

| Method |      Mean |    StdDev |       P67 |
|------- |---------- |---------- |---------- |
|    Foo | 1.0000 ns | 0.0000 ns | 1.0000 ns |
|    Bar | 1.0000 ns | 0.0000 ns | 1.0000 ns |

AndreyAkinshin:
@mattwarren, we already have such issue: https://github.com/dotnet/BenchmarkDotNet/issues/320
However, we decided to write approval tests first (we want to be sure that future fixes don't break anything).

mattwarren:
Ahh, I completely missed that issue. You're right, it's exactly the same :-)

AndreyAkinshin:
`CsvExporter.Default` and `CsvMeasurementsExporter.Default` use `CsvSeparator.CurrentCulture`. In tests, we should use an invariant version of these exporters (e.g. with `CsvSeparator.Comma`). Probably, it also makes sense to run all tests against different current cultures (e.g. `en-US` and `ru-RU`).

tests/BenchmarkDotNet.Tests/Mocks/MockHostEnvironmentInfo.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using BenchmarkDotNet.Environments;
+using BenchmarkDotNet.Horology;
+
+namespace BenchmarkDotNet.Tests.Mocks
+{
+    public class MockHostEnvironmentInfo : HostEnvironmentInfo
+    {
+        public MockHostEnvironmentInfo()
```
AndreyAkinshin: I suggest to rewrite this code in the following way:
```cs
public static readonly HostEnvironmentInfo Default = new MockHostEnvironmentInfo
{
    BenchmarkDotNetCaption = "BenchmarkDotNet",
    BenchmarkDotNetVersion = "0.x.y-mock",
    OsVersion = "MockOS",
    ProcessorName = new Lazy<string>(() => "MockProcessor CPU 42.0GHz"),
    ProcessorCount = 42,
    RuntimeVersion = "Clr 4.0.x.y",
    Architecture = "64bit",
    HasAttachedDebugger = false,
    HasRyuJit = true,
    Configuration = "RELEASE",
    JitModules = "clrjit-v4.6.x.y",
    DotNetCliVersion = new Lazy<string>(() => "1.0.0-mock-42"),
    ChronometerFrequency = new Frequency(42),
    ChronometerResolution = TimeInterval.FromNanoseconds(42),
    HardwareTimerKind = HardwareTimerKind.Tsc
};

private MockHostEnvironmentInfo()
{
}
```
1. It allows using a single instance of this class everywhere.
2. It allows introducing other instances of MockSummary.
3. The `Default` type is `HostEnvironmentInfo` instead of `MockHostEnvironmentInfo` which is better because we use `HostEnvironmentInfo` in the core method signatures.
4. It should be easy to see that we are working with the mock data. For example, if somebody accidentally prints the real environment data in an exporter, it shouldn't be equal to the mock data.
I made the above changes and found a bug. Here is my `ApprovalTest.Exporter.Md_StackOverflow.received.txt`:
```

    BenchmarkDotNet=v0.x.y-mock, OS=MockOS
    Processor=MockProcessor CPU 42.0GHz, ProcessorCount=42
    Frequency=42 Hz, Resolution=42.0000 ns, Timer=TSC
      [Host] : Clr 4.0.x.y, 64bit RyuJIT-v4.6.1586.0

    Job=LongRun  LaunchCount=3  TargetCount=100  
    WarmupCount=15  Allocated=N/A  

     Method |      Mean |    StdDev |       P67 |
    ------- |---------- |---------- |---------- |
        Foo | 1.0000 ns | 0.0000 ns | 1.0000 ns |
        Bar | 1.0000 ns | 0.0000 ns | 1.0000 ns |
```
As you can see, RyuJIT version is `4.6.1586.0` instead of `4.6.x.y` which is declared in `JitModules`. Thus, the test will fail for some version of the .NET Framework (with another RyuJIT version).

src/BenchmarkDotNet.Core/Environments/HostEnvironmentInfo.cs
```diff
@@ -11,53 +11,55 @@ namespace BenchmarkDotNet.Environments
...
         /// </summary>
-        public Lazy<string> DotNetCliVersion { get; }
+        public Lazy<string> DotNetCliVersion { get; protected set; }
 
         /// <summary>
         /// The frequency of the timer as the number of ticks per second.
         /// </summary>
-        public Frequency ChronometerFrequency { get; }
+        public Frequency ChronometerFrequency { get; protected set; }
+        public TimeInterval ChronometerResolution { get; protected set; }
```
AndreyAkinshin: It's better to have a method for `Resolution` instead of a property because we want to have the following invariant: `Resolution * Frequency = 1 second`. The implementation is trivial, just use `ChronometerFrequency.ToResolution()`.

tests/BenchmarkDotNet.IntegrationTests.Classic/Exporters/ApprovalTest.cs
```diff
@@ -0,0 +1,57 @@
...
+using BenchmarkDotNet.Exporters.Csv;
+using BenchmarkDotNet.Exporters.Json;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Tests.Mocks;
+using Xunit;
+
+namespace BenchmarkDotNet.IntegrationTests.Classic.Exporters
+{
+    [UseReporter(typeof(XUnit2Reporter))]
+    public class ApprovalTest
```
AndreyAkinshin: For now, `BenchmarkDotNet.IntegrationTests.Classic` is a good place for `ApprovalTests` because the corresponded NuGet package doesn't support .NET Core. In the future (with dotnet-preview3+), we will be able to move it to `BenchmarkDotNet.Tests` without pain. I suggest to put a comment here which is explain why do we keep the `ApprovalTest` in `IntegrationTests` and contains a `TODO` for the future.

AndreyAkinshin:
@alinasmirnova, it also makes sense to add the `*.received.txt` files to `.gitignore`.

tests/BenchmarkDotNet.IntegrationTests.Classic/Exporters/ApprovalTest.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            NamerFactory.AdditionalInformation = testName;
+
+            var logger = new AccumulationLogger();
+            exporter.ExportToLog(MockFactory.CreateSummary(config), logger);
+
+            Approvals.Verify(logger.GetLog());
+        }
+
+        private static List<object []> GetExporters()
```
AndreyAkinshin: Here we could use [`TheoryData<IExporter, string>`](https://github.com/xunit/xunit/blob/master/src/xunit.core/TheoryData.cs).

tests/BenchmarkDotNet.Tests/Mocks/MockFactory.cs
```diff
@@ -44,10 +51,45 @@ private static BenchmarkReport CreateReport(IConfig config)
...
+                HasRyuJit = true,
+                IsConcurrentGC = false,
+                IsServerGC = false,
+                JitInfo = "RyuJIT-v4.6.1637.0",
+                JitModules = "clrjit-v4.6.1637.0",
+                OsVersion = new Lazy<string>(() => "Microsoft Windows NT 10.0.14393.0"),
+                ProcessorCount = 8,
+                ProcessorName = new Lazy<string>(() => "Intel(R) Core(TM) i7-6700HQ CPU 2.60GHz"),
+                RuntimeVersion = "Clr 4.0.30319.42000"
+            };
```
AndreyAkinshin: Could we make these strings look like mock string. E.g. `0.10.3.20170408-develop` may confuse someone (it looks like an obsolete version). It's better to use something like `0.10.x-mock` here. Real versions of RyuJIT or Windows could accidentely coincide with real values and hide bugs, it's better use versions which are obvious fake (like `RyuJIT-v4.6.mock`).

tests/BenchmarkDotNet.Tests/Exporters/ApprovedFiles/ApprovalTest.Exporter.AsciiDocExporter_English (United States).approved.txt
```diff
@@ -0,0 +1,15 @@
+﻿....
```
AndreyAkinshin: Could we rename the `*Invariant Language (Invariant Country)*` files to just `*Invariant*`?

tests/BenchmarkDotNet.Tests/Exporters/ApprovedFiles/ApprovalTest.Exporter.MarkdownExporter-stackoverflow_Russian (Russia).approved.txt
```diff
@@ -0,0 +1,13 @@
+﻿
```
AndreyAkinshin: It also would be great to rename `Russian (Russia)` to `ru-ru` and `English (United States)` to `en-us`.

AndreyAkinshin:
@alinasmirnova, could you also add a few lines about Approval tests to docs?

tests/BenchmarkDotNet.Tests/Mocks/MockFactory.cs
alinasmirnova: Yeah, I think it's a good idea. I'll fix it today.

tests/BenchmarkDotNet.Tests/Exporters/ApprovedFiles/ApprovalTest.Exporter.AsciiDocExporter_English (United States).approved.txt
alinasmirnova: Not sure about it, but I'll try.

Issue #136, opened 2016-04-03
[Feature request] Fastcheck for correctness of benchmark implementations?


ig-sinicyn:
Use case: prevent users from comparing methods that do different things.

It's very easy to leave bug unnoticed in one of the benchmark methods. I just did it and had to spent some time to find that one of the competitors was exiting in the middle. 

Proposed solution is fragile, does not _proofs_ that all code is correct but it's still better than nothing.
**So, the proposal:**

If benchmark method has a return value, collect it and check that all another non-void methods return the same value. The value serves as a checksum, so, if it differs benchmark should fail.
Ensuring that result of the benchmark is calculated correctly and that the result does not produce false positives is responsibility of the author of the benchmark.

I think that this has to be opt-in feature and it's enough to check the results only on first run.

adamsitnik:
I have some idea, will try to implement it in form of a Validator that will be called before benchmark run.

adamsitnik:
@ig-sinicyn Some progress: I have implemented ExecutionValidator that checks whether it is possible to run all benchmarks first. If you do not want to continue execution if single benchmark fails validation you need to use ExecutionValidator.FailOnError

The feature is not done yet, I will have to store and compare the results. I need to consider collections as well, but I will most probably only derive CorrectnessValidator from ExecutionValidator .

ig-sinicyn:
@adamsitnik  ok, thanks for it! :)

mattwarren:
BTW, after using BenchmarkDotNet @fransbouma also asked for this feature, see https://twitter.com/FransBouma/status/723818104501997568

So consider that another +1 for it.

ig-sinicyn:
@adamsitnik, @AndreyAkinshin 

Upping. With #356 merged there's an API that can be easily extended with result reporting. 

Initial idea is following: last invocation result should be passed as a member of `RunResults` and should be exposed via `BenchmarkReport`. I'd suggest to use  `Convert.ChangeType`  or `TypeConverter.ConvertFromString()`, both should work in .Net Core.

adamsitnik:
@AndreyAkinshin I believe that we should not implement this feature. We should focus on performance related things, we don't have enough time to implement this one. 

People should use unit test frameworks to check the correctness, it's not our responsibility.

PR #356, opened 2017-02-03
Feature: host API interface

ig-sinicyn:
# Request for comments: Host API & no console dependency.

This PR is aimed to fix two major issues for BDN:
**First**, there should be a way to run perftests without dependency on the Console API.
This is critical for various scenarios, to name a few: 
* in-process benchmarks
* xUnit-hosted tests (xUnit does not support console output)

**Second**, there's no clear API to use for diagnoser notifications.
There's black magic with `Console.WriteLine(Signals.*)` but it is quite unreliable especially if the output is buffered.

This proposal solves (okay, tries to solve:)) both of them.

Note that current design is not a thing I'm going to insist on. Tests [are green](https://ci.appveyor.com/project/ig-sinicyn/benchmarkdotnet), suggestions are welcome:)

### 1. No hard dependency on Console output.
This was surprisingly easy. Just introduced `HostEnvironmentInfo.FallbackLogger` and use it where the config is unavailable.
There were only two such places, both at `BenchmarkConverter`.

### 2. Better API for diagnoser notifications.
The idea is pretty simple. There's an `IHostApi` interface to be used by `Engine` and by `Runnable`. 
And there is a simple implementation that uses `Console.Out` (or any another `TextWriter`) to notify the host.

Third-party toolchains may provide own host API that will call diagnosers directly (for in-process toolchains)
or will use cross-process synchronization primitives (Mutex, for example) to improve latency and to make diagnosers more accurate.

As far as I can see the only breaking changes are:
* Addition of `TextWriter` arg to the `ClockExtensions.PrintInfo` (was not used anyway).
* Addition of `TextWriter` arg to the `RunResults.Print()` (may be reverted, the body of the Print() must be copied into implementations of the `IHostApi` then).
* Addition of the `IHostApi` to the `IEngine `interface (may be removed as I left `IsDiagnoserAttached` and `WriteLine()` for compatibility).
* Addition of the `IHostApi` to the Engine parameters.
 
## Things to improve:
* entire `FallbackLogger` idea. 
  - It should be either removed (there cannot be null configs then, the config should be created only once at start and should be passed through all calls)
  - or the `FallbackLogger` should be settable (only once, all calls but first should throw to prevent possible races).
* This implementation hardcodes `ConsoleHostApi` into a `BenchmarkProgram.txt`. Must be replaced with `$HostApi$` placeholder, I guess.
* I've added a stub for `AfterAnythingElse` signal but did not add `IDiagnoser.AfterAnythingElse()` method as I'm not sure we need it at all.
It may be better to make 'all done' notification to be responsibility of the host. This way it will work even if child process dies unexpectedly.
* `IHostApi.WriteLine()` overloads. Do we need al three of them?
* Add to the IHostApi a method for passing return value. Will solve the #136.


adamsitnik:
@ig-sinicyn great job! I am on holidays now, will do a code review in the next week (after 15th of Feb)

ig-sinicyn:
@adamsitnik ok, I'm not in hurry. Will push the code for review in a day or two.

ig-sinicyn:
Up.
I'm not hurry still, but will be happy to hear something about the PR :)

AndreyAkinshin:
@ig-sinicyn, sorry for the late reply. We were too busy on new csproj toolchain support, now it's done (thanks to @adamsitnik). Your PR is the next one. =) Could you rebase your changes on master?

ig-sinicyn:
@AndreyAkinshin 
It was done already:)

Congrats with transition to csproj! Finally can uninstall vs2015:)

AndreyAkinshin:
> It was done already:)

Awesome! Will review it soon. @adamsitnik, could you also take a look?

src/BenchmarkDotNet.Core/Engines/ConsoleHostApi.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.IO;
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Engines
+{
+    public sealed class ConsoleHostApi: IHostApi
+    {
+        private readonly TextWriter _outWriter;
```
AndreyAkinshin: In BenchmarkDotNet, we don't use `_` as the private field prefix. Please, remove it.

src/BenchmarkDotNet.Core/Engines/ConsoleHostApi.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public bool IsDiagnoserAttached { get; }
+
+        public void Write(string message) => _outWriter.Write(message);
+
+        public void WriteLine() => _outWriter.WriteLine();
+
+        public void WriteLine(string message) => _outWriter.WriteLine(message);
+
+        public void WriteLine(string messageFormat, params object[] args) => _outWriter.WriteLine(messageFormat, args);
```
AndreyAkinshin: It makes sense to add `JetBrains.Annotations.StringFormatMethodAttribute` here.

src/BenchmarkDotNet.Core/Engines/IHostApi.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace BenchmarkDotNet.Engines
+{
+    public interface IHostApi
```
AndreyAkinshin: Are you sure that we need all these methods in the interface? For example, it's easy to implement `WriteLine` and his friends as extension methods via `Write`. The signal methods also could be moved out from the interface. If we are going to create other implementation of `IHostApi` (e.g. via sockets), such extension method will help to avoid code duplications.

AndreyAkinshin:
LGTM. I wrote a few minor comments, but I like it in general. Thanks, @ig-sinicyn! Sorry for the late review.

ig-sinicyn:
@AndreyAkinshin 

Ok, will do.

> Are you sure that we need all these methods in the interface?

Nope:) Will left only these overloads
```c#
        void Write(string message); 
        void WriteLine(); 
        void WriteLine(string message); 
```
to keep in sync with ILogger methods.

> The signal methods also could be moved out from the interface.

Well, may be something like
```c#
enum HostEventKind { ... }
interface IHostApi
{
    ...

    void Notify(HostEventKind kind);
}
```
?
This way it will not be a breaking change to add (or remove) new signals.
If not, can you list all members that should be in `IHostApi`?

AndreyAkinshin:
I'm thinking about something like this:
```cs
public enum HostSignal
{
    BeforeAnythingElse,
    AfterSetup,
    BeforeCleanup,
    AfterAnythingElse
}
public enum HostProperty
{
    IsDiagnoserAttached
}
public interface IHost
{
    void SendMessage(string message);
    void SendResults(RunResults runResults);
    void SendSignal(HostSignal hostSignal);
    void SendProperty(HostProperty hostProperty, string value)
}
```

Issue #358, opened 2017-02-06
[Need help] InProcessToolchain PR, some design questions


ig-sinicyn:
**UPD** [Completed](https://github.com/ig-sinicyn/BenchmarkDotNet/tree/feature-inprocess/src/BenchmarkDotNet.Core/Toolchains/InProcess), will do PR after #356 merged. 
 
I'm preparing PR of the InProcess toolchain that allows to run benchmark code without compiling binaries / running another process.

The main benefit of it is ability to run a lot of tests very quickly in cost of less accurate results. In practice we're getting stable results with 2-3 seconds per benchmark time limit (fast enough if compared to typical 20-60 secs for outofproc toolchain).

I'm going to keep implementation in sync with the code in `BenchmarkProgram.txt` but I want to clarify some moments at first.

1. Codegen. Currently I'm using Reflection.Emit for loop unroll and there is fallback code that uses `Delegate.Combine()` instead of emitting the code. It produces less accurate results due to additional overhead but it works fine with .Net Native. Is it ok to have both?

2. `Consumer.Consume()` call for non-void methods. Is it ok to emit stfld instead of calling the method? Should the return value be stored each time the target method is invoked or it will be better to save the return value for last call only?

3. `Dummy()` methods. Is the `@DummyUnroll@` number a constant? If it is not - is it ok to fallback to method with hardcoded 64 lines of `dummyVar++;` for cases when `Reflection.Emit` is unavailable?
Should be there three different dummy methods or it's enough to have just one?

---

**UPD, More questions:**

**BenchmarkProgram.txt**
* Setup/Cleanup method are invoked by Runnable. Shouldn't it be responsibility of the Engine?
* `Target.AdditionalLogic` is unused for now. Should it be removed? 
  If not - is it ok if InProcessValidator will warn when AdditionalLogic is not empty?
* `InvokeMultiple` loop does not cache mainAction/consumer in local variables. Is this intentional?

**Diagnosers**
* Shouldn't notification signals for diagnosers include a native thread id?
* `Diagnoser.BeforeCleanup()` has no args - is it ok?

**Current impl drawbacks**
* Current implementation redirects output directly into the logger so there's nothing to pass into 
`ExecuteResult.ExtraOutput`. Is it ok? Should I use intermediate buffer (`AccumulationLogger`) instead?
* Thread.SetApartmentState() and Thread.Priority are not avalable in .net core FW (at least until netstandard 2), so there's no way to set up them.
[...]

PR #379, opened 2017-03-03
Feature: in-process benchmarks

ig-sinicyn:
This PR adds toolchain that enables in-process benchmark running.

The main benefit of it is ability to run a lot of benchmarks very quickly in a cost of less accurate results.

In practice we're getting stable results with 2-3 seconds per benchmark time limit (fast enough if compared to typical 20-60 secs for outofproc toolchain).
*Note: we use custom config like this:
```c#
		public static readonly Job DefaultJob = new Job("Competition")
		{
			Env =
			{
				Gc =
				{
					Force = false
				}
			},
			Run =
			{
				LaunchCount = 1,
				WarmupCount = 100,
				TargetCount = 300,
				RunStrategy = RunStrategy.Throughput,
				UnrollFactor = 16,
				InvocationCount = 256
			},
			Infrastructure =
			{
				Toolchain = InProcessToolchain.DontLogOutput
			}
		}.Freeze();
```

The implementation is in sync of content of `BenchmarkProgram.txt` so with same setup results are expected to be pretty close to results obtained with default toolchain (and they are).

By default toolchain uses Reflection.Emit to emit unrolled body of the loops but in case the benchmark run in platform that does not support dynamic code generation (e.g. .net native)
there's a fallback code that uses Delegate.Combine to unroll the loop (can be set via BenchmarkActionCodegen enum).
Because of the latter, the body of the Dummy method (BenchmarkActionFactory.DummyInstance) is hardcoded.

Also, there's an option that suppress output from the toolshain's executor. It makes the log much more shorter:)

Finally, there is a InProcessValidator that helps to ensure that surrent envirement matches to the Job settings.

More details and things to improve - at #358.


AndreyAkinshin:
@ig-sinicyn, could you add an example of an in-process benchmark in the `BenchmarkDotNet.Samples` project? (For example, `IntroInProcess`)

ig-sinicyn:
@AndreyAkinshin yes, of course.

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessExecutor.cs
```diff
@@ -0,0 +1,150 @@
...
+    [PublicAPI]
+    [SuppressMessage("ReSharper", "ArrangeBraces_using")]
+    public class InProcessExecutor : IExecutor
+    {
+        private static readonly TimeSpan UnderDebuggerTimeout = TimeSpan.FromDays(1);
+
+        /// <summary> Default timeout for in-process benchmarks. </summary>
+        public static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(5);
+
+        #region .ctor & properties
```
adamsitnik: nit: we don't use regions in BenchmarkDotNet, they are ugly ;)

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessExecutor.cs
```diff
@@ -0,0 +1,150 @@
...
+            BuildResult buildResult, Benchmark benchmark, ILogger logger, IResolver resolver, IDiagnoser diagnoser = null)
+        {
+            // TODO: preallocate buffer for output (no direct logging)?
+            var host = new InProcessHost(benchmark, LogOutput ? logger : null, diagnoser);
+
+            int exitCode = -1;
+            var runThread = new Thread(() => exitCode = ExecuteCore(host, benchmark, logger));
+            if (benchmark.Target.Method.GetCustomAttributes<STAThreadAttribute>(false).Any())
+            {
+                // TODO: runThread.SetApartmentState(ApartmentState.STA);
```
adamsitnik: it's great that you remember that we support STA, may I ask why is this commented out?

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessExecutor.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <param name="benchmark">The benchmark.</param>
+        /// <param name="logger">The logger.</param>
+        /// <param name="resolver">The resolver.</param>
+        /// <param name="diagnoser">The diagnoser.</param>
+        /// <returns>Execution result.</returns>
+        public ExecuteResult Execute(
+            BuildResult buildResult, Benchmark benchmark, ILogger logger, IResolver resolver, IDiagnoser diagnoser = null)
+        {
+            // TODO: preallocate buffer for output (no direct logging)?
+            var host = new InProcessHost(benchmark, LogOutput ? logger : null, diagnoser);
```
adamsitnik: maybe instead of passing null you could implement some Null Object pattern here (logger that does nothing)?

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessExecutor.cs
```diff
@@ -0,0 +1,150 @@
...
+            if (benchmark.Target.Method.GetCustomAttributes<STAThreadAttribute>(false).Any())
+            {
+                // TODO: runThread.SetApartmentState(ApartmentState.STA);
+            }
+            runThread.IsBackground = true;
+
+            var timeout = HostEnvironmentInfo.GetCurrent().HasAttachedDebugger ? UnderDebuggerTimeout : ExecutionTimeout;
+
+            runThread.Start();
+            // TODO: to timespan overload (not available for .Net Core for now).
```
adamsitnik: you might use conditional switches if something is not available for .NET Core
#if !CORE
#else
#endif

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessExecutor.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            var affinity = benchmark.Job.ResolveValueAsNullable(EnvMode.AffinityCharacteristic);
+            try
+            {
+                process.TrySetPriority(ProcessPriorityClass.High, logger);
+                if (affinity != null)
+                {
+                    process.TrySetAffinity(affinity.Value, logger);
+                }
+                // TODO: set thread priority to highest
```
adamsitnik: why setting the priority was left as todo? is there some contract mismatch (.net core and stuff)?

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessExecutor.cs
```diff
@@ -0,0 +1,150 @@
...
+            {
+                return new ExecuteResult(true, exitCode, new string[0], new string[0]);
+            }
+
+            var lines = new List<string>();
+            foreach (var measurement in runResults.GetMeasurements())
+            {
+                lines.Add(measurement.ToOutputLine());
+            }
+            lines.Add(runResults.GCStats.WithTotalOperations(runResults.TotalOperationsCount).ToOutputLine());
```
adamsitnik: 👍 thanks for keeping the MemoryDiagnoser support!

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessRunner.cs
```diff
@@ -0,0 +1,137 @@
...
+            bool isDiagnoserAttached = host.IsDiagnoserAttached;
+
+            // the first thing to do is to let diagnosers hook in before anything happens
+            // so all jit-related diagnosers can catch first jit compilation!
+            if (isDiagnoserAttached)
+                host.BeforeAnythingElse();
+
+            try
+            {
+                // we are not using Runnable here in any direct way in order to avoid strong dependency Main<=>Runnable
```
adamsitnik: 👍 thanks for keeping the old behaviour!

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessRunner.cs
```diff
@@ -0,0 +1,137 @@
...
+                var engine = job
+                    .ResolveValue(InfrastructureMode.EngineFactoryCharacteristic, InfrastructureResolver.Instance)
+                    .Create(engineParameters);
+
+                engine.PreAllocate();
+
+                setupAction.InvokeSingle();
+
+                if (job.ResolveValue(RunMode.RunStrategyCharacteristic, EngineResolver.Instance).NeedsJitting())
+                    engine.Jitting(); // does first call to main action, must be executed after setup()!
```
adamsitnik: 👍 thanks for keeping another detail in mind!

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessRunner.cs
```diff
@@ -0,0 +1,137 @@
...
+            /// <param name="benchmark">The benchmark.</param>
+            private static void FillProperties(object instance, Benchmark benchmark)
+            {
+                foreach (var parameter in benchmark.Parameters.Items)
+                {
+                    var flags = BindingFlags.Public;
+                    flags |= parameter.IsStatic ? BindingFlags.Static : BindingFlags.Instance;
+
+                    var targetType = benchmark.Target.Type;
+                    var paramProperty = targetType.GetProperty(parameter.Name, flags);
```
adamsitnik: does this method support public fields? (not only properties)

ig-sinicyn:
@AndreyAkinshin 
Done.

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessToolchain.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </summary>
+    /// <seealso cref="IToolchain" />
+    [PublicAPI]
+    public sealed class InProcessToolchain : IToolchain
+    {
+        /// <summary>The default toolchain instance.</summary>
+        public static readonly IToolchain Instance = new InProcessToolchain(true);
+
+        /// <summary>The toolchain instance without output logging.</summary>
+        public static readonly IToolchain DontLogOutput = new InProcessToolchain(false);
```
adamsitnik: 👍 that's a very nice and desired feature!

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessToolchain.cs
```diff
@@ -0,0 +1,69 @@
...
+            Builder = new InProcessBuilder();
+            Executor = new InProcessExecutor(timeout, codegenMode, logOutput);
+        }
+
+        /// <summary>Determines whether the specified benchmark is supported.</summary>
+        /// <param name="benchmark">The benchmark.</param>
+        /// <param name="logger">The logger.</param>
+        /// <param name="resolver">The resolver.</param>
+        /// <returns><c>true</c> if the benchmark can be run with the toolchain.</returns>
+        public bool IsSupported(Benchmark benchmark, ILogger logger, IResolver resolver) => true;
```
adamsitnik: I believe that there are few scenarios that you can not support. For example: architecture/jit different than host/current

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessValidator.cs
```diff
@@ -0,0 +1,145 @@
...
+                { AccuracyMode.MinIterationTimeCharacteristic, DontValidate },
+                { AccuracyMode.RemoveOutliersCharacteristic, DontValidate },
+                { InfrastructureMode.ClockCharacteristic, DontValidate },
+                { InfrastructureMode.EngineFactoryCharacteristic, DontValidate },
+                { InfrastructureMode.ToolchainCharacteristic, ValidateToolchain }
+            };
+
+        // ReSharper restore HeapView.DelegateAllocation
+
+        private static string DontValidate(Job job, Characteristic characteristic) => null;
```
adamsitnik: 👍 very smart way to show that you don't validate something on purpose!

src/BenchmarkDotNet.Core/Toolchains/InProcess/[BenchmnarkActions]/BenchmarkActionFactory.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                var genericType = resultType.GetGenericTypeDefinition();
+                var argType = resultType.GenericTypeArguments[0];
+                if (typeof(Task<>) == genericType)
+                    return Create(
+                        typeof(BenchmarkActionTask<>).MakeGenericType(argType),
+                        resultInstance, targetMethod,
+                        codegenMode, unrollFactor);
+
+                if (typeof(ValueTask<>).IsAssignableFrom(genericType))
```
Comment (adamsitnik) [inline src/BenchmarkDotNet.Core/Toolchains/InProcess/[BenchmnarkActions]/BenchmarkActionFactory.cs]: 👍 thanks for keeping another nasty detail in mind!

src/BenchmarkDotNet.Core/Extensions/ProcessExtensions.cs
```diff
@@ -25,5 +26,55 @@ public static void EnsureProcessorAffinity(this Process process, IntPtr value)
...
+        {
+            if (process == null)
+                throw new ArgumentNullException(nameof(process));
+            if (logger == null)
+                throw new ArgumentNullException(nameof(logger));
+
+            try
+            {
+                var cpuMask = (1 << Environment.ProcessorCount) - 1;
+                process.ProcessorAffinity = new IntPtr(processorAffinity.ToInt32() & cpuMask);
```
AndreyAkinshin: Nice trick with `cpuMask`! Could we introduce a helper function and use it in `EnsureProcessorAffinity`?

ig-sinicyn: Yep, will add.

src/BenchmarkDotNet.Core/Toolchains/InProcess/[BenchmnarkActions]/BenchmarkActionFactory_Base.cs
```diff
@@ -0,0 +1,228 @@
...
+
+            protected static TDelegate Unroll<TDelegate>([NotNull] TDelegate callback, int unrollFactor)
+            {
+                if (callback == null)
+                    throw new ArgumentNullException(nameof(callback));
+
+                if (unrollFactor <= 1)
+                    return callback;
+
+                return (TDelegate)(object)Delegate.Combine(
```
Comment (adamsitnik) [inline src/BenchmarkDotNet.Core/Toolchains/InProcess/[BenchmnarkActions]/BenchmarkActionFactory_Base.cs]: 👍 another smart idea!

tests/BenchmarkDotNet.IntegrationTests/InProcessTest.cs
```diff
@@ -0,0 +1,388 @@
...
+        public void TestBenchmarkActionCleanup() => TestInvoke(x => x.Cleanup(), UnrollFactor);
+
+        [Fact]
+        public void TestBenchmarkActionVoid() => TestInvoke(x => x.InvokeOnceVoid(), UnrollFactor);
+        [Fact]
+        public void TestBenchmarkActionTask() => TestInvoke(x => x.InvokeOnceTaskAsync(), UnrollFactor, null);
+        [Fact]
+        public void TestBenchmarkActionRefType() => TestInvoke(x => x.InvokeOnceRefType(), UnrollFactor, StringResult);
+        [Fact]
+        public void TestBenchmarkActionValueType() => TestInvoke(x => x.InvokeOnceValueType(), UnrollFactor, DecimalResult);
```
adamsitnik: personally, I don't like to use [Facts] and not following the naming convention. For example `TestBenchmarkActionValueType` is not a fact, `BenchmarksReturningValueTypesAreSupported` is

tests/BenchmarkDotNet.IntegrationTests/InProcessTest.cs
```diff
@@ -0,0 +1,388 @@
...
+                .With(logger)
+                .With(DefaultColumnProviders.Instance);
+            try
+            {
+                BenchmarkAllCases.Counter = 0;
+
+                var summary = CanExecute<BenchmarkAllocates>(config);
+
+                var testLog = logger.GetLog();
+                Assert.Contains("// Benchmark: BenchmarkAllocates.Allocates:", testLog);
```
adamsitnik: it would be better if you could use the strong-typed properties of the summary instead of looking for strings in the output

src/BenchmarkDotNet.Core/Extensions/ProcessExtensions.cs
ig-sinicyn: Added. Also, fixed dumb mistake with processorAffinity.ToInt32().

AndreyAkinshin:
@ig-sinicyn, please don't use square brackets in folder names (`[BenchmarkActions]`).

src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessHost.cs
```diff
@@ -0,0 +1,115 @@
...
+        private readonly Process currentProcess;
+
+        /// <summary>Creates a new instance of <see cref="InProcessHost"/>.</summary>
+        /// <param name="benchmark">Current benchmark.</param>
+        /// <param name="logger">Optional logger for informational output.</param>
+        /// <param name="diagnoser">Diagnosers, if attached.</param>
+        public InProcessHost([NotNull] Benchmark benchmark, [CanBeNull] ILogger logger, [CanBeNull] IDiagnoser diagnoser)
+        {
+            if (benchmark == null)
+                throw new ArgumentNullException(nameof(benchmark));
```
AndreyAkinshin: Remove this check or replace `[NotNull]` by `[CanBeNull]`.

ig-sinicyn:
@adamsitnik 

Wow!:)

Summarized to keep the discussion simple :)

0. **what is the benefit of using the mask here?**
Process.ProcessorAffinity setter throws if the value is larger than avalable processors mask :)
I've learned it the hard way:)

1. **XML comments**: ported as is, sorry:)
Should I remove these 'dummy' comments?
If yes - all of them?
Explanation: we're using sandcastle to build helpfile on each build, so it's much easier to add a stub for help than re-comment all code later.

2. **#regions** - ok, they are gone:)

3. **InProcessHost:** logger that does nothing)?
+1, replaced with empty CompositeLogger.

4. **STA, why is this commented out?** and **why setting the priority was left as todo?**
I'm afraid .Net core has no API for it:(

5. **InProcessRunner.FillProperties:** 
nice catch! Fields are supported now.

6. **Toolchain.IsSupported:** 
this is already checked via in-process validator:)
Should I call the validator from InProcessToolchain.IsSupported?

7. **Personally, I don't like to use [Facts]** 
Fixed.

8. **it would be better if you could use the strong-typed properties of the summary** 
They are checked too. 
```c#
Assert.True(summary.Reports.Sum(r => r.GcStats.AllocatedBytes) > 0);
```
The output is checked to ensure that it is not broken:)

Will fix (2) and (6) as answered, others are fixed.


@AndreyAkinshin yours are not done, will commit later.

adamsitnik:
@ig-sinicyn thanks for applying the fixes and providing the answers! 

> Should I remove these 'dummy' comments?

No, but please don't add more in the future.

> STA, why is this commented out? and why setting the priority was left as todo?

I think that it would be better to do what we can for the classic .NET like
```cs
#if CLASSIC
setSta()
#endif
```

> this is already checked via in-process validator:)
> Should I call the validator from InProcessToolchain.IsSupported?

Thanks for the explanation. Is it possible to make sure that if user is using `InProcessToolchain` the `InProcessValidator` is always enabled? You did a great job with this validator, I just want to make sure people always get clear message what is possible and what is not.

btw do you have a twitter account so I can mention you in some conversations?

adamsitnik:
@ig-sinicyn the build has failed, can you take a look?

Otherwise LGTM and ready to merge!

ig-sinicyn:
@adamsitnik 

Sorry for broken build, missed a file. Fixed.

---

> No, but please don't add more in the future.

Ok, I've got it:)

>I think that it would be better to do what we can for the classic .NET like

Done.

>Thanks for the explanation. Is it possible to make sure that if user is using InProcessToolchain the InProcessValidator is always enabled?

Ok, there's what we can do:

1. Leave as is - no warnings, it silently works even if the environment do not match config.

2. Remove the validator and move the validator logic into `Toolchain.IsSuported()` - ugly. Validation errors will not be exposed by `Summary.ValidationErrors`.

3. Call validator from toolchain and log validation results as verbose (with `// ` prefix). Allow users to add validators if they want explicit validation.

I'm going to do the third. Better ideas, anyone? :)

>btw do you have a twitter account so I can mention you in some conversations?

Very old one https://twitter.com/Sinix . I almost do not use it, but will answer if there will be any questions.

ig-sinicyn:
@adamsitnik 

> Is it possible to make sure that if user is using InProcessToolchain the InProcessValidator is always enabled?

Done (see prev post for details).

AndreyAkinshin:
@adamsitnik  I think that you hurry to do it, I still have some concerns. The biggest one: here is my output for `IntroInProcess`:
```
// * Detailed results *
IntroInProcess.'new byte[10kB]': InProcess(Toolchain=InProcessToolchain, LaunchCount=1, TargetCount=15, WarmupCount=10)
Runtime = ; GC =
Mean = 533.4784 ns, StdErr = 6.2640 ns (1.17%); N = 15, StdDev = 24.2602 ns
Min = 503.6674 ns, Q1 = 511.8996 ns, Median = 530.5997 ns, Q3 = 548.4981 ns, Max = 587.9064 ns
IQR = 36.5985 ns, LowerFence = 457.0019 ns, UpperFence = 603.3958 ns
ConfidenceInterval = [521.2010 ns; 545.7558 ns] (CI 95%)
Skewness = 0.74, Kurtosis = 2.44


IntroInProcess.'stackalloc byte[10kB]': InProcess(Toolchain=InProcessToolchain, LaunchCount=1, TargetCount=15, WarmupCount=10)
Runtime = ; GC =
Mean = 5.6903 ns, StdErr = 0.0193 ns (0.34%); N = 15, StdDev = 0.0747 ns
Min = 5.5570 ns, Q1 = 5.6381 ns, Median = 5.7103 ns, Q3 = 5.7320 ns, Max = 5.8272 ns
IQR = 0.0939 ns, LowerFence = 5.4973 ns, UpperFence = 5.8728 ns
ConfidenceInterval = [5.6525 ns; 5.7281 ns] (CI 95%)
Skewness = -0.05, Kurtosis = 2.03


IntroInProcess.'new byte[10kB]': OutOfProc(LaunchCount=1, TargetCount=15, WarmupCount=10)
Runtime = .NET Core 4.6.24628.01, 64bit RyuJIT; GC = Concurrent Workstation
Mean = 438.6340 ns, StdErr = 3.7297 ns (0.85%); N = 15, StdDev = 14.4451 ns
Min = 424.7953 ns, Q1 = 425.9633 ns, Median = 433.1360 ns, Q3 = 449.5959 ns, Max = 469.3894 ns
IQR = 23.6327 ns, LowerFence = 390.5143 ns, UpperFence = 485.0449 ns
ConfidenceInterval = [431.3238 ns; 445.9443 ns] (CI 95%)
Skewness = 0.79, Kurtosis = 2.08


IntroInProcess.'stackalloc byte[10kB]': OutOfProc(LaunchCount=1, TargetCount=15, WarmupCount=10)
Runtime = .NET Core 4.6.24628.01, 64bit RyuJIT; GC = Concurrent Workstation
Mean = 5.7449 ns, StdErr = 0.0895 ns (1.56%); N = 15, StdDev = 0.3465 ns
Min = 5.4198 ns, Q1 = 5.4567 ns, Median = 5.5326 ns, Q3 = 6.1673 ns, Max = 6.3062 ns
IQR = 0.7107 ns, LowerFence = 4.3907 ns, UpperFence = 7.2333 ns
ConfidenceInterval = [5.5696 ns; 5.9203 ns] (CI 95%)
Skewness = 0.6, Kurtosis = 1.52


Total time: 00:01:17 (77.6 sec)

// * Summary *

BenchmarkDotNet=v0.10.3.20170304-develop, OS=Microsoft Windows 10.0.14393
Processor=Intel(R) Core(TM) i7-6700HQ CPU 2.60GHz, ProcessorCount=8
Frequency=2531250 Hz, Resolution=395.0617 ns, Timer=TSC
dotnet cli version=2.0.0-alpha-004853
  [Host]    : .NET Core 4.6.24628.01, 64bit RyuJIT
  OutOfProc : .NET Core 4.6.24628.01, 64bit RyuJIT

LaunchCount=1  TargetCount=15  WarmupCount=10

                  Method |       Job |          Toolchain |        Mean |    StdErr |     StdDev |  Gen 0 | Allocated |
------------------------ |---------- |------------------- |------------ |---------- |----------- |------- |---------- |
 'stackalloc byte[10kB]' | InProcess | InProcessToolchain |   5.6903 ns | 0.0193 ns |  0.0747 ns |      - |       0 B |
 'stackalloc byte[10kB]' | OutOfProc |            Default |   5.7449 ns | 0.0895 ns |  0.3465 ns |      - |       0 B |
        'new byte[10kB]' | OutOfProc |            Default | 438.6340 ns | 3.7297 ns | 14.4451 ns | 3.1590 |  10.02 kB |
        'new byte[10kB]' | InProcess | InProcessToolchain | 533.4784 ns | 6.2640 ns | 24.2602 ns | 3.1574 |  10.02 kB |
```
Why do we have a huge difference for 'new byte[10kB]' in `InProcess` and `OutOfProc`?

adamsitnik:
>  I think that you hurry to do it, I still have some concerns

I am sorry, I did not know that. I can revert the merge commit if you want.

> Why do we have a huge difference for 'new byte[10kB]' in InProcess and OutOfProc?

Do other benchmarks have also such huge difference?

AndreyAkinshin:
And some other minor concerns:
1. InProcess benchmark has empty `Runtime` and `GC` values in the "Detailed results" section. They should be filled by the host values.
2. We should print `InProcess` instead of `InProcessToolchain` in the Toolchain columns of the summary table.
3. Eventually, I want to transform `BenchmarkDotNet.Samples` into a gallery of benchmark which teach people how to use our API. `IntroInProcessWrongEnv` shows how to get an error message. I think it should be an integration test instead which verifies that we print this message.
4. It makes sense to creaate a nice attribute for in-process benchmark. For example, we can introduce `InProcessJobAttribute` which is based on `SimpleJobAttribute` and always uses InProcessToolchain. So, we will be able to declare such benchmarks with a single attribute line without of a ManualConfig requirement.
5. It would be also nice to add some info in our documentation which explains the difference between default toolchain and InProcessToolchain and shows how to use it.

AndreyAkinshin:
> I am sorry, I did not know that. I can revert the merge commit if you want.

There's nothing so terrible in this merge, we can try to fix it in master. A note for a future: it's ok to quickly merge simple PRs, but huge PRs which contains a lot of tricky logic should be carefully tests for a merge.

> Do other benchmarks have also such huge difference?

I hadn't a lot of time for testing, but it seems that CPU-bound benchmarks work fine. I guess that the problem is about huge memory traffic, maybe it's a GC-related issue. So, we have to look at all the GC settings and check if there is a difference between Default and InProcess toolchains.

ig-sinicyn:
@AndreyAkinshin 
> Why do we have a huge difference 

I'm almost sure It's because of GC. More stuff to collect => more time spent in GC. Sadly, there's no diagnoser for % GC time and I cannot proof it easily

AndreyAkinshin:
@ig-sinicyn, it seems that we still have the same GC-related problem in `IntroInProcess`.

PR #381, opened 2017-03-04
Implement UAP toolchain

zabulus:


src/BenchmarkDotNet.Core/Attributes/Jobs/UapJobAttribute.cs
```diff
@@ -0,0 +1,18 @@
+﻿using BenchmarkDotNet.Environments;
+using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Attributes.Jobs
+{
+    public class UapJobAttribute : JobConfigBaseAttribute
+    {
+        public UapJobAttribute(string devicePortalUri, string csfrCookie, string wmidCookie, string uapBinariesPath)
```
adamsitnik: could you add some `/// <summary>` here so the users know what to pass?

src/BenchmarkDotNet.Core/Configs/DefaultConfig.cs
```diff
@@ -69,7 +69,7 @@ public IEnumerable<IValidator> GetValidators()
 
         private static IDiagnoser[] LoadDiagnosers()
         {
-#if !CORE
+#if !CORE && !UAP
```
adamsitnik: nit: you can write `#if CLASSIC` instead of `#if !CORE && !UAP`

src/BenchmarkDotNet.Core/Diagnosers/CompositeDiagnoser.cs
```diff
@@ -19,13 +19,13 @@ public CompositeDiagnoser(params IDiagnoser[] diagnosers)
 
         public IColumnProvider GetColumnProvider() 
             => new CompositeColumnProvider(diagnosers.Select(d => d.GetColumnProvider()).ToArray());
-
+#if !UAP
```
adamsitnik: is the `Process` class not available for UAP? If so I have one idea how we could get rid of most of the `#if`. Maybe we could define a class that is called Process and give it the same api but just throw NotImplementedException for every method? Then the rest of the BDN code could remain untouched? Sth like

```cs
#if UWP
namespace System.Diagnostics
{
   public class Process
   {

   }
}
#endif
```

src/BenchmarkDotNet.Toolchains.Uap/BenchmarkDotNet.Toolchains.Uap.csproj
```diff
@@ -0,0 +1,45 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+  
+  <PropertyGroup>
+    <AssemblyTitle>BenchmarkDotNet.Toolchains.Uap</AssemblyTitle>
+    <VersionPrefix>0.10.1-develop</VersionPrefix>
```
adamsitnik: VersionPrefix is not needed here, common.props does everything for versioning

src/BenchmarkDotNet.Toolchains.Uap/Helpers/ResourceHelper.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.IO;
+using System.Reflection;
+
+namespace BenchmarkDotNet.Toolchains.Uap.Helpers
+{
+    internal static class ResourceHelper
```
adamsitnik: can it be refactored to avoid having two copies of ResourceHelper class?

src/BenchmarkDotNet.Toolchains.Uap/Templates/UapBenchmarkProgram.txt
```diff
@@ -0,0 +1,285 @@
...
+                    finally
+                    {
+                        lc.Dispose();
+                        Application.Current.Exit();
+                    }
+                };
+
+                Task.Factory.StartNew(action);
+            }
+            catch (Exception ex)
```
adamsitnik: the second catch will never be raised. You start the async job and as long as you don't await/wait the task it will not throw in this context. Suggestion: just remove the second catch block

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/DevicePortalApiWrapper.cs
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        private static long GetTime()
+        {
+            long retval = 0;
+            var st = new DateTime(1970, 1, 1);
+            TimeSpan t = (DateTime.Now.ToUniversalTime() - st);
+            retval = (long)(t.TotalMilliseconds + 0.5);
+            return retval;
+        }
```
adamsitnik: could you explain this logic?

zabulus:
Build failure caused by "dotnet restore" command. 
Current tooling is unable to work with OS-specific projects (particularly UWP). Issue: https://github.com/dotnet/cli/issues/5846. I can confirm, that suggested replace to "msbuild /t:Restore" should work out.

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/DevicePortalApiWrapper.cs
zabulus: Not sure, I think it was replicated with what DevicePortal client sends.

src/BenchmarkDotNet.Core/Diagnosers/CompositeDiagnoser.cs
zabulus: The Process class is available as ref nuget package. But there is no aop runtime implementation of Process and this causes app build failure with smth like "unresolved class" error.

src/BenchmarkDotNet.Core/Attributes/Jobs/UapJobAttribute.cs
zabulus: I plan to write separate documentation about how this supposed to work. Process of bringing it all ready not so straightforward that can be described in several lines of summary.

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/DevicePortalApiWrapper.cs
zabulus: On second look, I only don't understand what "+0.5" means.
Though, in API reference I don't see that parameter +"_+ is mandatory.

src/BenchmarkDotNet.Core/Diagnosers/CompositeDiagnoser.cs
zabulus: What if Diagnoser's interface exploit something like IHost interface, that would be implemented in OS-specific terms.

zabulus:
Can't figure out what cause a build to fail:
> [00:02:10]      4>Toolchains\InProcess\BenchmarkActionFactory_Base.cs(125,17): error CS0246: The type or namespace name 'DynamicMethod' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\benchmarkdotnet\src\BenchmarkDotNet.Core\BenchmarkDotNet.Core.csproj]
[00:02:10]      4>Toolchains\InProcess\InProcessHost.cs(27,26): error CS0246: The type or namespace name 'Process' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\benchmarkdotnet\src\BenchmarkDotNet.Core\BenchmarkDotNet.Core.csproj]

Can't find those files

ig-sinicyn:
@zabulus 

It's in `System.Reflection.Emit.Lightweight ` package. AFAIR it is supported on .net native.

zabulus:
Build is passing.
Documentation I've added.

src/BenchmarkDotNet.Core/Helpers/ResourceHelper.cs
```diff
@@ -4,11 +4,24 @@
 
 namespace BenchmarkDotNet.Helpers
 {
-    internal static class ResourceHelper
+    internal class ResourceHelper
     {
-        internal static string LoadTemplate(string name)
+        private readonly string prefix1;
```
adamsitnik: the names prefix1 and prefix2 don't tell anything to the user.
I looked at the code and it seems that there is no need for having two prefixes.

Suggestion: 

- please rename prefix1 to namespacePrefix
- please remove prefix2
- please add a `/// <summary>` comment: the namespace of searched resource file is required to read it from dll

src/BenchmarkDotNet.Core/Attributes/Jobs/UapJobAttribute.cs
adamsitnik: ok, so could you put a link to the documentation in the summary header? You have done a great job and if you  want other people to use 100% of the feature you implemented you need to let them know how to use it properly

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -153,6 +158,11 @@ internal static string GetRuntimeVersion()
             return $"Clr {System.Environment.Version}";
 #elif CORE
             return System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription;
+#elif UAP
+            // System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription throws on UWP release build.
```
adamsitnik: 👍 thanks for leaving the comment!

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/UapBuilder.cs
```diff
@@ -0,0 +1,62 @@
...
+
+                return process.ExitCode <= 0;
+            }
+        }
+
+        private static ProcessStartInfo BuildStartInfo(string workingDirectory, string arguments)
+        {
+            return new ProcessStartInfo
+            {
+                FileName = arguments.Split(' ').First(),
```
adamsitnik: it would be better to have two parameters: commandName and arguments just to avoid the split & skip magic

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/UapBuilder.cs
```diff
@@ -0,0 +1,62 @@
...
+            return BuildResult.Success(generateResult);
+        }
+
+        internal static bool ExecuteCommand(string commandWithArguments, string workingDirectory, TimeSpan timeout)
+        {
+            using (var process = new Process { StartInfo = BuildStartInfo(workingDirectory, commandWithArguments) })
+            {
+                process.Start();
+
+                process.WaitForExit((int)timeout.TotalMilliseconds);
```
adamsitnik: what about the error output of the command when it fails? If we don't print it to the users they will come back to you with every single problem with the new toolchain ;)

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/DevicePortalApiWrapper.cs
adamsitnik: it does not sound very convincing ;)

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/UapBuilder.cs
adamsitnik: you can use `process.StandardError.ReadToEnd();` to obtain the error output

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/DevicePortalApiWrapper.cs
zabulus: OK, I've figured this out. It's replication of JQuery logic, that adds additional parameter "_" to force servers to response with not cached data. I think this could be removed.

adamsitnik:
@zabulus LGTM! Thanks!

AndreyAkinshin:
Hi @zabulus, thanks for the amazing PR. However, I have some major concerns about these changes:
* Why does it support WindowsMobile only? In this case, developers without a device will not be able to use `UapJob`. It's also impossible to write independent integration tests for this feature.
* `UapBenchmarkProgram.txt` shouldn't contain a copy of the main template, both toolchains (Default and Uap) should share the same core template somehow. We already have two implementations of this template (Default and InProcess) which is already hard to maintain. And it would be really hard to maintain 3 of them (especially for people without a device who want to make some changes in the template).
* We already have some complaints from our users about the packages folder size. So, I don't want to add additional dependencies in the main package. I guess only a few percents of our users will actually use remote Windows Phone benchmarking; other people will be unhappy because of many excess uap10 dependencies. So, I think that the main BenchmarkDotNet package shouldn't target uap10; all the UapToolchain code should be in a separate package.
* The amount of the `#if !UAP` directives scare me. It's really hard to make some source code of modification when you have a lot of such directives. Probably we could do a refactoring and introduce a set of platform-specific packages (e.g. `BenchmarkDotNet.Core.NetStandard`, `BenchmarkDotNet.Core.NetCoreApp`, `BenchmarkDotNet.Core.Full`, `BenchmarkDotNet.Core.Uap`). @adamsitnik, what do you think?

zabulus:
> Why does it support WindowsMobile only? In this case, developers without a device will not be able to use UapJob. It's also impossible to write independent integration tests for this feature.

Good concern. Main problem is that each DP variation uses different authentication schemes. For mobile it's PIN, that I've workarounded via cookies, on PC it's basic schema. I plan to add support for running on PC after merge of this. 

> UapBenchmarkProgram.txt shouldn't contain a copy of the main template, both toolchains (Default and Uap) should share the same core template somehow. We already have two implementations of this template (Default and InProcess) which is already hard to maintain. And it would be really hard to maintain 3 of them (especially for people without a device who want to make some changes in the template).

Agree. Do you have any suggestions how to achieve this? I have several ideas, but I think this should be discussed as separate issue.

> We already have some complaints from our users about the packages folder size. So, I don't want to add additional dependencies in the main package. I guess only a few percents of our users will actually use remote Windows Phone benchmarking; other people will be unhappy because of many excess uap10 dependencies. So, I think that the main BenchmarkDotNet package shouldn't target uap10; all the UapToolchain code should be in a separate package.

UapToolchain is packed in the same way Roslyn did. My view is to break BDN in several separate packages similar to: Runner, Target, Toolchain as addins (loaded on reference demand, Job attributes should be moved here as well). In this case end-user could add only that packages that he really need.
But again, this should be discussed as separate issue, and maybe implemented/merged before this PR.

> The amount of the #if !UAP directives scare me. It's really hard to make some source code of modification when you have a lot of such directives. Probably we could do a refactoring and introduce a set of platform-specific packages (e.g. BenchmarkDotNet.Core.NetStandard, BenchmarkDotNet.Core.NetCoreApp, BenchmarkDotNet.Core.Full, BenchmarkDotNet.Core.Uap). 

Everything that excluded via UAP is about Runner part of my above distinction, not Target. Exception is only for Runtime helpers. 
I'd wanted to not include BDN.dll assembly (and therefore not build it as uap10.0 target), but there is code in template:
https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Templates/BenchmarkProgram.txt#L68
```$JobSetDefinition$;```
That is used in EngineParameters for InvokeCount and so on. If we could make this $JobSetDefinition$ portable, or at least not dependable on BDN.dll assembly, then I would be able to remove both dependency and uap10.0 target in BDN.dll and BDN.UAP toolchain assemblies.

**Comment by adamsitnik** [comment]
> I plan to add support for running on PC after merge of this.

@AndreyAkinshin I think that we could merge this, add the PC support when @zabulus implments it and then I could do some refactoring to help with the `#if` and common part of the code templates.

For me the size of our package is not a problem. BenchmarkDotNet is not a lib that goes as the dependency with the product that is using it to the production. 

I will think about breaking into more packages (especially to the part that is independent from runtimes), but I don't have a ready solution now.

**Comment by zabulus** [comment]
> I will think about breaking into more packages (especially to the part that is independent from runtimes), but I don't have a ready solution now.

Could you create issues about breaking and template refactoring (I need more formal problem description), so I could provide ideas I have after the implementation, and we could discuss them?

**Comment by AndreyAkinshin** [comment]
Let me think for a few days about this refactoring, after that, I will create an issue and write down all my thoughts. I want to formalize all the problems that we want to solve and how they can be solved.
Anyway, I think that we are not ready to the merge right now. There are two many reasons:
* As I mentioned before, it's impossible to check how the UAP works without a device, and we don't have independent integration tests which check it automatically. So, I don't want to have such big uncovered feature in the master branch because someone can easily break it and it will be hard to detect the problem. I think that we need UAP-Desktop support at the same merge.
* I think that it will be much easy to refactor all the source code without dozens of additional UAP directives.

@zabulus, once again, thanks for this PR. I believe that it would be a great plugin for BenchmarkDotNet. Hopefully, it will be merged in the nearest future, but we just don't ready to do it right now (because we will have a lot of maintenance problems in this case).

**Comment by zabulus** [comment]
Ok. Meanwhile I'll implement PC DP support, and add integration tests

**Comment by AndreyAkinshin** [comment]
> Ok. Meanwhile I'll implement PC DP support, and add integration tests

Awesome! Looking forward to check it out!

**Comment by zabulus** [comment]
Side-by-side running on different UAP devices:
``` ini

BenchmarkDotNet=v0.10.3.20170309-develop, OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-6500U CPU 2.50GHz, ProcessorCount=4
Frequency=2531249 Hz, Resolution=395.0619 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1586.0
  AcerAspire : .NET Native 4.6.24903.00, 64bit RyuJIT
  Lumia 640  : .NET Native 4.6.24903.00, 32bit RyuJIT


```
 |        Method |        Job |    Runtime |          Mean |     StdErr |     StdDev |
 |-------------- |----------- |----------- |-------------- |----------- |----------- |
 | UtcNowLatency | AcerAspire | AcerAspire |    15.4756 ns |  0.1819 ns |  0.9276 ns |
 |    NowLatency | AcerAspire | AcerAspire |   327.8216 ns |  3.2954 ns | 17.1236 ns |
 | UtcNowLatency |  Lumia 640 |  Lumia 640 |   212.1993 ns |  0.9043 ns |  3.5023 ns |
 |    NowLatency |  Lumia 640 |  Lumia 640 | 7,258.3506 ns | 10.2757 ns | 35.5959 ns |

Following issues of reporting, that could be resolved separately:
1. Wrong detection of platform for ARM. Consider using of https://docs.microsoft.com/de-de/dotnet/core/api/system.runtime.interopservices.runtimeinformation#System_Runtime_InteropServices_RuntimeInformation_ProcessArchitecture
2. Summary header contains processor specification of host application, not target. Consider move this information to Runtime parts.

AndreyAkinshin:
@zabulus,
> Following issues of reporting, that could be resolved separately:

Could you create separate issues?

src/BenchmarkDotNet.Toolchains.Uap/Toolchains/Uap/DevicePortalApiWrapper.cs
```diff
@@ -270,32 +307,40 @@ private static Task AsTask(CancellationToken cancellationToken)
...
+                        sb.Append(str);
+                        if (res.EndOfMessage)
+                        {
+                            var @object = SimpleJson.SimpleJson.DeserializeObject<RootObject>(sb.ToString());
+                            ret.AddRange(@object.Events.Select(x => x.StringMessage));
+                            sb.Clear();
+                        }
                     }
-                    else
+                    catch (AggregateException ex)
```
adamsitnik: you can simplify the code by using "when" keyword:

catch (AggregateException ex) when (ex.InnerException is OperationCanceledException)
{
   // empty on purpose
}

zabulus:
Cause of build fail:
https://github.com/dotnet/coreclr/issues/9694
Though, uap integration works if ignore the error.

zabulus:
Build is green. Integration test script is added. Anything else I should do from here?
@adamsitnik

src/BenchmarkDotNet.Core/BenchmarkDotNet.Core.csproj
```diff
@@ -23,6 +34,17 @@
...
+  <ItemGroup Condition=" '$(TargetFramework)' == 'uap10.0' ">
+    <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="5.3.2" />
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.3.0" />
+    <PackageReference Include="System.Xml.XPath.XmlDocument" Version="4.3.0" />
+    <PackageReference Include="System.Reflection.Extensions" Version="4.3.0" />
+    <PackageReference Include="System.Threading" Value="4.3.0" />
+    <PackageReference Include="System.Reflection" Value="4.3.0" />
+    <PackageReference Include="System.Console" Value="4.3.0" />
+    <PackageReference Include="System.Reflection.Emit.Lightweight" Value="4.3.0" />
```
AndreyAkinshin: Value is not a valid property in PackageReference. You have to use Version instead of Value for these 4 packages.

adamsitnik:
@zabulus big thanks once again! I will work on #420 on the uap branch. Once I am done and @AndreyAkinshin accepts it, we merge it to the master branch.

I should start working on during after 27th of April (I need few hours of really good flow ;) )

AndreyAkinshin:
@adamsitnik, remember that we shouldn't target `uap10.0` in the BenchmarkDotNet.Core package! Our users will not be happy in this case. Uap support is an awesome feature, but it should be activated by installing an additional package.

AndreyAkinshin:
@adamsitnik another note: we should share the BenchmarkProgram template somehow between the Roslyn and Uap toolchains. I'm going to implement multithreading benchmarks one day and it requires changes in the template. It will not be easy because I have to update 2 toolchains (Roslyn and InProcess). In the case of 3 templates, it would be hard to maintain (I also don't have and Windows mobile device, so I will not be able how it works with the new template).

Issue #397, opened 2017-03-22
Right-justify/align number columns for GitHub-flavored Markdown


stevedesmond-ca:
See https://github.github.com/gfm/#delimiter-row: right-justified columns should have their delimiter row ending in a colon (e.g. `|-----:|`) -- otherwise, if not displayed using a fixed-width font, numbers won't be aligned properly.

```
| Not right-justified | Right-justified |
|-------------------- |----------------:|
|              0.1 ms |          0.2 ms |
|            100.5 ms |        100.6 ms |
```

Becomes:

| Not right-justified | Right-justified |
|-------------------- |----------------:|
|              0.1 ms |          0.2 ms |
|            100.5 ms |        100.6 ms |

PR #396, opened 2017-03-22
Allow users to pick, show and hide measurement units in the reports and exports.

AmadeusW:
Introduces `ISummaryStyle` which can be added to the `IConfig` and to participating exporters (currently `CsvExporter`). 

The style controls whether units appear in headers, whether they appear in the data and what are the units. If user did not specify units in the style, `SummaryTable` will set recommended units based on report results. 

Participating implementations of `IColumn` (currently `StatisticColumn` and two `AllocationColumn`) render their headers and values based on `ISummaryStyle`

Closes #146 - Ability to specify units / easier comparison
Closes #118 - Raw data in CSV exports
Resolves a TODO in Windows.MemoryDiagnoser

This sample program
```
            var config = ManualConfig.Create(DefaultConfig.Instance);
            config.Set(new Reports.SummaryStyle()
            {
                PrintUnitsInHeader = false,
                PrintUnitsInContent = true,
                TimeUnit = Horology.TimeUnit.Second,
                SizeUnit = Columns.SizeUnit.B
            });
            config.Add(new MemoryDiagnoser());
            config.Add(new CsvExporter(
                CsvSeparator.CurrentCulture,
                new Reports.SummaryStyle()
                {
                    PrintUnitsInHeader = true,
                    PrintUnitsInContent = false,
                    TimeUnit = Horology.TimeUnit.Millisecond,
                    SizeUnit = Columns.SizeUnit.kB
                }
            ));
            BenchmarkRunner.Run<Intro.IntroColumns>(config);
```
produces **.md** output:

 |    Method |     Mean |   StdErr |   StdDev |      Min |      Max | Allocated |
 |---------- |--------- |--------- |--------- |--------- |--------- |---------- |
 | Benchmark | 0.0305 s | 0.0066 s | 0.0209 s | 0.0101 s | 0.0507 s |    1546 B |

and **.csv** output:

image

PR #398, opened 2017-03-22
Markdown exporter makes all columns after the first right-justified

stevedesmond-ca:
This is a potential quick-fix for #397.

As far as I can tell, at least in all the results I have seen, all columns after the first are numeric, and therefore should be right-justified.

Alternatively, the fixed-width versions of these tables have every column right-justified, so if we want to go that route, the inline ternary could be removed and all columns could be delimited by `:|`.

A more complete solution would be for columns to have an `IsNumeric` property, or something like it, that would help identify intended alignment in display tables.

Let me know which you think would be best!

### DIFF
--- a/src/BenchmarkDotNet.Core/Exporters/MarkdownExporter.cs
+++ b/src/BenchmarkDotNet.Core/Exporters/MarkdownExporter.cs
@@ -139,7 +139,7 @@ namespace BenchmarkDotNet.Exporters
                     logger.Write(tableHeaderSeparator);
                 }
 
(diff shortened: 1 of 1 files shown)


AmadeusW:
> A more complete solution would be for columns to have an IsNumeric property, or something like it, that would help identify intended alignment in display tables.

FYI I've introduced `QuantityType` property on `IColumn` in PR #396

stevedesmond-ca:
Perfect -- once that gets merged in, I can rebase this and use `QuantityType` as the differentiator.

AndreyAkinshin:
@stevedesmond-ca, thanks for the PR. However, it doesn't solve the problem in general case: we don't know in advance the column order: user can remove some default columns or add custom columns. Let's discuss a better solution for #398 after #396 will be merged.

PR #400, opened 2017-03-23
fix spelling error: misspredict => mispredict

stevedesmond-ca:
I was alerted to this by @AmadeusW's most recent comment in #396


AndreyAkinshin:
@stevedesmond-ca, thanks for the PR!

PR #405, opened 2017-03-29
Make RoslynToolchain types public

cdmihai:
I need to extend the Roslyn Generator and Builder with custom build steps for our codebase, so I made the types public. Now they have the same visibility as the dotnet CLI generators and builders.

Also considered more general solutions that would allow API users to extend any toolchain without caring what it is:
1. make the GeneratorBase's protected hook methods public, so I can then decorate a GeneratorBase with my own hook implementations which would call the inner hooks and then do something extra
2. replace GeneratorBases's protected hook methods with public delegates, thus transforming inheritance to composition. API users could then start with some default compositions (roslyn / CLI) and add to the existing delegates.

In the end I went with this implementation since it's small, simple, and does not break backcompat.

### DIFF
--- a/src/BenchmarkDotNet.Toolchains.Roslyn/Builder.cs
+++ b/src/BenchmarkDotNet.Toolchains.Roslyn/Builder.cs
@@ -6,16 +6,19 @@ using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.Results;
+using JetBrains.Annotations;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using OurPlatform = BenchmarkDotNet.Environments.Platform;
 
 namespace BenchmarkDotNet.Toolchains.Roslyn
 {
+    [PublicAPI]
     public class Builder : IBuilder
     {
         private static readonly Lazy<AssemblyMetadata[]> FrameworkAssembliesMetadata = new Lazy<AssemblyMetadata[]>(GetFrameworkAssembliesMetadata);
 
+        [PublicAPI]
         public BuildResult Build(GenerateResult generateResult, ILogger logger, Benchmark benchmark, IResolver resolver)
         {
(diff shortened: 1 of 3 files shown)


cdmihai:
In the end I went with decorating IToolChain and IGenerator. Turned out I do not need the extra resolution of the GeneratorBase hook methods. Feel free to either close or merge this PR.

AndreyAkinshin:
@adamsitnik, could you take a look?

src/BenchmarkDotNet.Toolchains.Roslyn/Builder.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace BenchmarkDotNet.Toolchains.Roslyn
 {
-    internal class Builder : IBuilder
+    public class Builder : IBuilder
```
AndreyAkinshin: @cdmihai, if you want to make something a part of public API, please mark it with `[PublicAPI]`.

cdmihai:
Negating my previous comment saying I do not need the hooks. :) 
After using my IToolChainWrapper, it turns out it kinda works, but it would be even better if I had access to the hooks.

cdmihai:
I want to write some perf tests to hopefully detect when we regress. I am writing a Config that, given multiple versions of our codebase (in form of either git commits, repo paths, or already built bin dir paths), fetches the commits, builds the repos, and creates a new job for each codebase variant. Each job has an IToolChain wrapper (does not care what type of toolchain, for now) which during generation copies over the freshly built binaries in the `ArtifactsPaths.BinariesDirectoryPath`.

Usage scenarios / features:
- CI. Each PR would run a set of small, fast tests. Compares branch head and master. Or maybe hardcode a "last known good perf" commit and test against that to avoid slow perf drifts.
- Local Dev analysis: given arbitrary commits, run some collection of tests on them to catch drifting perf 
  - last N github release commits
  - last N commits on master

After I'm done with my concrete tooling, I could look at what abstractions to extract out. Probably everyone will have to write down the logic required to build their repos, and maybe how to copy the binaries into the perf test bin path. On the other hand the git fetchers, repo abstractions, and job injection should be common.

adamsitnik:
@cdmihai sounds very interesting! Please let me know if you need some help

PR #407, opened 2017-04-01
Propagate benchmark to hooks

cdmihai:
`Cleanup` and `CopyRequiredFiles` did not receive the benchmark. Now they do. Unfortunately this is a breaking change, but I think it's worth it, since now all hook signatures are uniform (these were the only two hooks not receiving the benchmark object).

In my case, I am copying in different versions of a benchmark's dependencies, so I need to move the `BuildArtifactsDirectoryPath` outside of the benchmark's bindir, where the `BuildArtifactsDirectoryPath` defaults to for the Roslyn Generator, and then copy everything from the benchmark bindir there. This means that in my case, both `Cleanup` and `CopyRequiredFiles` hooks need to know where the benchmark type is located.


adamsitnik:
thanks!

PR #413, opened 2017-04-07
Allows picking showing and hiding measurement units in reports

### LINKED ISSUE #118: Raw data in CSV reports
CSV files is target for tools. So values in columns suffixed  with unit (like `ns`) is an inconvenience.
Can the unit be moved to column names in the header?

### LINKED ISSUE #146: Ability to specify units / easier comparison
I noticed that a report I've generated uses `us` and `ns`. While I can multiply microseconds by 1000 to get nanoseconds, it'd be easier if I could specify the units to use in the report.

I'm currently not able to install the R package to view the plots so I'm not sure if they're better.

I'm using BenchmarkDotNet 0.9.4 from NuGet.

AmadeusW:
Introduces ISummaryStyle which can be added to the IConfig and to participating exporters (currently CsvExporter).

The style controls whether units appear in headers, whether they appear in the data and what are the units. If user did not specify units in the style, SummaryTable will set recommended units based on report results.

Participating implementations of IColumn (currently StatisticColumn and two AllocationColumn) render their headers and values based on ISummaryStyle

Closes #146 - Ability to specify units / easier comparison
Closes #118 - Raw data in CSV exports
Resolves a TODO in Windows.MemoryDiagnoser

This PR is a squashed version of #396

### DIFF
--- a/src/BenchmarkDotNet.Core/Columns/BaselineScaledColumn.cs
+++ b/src/BenchmarkDotNet.Core/Columns/BaselineScaledColumn.cs
@@ -91,6 +91,8 @@ namespace BenchmarkDotNet.Columns
         public bool AlwaysShow => true;
         public ColumnCategory Category => ColumnCategory.Baseline;
         public int PriorityInCategory => (int) Kind;
+        public UnitType UnitType => UnitType.Dimensionless;
+        public string GetValue(Summary summary, Benchmark benchmark, ISummaryStyle style) => GetValue(summary, benchmark);
         public override string ToString() => ColumnName;
         public bool IsDefault(Summary summary, Benchmark benchmark) => false;
     }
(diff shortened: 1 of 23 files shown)


AndreyAkinshin:
@AmadeusW. you shouldn't create another PR for the squashed version. It's better to squash the `summaryStyle` branch and run `git push -f`; https://github.com/dotnet/BenchmarkDotNet/pull/396 will be automatically updated. It's better because it allows keeping the PR source code and the discussion in the same place. By the way, GitHub allows squashing on merge which is also a great feature.
It seems that we have some confilct in the `ManualConfig` file (probably because of recent @adamsitnik  fixes). Could you resolve it?

AmadeusW:
Ok I'll go to #396, fix the conflict and you should be able to squash-merge, right?

Issue #377, opened 2017-03-01
Simple API for path customization of BenchmarkDotNet.Artifact and job folders


AndreyAkinshin:
Sometimes we have troubles with `UnauthorizedAccessException` (we don't have access to the current directories) or long paths. I think, it would be great to provide a nice API which allows to customzie BenchmarkDotNet working directory (which will contain all the generated files and folders).

adamsitnik:
> which will contain all the generated files and folders

@AndreyAkinshin do you mean only the results (`.csc`, `.html` and etc) and logs (`.log`)? then such change is easy

but if you also mean (`.exe`, `.notcs` and `.config`) then not so easy because today we use `benchmark.Target.Type.GetTypeInfo().Assembly.Location` where something (like msbuild) has produced the exe which makes assembly loading very easy for us (everything is in the current directory). If we want to produce the .exe somewhere else then most probably we would need to copy some files there or play with the CurrentDirectory of the benchmark process

AndreyAkinshin:
Hmm. In the perfect world, I want to have an ability to specify a folder for *all* generated files. But we can start with the artifacts folder.

daveaglick:
I hit this trying to set up some benchmarks in LINQPad. Specifically, it's trying to write to "C:\Program Files (x86)\LINQPad5\BenchmarkDotNet.Artifacts" which is locked down since it's in the "Program Files" folder. I can work around it, but a simple API to direct file writes elsewhere would be helpful.

jhorv:
I hit upon this problem in LINQPad.  In lieu of a BenchmarkDotNet API for configuring, I found that setting `System.Environment.CurrentDirectory` to another existing directory will change the benchmark output location.  Below is some general code for setting this to the same location and name as your LINQPad query:

```
void Main()
{
	var logPath = GetBenchmarkOutputPath();
	Directory.CreateDirectory(logPath);
	Environment.CurrentDirectory = logPath;

	// ...
}

static string GetBenchmarkOutputPath()
{
	var currentQueryDirectory = Path.GetDirectoryName(LINQPad.Util.CurrentQueryPath);
	var currentQueryName = Path.GetFileNameWithoutExtension(LINQPad.Util.CurrentQueryPath);
	var result = Path.Combine(currentQueryDirectory, currentQueryName);
	result += "_" + DateTime.Now.ToString("yyyyMMdd-HHmmss");
	return result;
}

```

seesharper:
[...]

PR #416, opened 2017-04-09
Exports file to temporary location if target is locked

### LINKED ISSUE #414: BDN crashes with IOException when it can't save a report file
Version used: **0.10.3**

To repro:
- Create a benchmark with `config.Add(CsvExporter.Default);`
- Run benchmark
- Open .csv in Excel. Excel gets a lock on the file
- Run benchmark again
- BDN crashes with `System.IO.IOException`

Potential resolution:
- Ignore the error and continue
- Try to save under another name, for example `MyTest-report (2).csv`. What's the preferred naming convention?

AmadeusW:
Fixes #414 
When a target file exists and is locked, currently BDN crashes and some of the data is lost.
This PR makes BDN export data under randomly-unique name to a temporary location and print a message to the log:
image
```
// * Export *
Could not overwrite file C:\src\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Artifacts\results\Algo_BitCount-report.csv. Exporting to C:\Users\amadeusz\AppData\Local\Temp\uvkf5nub-Algo_BitCount-report.csv
  C:\Users\amadeusz\AppData\Local\Temp\uvkf5nub-Algo_BitCount-report.csv
  BenchmarkDotNet.Artifacts\results\Algo_BitCount-report-github.md
  BenchmarkDotNet.Artifacts\results\Algo_BitCount-report.html
```

I'm not fully convinced that this sufficiently notifies user that the report in expected location is out of date, but I think this is better than data loss after a long running benchmark.

### DIFF
--- a/src/BenchmarkDotNet.Core/Exporters/ExporterBase.cs
+++ b/src/BenchmarkDotNet.Core/Exporters/ExporterBase.cs
@@ -26,7 +26,8 @@ namespace BenchmarkDotNet.Exporters
                 }
                 catch (IOException)
                 {
-                    var alternativeFilePath = Path.GetTempPath() + Path.GetFileNameWithoutExtension(Path.GetRandomFileName()) + $"-{summary.Title}-{FileCaption}{FileNameSuffix}.{FileExtension}";
+                    var uniqueString = System.DateTime.Now.ToString("yyyyMMdd-HHmmss");
+                    var alternativeFilePath = $"{Path.Combine(summary.ResultsDirectoryPath, summary.Title)}-{FileCaption}{FileNameSuffix}-{uniqueString}.{FileExtension}";
                     consoleLogger.WriteLineError($"Could not overwrite file {filePath}. Exporting to {alternativeFilePath}");
                     filePath = alternativeFilePath;
                 }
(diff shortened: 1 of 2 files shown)


adamsitnik:
I like the idea, but I think that we should save the file to `summary.ResultsDirectoryPath`, not `Path.GetTempPath()`.

Perhaps the name should be time-dependent by default? So our users could compare the results easily.

AmadeusW:
How about this? It yields a unique representation and follows [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

image

Is the message superfluous, and we could remove the second sentence?

```
// * Export *
Could not overwrite file C:\src\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Artifacts\results\Algo_BitCount-report.csv. Exporting to C:\src\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Artifacts\results\Algo_BitCount-report-20170409-192535.csv
  BenchmarkDotNet.Artifacts\results\Algo_BitCount-report-20170409-192535.csv
  BenchmarkDotNet.Artifacts\results\Algo_BitCount-report-github.md
  BenchmarkDotNet.Artifacts\results\Algo_BitCount-report.html
```

adamsitnik:
@AmadeusW I really like it!

@AndreyAkinshin what is your opinion on that?

AndreyAkinshin:
It's a common problem, so I really like that now we will have a solution for such cases.
@AmadeusW, thanks for the PR! Could you also write an integration test for it? It should be easy to write: just create `*-report*.csv` in advance, lock it (e.g. start to read from it), run a benchmark, check that we created this report with another name. I'm asking about it because we are going to implement other file creation issues (see a related issue: https://github.com/dotnet/BenchmarkDotNet/issues/377) and I want to be sure that future fixes will not break this one.

AmadeusW:
For sure

adamsitnik:
@AmadeusW once again big thanks!

PR #421, opened 2017-04-13
MarkdownExporter right-justifies numeric columns

### LINKED ISSUE #397: Right-justify/align number columns for GitHub-flavored Markdown

stevedesmond-ca:
Now that #396 is complete, I've made a more complete pass at #398, to right-justify numeric columns in the `MarkdownExporter`, which would resolve #397.

1. `IColumn` has a new `IsNumeric` property, which is set (what I believe to be) appropriately in all its implementations.
1. `SummaryTableColumn` uses `IsNumeric` to set a new `Justify` property to either `Left` or `Right`
1. `MarkdownExporter` uses `Justify` to determine whether to use `:` or ` ` as the justification indicator

### DIFF
--- a/src/BenchmarkDotNet.Core/Exporters/MarkdownExporter.cs
+++ b/src/BenchmarkDotNet.Core/Exporters/MarkdownExporter.cs
@@ -139,7 +139,7 @@ namespace BenchmarkDotNet.Exporters
                     logger.Write(tableHeaderSeparator);
                 }
 
-                logger.WriteLineStatistic(string.Join("", table.Columns.Where(c => c.NeedToShow).Select(c => new string('-', c.Width) + " |")));
+                logger.WriteLineStatistic(string.Join("", table.Columns.Where(c => c.NeedToShow).Select(getDelimiterLineForColumn)));
             }
             var rowCounter = 0;
             var highlightRow = false;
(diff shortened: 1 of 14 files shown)


stevedesmond-ca:
my bad, apparently missed some implementations

src/BenchmarkDotNet.Core/Exporters/MarkdownExporter.cs
```diff
@@ -160,5 +160,11 @@ private void PrintTable(SummaryTable table, ILogger logger)
                 rowCounter++;
             }
         }
+
+        private static string getDelimiterLineForColumn(SummaryTable.SummaryTableColumn column)
+        {
+            var justifier = column.Justify == SummaryTable.SummaryTableColumn.TextJustification.Left ? " " : ":";
```
AndreyAkinshin: Could we use a `switch` statement here? With the `? :` operator, it would be tricky to find a bug when we decided to add another value in the enum (e.g. `Center`).

src/BenchmarkDotNet.Core/Reports/SummaryTable.cs
```diff
@@ -80,18 +80,27 @@ public class SummaryTableColumn
             public bool NeedToShow { get; }
             public int Width { get; }
             public bool IsDefault { get; }
+            public TextJustification Justify { get; }
 
-            public SummaryTableColumn(SummaryTable table, int index, bool alwaysShow)
+            public SummaryTableColumn(SummaryTable table, int index, IColumn column)
```
AndreyAkinshin: I like the new signature, it looks much better.

AndreyAkinshin:
@stevedesmond-ca, thanks!

PR #423, opened 2017-04-16
Updated Microsoft.Net.Test.Sdk

alinasmirnova:
Approval tests doesn't work with preview version — it can't get source path correctly.


AndreyAkinshin:
@alinasmirnova, thanks!

PR #430, opened 2017-04-21
Fix unix OS detection.

mfilippov:
Mono returns PlatformID.Unix on Linux and macOS. It's not enough for detect OS version. I use native call uname from libc for fix it.

### DIFF
--- a/src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
@@ -46,7 +46,8 @@ namespace BenchmarkDotNet.Portability
         internal static bool IsLinux()
         {
 #if CLASSIC
-            return System.Environment.OSVersion.Platform == PlatformID.Unix;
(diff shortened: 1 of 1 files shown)


src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -263,5 +265,31 @@ public JitModule(string name, string version)
                 Version = version;
             }
         }
+
+        [DllImport("libc", SetLastError=true)]
+        private static extern int uname(IntPtr buf);
+
+        private static string GetSysnameFromUname()
+        {
+          var buf = IntPtr.Zero;
```
AndreyAkinshin: @mfilippov, could you increase the indent here?

mfilippov: fixed

mfilippov:
Build failed, looks like NuGet server problem not depended with my change.

AndreyAkinshin:
@mfilippov, thanks!

PR #436, opened 2017-04-24
Fix false allocations detection

ig-sinicyn:
` Host.BeforeMainRun();` should run _before_ gc stats init as some host implementations (in-process, as example) may allocate during notification.

**UPD:** Allocation in InProcessHost is triggered by [WriteLine method](https://github.com/dotnet/BenchmarkDotNet/blob/610f3b72808bd76837facf6cd75acf70a30eb772/src/BenchmarkDotNet.Core/Toolchains/InProcess/InProcessHost.cs#L84):
```cs
                case HostSignal.BeforeMainRun:
                    diagnoser?.BeforeMainRun(diagnoserActionParameters);
                    WriteLine(Engine.Signals.BeforeMainRun);
                    break;
```


AndreyAkinshin:
@adamsitnik, could you take a look?

adamsitnik:
good catch!

Issue #376, opened 2017-02-28
Pick better background colors for output


davidfowl:
![image](https://cloud.githubusercontent.com/assets/95136/23394968/d3b3a35e-fd40-11e6-8719-2247fe27b76e.png)

AndreyAkinshin:
Makes sense. Usually, I use `cmd`. In PowerShell it doesn't look great. Any color recommendation?

davidfowl:
Maybe always set the background color as well as foreground color? OR do something fancy like this https://github.com/Eilon/ProjectKIssueList/blob/5040df74b41dcd4f3bbf0470f851a6b924b0af06/src/Hubbup.Web/Utils/ColorMath.cs 😄

AndreyAkinshin:
Looks interesting, I like an idea of adaptive colors. Is it possible to specify custom ConsoleColor for both `net45` and `netcoreapp1.1` (it should be xplat and work on Mono). Now we are using `Console.ForegroundColor` which should have values from [ConsoleColor](https://msdn.microsoft.com/en-us/library/system.consolecolor(v=vs.110).aspx) (this set is very poorly =( ).

AndreyAkinshin:
I found this awesome library: [Colorful.Console](http://colorfulconsole.com/). Let me check how it works on different operation systems.

AndreyAkinshin:
Colorful.Console doesn't work on Linux+Mono. =(
@davidfowl, will it be ok if we introduce advanced color schemes only for Windows?

mattwarren:
BTW, even with `cmd` the output from the 'Outliers' is hard to read, e.g.

![image](https://cloud.githubusercontent.com/assets/157298/23400273/5c050022-fd9b-11e6-8081-537a0e1b0268.png)

mattwarren:
FWIW, a while back I wrote some code that shows all the possible foreground/background colours when running on Windows, here's the screenshot:

![clr or net - console colours](https://cloud.githubusercontent.com/assets/157298/23400332/90791334-fd9b-11e6-8574-572fb74712e3.png)

AndreyAkinshin:
Ok, I changed the hint color to DarkCyan, now it looks much better.

adamsitnik:
I am closing since Andrey has already changed the colors

PR #440, opened 2017-05-03
Make our Core package Runtime-independent

adamsitnik:
Changes:

1. I have renamed BenchmarkDotNet.Core to BenchmarkDotNet.Runtime
2. BenchmarkDotNet.Runtime targets now .NET Standard 1.3 and is runtime-independent
3. I have introduced BenchmarkDotNet.Runtime.Core/Uap/Classic which contain runtime-specific implementations of the stuff required to run
4. I have made BenchmarkDotNet targeting .NET 4.6 and .NET Core 1.1 only, not Uap
5. I have renamed BenchmarkDotNet.Toolchains.Uap to BenchmarkDotNet.Uap and implemented Runner for it.
6. I moved common part of the generated file to single txt file, so we have less things to sync now

Pros:

1. When users download BenchmarkDotNet package they don't have Uap support, but the package is smaller. (NuGet downloads all the files and dependencies for the package, no matter what you want to use)
2. No #if defines in our main package
3. Single dedicated package for UAP

Cons:

1. The dynamic loading of runtime-specific implementation is implemented in reflection-heavy way. We need to load B from A, when A is not dependent on B, but B is on A. Ofc we should change architecture to avoid that but with current common usage of static fields this is nearly impossible...

Workarounds:

1. F# tools don't support  .NET Standard yet ([issue](https://github.com/Microsoft/visualfsharp/issues/2461)) so I re-added .NET 4.6 target to our main package just to keep F# support.

@AndreyAkinshin @zabulus @ig-sinicyn I would love to hear your comments


adamsitnik:
@zabulus I did not test the Uap yet, I had some troubles running it.

Could you fetch the refactor branch and give it a try?

Also I have noticed that when you generate the .csproj file you reference few assemblies by path. The dll that contains the benchmarks, BenchmarkDotNet.dll and BenchmarkDotNet.Core.dll.

What if you would reference .csproj that contains the benchmarks? I do sth like this in our CsProjGenerator. The advantage is that I get all the settings, references and stuff from this project, I don't need to worry about the paths etc. Would that work with Uap? If so I think that it could simplify the toolchain. What do you think?

mattwarren:
@adamsitnik this is great work, I've been asked a few times about the *amount* of stuff that BenchmarkDotNet downloads via NuGet, so anything that cuts that down is v. useful!!

> The dynamic loading of runtime-specific implementation is implemented in reflection-heavy way. We need to load B from A, when A is not dependent on B, but B is on A. Ofc we should change architecture to avoid that but with current common usage of static fields this is nearly impossible...

I've not looked at what's happening, so I might be wrong, but can [the `AppDomain.AssemblyResolve` event](https://msdn.microsoft.com/en-us/library/system.appdomain.assemblyresolve(v=vs.110).aspx) help at all? (I have no idea if CoreCLR supports it, which might be a problem!)

zabulus:
According to this:
https://apisof.net/catalog/System.AppDomain.AssemblyResolve
It would be brought back only in netstandart 2.0.

adamsitnik:
@zabulus I have fixed the problem with loading assembly with [this commit](https://github.com/dotnet/BenchmarkDotNet/commit/6566f6533d2c2b548d0c5e96a1d0f976633c3f15) could you check it once again?

zabulus:
> @zabulus I have fixed the problem with loading assembly with this commit could you check it once again

Looks like it works. Do you still have a problem with ```runUapTests.cmd```?

src/BenchmarkDotNet.Runtime/Portability/RuntimeInformation.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        internal const string Unknown = "?";
+        internal const string DebugConfigurationName = "DEBUG";
+        internal const string ReleaseConfigurationName = "RELEASE";
+
+        public virtual bool IsMono => false;
+
+        public abstract bool IsWindows { get; }
+        public abstract bool IsLinux { get; }
+        public abstract bool IsMac { get; }
```
AndreyAkinshin: I think it makes sense to introduce own `OS` enum instead of 3 boolean properties.

src/BenchmarkDotNet.Runtime/Portability/RuntimeInformation.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace BenchmarkDotNet.Portability
+{
+    public abstract class RuntimeInformation
+    {
+        internal const string Unknown = "?";
+        internal const string DebugConfigurationName = "DEBUG";
+        internal const string ReleaseConfigurationName = "RELEASE";
+
+        public virtual bool IsMono => false;
```
AndreyAkinshin: Probably (I'm not sure), it would be better to use another `Runtime` enum here (which can be `Mono`, `Clr`, or `Core`).

adamsitnik:
@AndreyAkinshin what are your general thoughts on this PR?

AndreyAkinshin:
@adamsitnik, I like it! You did an awesome job!
This PR includes a lot of changes, so I need some time to review it. Will back to you with detailed feedback in a few days!

AndreyAkinshin:
@adamsitnik, by the way, could you introduce another project for classic-only `BenchmarkDotNet.Tests`? It will allow writing ApprovalTests (which doesn't support Core) without another `#if`.

src/BenchmarkDotNet.Runtime/Portability/RuntimeInformation.cs
adamsitnik: You are right. We even have such property: CurrentRuntime. I can add IsMono as an extension method.

adamsitnik:
@AndreyAkinshin why `ApprovalTests` are not running for .Core? do they have some dependency to thing available for classic .NET only?

If so I can move these tests to separate project, but I am not sure if it will help. It's easier to forget about running the tests if they are separated.

AndreyAkinshin:
> why ApprovalTests are not running for .Core? do they have some dependency to thing available for classic .NET only?

Here is the issue: https://github.com/approvals/ApprovalTests.Net/issues/162

> If so I can move these tests to separate project, but I am not sure if it will help. It's easier to forget about running the tests if they are separated.

Yeah, I think, we need a separate project, it will be easily to maintain. Our CI will not fotget about these tests. =)

src/BenchmarkDotNet.Runtime.Classic/InProcessExecutor.cs
```diff
@@ -101,13 +101,13 @@ private int ExecuteCore(IHost host, Benchmark benchmark, ILogger logger)
             var affinity = benchmark.Job.ResolveValueAsNullable(EnvMode.AffinityCharacteristic);
             try
             {
-                process.TrySetPriority(ProcessPriorityClass.High, logger);
+                ProcessExtensions.TrySetPriority(process, ProcessPriorityClass.High, logger);
 #if CLASSIC
```
AndreyAkinshin: Could we completely remove the `#if CLASSIC` line here and hide them inside some portability API?

adamsitnik: @AndreyAkinshin most probably yes, I will try it

AndreyAkinshin:
A general comment about `ServicesProvider`: could we make it more flexibale? Right now it has hardcoded list of misc properties (which is grouped in the `Services` class) which is not a good idea. Maybe we can came up with a style like this:
```cs
static class RuntimeInformationProvider {
  static RuntimeInformation Instance => ServicesProvider.GetService<RuntimeInformation>();
}
```
Here ServicesProvider is a some kind of DI container.
A use case from real life: I want to write a package (let's say `BenchmarkDotNet.Loggers.Windows`) which allows better output coloring on Windows (see https://github.com/dotnet/BenchmarkDotNet/issues/376#issuecomment-282964316) via [Colorful.Console](http://colorfulconsole.com/). It should be a separate package because it has an extra dependency and it works only on Windows. If user install this package and use BDN on Windows, it should override default `ConsoleLogger` in `DefaultConfig`. Thus, I need an extension point which should be used in core `BenchmarkDotNet.Runtime` and implemented (or not implemented) in an optional package.
It's just one example, I believe we will have a lot of such cases in the future (in fact, I already have several ideas). And it would be great, if the new architecture will support it.

adamsitnik:
@AndreyAkinshin I see your point, I will implement it

AndreyAkinshin:
@adamsitnik, awesome, thanks!

adamsitnik:
@AndreyAkinshin I have updated the code. Please take a look ;)

I did not separate the .Test.Classic project yet, I want to do this after we merge refactor to UAP and master to UAP, before UAP to master ;)

AndreyAkinshin:
@adamsitnik, now it's much better! A few more comments/questions:
* Currently appveyor build is red.
* Now `BenchmarkDotNet.Runtime` targets `netstandard1.3;net46`. Why not only `netstandard1.3`?
* Why do we still have the `BenchmarkRunner`, `BenchmarkSwitcher`, and `ToolchainExtensions` packages in the `BenchmarkDotNet` package. Could we move it to the `BenchmarkDotNet.Runtime` package?
* Let's say I want to create a classic application which targets net46 only. Is it possible to install BenchmarkDotNet without netcoreapp stuff?

ig-sinicyn:
@adamsitnik, @AndreyAkinshin 

+1000 on suggestion of having no .net core dependencies when targeting full .net FW, at least for code referencing just `BenchmarkDotNet.Core` package (or to what it was renamed). I hope this scenario will not break with BDN vNext?

adamsitnik:
@AndreyAkinshin 

> Currently appveyor build is red.

One of the builds failed because it lasted more than hour. I think that the reason was that this particular build agent did not have the UAP packages preinstalled. The other build failed for no reason, I suspect random failure. The builds for merging refactor into uap branch have passed, so I would ignore these two failures.

> Now BenchmarkDotNet.Runtime targets netstandard1.3;net46. Why not only netstandard1.3?

F# tools don't support  .NET Standard yet ([issue](https://github.com/Microsoft/visualfsharp/issues/2461)) so I re-added .NET 4.6 target to our main package just to keep F# support.

> Why do we still have the BenchmarkRunner, BenchmarkSwitcher, and ToolchainExtensions packages in the BenchmarkDotNet package. Could we move it to the BenchmarkDotNet.Runtime package?

`BenchmarkDotNet.Runtime` has no idea about the existence of our RoslynToolchain, which is kept in separate package. This was done mostly because Roslyn prior to 2.0 did not support .NET Core, so we were keeping it in a separate package. Now we could move our RoslynToolchain to `BenchmarkDotNet.Runtime`, Roslyn dependency would be mandatory. Why it's good to avoid that: some people (like @ig-sinicyn ) don't want to have this heavy dependency due to the size. Roslyn team could not benchmark the code they generate with their latest build. Now the Roslyn dependency is optional, so they can build their custom toolchain and benchmark the code they generate.

> Let's say I want to create a classic application which targets net46 only. Is it possible to install BenchmarkDotNet without netcoreapp stuff?

It's NuGet client limitation. If somebody wants to workaround that, then he or she needs to install `BenchmarkDotNet.Toolchains.Roslyn` and call `BenchmarkRunnerCore` like this `BenchmarkRunnerCore(benchmarks, config, () => new RoslynToolchain());` We could create another package for this, but I am not sure if it would make any sense.

> +1000 on suggestion of having no .net core dependencies when targeting full .net FW, at least for code referencing just BenchmarkDotNet.Core package (or to what it was renamed). I hope this scenario will not break with BDN vNext?

The current name is `BenchmarkDotNet.Runtime`, it has only .NET Standard 1.3 dependencies. @ig-sinicyn if you have custom toolchain which we discussed recently, then you just need to install `BenchmarkDotNet.Runtime.Classic` and call `BenchmarkRunnerCore` like this `BenchmarkRunnerCore(benchmarks, config, () => new CustomToolchain());` In that case you have no .NET Core and no Roslyn dependencies.

I know that all of that might sound crazy, but I believe that it allows for all scenarios. At least this is the best I could get.

ig-sinicyn:
@adamsitnik  Thanks! That's exactly what I've expected:)

AndreyAkinshin:
@adamsitnik can we close this PR (because it's obsolete)?

Issue #157, opened 2016-05-04
Implement export to xml


mattwarren:
**Note:** this should be done without introducing a .dll or Nuget dependency to the core of BenchmarkDotNet. For example the [Json export](https://github.com/PerfDotNet/BenchmarkDotNet/issues/84) was done with [SimpleJson](https://github.com/facebook-csharp-sdk/simple-json/blob/master/src/SimpleJson/SimpleJson.cs) as it is a single `.cs` file

Teknikaali:
I can do this.

The previously mentioned [Json export](https://github.com/dotnet/BenchmarkDotNet/issues/84) seems to be a good example to follow how to implement this.

Some steps I came up with after exploring the repo:
- [ ] Figure out an alternative way to serialize **Summary**-object to XML without **XmlSerializer**
- [ ] Create the new **XmlExporter**-class by inheriting from **ExporterBase**
- [ ] Include **XmlExporter** in **ApprovalTests**
- [ ] Define how the XML output should look like

@mattwarren @AndreyAkinshin are there any other things I should take into account before diving in too deep?

Taking the lower level path by using **XmlWriter** rather than the "now-unavailable-in-netcoreapp1.1" **XmlSerializer** could be a viable way here.

Currently no XML-alternative exists for [SimpleJson](https://github.com/facebook-csharp-sdk/simple-json/blob/master/src/SimpleJson/SimpleJson.cs) that I'd know of, so that leaves me to create a strict XML serialization of the **Summary**-type. I'll make the serialization part "easy to change", if we'll some day see re-addition of **XmlSerializer**-class to .NET Core without the need of the extra NuGet dependency.

I start setting up the dev environment!

adamsitnik:
> are there any other things I should take into account before diving in too deep?

@Teknikaali I am currently working on making our .Core package .NET Standard compatible, so if you have to add any new dependency please make sure that it has NuGet package that targets .NET Standard 1.3 or lower

good luck!

Teknikaali:
@adamsitnik Thanks for the heads-up!

The only NuGet package that (_in theory_) could make the implementation a bit easier, would be [System.Xml.XmlSerializer](https://www.nuget.org/packages/System.Xml.XmlSerializer/4.3.0). It should support .NET Standard 1.3 looking at its dependencies list if I'm not all mistaken.

But there's a small catch why the `XmlWriter`-path seems a bit better than using the `XmlSerializer` NuGet package:
[...]

PR #451, opened 2017-05-21
Fix minor bugs in JsonExporter

Teknikaali:
This is a pull request for some minor bugs found from JsonExporter behavior while working on the [XmlExporter](https://github.com/dotnet/BenchmarkDotNet/issues/157). Here is some more context from the [comments](https://github.com/dotnet/BenchmarkDotNet/issues/157#issuecomment-300711076)

Fixes:

1. JsonExporter.BriefCompressed didn't correctly exclude the `Measurements`-property
2. JsonExport was showing incorrectly Lazy-type's `IsValueCreated` and `Value` properties in output instead just showing the actual value
3. SimpleJson uses `"\n"` as a line change, so I changed it to `Environment.NewLine` to match the rest of the approval test file line change style. Otherwise I couldn't get a `pass` from the exporter tests. I was getting `LF` line changes for the Json parts even though the approved file had `CRLF` line changes.
This change is based on the fact that the `AccumulationLogger` used in the exporter tests uses [`StringBuilder.AppendLine`](https://msdn.microsoft.com/en-us/library/cb3sbadh(v=vs.110).aspx) internally that uses `Environment.NewLine` as the default line change. I don't really know if this change could affect people running unix systems.. Is this something @alinasmirnova or maybe @mattwarren could comment on?

I ran both the "Classic" tests and "Core" tests like the contribution document suggested and both passed
* Core tests skipped 2 tests: `CanEnableServerGcMode` and `Test`, is this normal?

This is my first pull request so I'm gladly taking any feedback!
Pinging also @AndreyAkinshin and @adamsitnik

### DIFF
--- a/src/BenchmarkDotNet.Core/Exporters/Json/JsonExporter.cs
+++ b/src/BenchmarkDotNet.Core/Exporters/Json/JsonExporter.cs
@@ -4,7 +4,7 @@
     {
         public static readonly IExporter Brief = new JsonExporter("-brief", indentJson: true, excludeMeasurements: true);
         public static readonly IExporter Full = new JsonExporter("-full", indentJson: true, excludeMeasurements: false);
-        public static readonly IExporter BriefCompressed = new JsonExporter("-brief-compressed", indentJson: false, excludeMeasurements: false);
+        public static readonly IExporter BriefCompressed = new JsonExporter("-brief-compressed", indentJson: false, excludeMeasurements: true);
         public static readonly IExporter FullCompressed = new JsonExporter("-full-compressed", indentJson: false, excludeMeasurements: false);
 
         public static readonly IExporter Default = FullCompressed;
--- a/src/BenchmarkDotNet.Core/Exporters/Json/JsonExporterBase.cs
+++ b/src/BenchmarkDotNet.Core/Exporters/Json/JsonExporterBase.cs
@@ -28,8 +28,8 @@ namespace BenchmarkDotNet.Exporters.Json
             {
                 HostEnvironmentInfo.BenchmarkDotNetCaption,
                 summary.HostEnvironmentInfo.BenchmarkDotNetVersion,
-                summary.HostEnvironmentInfo.OsVersion,
-                summary.HostEnvironmentInfo.ProcessorName,
+                OsVersion = summary.HostEnvironmentInfo.OsVersion.Value,
+                ProcessorName = summary.HostEnvironmentInfo.ProcessorName.Value,
                 summary.HostEnvironmentInfo.ProcessorCount,
                 summary.HostEnvironmentInfo.RuntimeVersion,
                 summary.HostEnvironmentInfo.Architecture,
--- a/src/BenchmarkDotNet.Core/Exporters/Json/SimpleJson.cs
+++ b/src/BenchmarkDotNet.Core/Exporters/Json/SimpleJson.cs
@@ -1212,13 +1212,13 @@ namespace SimpleJson
         private static void ResetIndentationLevel()
         {
             indentationLevel = 0;
-            indentationText = "\n" + new string(' ', indentationLevel * spacesPerIndent);
+            indentationText = Environment.NewLine + new string(' ', indentationLevel * spacesPerIndent);
         }
 
         private static void HandleIndent(int change)
         {
             indentationLevel += change;
-            indentationText = "\n" + new string(' ', indentationLevel * spacesPerIndent);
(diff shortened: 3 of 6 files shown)


adamsitnik:
@Teknikaali big thanks!

AndreyAkinshin:
Hey @Teknikaali. Thanks for the PR and sorry for the late reply.

> I changed it to Environment.NewLine to match the rest of the approval test file line change style

It's great that now we have a consistent style for line ending. Unfortunately, `Environment.NewLine` is a bad choice because it will not work on Linux and MacOS, all unit tests will be failed. I think we can use the same approach as in the `CsvSeparator` case: we can introduce an enum like this:
```cs
public enum LineSeparator
{
  Windows, // \r\n
  Unix, // \n
  CurrentOperationSystem // Environment.NewLine
}
```
By default we can use `CurrentOperationSystem` for usual BenchmarkDotNet users and `Unix` for approval tests.
@adamsitnik, @Teknikaali, @adamsitnik, what do you think?

Teknikaali:
> By default we can use CurrentOperationSystem for usual BenchmarkDotNet users and Unix for approval tests.

@AndreyAkinshin this sounds like a good plan to keep up with the consistency.

Apparently someone else was having same problems with ApprovalTests.Net, line endings and tests and posted a new [issue ](https://github.com/approvals/ApprovalTests.Net/issues/166)to ApprovalTests.Net's GitHub page.

> we can introduce an enum

I've been fiddling around a bit with the `LineSeparator` idea and this is what I realized:
If `\n` line endings are _only_ needed for the approval tests, that could be the actual place to introduce this feature, instead of introducing a new enum and its handling in the exporters. I was thinking that there should be "one point of change" for this thing to work easily without messing with all the exporters too much..

**This is the simplest fix I came up with after looking a bit more into the `LineSeparator` enum:**
Currently `BenchmarkDotNet.Tests.Exporters.ApprovalTests.Exporters`-test uses `Approvals.Verify(string text)`-method for verifying the approved and received outputs.
`ApprovalTests.Net` allows one to "scrub" the given text _before_ verifying it. We could use this for our leverage and scrub some line separators:

`Approvals.Verify(string text, Func<string, string> scrubber)`  could be used instead to replace all `\r\n` line separators with the Unix alternative `\n`.

This is what the change would look like:
```csharp
    // ...
    Approvals.Verify(logger.GetLog(), lineSeparatorScrubber);
}

private Func<string, string> lineSeparatorScrubber =>
    input => input.Replace("\r\n", "\n");
```

I tested this method by converting all line endings of the `.approved.txt`-files to Unix `\n` and ran them against `Environment.NewLine` (`\r\n`) on a Windows system and the exporter tests passed.

The only con-side of this would be the idea of fiddling with the output result before it is actually being verified.. hmm.. it might be something that could be ignored as "it is just line endings we're talking about here 😅"

How does this sound like? We could still introduce the `LineSeparator`-enum to get rid of those "magic strings" `"\r\n"` and `"\n"` ofcourse.

AndreyAkinshin:
@alinasmirnova, what do you think?

Teknikaali:
Today I poked around the ApprovalTests repo a bit more and noticed something I had missed before:
> If you would like to verify line endings, simply add [assembly: IgnoreLineEndingsAttribute(false)] to your AssemblyInfo.cs

This is stated at the `readme.md` and I somehow totally missed that one sentence on the main page.

So apparently we're currently **not** verifying the line endings (which should be ok), as I couldn't find `IgnoreLineEndingsAttribute` usage from the `AssemblyInfo.cs`-file.

This is what the output line endings  _should_ be currently after this PR:

... | Windows | Unix
-|-|-
**Exported file** | `\r\n` | `\n`
**ApprovalTests** | `ignored` | `ignored`

All _should_ be fine with the line endings: Windows users would get `\r\n` and Unix users `\n` line endings accordingly when doing actual exporting and when testing with `ApprovalTests` line endings would be ignored completely.

Someone should probably still verify the tests with `Unix`-system as I don't have one at my disposal currently and AppVeyor doesn't seem to run the `ApprovalTests` automatically to see if we'd need to introduce that enum/scrubbing solution.

.. or did you @AndreyAkinshin mean other tests than `ApprovalTests`? If that's the case then we'd probably need that enum anyways.

_I wonder how I still received those differences in the line endings, even though they should have been ignored completely already while I was doing this PR.. weird._

AndreyAkinshin:
Looks awesome, `IgnoreLineEndingAttribute` is what we need.

PR #455, opened 2017-06-01
Wrong xml doc comment

ig-sinicyn:
Wrong xml doc comment 
```c# 
     /// <summary>
    /// Marks method to be executed >after< each benchmark iteration. 
    /// </summary> 
    public class IterationSetupAttribute: Attribute  { }
```
It was copied from IterationCleanupAttribute, I guess.


adamsitnik:
good catch! thanks!

Issue #457, opened 2017-06-01
Track Native Memory Allocations and more informations with our ETW Memory Diagnoser


adamsitnik:
We have two Memory Diagnosers. One of them is using ETW and is deprecated because it's less accurate than the other one. But this diagnoser could actually do more than just tracking allocated managed memory and GC collections.

This issue is `up-for-grabs`. The person that grabs it should:
* do an ETW research and answer following questions:
   - is it possible to track native memory allocations with ETW? I know that [dotTrace](https://blog.jetbrains.com/dotnet/2017/01/23/analyzing-native-memory-allocation-with-dottrace-2016-3/) can do this, many people asked for it including @JimBobSquarePants
   - is it possible to track the types of the allocated managed objects? (I know it is, some people asked for it)
   - is there anything more that is worth of tracking and showing to the users?
* actual implementation based on extending existing ETW [MemoryDiagnoser](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Diagnostics.Windows/MemoryDiagnoser.cs)
* testing, comparing the results with some 3rd party profiler like dotTrace

goldshtn:
You can track native memory allocations with ETW. However, it would probably slow down your benchmark by introducing ETW events for each allocation and free. You can also track each .NET memory allocation using ETW events. And that would probably slow down your benchmark even more. 

TBQH, I don't think this belongs in a benchmarking library. You can profile for allocations while running the benchmark using existing, proven tools.

adamsitnik:
>  And that would probably slow down your benchmark even more.

You are 100% right. This is why we perform separate run if any diagnoser is present. In that case we just use the diagnoser's output and ignore benchmark results (which could get affected).

>I don't think this belongs in a benchmarking library. You can profile for allocations while running the benchmark using existing, proven tools.

Yes, but on the other hand using our existing ETW Memory Diagnoser to extend it with such features should be relatively easy now. The existing code base is very good, just sign up for a few events and do the math. I believe that with relatively low effort we can add really nice feature.

I see few use cases:

- I know that my code allocates, but I would like to find out what objects are allocated without installing/buying any extra tool.
- I have some weird results, the benchmark should not allocate managed memory, but shows that something is allocated. What is that? 
- The same like above but 2 different memory profilers show 0 allocations. I need to check if BenchmarkDotNet is correct or not. (as the maintainer, to possibly find a bug in our Engine which should never allocate)
[...]

PR #460, opened 2017-06-06
PerfView-based ETW log diagnoser

exyi:
I have added a concept of PerfView-based ETW log diagnoser. It collects an ETL trace using PerfView, analyses that after the run and reports how much was time spent in specified methods.

**Don't merge it, please**, it's not tested. Take it as an issue with attached code ;)

### Rationale
I'm invoking entire HTTP requests in my benchmarks, so it's not very clear what is the slow thing there. That would force me to profile each benchmark separately, which would be very annoying. Or to see how much would optimizing certain method help, I'd have to profile all benchmarks, which takes few hours and would overrun my hard-drive. So I tried to profile it automatically - start an ETW collection on every benchmark, then load the stacks and sum them for each method. This way I have split each request processing into several stages and separated certain core methods, so I can easily see, what slows it down. You can have a look at [a report](https://ipfs.io/ipfs/QmScnYdY8xoPeHPN85edPdLPbi3GvHrUGicvHAuyMdrAQE/reports/BenchmarkRun-001-2017-05-31-10-34-59/report.html). It's still pretty large and hard-to-read table, but much better and faster to compare than a bunch of open Perfviews.
I'm not very sure that the numbers are right, but they seem to be both reproducible and reasonable. It also seems that normal time measurements are not influenced much by the profiler. (compare a run [without profiler](https://ipfs.io/ipfs/QmUs3zeXJqCtoUhYfmGdhRxhcjKdJDcrcbjuKohZ6n5ogf/BenchmarkRun-001-2017-05-24-09-32-19-report.html) and [with profiler](https://ipfs.io/ipfs/QmQPf8p7ppyZeChRrqHyJ8UwveHQkEE2NKBhxGM2zQPrJw/BenchmarkRun-001-2017-05-25-04-51-30-report.html))

This report comes from [dotvvm-benchmarks](https://github.com/riganti/dotvvm-benchmarks) project where I originally wrote this diagnoser. In contrast to what was said in the discussion in #457, I actually wrote it in my "business time" for real project and for real value ;) (thanks @tomasherceg for paying for that stuff). I understand that BenchmarkDotNet is benchmarking library, not a profiler, but this is the information I want to have from my benchmarks. And I believe that are some other people that would like that info :)

### Issues
* The PerfView was quite hard to convince to collect the trace and I didn't manage to use it for trace analysis. It is referenced as .exe file and depends on its own version of Microsoft.Diagnostics.Tracing.TraceEvent bundled as an embedded resource, so it's not usable from my code. This causes collisions with NuGet version of the TraceEvent and forced me to put PerfView invocations into separate AppDomain. It also completely prevents any exchange of Microsoft.Diagnostics.Tracing.TraceEvent instances with PerfView code. I'm not very sure how to overcome this limitation in a cleaner way. I could probably copy the code from PerfView, because I'm not using much of it, but it would be nice to use more capabilities ;). Maybe it would be possible to fork it, edit the build process to use the NuGet version of `TraceEvent` and produce a `dll` instead of `exe`.

* Some tests and stuff like that are missing. I'm not sure how to do that, do you have any test for something simmilar I could look at?

### More cool stuff
IMO it also a has greater potential than just showing CPU time spent in certain methods, this is just the beginning - there is a lot of thing the diagnoser could use from the PerfView - report GcStats, count memory allocations (#457) or even render some graphs quite simply.

What do you think? Any idea what to do with the integration?
cc @adamsitnik

### DIFF
--- a/BenchmarkDotNet.sln
+++ b/BenchmarkDotNet.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.10
+VisualStudioVersion = 15.0.26430.6
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{D6597E3A-6892-4A68-8E14-042FC941FDA2}"
 EndProject
@@ -54,6 +54,8 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BenchmarkDotNet.IntegrationTests", "tests\BenchmarkDotNet.IntegrationTests\BenchmarkDotNet.IntegrationTests.csproj", "{6A3CBB07-E337-488E-BDAC-ED96AF8ED608}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BenchmarkDotNet.IntegrationTests.ConfigPerAssembly", "tests\BenchmarkDotNet.IntegrationTests.ConfigPerAssembly\BenchmarkDotNet.IntegrationTests.ConfigPerAssembly.csproj", "{043F1DA4-CD51-45FD-805E-6571D67AA661}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BenchmarkDotNet.Diagnostics.PerfView", "src\BenchmarkDotNet.Diagnostics.PerfView\BenchmarkDotNet.Diagnostics.PerfView.csproj", "{4C32FC7D-244B-4BD0-B5FB-08ACC41EB240}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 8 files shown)


AndreyAkinshin:
Hey @exyi, thanks for the PR. Will do a review on the weekend.

adamsitnik:
I have been in touch with @exyi and I have encouraged him to post this to GitHub.

I will finally have some free time after 16th of June, then I am planning to:
1. Take a look at the PerfView dependency. Is it possible to do it only with out NuGet dependency to ETW?
2. Investigate Flame Graphs option as an output. I would love to see the results of the profiler in a nice chart.

adamsitnik:
I forgot to mention that I really like this idea!

mattwarren:
>  The PerfView was quite hard to convince to collect the trace and I didn't manage to use it for trace analysis. It is referenced as .exe file and depends on its own version of Microsoft.Diagnostics.Tracing.TraceEvent bundled as an embedded resource, so it's not usable from my code. This causes collisions with NuGet version of the TraceEvent and forced me to put PerfView invocations into separate AppDomain. It also completely prevents any exchange of Microsoft.Diagnostics.Tracing.TraceEvent instances with PerfView code. I'm not very sure how to overcome this limitation in a cleaner way.

@exyi One question, what does using `PerfView` directly give us that we can't do with the `TraceEvent` library instead (i.e. the [one on NuGet](https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/))? I always thought that `PerfView` was basically a UI wrapper on top of the `TraceEvent`, i.e. `TraceEvent` does the heavy lifting, collection of ETW events, etc. (for instance we already collect ETW events in the [`InliningDiagnoser`](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Diagnostics.Windows/InliningDiagnoser.cs) without shelling out to PerfView)

I guess it's worth looking at the [TraceEvent samples](https://github.com/Microsoft/dotnet-samples/tree/master/Microsoft.Diagnostics.Tracing/TraceEvent/TraceEvent) (also available as a [NuGet package](https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent.Samples/)) or even the [PerfView source](https://github.com/Microsoft/perfview/), to see if you can just use `TraceEvent` to mimic what's happening when you shell out to `PerfView`.

exyi:
@mattwarren I’m sure it is possible to write it using only TraceEvent, but the simple collection as shown in the samples simply did not work for me, so I decided to reuse PerfView’s code. I also looked at the source codes, what it does but there is quite a lot of stuff I don’t understand much (kernel/user space collection, symbols, rundowns, the ETW providers). Possibly, it is actually simple to run the collection, but I simply could not find the right way.
The second reason to integrate PerfView is, that it has some analysis features that are not present in the EventTrace but are also almost impossible to use with the AppDomain isolation. Because I could not reuse it, I only have a simple CPU time counter, which I implemented and is maybe correct. But it had a bug last week and it took me some time to figure out that I was counting recursive methods multiple time, so I’d be careful with reinventing stuff 😉 PerfView can do much more than just this, e.g. method grouping, GC stats and JIT stats are very cool; and implemented by some magic in the PerfView assembly.
 
However, I don’t say, that copying these few wanted snippets would not be the simplest solution 😉

exyi:
Hi, I'm sorry that I have not made any progress on this for a long time, but I basically stopped using Windows in the meantime, so I am not able to play with ETW anymore.

But, as you might have guessed, Linux also has it's tracing tools and I finally made some time to have a look at it. And I have migrated my [benchmarks project](https://github.com/riganti/dotvvm-benchmarks/tree/linux-perf-tracing/DotVVM.Benchmarks) to use Linux `perf` tool instead of perfview (you may have a look). The point is that I convert the collected stack traces to a simple text format (one stack per line) using Perl scripts in https://github.com/brendangregg/FlameGraph, and it should be also possible with Windows and OSX profilers, although I have not tested that. I think that it could be possible to create a common interface for different profilers and various diagnostics based on this and run the right profiler based on platform.

The only problem is that these are Perl scripts, and I think that it might be a bit problematic to run a Perl script in a library, especially on Windows. On the other hand Perl seems to pretty standard on Linux, so it should not that problematic to rewrite the windows related stackcollapse.pl to .NET and use Perl on Linux and OSX.

Sidenote: Yes, you noticed correctly, it's a FlameGraph library ;) With just one more Perl invocation you have nice interactive flamegraph in SVG. And because the stack trace format is pretty simple, you can simply combine traces from multiple benchmarks together, dig into specific methods and so on...

adamsitnik:
hi @exyi 

Sorry for a very late reply! Great to hear about your progress! Sounds very impressive!

I have just implemented the Flame Graphs for PerfView. I also want to implement an easier integration to any profilers for BenchmarkDotNet. Then I want to give make sure BDN works well with PerfView. When I get there we should have nice on-demand flame graphs on Windows. Once I have it I will contact you at try to get your Linux magic work for us ;p

At least this is my plan for now ;)

exyi:
> @adamsitnik, it seems that #878 should resolve the same problem. Do we still need this PR?

Wow, that looks pretty nice. I will close this PR.

I just think that it would be nice to have the possibility to display the statistics computed from the ETW traces (how much time was spend in certain methods, ...). But implementing that on top of Adam's implementation is IMO much better idea than having two implementations of the tracing. Should I file a new issue or is it already discussed somewhere?

adamsitnik:
@exyi great idea! I think that it could be just one more additional exporter for `EtwProfiler`. Please feel free to open a new issue. Or even send a PR :D

btw @exyi I hope to meet you at Update Conference in Prague!

PR #465, opened 2017-06-10
Small correction in Filters.md

mtschneiders:
I was reading through the docs and noticed this small error.

### DIFF
--- a/docs/guide/Configs/Filters.md
+++ b/docs/guide/Configs/Filters.md
@@ -20,7 +20,7 @@ public class IntroFilters
 {
     private class Config : ManualConfig
     {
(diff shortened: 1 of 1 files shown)


adamsitnik:
Thank you @mtschneiders !

AndreyAkinshin:
@mtschneiders, thanks!
We have the same error here: https://github.com/dotnet/BenchmarkDotNet/blob/1a8559f24c1e36bf235943d89e67ad83e5fcd9ce/samples/BenchmarkDotNet.Samples/Intro/IntroFilters.cs#L15 Could you make another PR with the same fix?

In the future, I want to make IntroSamples a part of the documentation. Unfortunately, now I have to copy-paste the samples source code (with typos).

Issue #453, opened 2017-05-25
MemoryDiagnoser and JsonExporter


ipjohnson:
I'm trying to get memory diagnostic results to be exported as part of JsonExporter.* and I can't seem to get it to work. What am I missing? 

```
    public class BenchmarkConfig : ManualConfig
    {
        public BenchmarkConfig()
        {
            Add(Job.Clr, Job.Core);
            Add(new MemoryDiagnoser());
            Add(new CompositeExporter(JsonExporter.Full));
        }
    }
```

Teknikaali:
It seems like the `JsonExporter` doesn't include the `MemoryDiagnoser`'s results yet, so @ipjohnson you weren't missing anything at all, _per se_.

This is a sample output for the `MemoryDiagnoser` in Json, if we simply include the `GcStats` property:
```json
...
"Memory":{
   "Gen0Collections":37628,
   "Gen1Collections":0,
   "Gen2Collections":0,
   "TotalOperations":15728640,
   "BytesAllocatedPerOperation":10048
},
...
```

One con side of adding the `GcStats`-property straight into `JsonExporter`'s output would be that _the operation isn't conditional_:
If there is no `MemoryDiagnoser` attached we will still get that "Memory"-entry in the Brief/Full Json export but with default values (it's a `struct`).

To overcome this con, we could check for the `MemoryDiagnoser` in `JsonExporterBase.ExportToLog`-method like so:

```csharp
// Initialize the normal data for export
var data = new Dictionary<string, object>
{
    { "DisplayInfo", r.Benchmark.DisplayInfo },
    ...
};

// Add the "Memory" entry only if a MemoryDiagnoser has been configured for the benchmark
if(summary.Config.GetDiagnosers().Any(diagnoser => diagnoser is MemoryDiagnoser))
{
    data.Add("Memory", r.GcStats);
}
```

@adamsitnik is this the correct way to perform the check if a `MemoryDiagnoser` is defined for the benchmark? Or is this kind of operation "forbidden" because of specific relationships between modules?

As I don't have any deep knowledge of what kind of relationships exist between exporters and diagnosers, but I could only make a good guess, that _it's the exporter's job to handle different diagnosers' output data inside the `ExporterBase.ExportToLog(...)` method._

[...]

PR #478, opened 2017-06-19
Add MemoryDiagnoser results to JsonExporter output

Teknikaali:
This PR fixes issue #453 by adding `MemoryDiagnoser` results to `JsonExporter` output.

## Related issue
#453 

## Example
```json
...
"Memory":{
   "Gen0Collections":37628,
   "Gen1Collections":0,
   "Gen2Collections":0,
   "TotalOperations":15728640,
   "BytesAllocatedPerOperation":10048
},
...
```

## Points of interest
- [This comment](https://github.com/dotnet/BenchmarkDotNet/issues/453#issuecomment-309508373) explains the thoughts behind the implementation
- Memory results will be added to the `JsonExporter`'s output only if `[MemoryDiagnoser]`-attribute has been configured for the benchmark. If it hasn't been defined no memory results will be added to the output.
- Both `Full` and `Brief` options are able to show the memory results

## Something to consider/look into
- Memory results do not appear in `ApprovalTests` comparison, as no `[MemoryDiagnoser]` has been defined in the config. I'm not really sure how/if it should be implemented there.


adamsitnik:
@Teknikaali thank you very much for solving this problem!

PR #480, opened 2017-06-22
Fix links to Overview/FAQ

davkean:
These were pointing to the wrong location.

I'm not a web guy, so not sure if relative links are frowned upon, if so, tell me and I'll change them to whatever you want.


adamsitnik:
Thank you!

PR #483, opened 2017-06-24
Update jobs docs

aarondandy:
Fixed a type name and changed some wording on the Jobs doc page.


adamsitnik:
@aarondandy thank you very much!

PR #488, opened 2017-06-30
Improve XmlExporter

Teknikaali:
This PR improves the `XmlExporter` by adding some features, bugfixes and unit tests.

## Incentive
My [Gitter comment ](https://gitter.im/dotnet/BenchmarkDotNet?at=5952b5c1c101bc4e3a293804)after realizing `XmlExporter` doesn't work 100% like it should:

> The XmlExporter currently throws if Statistics.Outliers have any results, but I already identified the problem: The exporter is using an indexer without first checking ContainsKey(...)..
I'll have it fixed in my upcoming pull requests. I'll have to check out if the Summary used in the ApprovalTests misses filling Outlierswith valid test data for all exporters..
I've also tested to get the Memory-info (GcStats) to show up for XmlExporter just like JsonExporter has it and it seems to be working fine.
I was also thinking to make sure the XmlSerializer works right in the background and add some unit tests to prove that.

There are many small changes included in this PR, so hang on tight!

### Add interfaces to improve `XmlSerializer`'s testability
`IXmlSerializer` interface has been added to improve the testability of `XmlSerializer`. Also an interface has been added for `XmlWriter` to enable mocking .
"Fluent-ish"-interface `IFluentXmlSerializer` has been added to deal with the serialization customization.

### Implement `IXmlWriter`
To enable clean mocking `SimpleXmlWriter : IXmlWriter` has been added. It wraps the actual `XmlWriter` and exposes only the necessary methods needed.

❔ `XmlWriter` is disposable so I implemented the `IDisposable` interface. I'm a bit unsure if I handled the disposing correctly.. is the [implementation](https://github.com/dotnet/BenchmarkDotNet/compare/master...Teknikaali:xmlexporter-memory?expand=1#diff-feb073088d064beaecaa0c7f55b9c4abR44) ok?

### Include `GcStats` in xml output as "Memory" element
[If a `MemoryDiagnoser` is defined](https://github.com/dotnet/BenchmarkDotNet/compare/master...Teknikaali:xmlexporter-memory?expand=1#diff-ff89785d11fdb56d0395e3087e75a224R35) we add the `GcStats` results in an element called "[Memory](https://github.com/dotnet/BenchmarkDotNet/compare/master...Teknikaali:xmlexporter-memory?expand=1#diff-9e58aa7b2561fc7e94a1b5f4c9764078R76)".

### Define `Outliers` collection's item name in xml output as "Outlier"
`Outliers` collection's item names [are now defined](https://github.com/dotnet/BenchmarkDotNet/compare/master...Teknikaali:xmlexporter-memory?expand=1#diff-ff89785d11fdb56d0395e3087e75a224R33) and it doesn't throw exception anymore if the collection contains any items:
```xml
<Outliers>
  <Outlier>372.670582203865</Outlier>
  <Outlier>373.025599908829</Outlier>
</Outliers>
```

❔ Is the amount of decimals ok? I think I saw either a commit or a comment regarding this..

### Replace collection item name customization by item type with collection name
Before collection's item element name was defined by giving the item type as parameter to `WithCollectionItemName` method. Now one can give the parent collection's property name as parameter instead the type to make the syntax more readable and flexible:
```csharp
// Before:
.WithCollectionItemName(typeof(double), "MyDoubleItem");
// All double type items are called "MyDoubleItem", even if we want to only affect one property's item naming

// After:
.WithCollectionItemName(nameof(SourceType.MyDoubleCollection), "MyDoubleItem");
// Now we have specified that only MyDoubleCollection items are called "MyDoubleItem"
// e.g. "Outliers" collection with "Outlier" as item name
```

### Minor fixes to `XmlSerializer`
* Changes some exception types
* Changes specific used types to interfaces to improve testability e.g. `XmlWriter` -> IXmlWriter`
* If no item name was defined an exception would have been thrown. Now we use `DefaultItemName`-property on `XmlSerializer` instead of throwing. Defaults to "Item" if user hasn't defined the collection's item names with `WithCollectionItemName` method

### Unit tests
`XmlSerializerTests.cs` contains tests for all `XmlSerializer`'s public methods.
I ran the tests locally with [OpenCover](https://github.com/OpenCover/opencover) to get 100% coverage.

### Notes
* `XmlSerializer` is missing documentation from e.g. [Exporters](http://benchmarkdotnet.org/Configs/Exporters.htm) page
* `XmlSerialize` doesn't have any samples yet
* `Outliers` and `Memory` aren't filled with any test data in `ApprovalTests`. This might be something to check out later..

Comments and improvement suggestions are always welcome!


src/BenchmarkDotNet.Core/Exporters/Xml/IXmlSerializer.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace BenchmarkDotNet.Exporters.Xml
+{
+    public interface IXmlSerializer : IFluentXmlSerializer
```
adamsitnik: I would expect that every  `IFluentXmlSerializer` is `IXmlSerializer` and not every `IXmlSerializer` is `IFluentXmlSerializer`.

do we need to have two interfaces here?

src/BenchmarkDotNet.Core/Exporters/Xml/SimpleXmlWriter.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.IO;
+using System.Xml;
+
+namespace BenchmarkDotNet.Exporters.Xml
+{
+    public class SimpleXmlWriter : IXmlWriter, IDisposable
```
adamsitnik: I suggest that only `XmlExporter` should be public. Our xml serializers, writers and corresponding interfaces should be internal. Our users don't need to know about their existence and the variety of choice you give them with `Full` and `Simple` version should prevent anybody from deriving a custom class.

src/BenchmarkDotNet.Core/Exporters/Xml/SimpleXmlWriter.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            if (writer == null)
+                throw new ArgumentNullException(nameof(writer));
+
+            this.writer = XmlWriter.Create(writer, new XmlWriterSettings() { Indent = indent });
+        }
+
+        public void WriteElementString(string localName, string value)
+        {
+            writer.WriteElementString(localName, value);
```
adamsitnik: could you use expression bodies here? it would simplify the code

src/BenchmarkDotNet.Core/Exporters/Xml/SimpleXmlWriter.cs
```diff
@@ -0,0 +1,65 @@
...
+                if (disposing)
+                {
+                    writer.Dispose();
+                }
+
+                disposedValue = true;
+            }
+        }
+
+        public void Dispose()
```
adamsitnik: the implementation of Dispose pattern should simply delegate the responsibility of disposal to the only disposable field in this class (writer). It should take care of things like throwing `ObjectDisposedException` after disposal etc.

So following code will be just enough:
`public void Dispose() => writer.Dispose()`

src/BenchmarkDotNet.Core/Exporters/Xml/XmlExporterBase.cs
```diff
@@ -10,38 +12,41 @@ public abstract class XmlExporterBase : ExporterBase
...
         {
-            var serializer = new XmlSerializer(typeof(SummaryDto))
-                                    .WithRootName(nameof(Summary))
-                                    .WithCollectionItemName(typeof(Measurement),
-                                                            nameof(Measurement))
-                                    .WithCollectionItemName(typeof(BenchmarkReportDto),
-                                                            nameof(BenchmarkReport.Benchmark));
+            IXmlSerializer serializer =
+                new XmlSerializer(typeof(SummaryDto))
+                    .WithRootName(nameof(Summary))
```
adamsitnik: I really like this fluent api! 👍 

Btw the next step to make it even better would be to create a fluent `XmlSerializerBuilder` with mutable private fields accessed via your public fluent api. Then at the end you could call `builder.Build()` and receive immutable serializer with `Serialize` method only.

tests/BenchmarkDotNet.Tests/Exporters/XmlSerializerTests.cs
```diff
@@ -0,0 +1,245 @@
...
+using Xunit;
+
+namespace BenchmarkDotNet.Tests.Exporters
+{
+    public class XmlSerializerTests
+    {
+        [Fact]
+        public void Serialize()
+        {
+            string expected =
```
adamsitnik: nit: you can make `expected` output a `const` if it's a string. Then you are sure that your code is not changing the value.

tests/BenchmarkDotNet.Tests/Exporters/XmlSerializerTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                "MockSource " +
+                "End";
+            var source = new MockSource();
+            var writer = new MockXmlWriter();
+            IXmlSerializer serializer = new XmlSerializer(typeof(MockSource));
+
+            serializer.Serialize(writer, source);
+            string actual = writer.ToString();
+
+            Assert.Equal(expected, actual);
```
adamsitnik: it took me some time to understand what this method is doing ;) You are testing xml serializer without using xml.

Have you run into white spaces comparison issue? I don't know what's the perfect way of testing xml, but [here](https://github.com/dotnet/BenchmarkDotNet/blob/master/tests/BenchmarkDotNet.Tests/AppConfigGeneratorTests.cs#L30) I am comparing the output (xml) with expected (xml) with whitespaces removal.

tests/BenchmarkDotNet.Tests/Exporters/XmlSerializerTests.cs
```diff
@@ -0,0 +1,245 @@
...
+        public void DoesNotWriteUnwriteableCollection()
+        {
+            IXmlWriter writer = new MockXmlWriter();
+            IXmlSerializer serializer = new XmlSerializer(typeof(UnwriteableCollectionSource));
+
+            serializer.Serialize(writer, new UnwriteableCollectionSource());
+            string actual = writer.ToString();
+
+            Assert.DoesNotContain(actual, "Items");
+        }
```
adamsitnik: a lot of great tests! We might even have better xml tests than the `corefx` repo now ;) 👍

src/BenchmarkDotNet.Core/Exporters/Xml/SimpleXmlWriter.cs
Teknikaali: I went through all the `XmlExporter`'s classes and I think I got them all changed to `internal` apart from the actual Exporter.

Teknikaali:
Nice comprehensive review @adamsitnik ! I've pushed changes and added comments to the review bits.

src/BenchmarkDotNet.Core/Exporters/Xml/SimpleXmlWriter.cs
Teknikaali: Adding expression bodies really made the class look cleaner.

src/BenchmarkDotNet.Core/Exporters/Xml/SimpleXmlWriter.cs
Teknikaali: Thanks for the guidance, I haven't had to write many Dispose patterns in the past so this was a good thing to learn! I replaced the more complex way with delegation to the writer.

tests/BenchmarkDotNet.Tests/Exporters/XmlSerializerTests.cs
Teknikaali: "There must be something bonkers with the code if it seems complicated and not self-explanatory" ;) But yeah you're right, I had a bit hard time figuring out the right way to test this method and apparently it shows too. The problem is probably just in the methodology behind "how to test the serialization".

I'm not really sure if I "went too far" with the unit testing methodology by _not allowing_ the serializer's tests to depend on any other implementation e.g. `SimpleXmlWriter`. That's why I'm using the `MockXmlWriter`, but its implementation is almost the same as `SimpleXmlWriter`'s, just without all the xml..

I tried testing the `Serialize` method just by storing each element tag call and item property addition etc. to collections inside the mock writer, and then compared those collections to expected values to define if the serialization process was successful. But after trying that out it seemed like using a string comparison would be more clean to read and work with.

One way to make this test method a bit more clear would be to actually test the xml serializer with some real xml data. That would mean `XmlSerializerTests` would have to implement `IDisposable` (`SimpleXmlWriter` is disposable) and then handle disposing accordingly.

If your first reaction when seeing the `Serialize` test was "_Where's the xml?_", then it's really an easy decision to change the expected value of `Serialize` test from current custom string to xml before accepting this PR. Should I do it or leave it as is?

(I could have had use for your white space implementation here, now that I have looked into it)

adamsitnik: @Teknikaali I think that we should leave it the way it is. I am very grateful that you wrote these tests.

PR #503, opened 2017-07-25
Make sure IterationCleanup is run after Jitting

### LINKED ISSUE #481: Iteration cleanup runs before the benchmark
Example:
```
  class Program
    {

        {      
            BenchmarkDotNet.Running.BenchmarkRunner.Run<BadBench>();
            Console.ReadLine();
        }
    }

   

    
    public class BadBench
    {
        [Benchmark]
        public void Test()
        {
            Console.WriteLine("Test");
            Thread.Sleep(1000);
        }


        [IterationSetup]
        public void Init()
        {
            Console.WriteLine("Init");
        }

        [IterationCleanup]
        public void Clean()
        {
            Console.WriteLine("Clean");
        }
    }
```
I'm expecting  to have "Test" before "Clean", am I? Or do I misunderstand something? Right now it goes right after "Init"
Version 0.10.8

smitpatel:
Resolves #481


adamsitnik:
@smitpatel thanks!

PR #506, opened 2017-07-27
Removes a temporary solution related to connectivity issues to nuget

Ky7m:
Removes a temporary solution related to connectivity issues to nuget.org https://appveyor.statuspage.io/incidents/m2vdvw39kdk8


AndreyAkinshin:
@Ky7m, thanks!

Issue #364, opened 2017-02-19
Benchmark just stops (Antivirus problem)


KonkLaw:
First of all, thanks for great job.

I've got next problem: sometimes after running benchmark I've been geting this issue:
![error](https://cloud.githubusercontent.com/assets/22473930/23103774/c5050192-f6d1-11e6-9f10-4a2104148962.png)
Console benchmark is just stopping with standard system message: "Press any key to continue ...". No exceptions or warnings in console or log file. As I see it doesn't depends on particular test. I really spend a bunch of time trying to find an error on my side. And it seems like I've found out what was wrong.
The thing is, unfortunately, environment where we are running tests is not always depends on us. And I've got one with antivirus which is working on back-back-baground. It simply blocks running test on the halfway. And that's not real problem. Benchmarking with running antivirus is problem of developer. The real problem is that I just see no any errors or warnings at all. Just "Press any key to continue ...". And googling doen't help at all.
I realize that handling antivirus cases may be quite tough call. May be, at least some warning in documentation could help me.

AndreyAkinshin:
Hey @KonkLaw, thanks for the report. You are absolutely right, we definitely should print a warning. Any clues how to detect we if have problems with antivirus (btw, may I ask which antivirus do you use)?

KonkLaw:
Bitdefender Antivirus Free Edition (updating works)
on Win7 SP1 x64
In some cases antivirus hides gui, but still works (I see it in task manager), so it stops my benchmark with no warnings. If I will run its gui, I will see correspond message ("process is blocked" or like that).

I've prepared small test with bug in my draft repository: https://github.com/KonkLaw/Tests
Also text in log file (bin\Release\BenchmarkDotNet.Artifacts\) in faulty cases always is less that output in console.
The interesting fact is that if I run two tests (two methods with attributes, for example in test given above) test ends successfully. But, at least, three methods is enough to get a fault. May be it is connected with log file size.

AndreyAkinshin:
Related issue: https://github.com/dotnet/BenchmarkDotNet/issues/390

guidmaster:
Have the same problem. Running in a corp environment means that there is no way to disable AV :(

Getting the error by just running the Getting Started sample.

adamsitnik:
Description for people interested in grabbing this issue:

- We generate, build and run new process for every benchmark
- Anti Virus is blocking this by default.
- My suggestion:
    - Detect the situation where our process is immediately killed
[...]

PR #510, opened 2017-07-30
EnvironmentAnalyser - Warning about installed AV software 

lukasz-pyrzyk:
According  to the #364

EnvironmentAnalyzer prints warning about any antivirus software and tip for user about InProcessToolchain

### DIFF
--- a/docs/guide/Advanced/Baseline.md
+++ b/docs/guide/Advanced/Baseline.md
@@ -1,6 +1,6 @@
 # Baseline
 
-In order to scale your results you need to mark one of your benchmark methods as a baseline. Only one method in class can have `Baseline = true` applied.
+In order to scale your results, you need to mark one of your benchmark methods as a baseline. Only one method in class can have `Baseline = true` applied.
 
 ## Example
 
--- a/docs/guide/Advanced/Percentiles.md
+++ b/docs/guide/Advanced/Percentiles.md
@@ -1,6 +1,6 @@
 # Percentiles
 
-The percentile represents a higher boundary for specified percengage of a measurements.
+The percentile represents a higher boundary for specified percentage of the measurements.
 For example, 95th percentile = 500ms means that 95% of all samples are not slower than 500ms.
 This metric is not very useful in microbenchmarks, as the values from consequent runs have a very narrow distribution.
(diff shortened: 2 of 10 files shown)


adamsitnik:
@lukasz-pyrzyk thanks!

@AndreyAkinshin LGTM. Please merge if you agree.

AndreyAkinshin:
1. This PR includes two commits (AV+typo fixes). Please, create a separate feature branch for the PR in your fork based on the current master branch in `dotnet/BenchmarkDotNet`. The feature branch should include only this single commit.
2. The code with ManagementObjectSearcher makes sense only on Windows. Please, add the corresponded check.
3. I'm not sure that we *always* should print *a warning* about antiviruses. A lot of people have enabled WindowsDefender, but it doesn't affect benchmarks in most cases. If BenchmarkDotNet will always print a warning about it, probably it is not a good warning. It would be better to detect "suspicious" situation and warn user only in case of problems. @adamsitnik, what do you think?
4. It also makes sense to print path to the antivirus product exe. For example, if I run
```
wmic /NAMESPACE:\\root\SecurityCenter2 PATH AntiVirusProduct GET 
```
I get the following on my Windows machine:
```
displayName       instanceGuid                            pathToSignedProductExe                       pathToSignedReportingExe                     productState  timestamp
Windows Defender  {D68DDC3A-831F-4fae-9E44-DA132C1ACF46}  %ProgramFiles%\Windows Defender\MSASCui.exe  %ProgramFiles%\Windows Defender\MsMpeng.exe  397568        Fri, 28 Jul 2017 13:17:20 GMT
```
`pathToSignedProductExe` looks like a good option.

lukasz-pyrzyk:
Hi @AndreyAkinshin, @adamsitnik, thanks for comments.
a) Sure, my fault
b) Totally agree, will do
c) I can walk through Summary and check if any benchmark exited with error code != 0. If yes, print a warning. What do you think about it? 
d) Will add

AndreyAkinshin:
a,b,d) Ok, great.
c) LGTM. @adamsitnik?

adamsitnik:
>  I can walk through Summary and check if any benchmark exited with error code != 0. If yes, print a warning. What do you think about it?

good idea!

PR #512, opened 2017-08-01
Warning about antivirus software after benchmark failure

lukasz-pyrzyk:
According to the #364 

In case of invalid error code, analyzer prints warning about installed antivirus software and InProcessToolchain. 

Tested with Symantec and Windows Defender.

### DIFF
--- a/src/BenchmarkDotNet.Core/Analysers/EnvironmentAnalyser.cs
+++ b/src/BenchmarkDotNet.Core/Analysers/EnvironmentAnalyser.cs
@@ -1,7 +1,10 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Portability;
 using BenchmarkDotNet.Reports;
 using BenchmarkDotNet.Toolchains.InProcess;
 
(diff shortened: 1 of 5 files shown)


src/BenchmarkDotNet.Core/Analysers/EnvironmentAnalyser.cs
```diff
@@ -24,6 +29,28 @@ public override IEnumerable<Conclusion> AnalyseSummary(Summary summary)
...
+        }
+
+        private static string CreateWarningAboutAntivirus(ICollection<Antivirus> avProducts)
+        {
+            var sb = new StringBuilder("Detected error exit code from one of the benchmarks. " +
+                                       "It might be caused by following antivirus software:");
+            sb.AppendLine();
+
+            foreach (var av in avProducts)
+                sb.AppendLine($"\t - {av}");
```
AndreyAkinshin: Could we use spaces instead of `\t`?

AndreyAkinshin:
It looks much better now.
Could you also add a note about it in `docs/guide/FAQ.md`?

src/BenchmarkDotNet.Core/Analysers/EnvironmentAnalyser.cs
lukasz-pyrzyk: Fixed

lukasz-pyrzyk:
Changed to spaces and added a FAQ note. What do you think?

AndreyAkinshin:
Great, thanks!

mbjasonayer:
My AV is controlled by my work's IT so disabling AV is out for me. Does anyone have a work around for this issue? I tried a suggestion from another thread about running Command Prompt with Administrator rights from the executable's location, but that did not help.

Issue #495, opened 2017-07-18
Attributes put on base methods are not considered in derived class


smitpatel:
```C#
class Program
{
    static void Main(string[] args)
    {
        BenchmarkRunner.Run<Derived>();
    }
}

public abstract class Base
{
    protected int SleepTime;

    [IterationSetup]
    public abstract void Setup();
        
    [Benchmark]
    public void Test()
    {
        Thread.Sleep(SleepTime);
    }
}

public class Derived : Base
{
    //[IterationSetup]
    public override void Setup()
    {
        SleepTime = 100;
    }
}
```
In above code I have annotated method `Base.Setup` with `IterationSetup`. When I override the method to provide implementation in derived class but don't put attribute (assuming it will come from base class automatically), then the method is not called. This makes SleepTime = `default(int)` which is 0. 
If I put attribute in derived class then it sets proper value.
Without attribute in derived class
```
Method |     Mean |    Error |   StdDev |
------- |---------:|---------:|---------:|
   Test | 164.9 ns | 3.194 ns | 3.802 ns |
```
With attribute in derived class
```
 Method |     Mean |     Error |    StdDev |
------- |---------:|----------:|----------:|
   Test | 100.5 ms | 0.0955 ms | 0.0893 ms |
```

The framework should consider attributes put on base methods.

adamsitnik:
hi @smitpatel !

Thanks for a good report. I think that you are right, the expected behavior is different than what we have today.

@AndreyAkinshin do you agree? It should be easy to implement

AndreyAkinshin:
@adamsitnik, LGTM.

lukasz-pyrzyk:
@AndreyAkinshin can you assign it to me? 

I have done it locally, but I think of a suitable unit test.

AndreyAkinshin:
@lukasz-pyrzyk, assigned.

A short note about unit tests: usually we check such stuff via integration tests which is wrong (we do because it's the easiest way). The main problem here is that it significantly increases the total time of the whole test suite (especially for .NET Core profile).
In fact, we should introduce tests which build the benchmark sets via our API (based on the class description) and check it manually (instead of parsing logs in integration tests).

PR #514, opened 2017-08-02
#495 - Unit test for reading attributes from the base class

lukasz-pyrzyk:
#495 was fixed by @ipjohnson in 557246db by using `GetCustomAttributes` instead of `ReflectionExtensions::HasAttribute`. Version `v0.10.9` has this fix.

To be sure, I've added a unit test for expected behaviour

### DIFF
--- a/tests/BenchmarkDotNet.Tests/Running/BenchmarkConverterTests.cs
+++ b/tests/BenchmarkDotNet.Tests/Running/BenchmarkConverterTests.cs
@@ -13,14 +13,23 @@ namespace BenchmarkDotNet.Tests.Running
         [Fact]
         public void ReadsAttributesFromBaseClass()
         {
-            Benchmark benchmark = BenchmarkConverter.TypeToBenchmarks(typeof(Derived)).Single();
+            var derivedType = typeof(Derived);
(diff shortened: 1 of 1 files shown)


tests/BenchmarkDotNet.Tests/Running/BenchmarkConverterTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            Benchmark benchmark = BenchmarkConverter.TypeToBenchmarks(typeof(Derived)).Single();
+           
+            Assert.NotNull(benchmark);
+            Assert.NotNull(benchmark.Target);
+            Assert.NotNull(benchmark.Target.IterationSetupMethod);
+            Assert.NotNull(benchmark.Target.IterationCleanupMethod);
+            Assert.NotNull(benchmark.Target.GlobalCleanupMethod);
+            Assert.NotNull(benchmark.Target.GlobalSetupMethod);
+        }
```
adamsitnik: this benchmark should not only Assert that given method is not null, but also that given method points to derived one, not the base one

lukasz-pyrzyk: Agree, this will be more strict

PR #518, opened 2017-08-04
Fixed information about MemoryDiagnoser.

cincuranet:
Same as in `Diagnosers.md`.

Also maybe instead of having same information on two places, the `Overview.md` could just contain link to the `Diagnosers.md`, where all (more anyway) the detailed information is.


adamsitnik:
@cincuranet thanks!

Issue #511, opened 2017-07-30
[bug] Bug in GetTargetedMatchingMethod() logic


ig-sinicyn:
BDN 0.10.9.

The `BenchmarkConverter.GetTargetedMatchingMethod()` [method](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Running/BenchmarkConverter.cs#L93-L102) depends on order of methods passed to it, first match wins.

The issue is, there is no warranty that setup/cleanup method with matching `Target` will be enumerated before  setup/cleanup method with empty target. As a proof:
```cs
		[GlobalSetup]
		public static void SetupAlwaysWins() { ... }

		[GlobalSetup(Target = nameof(MyMethod))]
		public static void SetupNeverCalled() { ... }
```

ipjohnson:
I hadn't gone down the road of Targeted and non Targeted in the same benchmark.

It seems like there are two ways to handle this
1) throw an exception as not supported (go one way or the other)
2) treat it as a fallback like what @ig-sinicyn filled a bug on.

Let me know which behavior is preferred and I can put in a pr in the coming days

AndreyAkinshin:
I think that benchmarks which have Target and NonTarget setups at the same time look confusing. So, I suggest to show error about invalid configuration and don't run any of benchmarks.
@ig-sinicyn, @adamsitnik, what do you think?

ig-sinicyn:
I'd prefer to follow standard approach: most specific setup wins.
This way there will be no need to include 10 setup/cleanup methods if there are 10 benchmark methods, 9 methods reuse same logic and only one needs special handling.

AndreyAkinshin:
> This way there will be no need to include 10 setup/cleanup methods if there are 10 benchmark methods, 9 methods reuse same logic and only one needs special handling.

Ok, makes sense.

> Also, there may be more than one targeted attribute applied to the setup/cleanup method.

It looks too strange.

ipjohnson:
@AndreyAkinshin to summarize I'll put together a PR that allows a targeted setup/cleanup to override a non targeted. 

To make things more deterministic I'll add code to throw an exception if there are to many setup/cleanup methods found.

adamsitnik:
I believe that:

1. More than one targeted setup/cleanup matches the benchmark = fail with error
2. One targeted and one general setup/cleanup matches the benchmark = choose the targeted

[...]

PR #525, opened 2017-08-11
adding validator for setup cleanup attributes

ipjohnson:
this is to address issue #511 adds a validator for setup cleanup attributes as well as prefers targeted setup and cleanup methods over non targeted.


adamsitnik:
@ipjohnson thanks a lot!

Issue #167, opened 2016-05-06
Detect virtual machine environment


AndreyAkinshin:
Some people try to run benchmark on virtual machines. It can affect benchmark results because of virtualization effects and incorrect `Stopwatch.Frequency` value. We should definitely add a line about VM in the environment info. Possibly we also should add a warning.

See also: http://stackoverflow.com/questions/498371/how-to-detect-if-my-application-is-running-in-a-virtual-machine

adamsitnik:
I am one of these people ;) We could add it to the [EnvironmentAnalyser](https://github.com/PerfDotNet/BenchmarkDotNet/blob/e31872d1e521da27d46a919a58fb29253562e56a/BenchmarkDotNet/Analysers/EnvironmentAnalyser.cs)

lukasz-pyrzyk:
I have an idea how to do it,. Can you assign me to this feature?

AndreyAkinshin:
@lukasz-pyrzyk, great! I removed the up-for-grabs issue and send you an invitation to be a collaborator (I need it to put you in the assignees field). Looking forward to the PR.

PR #527, opened 2017-08-14
Detecting virtual machine hypervisor, #167

lukasz-pyrzyk:
According to the #167 - **Detect virtual machine environment**

Please leave it PR open till I finish following tasks.

- [x] - Detect Hyper-V
- [x] - Detect VMWare
- [x] - Detect VirtualBox
- [x] - Update of the FAQ


AndreyAkinshin:
LGTM

lukasz-pyrzyk:
I'm done making changes, @adamsitnik feel free to review :)

src/BenchmarkDotNet.Core/Portability/StringExtensions.cs
```diff
@@ -13,5 +13,10 @@ internal static bool EqualsWithIgnoreCase(this string left, string right)
             return left.Equals(right, StringComparison.OrdinalIgnoreCase);
 #endif
         }
+
+        internal static bool ContainsWithIgnoreCase(this string text, string word)
+        {
+            return text.IndexOf(word, StringComparison.OrdinalIgnoreCase) >= 0;
+        }
```
AndreyAkinshin: It would be better to introduce a private property:
```cs
        private static StringComparison IgnoreCaseStringComparison
        {
            get
            {
#if !CORE
                return StringComparison.InvariantCultureIgnoreCase;
#else
// http://stackoverflow.com/questions/14600694/where-has-stringcomparison-invariantcultureignorecase-gone
                return StringComparison.OrdinalIgnoreCase;
#endif
            }
        }
```
and use it in both `EqualsWithIgnoreCase` and `ContainsWithIgnoreCase`.

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -321,5 +321,35 @@ internal static ICollection<Antivirus> GetAntivirusProducts()
             return Array.Empty<Antivirus>();
 #endif
         }
+
+
+        internal static VirtualMachineHypervisor GetVirtualMachineHypervisor()
+        {
+            VirtualMachineHypervisor[] hypervisors = { new HyperV(), new VirtualBox(), new VMware() };
```
AndreyAkinshin: It would be better to make constructors of `HyperV`, `VirtualBox`, `VMware` private; introduce static instances in each class, and use them here.

src/BenchmarkDotNet.Core/Portability/HyperV.cs
```diff
@@ -0,0 +1,13 @@
+namespace BenchmarkDotNet.Portability
+{
+    public class HyperV : VirtualMachineHypervisor
+    {
+        public override string Name => "Hyper-V";
+
+        public override bool IsVirtualMachine(string manufacturer, string model)
+        {
+            return manufacturer != null && model != null &&
```
AndreyAkinshin: I don't like that we have these null checks in each vm class. Could we do it once in `VirtualMachineHypervisor` and just call protected virtual method (which will be implemented in `HyperV`, `VirtualBox`, `VMware`) with the `([NotNull] string manufacturer, [NotNull] string model)` signature?

AndreyAkinshin:
I think it makes sense to also add info about vm in the summary header (because a lot of people don't share their warnings). E.g. on the processor line:
```
BenchmarkDotNet=v0.10.9.279-nightly, OS=Windows 10 Redstone 1 (10.0.14393)
Processor=Intel Xeon CPU E5-2680 v3 2.50GHz, ProcessorCount=2 [NAME_OF_VM]
```

AndreyAkinshin:
@lukasz-pyrzyk another question about appveyor: can we detect which vm do you use during an appveyor build? It would be nice to have a test for it.

src/BenchmarkDotNet.Core/Portability/StringExtensions.cs
lukasz-pyrzyk: Good catch, done

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
lukasz-pyrzyk: Naturally. Do you prefer this because of the readability or something else?

AndreyAkinshin: It's a "best practice" in BenchmarkDotNet.Core mainly because we care about allocations. Internal BenchmarkDotNet allocations should never affect memory traffic in user code, so we are trying to avoid direct constructors if it's possible.

src/BenchmarkDotNet.Core/Portability/HyperV.cs
lukasz-pyrzyk: Sometimes null is valid for VM detector. For example, `VMware` does not care about `manufacturer` value.

To avoid so many manual null checks, I can provide base method like this one
```csharp
protected bool ContainsVmIdentifier(string systemInformation, string vmIdentifier)
{
    return systemInformation != null && systemInformation.ContainsWithIgnoreCase(vmIdentifier);
}
```

with following usage in VMware:
```csharp
public override bool IsVirtualMachine(string manufacturer, string model)
{
    return ContainsVmIdentifier(model, "vmware");
}
```
What do you think @AndreyAkinshin ?

AndreyAkinshin: LGTM.
Also please use `[NotNull]` and `[CanBeNull]` attributes where it's possible.

lukasz-pyrzyk:
> another question about appveyor: can we detect which vm do you use during an appveyor build? It would be nice to have a test for it.

I think it won't be a problem. However, test should be ignored on locally. Do we have any code which can enable/run test only on appveyor / travis?

> I think it makes sense to also add info about vm in the summary header (because a lot of people don't share their warnings). E.g. on the processor line:

Really good idea. I would like to suggest doing it in next PR, because i won't be able to do it in following week. Maybes somebody will pick it up as `up for grabs` :)

AndreyAkinshin:
> I think it won't be a problem. However, test should be ignored on locally. Do we have any code which can enable/run test only on appveyor / travis?

On appveyor, we have defined `APPVEYOR_BUILD_NUMBER` const, you can use it as a `#if` hack.

> Really good idea. I would like to suggest doing it in next PR, because i won't be able to do it in following week.

Ok, I will create another issue as soon as I merge this one.

AndreyAkinshin:
@lukasz-pyrzyk, the `DetectsHyperVOnAppveyor` is red on appveyor. =(
See https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/build/0.10.9.287-cuaybuay/tests

lukasz-pyrzyk:
I will connect to some AppVeyor node with remote and I'll check vmi details

tests/BenchmarkDotNet.Tests/XUnit/AppveyorWithClassicFrameworkOnlyFactAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public class AppVeyorWithClassicFrameworkOnlyFactAttribute : AppVeyorOnlyFactAttribute
+    {
+        private const string Message = "Test requires classic .NET Framework";
+
+        private bool classicDotNet
+        {
+            get
+            {
+#if !CORE
```
AndreyAkinshin: Please, try to avoid `#if !CORE` constructions outside `BenchmarkDotNet.Portability`. E.g., you can introduce a public `RuntimeInformation.IsCore()` method.

tests/BenchmarkDotNet.Tests/XUnit/AppveyorWithClassicFrameworkOnlyFactAttribute.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace BenchmarkDotNet.Tests.XUnit
+{
+    public class AppVeyorWithClassicFrameworkOnlyFactAttribute : AppVeyorOnlyFactAttribute
```
AndreyAkinshin: You can use `Clr` instead of `ClassicFramework`.

lukasz-pyrzyk:
Just thinking. Maybe `BenchmarkDotNet.IntegrationTests` is a better place for the `DetectsHyperVOnAppveyor`? It uses WMI under the hood and requires Appveyor, so for me it looks more like integration test.

AndreyAkinshin:
@lukasz-pyrzyk, `IntegrationTests` should contain only long-running tests which actually run some benchmarks.

AndreyAkinshin:
Now everything looks good to me.
@adamsitnik do you have any comments?

adamsitnik:
@lukasz-pyrzyk big thanks!

PR #548, opened 2017-09-14
Fix condition in package reference list and update dotnet cli version from 1.0.4 to 2.0.0 for non-Windows system

Ky7m:


AndreyAkinshin:
@Ky7m, thanks!

src/BenchmarkDotNet.Core/BenchmarkDotNet.Core.csproj
```diff
@@ -23,7 +23,7 @@
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' Or '$(TargetFramework)' == 'netcoreapp1.1' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' Or '$(TargetFramework)' == 'netcoreapp2.0' ">
```
adamsitnik: good catch! I have no idea how did it work without it ;)

Ky7m: @adamsitnik, does TargetFrameworkIdentifier in condition can simplify clause, example
https://github.com/aspnet/MusicStore/blob/dev/samples/MusicStore/MusicStore.csproj#L20,L24
what do you think?

AndreyAkinshin: It makes much more sense to me. I think it's a good idea to use `TargetFrameworkIdentifier` everywhere (not only for `.NETCoreApp`, but also for `.NETFramework`).

adamsitnik: @Ky7m very good idea!

Ky7m: Great, let me grab this item and I'll send a PR

Issue #529, opened 2017-08-19
No namespace in export filenames can lead to data loss


jskeet:
Yesterday I added benchmarks to NodaTime for a type, NodaTime.TimeZones.StandardDaylightAlternatingMap. I put the benchmarks in  the NodaTime.Benchmarks.NodaTimeTests.TimeZones namespace.

As it happens, I already *had* some benchmarks for that type, in the NodaTime.Benchmarks.NodaTimeTests namespace. When I looked at the results of my overnight run, only the new tests existed. That's because the JSON file that's generated only includes the short type name, with no namespace.

That's probably great for cases when you're looking at results by hand and don't want long filenames, but in my case I'd rather have long filenames (which are being consumed automatically anyway) and no loss of data.

Perhaps this should be an exporter option?

AndreyAkinshin:
Hey @jskeet, thanks for the report.

I think we always should use fully qualified names for all benchmarks without additional exporter options.

Teknikaali:
I dug around the code and this is what I discovered about exported files' naming conventions:

`Summary.Title` property is being used for **both** the _filename_ and the actual _title of the benchmark_.
Currently `Summary.Title` is generated by using `Type.Name`, but it could be change to `Type.FullName` to enable fully qualified filenames. 

This change would cause a small side effect to the actual output data by making the `Title` value a bit longer. If this is a wanted result then it's ok and nothing to worry about.. The results of this change would look like this:

Property used | Filename | Example JSON output snip
---------- | ---------- | ------
**Type.Name** | IntroExporters-report-brief.json | `"Title":"IntroExporters",` 
**Type.FullName** | BenchmarkDotNet.Samples.Intro.IntroExporters-report-brief.json | `"Title":"BenchmarkDotNet.Samples.Intro.IntroExporters",`

I can start fixing this, if these changes look good to you!

adamsitnik:
@Teknikaali IMHO the best option would be to leave the `Summary.Title` untouched and use the `benchmar.Target.FullName` when creating files

@AndreyAkinshin what do you think?

Teknikaali:
@adamsitnik I think you're right, it should work nicely without any side effects if the `Benchmark.Target.Type.FullName` property would be used. I totally missed this this one!

This then leaves only one more question about the filename handling inside the `ExporterBase` class:
` BenchmarkRunnerCore.GetTitle(...)` is a method that parses the title from a _collection_ of `Benchmark`s.
[...]

PR #552, opened 2017-09-21
Fix exporters to use fully qualified filenames

Teknikaali:
This PR fixes exporters to use fully qualified filenames instead of simple type names.
## Issue
#529

##  Features
- Changes how `ExporterBase` handles file naming:
  - Uses `Type.FullName` instead of `Type.Name` as part of the filename
  - Uses `Summary.Title` as the filename, if multiple benchmarks have been joined into a single `Summary`
    - This convention follows the way `Summary.Title` is generated in [`BenchmarkRunnerCore r60`](https://github.com/dotnet/BenchmarkDotNet/blob/174c19dced6883c9f5d810303b7e969739e1cc2b/src/BenchmarkDotNet.Core/Running/BenchmarkRunnerCore.cs#L60)
- Adds two integration tests for the aforementioned features inside `BenchmarkDotNet.IntegrationTests.ExporterIOTests`:
  1) Asserts the exporter uses fully qualified type name as part of the filename
  2) Asserts joined summaries use `Summary.Title` as filename instead of the fully qualified type name

### Example filename
Before: `IntroExporters-report-brief.json`
After: `BenchmarkDotNet.Samples.Intro.IntroExporters-report-brief.json`

## Other notes
- These changes won't take into account multiple classes with a same name inside different namespaces when _joining summaries of multiple benchmarks_
  - For example benchmarks inside `NamespaceA.ClassA` and `NamespaceB.ClassA` wouldn't be considered as two different benchmark classes _for the filename_ when joining. `ClassA-report.txt` would be used as the filename instead of `BenchmarkRun-001-2017-09-21-03-42-28-report.txt`
> I think this is a rare corner case so I left its handling out of for now

I think that's about it! ✔️
Let me know if I missed something or if something needs improvement.

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/ExporterIOTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ExporterIOTests.cs
@@ -5,6 +5,7 @@ using BenchmarkDotNet.Exporters;
 using BenchmarkDotNet.IntegrationTests.Xunit;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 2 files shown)


tests/BenchmarkDotNet.IntegrationTests/ExporterIOTests.cs
```diff
@@ -90,5 +91,47 @@ public override void ExportToLog(Summary summary, ILogger logger)
                 ExportCount++;
             }
         }
+
+        [Fact]
+        public void ExporterUsesFullyQualifiedTypeNameAsFileName()
+        {
+            var exporter = new MockExporter();
+            var summary = CanExecute<ClassA>();
```
adamsitnik: could you create some `Summary` without running the benchmarks?

We are currently having a problem with too long builds. We have too many integration tests that run benchmarks and we exceed the 1h we have in our current plan :/

Teknikaali: Yes I'll look into it! AppVeyor currently reports these durations for the added tests:

Test | Duration
----- | ---------
ExporterUsesFullyQualifiedTypeNameAsFileName | 28 sec 293 ms
ExporterUsesSummaryTitleAsFileNameWhenBenchmarksJoinedToSingleSummary | 3 min 51 sec 219 ms

I suppose its possible to get these down to a couple of milliseconds each, if I mock out the `Summary` correctly.

adamsitnik: > I suppose its possible to get these down to a couple of milliseconds each, if I mock out the Summary correctly.

exactly! thanks for understanding!

adamsitnik:
@Teknikaali thanks a lot!

Issue #556, opened 2017-09-26
Table in report is broken in VSCode markdown viewer


qbit86:
Steps to reproduce: open `*-report-github.md` in Visual Studio Code, open markdown preview (Ctrl+Shift+V by default on Windows).

Observed result: columns in table body are shifted right comparing to table header.

This is probably the bug in VSCode parsing/rendering — they treat leading space character in a row as empty column. But it can be fixed on your side, if you will not output leading space before table border.

Thanks!

PR #563, opened 2017-09-29
Remove leading space character in a MD table row, #556

rolshevsky:


adamsitnik:
thank you @rolshevsky  ! good job!

Issue #360, opened 2017-02-10
[Suggestion] Duplicates handling for IConfig


ig-sinicyn:
Current implementations of `CompositeXxx` wrappers (`CompositeLogger`, `CompositeDiagnoser` etc) do not have common policy for duplicate values.

For example, the `CompositeDiagnoser` [removes duplicates](https://github.com/dotnet/BenchmarkDotNet/blob/f1f2317dafc663a2fa854ea566ede8befc68ea6d/src/BenchmarkDotNet.Core/Diagnosers/CompositeDiagnoser.cs#L17) but the `CompositeAnalyser` [stores them as-is](https://github.com/dotnet/BenchmarkDotNet/blob/2f85847cdf077e695f0e65b71ed9f1b5faac7d5c/src/BenchmarkDotNet.Core/Analysers/CompositeAnalyser.cs#L14)

**Option one**
Add `.Distinct()` calls at `ConfigExtensions.GetCompositeXxx()` [methods](https://github.com/dotnet/BenchmarkDotNet/blob/24e944ae06101a8051955fcfb0ae6c4413950fdc/src/BenchmarkDotNet.Core/Configs/ConfigExtensions.cs#L16-L20).

**Option two**
Introduce explicit config preparation step that will create full config and remove duplicates at the end. Make config preparation to be responsibility of the `BenchmarkRunner`. 
The `BenchmarkRunnerCore` should receive a not-null prepared config parameter and should not call 
`BenchmarkConverter.GetFullConfig()` by itself. For null configs ArgumentNullException should be thrown.

I vote for option two with two hands as it'll solve console dependency issue at #356 (will allow to remove the `FallbackLogger`). Moreover it will simplify code for my own perftest project as it uses its own [configuration subsystem](https://github.com/rsdn/CodeJam/blob/master/PerfTests/docs/ConfigurationSystem.md) that conflicts with `IConfigSource` attributes.

PR #565, opened 2017-10-08
Single point of full config creation

ig-sinicyn:
This commit tries to solve the issue of full config creation. 
Some parts of current code create full configs [three times during benchmark run](https://github.com/dotnet/BenchmarkDotNet/blob/5fe5647a158b41d0c2c949b6e04bca44ec5ff1d7/tests/BenchmarkDotNet.IntegrationTests/BenchmarkTestExecutor.cs#L64), others [do not create full config at all](https://github.com/dotnet/BenchmarkDotNet/blob/f2b9c0750fe236d9f4dc80571a58392d0ec7660d/src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs#L77).

The fix is following:
* A new type, `BenchmarkRunInfo` is introduced. It keeps together benchmarks, a full config and a type for which the config was created. In case when there are multiple types in a single run (like with `join` switch) the current behavior is preserved: config is created using first type.
* `XxxToBenchmarks()` methods now return `BenchmarkRunInfo`, runner methods that accept `BenchmarkRunInfo` DO NOT create full config, all others use full config that returned from `XxxToBenchmarks()` call.

As a bonus I've added a `ReadOnlyConfig` class that allows to prevent errors like "passing a partial config to methods that expect full one". The class may be removed if not needed.

P.S. The build fails due to https://github.com/dotnet/BenchmarkDotNet/commit/9076a69f4e4e471d7820e12aef9b736c6d64a60f commit, branch build is fine, https://ci.appveyor.com/project/ig-sinicyn/benchmarkdotnet/build/0.10.9.47

### DIFF
--- a/src/BenchmarkDotNet.Core/Configs/ReadOnlyConfig.cs
+++ b/src/BenchmarkDotNet.Core/Configs/ReadOnlyConfig.cs
@@ -19,7 +19,6 @@ namespace BenchmarkDotNet.Configs
     [PublicAPI]
     public class ReadOnlyConfig : IConfig
     {
-        #region Fields & .ctor
         private readonly IConfig config;
 
         public ReadOnlyConfig([NotNull] IConfig config)
@@ -29,9 +28,7 @@ namespace BenchmarkDotNet.Configs
 
             this.config = config;
         }
-        #endregion
 
-        #region IConfig implementation
         public IEnumerable<IColumnProvider> GetColumnProviders() => config.GetColumnProviders();
         public IEnumerable<IExporter> GetExporters() => config.GetExporters();
         public IEnumerable<ILogger> GetLoggers() => config.GetLoggers();
@@ -48,7 +45,5 @@ namespace BenchmarkDotNet.Configs
         public ConfigUnionRule UnionRule => config.UnionRule;
 
         public bool KeepBenchmarkFiles => config.KeepBenchmarkFiles;
-
-        #endregion
     }
 }
--- a/src/BenchmarkDotNet.Core/Configs/ReadOnlyConfig.cs
+++ b/src/BenchmarkDotNet.Core/Configs/ReadOnlyConfig.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+
+using BenchmarkDotNet.Analysers;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Diagnosers;
+using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.Filters;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Order;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Validators;
+
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Configs
+{
+    [PublicAPI]
+    public class ReadOnlyConfig : IConfig
+    {
+        #region Fields & .ctor
+        private readonly IConfig config;
+
+        public ReadOnlyConfig([NotNull] IConfig config)
+        {
+            if (config == null)
+                throw new ArgumentNullException(nameof(config));
+
+            this.config = config;
+        }
+        #endregion
+
+        #region IConfig implementation
+        public IEnumerable<IColumnProvider> GetColumnProviders() => config.GetColumnProviders();
+        public IEnumerable<IExporter> GetExporters() => config.GetExporters();
+        public IEnumerable<ILogger> GetLoggers() => config.GetLoggers();
+        public IEnumerable<IDiagnoser> GetDiagnosers() => config.GetDiagnosers();
+        public IEnumerable<IAnalyser> GetAnalysers() => config.GetAnalysers();
+        public IEnumerable<Job> GetJobs() => config.GetJobs();
+        public IEnumerable<IValidator> GetValidators() => config.GetValidators();
+        public IEnumerable<HardwareCounter> GetHardwareCounters() => config.GetHardwareCounters();
+        public IEnumerable<IFilter> GetFilters() => config.GetFilters();
+
+        public IOrderProvider GetOrderProvider() => config.GetOrderProvider();
+        public ISummaryStyle GetSummaryStyle() => config.GetSummaryStyle();
+
+        public ConfigUnionRule UnionRule => config.UnionRule;
+
+        public bool KeepBenchmarkFiles => config.KeepBenchmarkFiles;
+
+        #endregion
+    }
+}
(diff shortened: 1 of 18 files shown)


src/BenchmarkDotNet.Core/Running/ClassicBenchmarkConverter.cs
```diff
@@ -66,18 +68,30 @@ public static Benchmark[] SourceToBenchmarks(string source, IConfig config = nul
...
+
+            var resultBenchmarks = new List<BenchmarkRunInfo>();
+            foreach (var type in types)
+            {
+                var runInfo = TypeToBenchmarks(type, config);
+                var benchmarks = runInfo.Benchmarks.Select(b =>
+                {
+                    var target = b.Target;
+                    return new Benchmark(
+                        new Target(target.Type, target.Method, target.GlobalSetupMethod, target.GlobalCleanupMethod,
```
ig-sinicyn: I'm not sure if the `new Target()` part is required at all, code left as is.

src/BenchmarkDotNet.Core/Configs/ReadOnlyConfig.cs
```diff
@@ -0,0 +1,54 @@
...
+using BenchmarkDotNet.Validators;
+
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Configs
+{
+    [PublicAPI]
+    public class ReadOnlyConfig : IConfig
+    {
+        #region Fields & .ctor
```
adamsitnik: what is the point of adding regions to such simple class?

src/BenchmarkDotNet.Core/Running/BenchmarkRunnerCore.cs
```diff
@@ -31,10 +30,21 @@ public static class BenchmarkRunnerCore
 
         internal static readonly IResolver DefaultResolver = new CompositeResolver(EnvResolver.Instance, InfrastructureResolver.Instance);
 
-        public static Summary Run(Benchmark[] benchmarks, IConfig config, Func<Job, IToolchain> toolchainProvider)
+        public static Summary Run(BenchmarkRunInfo[] benchmarkRunInfos, Func<Job, IToolchain> toolchainProvider)
+        {
+            var temp = benchmarkRunInfos.FirstOrDefault();
+            var benchmarkRunInfo = new BenchmarkRunInfo(
```
adamsitnik: I don't understand this idea.

You have an array of infos, each may have different Type and Config. But before you run all of them you choose the type and config of the first info.

What if every info has it's own config and/or different type?  Could you explain?

src/BenchmarkDotNet.Core/Running/ClassicBenchmarkConverter.cs
```diff
@@ -1,9 +1,11 @@
 ﻿#if CLASSIC
 using System;
 using System.CodeDom.Compiler;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Runtime.Remoting.Activation;
```
adamsitnik: what is the purpose of having the dependency to `System.Runtime.Remoting.Activation`?

src/BenchmarkDotNet/Running/BenchmarkRunner.cs
```diff
@@ -9,21 +10,36 @@ namespace BenchmarkDotNet.Running
...
 
         public static Summary Run(Type type, MethodInfo[] methods, IConfig config = null) =>
-            BenchmarkRunnerCore.Run(BenchmarkConverter.MethodsToBenchmarks(type, methods, config), config, ToolchainExtensions.GetToolchain);
+            BenchmarkRunnerCore.Run(BenchmarkConverter.MethodsToBenchmarks(type, methods, config), ToolchainExtensions.GetToolchain);
 
-        public static Summary Run(Benchmark[] benchmarks, IConfig config) =>
-            BenchmarkRunnerCore.Run(benchmarks, config, ToolchainExtensions.GetToolchain);
+        public static Summary Run(Benchmark[] benchmarks, IConfig config)
+        {
+            var targetType = benchmarks?.FirstOrDefault()?.Target.Type;
```
adamsitnik: again: why do you assume that all Benchmarks have the same Target?

src/BenchmarkDotNet/Running/BenchmarkRunner.cs
```diff
@@ -9,21 +10,36 @@ namespace BenchmarkDotNet.Running
...
+            return BenchmarkRunnerCore.Run(
+                new BenchmarkRunInfo(benchmarks, targetType, BenchmarkConverter.GetFullConfig(targetType, config)),
+                ToolchainExtensions.GetToolchain);
+        }
+
+        public static Summary Run(BenchmarkRunInfo benchmarks)
+        {
```
adamsitnik: Please use `=>` from C# 7.0 for such short methods

src/BenchmarkDotNet.Core/Running/BenchmarkRunnerCore.cs
ig-sinicyn: @adamsitnik I dislike it too but I cannot imagine anything better. Current code preserves behavior from Benchmark switcher with --join switch and BenchmarkConverter's `BenchmarksFromSource()`.

src/BenchmarkDotNet.Core/Configs/ReadOnlyConfig.cs
ig-sinicyn: Will fix:)

src/BenchmarkDotNet/Running/BenchmarkRunner.cs
ig-sinicyn: No problem)

ig-sinicyn:
@adamsitnik 
Pushed the fixes:)

>I am not sure if the fact that we created config few times was a problem. On the other hand you want to change a contract of public methods here.

Well, there is a problem with custom config attributes. Multiple calls of the same attribute will add multiple instances of logger/exporter/diagnoser etc into resulting config and there's no code that removes duplicates from resulting code. This may be fixed by altering the `GetFullConfig` so this is not a strong argument, definitely:)
Also, with current code there is no way to pass custom config into `BenchmarkRunnerCore` as is, without altering it via config attributes.

Technically, both of these may be fixed without changing the public API. But without explicit API contract it is very easy to broke fixed behavior accidentally, just by calling overload that calls `GetFullConfig(fullConfig)` or forgetting to call it on partial config.

>could you create an issue next time before creating the PR?

Ok.

src/BenchmarkDotNet.Core/Running/ClassicBenchmarkConverter.cs
ig-sinicyn: Fixed

adamsitnik:
@ig-sinicyn big thanks for the explanation and fixes!

> s. Multiple calls of the same attribute will add multiple instances of logger/exporter/diagnoser etc into resulting config and there's no code that removes duplicates from resulting code.

now I get it! Does it mean that this PR closes #360 ?

ig-sinicyn:
@adamsitnik 
Well, almost:)
There still be duplicates if there are multiple attributes that do add same exporters/diagnosers/etc but at least there's now is a single point (the `GetFullConfig()` method) where all duplicates may be removed.

I did not fix it as I'm not sure how to implement duplicate detection. Should all items expose smth like `Id` property or it will be better to implement IEquatable<T> on base types?

PR #569, opened 2017-10-18
Update cakebuild scripts

Ky7m:
- Add installation step to get the required .NET Core runtime to run backward compatibility tests (Windows, Linux).
- Clean up travis CI definition (since .NET Core 2.0 openssl prerequisite is not required).
- Replace CleanDirectories(delete and create) and  with DeleteDirectories(only delete) command.
- Updated Cake tool to version 0.23.0
- Fix prerequisite step that is required to run Tests tasks properly (e.g. build.ps1 -Target BackwardCompatibilityTests)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ os:
   - osx
 sudo: required
 dist: trusty
+osx_image: xcode9.1 # OS X 10.12
 mono: latest
 
 addons:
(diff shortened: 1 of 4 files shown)


adamsitnik:
@Ky7m big thanks for doing this!

Ky7m:
@adamsitnik, my pleasure!

PR #573, opened 2017-10-30
Сreate a tail call diagnoser

GeorgePlotnikov:
Added diagnoser for JIT Tail Call Events, based on JIT Inlining Events diagnoser.


src/BenchmarkDotNet.Diagnostics.Windows/TailCallDiagnoser.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        private readonly bool logFailuresOnly = true;
+        private readonly bool filterByNamespace = true;
+        private string expectedNamespace;
+
+        public TailCallDiagnoser()
+        {
+
+        }
+        /// <summary>
```
adamsitnik: nit:  new line missing

src/BenchmarkDotNet.Diagnostics.Windows/TailCallDiagnoser.cs
```diff
@@ -0,0 +1,75 @@
...
+                        Logger.WriteLineHelp($"Caller: {jitData.CallerNamespace}.{jitData.CallerName} - {jitData.CallerNameSignature}");
+                        Logger.WriteLineHelp($"Callee: {jitData.CalleeNamespace}.{jitData.CalleeName} - {jitData.CalleeNameSignature}");
+                        Logger.WriteLineError($"Fail Reason: {jitData.FailReason}");
+                        Logger.WriteLineHeader(LogSeparator);
+                    }
+                }
+            };
+        }
+
+        private bool ShouldPrintEventInfo(Tuple<string, string> namespaces) => !filterByNamespace || namespaces.Item1.StartsWith(expectedNamespace) || namespaces.Item2.StartsWith(expectedNamespace);
```
adamsitnik: nit: you could use ValueTuple here to avoid heap allocation and for better readability (no need to use ItemX anymore)

tests/BenchmarkDotNet.IntegrationTests/ValuesReturnedByBenchmarkTest.cs
```diff
@@ -30,6 +30,12 @@ public Diagnostics.Windows.InliningDiagnoser TypeFromCustomDependency()
             {
                 return new Diagnostics.Windows.InliningDiagnoser();
             }
+
+            [Benchmark]
+            public Diagnostics.Windows.TailCallDiagnoser TypeFromCustomDeoendency1()
```
adamsitnik: The test which returns `InliningDiagnoser` verifies that we can have a benchmark which returns type from custom dll. Your test checks the same thing. It does not add any value and it prolongs the build. As of today we are very close of reaching the max time per build now so we try not to duplicate the tests. Please remove it

GeorgePlotnikov:
@adamsitnik thank you very much for your comments, I'll take them into consideration to update the branch. Regarding the F# the case, I'm interested in it, no doubts. If you have some ideas where shall I start the research I would appreciate you a lot.

adamsitnik:
@GeorgePlotnikov [this](https://blogs.msdn.microsoft.com/fsharpteam/2011/07/08/tail-calls-in-f/) blog post gives a good introduction. However I don't have any link to a post which shows what decisions can be made by the F# compiler. 

@dsyme maybe you could provide us some hints for start?

src/BenchmarkDotNet.Diagnostics.Windows/TailCallDiagnoser.cs
```diff
@@ -0,0 +1,79 @@
...
+        private readonly bool filterByNamespace = true;
+        private string expectedNamespace;
+
+        public TailCallDiagnoser()
+        {
+
+        }
+
+        /// <summary>
+        /// creates the new TailCallDiagnoser instace
```
AndreyAkinshin: Typo: "instace" -> "insta**n**ce"

GeorgePlotnikov:
@AndreyAkinshin seems @adamsitnik refused the previously written integration test, because of it was the same as for inline diagnoser, so it was removed by the second commit.

AndreyAkinshin:
@GeorgePlotnikov if I understand correctly, the original test just checked that we can do some basic diagnostics stuff. But we need a test which checks the diagnoser output (with lines like Caller, Callee, Tail prefix, Tail call type, and so on). It will help to verify that diagnoser is still working in case of any changes. Right now I have to check it manually before each release, but it will be better to make it automatic and increase the coverage.
The total test suite duration is still an issue, but we have to find another way to solve it.

GeorgePlotnikov:
@AndreyAkinshin you mean something like BenchmarkDotNet.IntegrationTests.MemoryDiagnoserTests? I think what you are talking about should look like that. Does it make sense?

adamsitnik:
> you mean something like BenchmarkDotNet.IntegrationTests.MemoryDiagnoserTests?

yes, something similar. Run some benchmark, with the new diagnoser and check the output of the diagnoser (info about methods reported to the output).

> refused the previously written integration test, because of it was the same as for inline diagnoser

your previous test was just checking if we can have a benchmark, which returns new diagnoser. It was not using the diagnoser. `AnyValueCanBeReturned` test verifies that we build the right code for all possible use cases of benchmarks (returning different types)

GeorgePlotnikov:
@AndreyAkinshin , @adamsitnik thank you guys for the review. I'll add the test.

GeorgePlotnikov:
@AndreyAkinshin, @adamsitnik hi guys, I've just added the integration test, please review

tests/BenchmarkDotNet.IntegrationTests/TailCallDiagnoserTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        private LogCapture Execute<T>(Jit jit, Platform platform, Runtime runtime)
+        {
+            var tailCallDiagnoser = new TailCallDiagnoser(false, true);
+            CanExecute<T>(CreateConfig(jit, platform, runtime, tailCallDiagnoser));
+            var output = (LogCapture)(tailCallDiagnoser as EtwDiagnoser<object>).GetType().GetField("Logger", BindingFlags.Instance | BindingFlags.NonPublic).GetValue(tailCallDiagnoser);
+            return output;
+        }
+
+        private IConfig CreateConfig(Jit jit, Platform platform, Runtime runtime, TailCallDiagnoser diagnoser) => ManualConfig.CreateEmpty()
+            .With(Job.ShortRun.With(jit).With(platform).With(runtime))
```
AndreyAkinshin: Am I right that we don't need actual benchmark results here? Could we use the dry job here (instead of `ShortRun`)?

tests/BenchmarkDotNet.IntegrationTests/TailCallDiagnoserTests.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+#if NET46
+    public class TailCallDiagnoserTests : BenchmarkTestExecutor
```
AndreyAkinshin: Am I right that these tests should work only on Windows?
`#if NET46` requires only full .NET Framework (not .NET Core), but it's a valid profile for Mono+Linux/macOS.
If you have tests which should work only on Windows, please use `TheoryWindowsOnlyAttribute`
https://github.com/dotnet/BenchmarkDotNet/blob/696f89908e82118ecf9989c2484c23a6f52ba6b2/tests/BenchmarkDotNet.IntegrationTests/Xunit/TheoryWindowsOnlyAttribute.cs

AndreyAkinshin:
@GeorgePlotnikov, could you also add some information about this diagnoser in our documentation (the `docs` folder)? It would be nice to say a few words about what is it, when it can be useful, show a few benchmark examples (e.g. on C# and F#), and show an example of the diagnoser output.

tests/BenchmarkDotNet.IntegrationTests/TailCallDiagnoserTests.cs
GeorgePlotnikov: True, it's working only for windows with full .NET Framework, thus I may use `TheoryWindowsOnlyAttribute` plus `#if NET46` to fix compile time errors, related to `BenchmarkDotNet.Diagnostics.Windows.Configs` which does not support .NET CORE. Does it make sense?

tests/BenchmarkDotNet.IntegrationTests/TailCallDiagnoserTests.cs
GeorgePlotnikov: True, thanks for the advice

tests/BenchmarkDotNet.IntegrationTests/TailCallDiagnoserTests.cs
AndreyAkinshin: Yes.

GeorgePlotnikov:
Applied the request review comments to the integration test and pulled the [request](https://github.com/dotnet/BenchmarkDotNet/pull/598) to update the documentation

PR #575, opened 2017-11-02
xUnit runner upgrade and small tests refactoring

Ky7m:
The main goal of this PR is switching from `dotnet test` to `dotnet xunit` (**note**: it doesn't replace, it just adds an additional ability to run test using `dotnet xunit` command) and upgrade packages.

There are two main reasons:

1. New process (restore, build, test and pack) is a bit faster and we can save up to 5 minutes in the current configuration.
2. With `dotnet xunit` tool we can have more flexibility to configure test runs, e.g. specify **-stoponfail** switch to stop running tests after the first test failure (as result, we can get feedback more quickly).

In addition, NuGet packages were upgraded and consolidated around all projects:

- System.ComponentModel.EventBasedAsync -> 4.3.0
- Microsoft.DotNet.PlatformAbstractions -> 2.0.0
- Microsoft.Win32.Registry -> 4.4.0
- System.ValueTuple -> 4.4.0
- Microsoft.Diagnostics.Runtime -> 0.9.170809.3
- Microsoft.CodeAnalysis.CSharp -> 2.4.0
- Microsoft.NETCore.Platforms -> 2.0.0
- Microsoft.NET.Test.Sdk -> 15.3.0
- System.Runtime.CompilerServices.Unsafe -> 4.4.0
- xunit.runner.visualstudio -> 2.3.1
- xunit -> 2.3.1
- dotnet-xunit -> 2.3.1

Finally, for all test projects were added Roslyn analyzers for xUnit ([xunit.analyzers](https://xunit.github.io/xunit.analyzers/rules/)) and all violations of best practices/rules have been fixed.


adamsitnik:
> New process (restore, build, test and pack) is a bit faster and we can save up to 5 minutes in the current configuration.

This is great!

However I am not sure about updating the dependencies of BDN. As a library we should reference lowest version possible. But on the other hand, we are never a dependency for production app.

@AndreyAkinshin what do you think about updating our dependencies?

AndreyAkinshin:
There are no any problems to update dependencies in Integration Tests or Samples. However, why do we have to update dependencies of BenchmarkDotNet.Core? I agree with the idea about lowest possible versions. I'm also ready to discuss each dependency. It's ok to update it if we really need the newest version, if we have *a reason* for it. Otherwise, I prefer don't change it. It can be a problem if somebody want to compare performance between different versions of our dependencies (e.g., System.ValueTuple 4.3.0 vs. 4.4.0).

Ky7m:
It is no problem to revert back all changes in BenchmarkDotNet.Core related to packages. Should I do only for BenchmarkDotNet.Core project or for following projects:

- BenchmarkDotNet.Disassembler.*.csproj - Microsoft.Diagnostics.Runtime - from 0.9.170809.3 to 0.9.170626.1
- BenchmarkDotNet.Toolchains.Roslyn.csproj - Microsoft.CodeAnalysis.CSharp - from 2.4.0 to 2.0.0 

as well?

adamsitnik:
@Ky7m I think that it is good to update Roslyn dependency. We can get some benefits.

Ky7m:
Ok, give me some time to update PR (you know that Nuget is down due to Azure DC outage) and verify everything one more time. Thank you for your feedback folks.

adamsitnik:
@Ky7m it looks very good to me know! Please let me know when you remove the `WIP` term and I merge it!

Once again very big thanks!

Ky7m:
@adamsitnik, I'm done.

PR #581, opened 2017-11-16
Add Debug and Release configuration for *nix based systems.

mfilippov:


AndreyAkinshin:
@mfilippov, the only problem with the original Debug/Release configuration is VisualBasic project compilation. Could we transform it into .NET Core project instead?

PR #590, opened 2017-11-21
Add precision to Scaled Column

### LINKED ISSUE #509: Better formatting for the Scaled column
Subj: https://twitter.com/Nick_Craver/status/891321344364883969

If there are values in the Scaled column which are too small (like 0.02-0.01 or less), the amount of digits after the decimal point should be automatically increased.

Chrisgozd:
Fixes #509 

Is this all that needs to be done?


adamsitnik:
hi @Chrisgozd thanks for your PR!

@AndreyAkinshin is this everything we need?

Chrisgozd:
Thanks for the feedback.

1) Advice on how this should be done? It seems like the column calculation is independent of each other, so row 1 scale is calculated, then row 2 scale is calculated.  I could calculate the mean for all columns each time the scale column is calculated and if any of the means is < 0.01, use N3, otherwise N2....

AndreyAkinshin:
@Chrisgozd, unfortunately, we don't have a nice way to do it right now. Probably, we have to refactor columns.
However, I guess that it's fine to calculate the precision each time (for now).

samples/BenchmarkDotNet.Samples/Program.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.Reflection;
-using BenchmarkDotNet.Running;
-
+using BenchmarkDotNet.Running;
+
```
AndreyAkinshin: Please, don't change this file.

src/BenchmarkDotNet.Core/Columns/BaselineScaledColumn.cs
```diff
@@ -72,15 +72,23 @@ public string GetValue(Summary summary, Benchmark benchmark)
...
-                    if (baselineStat.N < 2 || targetStat.N < 2)
-                        return "NA";
-                    double pvalue = WelchTTest.Calc(baselineStat, targetStat).PValue;
-                    return pvalue > 0.0001 || pvalue < 1e-9 ? pvalue.ToStr("N4") : pvalue.ToStr("e2");
+                case DiffKind.WelchTTestPValue:
+                {
+                    if (baselineStat.N < 2 || targetStat.N < 2)
+                        return "NA";
+                    double pvalue = WelchTTest.Calc(baselineStat, targetStat).PValue;
+                    return pvalue > 0.0001 || pvalue < 1e-9 ? pvalue.ToStr("N4") : pvalue.ToStr("e2");
```
AndreyAkinshin: Please, don't change these files.

src/BenchmarkDotNet.Core/Columns/BaselineScaledColumn.cs
```diff
@@ -72,15 +72,23 @@ public string GetValue(Summary summary, Benchmark benchmark)
             switch (Kind)
             {
                 case DiffKind.Mean:
+                    var nonBaselines = summary.Benchmarks.
+                        Where(b => b.Job.DisplayInfo == benchmark.Job.DisplayInfo).
+                        Where(b => b.Parameters.DisplayInfo == benchmark.Parameters.DisplayInfo).
+                        Where(b => !b.Target.Baseline);
+                    if (nonBaselines.Any(x => Statistics.DivMean(summary[x].ResultStatistics, baselineStat) < 0.01))
```
AndreyAkinshin: Could we introduce a separate method for this check?

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public ScaledPrecisionTests(ITestOutputHelper output)
+        {
+            this.output = output;
+        }
+
+        [Theory]
+        [InlineData(140)]
+        [InlineData(50)]
+        public void ScaledPrecisionTests(int baselineValue)
```
AndreyAkinshin: I can't build it.
```
"W:\Work\BenchmarkDotNet\BenchmarkDotNet.sln" (default target) (1) ->
"W:\Work\BenchmarkDotNet\tests\BenchmarkDotNet.Tests\BenchmarkDotNet.Tests.csproj" (default target) (9) ->
"W:\Work\BenchmarkDotNet\tests\BenchmarkDotNet.Tests\BenchmarkDotNet.Tests.csproj" (Build target) (9:2) ->
(CoreCompile target) ->
  Reports\ScaledPrecisionTests.cs(34,21): error CS0542: 'ScaledPrecisionTests': member names cannot be the same as their enclosing type [W:\W ork\BenchmarkDotNet\tests\BenchmarkDotNet.Tests\BenchmarkDotNet.Tests.csproj]
```

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        private readonly ITestOutputHelper output;
+
+        public ScaledPrecisionTests(ITestOutputHelper output)
+        {
+            this.output = output;
+        }
+
+        [Theory]
+        [InlineData(140)]
+        [InlineData(50)]
```
AndreyAkinshin: `0` is also a valid baseline. Please, add this test and make sure that everything works fine.
Also, we need a test with 3 benchmark methods: a baseline, a method with scaled value > 0.01, and a method with scaled value < 0.01.

Chrisgozd: If the baseline is 0, would you expect all Scaled columns to be "?" ?

AndreyAkinshin: Yes, it's the expected output in the current version. The main thing that we should check: the new code doesn't throw any exception in such cases. It will help to refactor this logic in the future.

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -29,27 +29,41 @@ public ScaledPrecisionTests(ITestOutputHelper output)
...
+        [InlineData(new int[] { 140, 1, 50 })]
+        [InlineData(new int[] { 40, 1, 20 })]
+        [InlineData(new int[] { 0, 1, 20 })]
+        // First value is baseline, others are benchmark measurements
+        public void ScaledPrecisionTestWithBaseline(int[] values)
         {
-            var summary = CreateSummary(baselineValue);
+            var summary = CreateSummary(values);
             var scaledIndex = Array.FindIndex(summary.Table.FullHeader, c => c == "Scaled");
             var testNameColumn = Array.FindIndex(summary.Table.FullHeader, c => c == "Method");
```
AndreyAkinshin: Why do we need `testNameColumn` here? Could we remove this line?

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        [Theory]
+        [InlineData(new int[] { 140, 1, 50 })]
+        [InlineData(new int[] { 40, 1, 20 })]
+        [InlineData(new int[] { 0, 1, 20 })]
+        // First value is baseline, others are benchmark measurements
+        public void ScaledPrecisionTestWithBaseline(int[] values)
+        {
+            var summary = CreateSummary(values);
+            var scaledIndex = Array.FindIndex(summary.Table.FullHeader, c => c == "Scaled");
+            var testNameColumn = Array.FindIndex(summary.Table.FullHeader, c => c == "Method");
```
AndreyAkinshin: Why do we need `testNameColumn` here? Could we remove this line?

AndreyAkinshin:
@Chrisgozd, our appveyor problem is finally resolved. Could you merge master into your branch (thus, we will be able to check your PR on appveyor).

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            foreach (var row in summary.Table.FullContent)
+            {
+                // Check to see if scale value has a decimal point, otherwise its probably "?"
+                if (row[scaledIndex].Split('.').Count() > 0)
```
AndreyAkinshin: It's better to use `.Any()` instead of `.Count() > 0`

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        private readonly ITestOutputHelper output;
+
+        public ScaledPrecisionTests(ITestOutputHelper output)
+        {
+            this.output = output;
+        }
+
+        [Theory]
+        [InlineData(new int[] { 140, 1, 50 })]
```
AndreyAkinshin: You can skip `int` and write just `[InlineData(new[] { 140, 1, 50 })]`

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            output.WriteLine(logger.GetLog());
+            return summary;
+        }
+
+        private static BenchmarkReport CreateReport(Benchmark benchmark, int measurementValue)
+        {
+            var buildResult = BuildResult.Success(GenerateResult.Success(ArtifactsPaths.Empty, Array.Empty<string>()));
+            var executeResult = new ExecuteResult(true, 0, Array.Empty<string>(), Array.Empty<string>());
+            var measurements = new List<Measurement>();
+                measurements = new List<Measurement>
```
AndreyAkinshin: It makes sense to merge these two lines and assign `measurements` only once.

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Attributes.Jobs;
+using BenchmarkDotNet.Columns;
```
AndreyAkinshin: The `BenchmarkDotNet.Columns` can be safely removed.

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        [InlineData(new int[] { 0, 1, 20 })]
+        // First value is baseline, others are benchmark measurements
+        public void ScaledPrecisionTestWithBaseline(int[] values)
+        {
+            var summary = CreateSummary(values);
+            var scaledIndex = Array.FindIndex(summary.Table.FullHeader, c => c == "Scaled");
+            var testNameColumn = Array.FindIndex(summary.Table.FullHeader, c => c == "Method");
+            foreach (var row in summary.Table.FullContent)
+            {
+                // Check to see if scale value has a decimal point, otherwise its probably "?"
```
AndreyAkinshin: Please, introduce a method with a meaningful name instead of this comment.

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
AndreyAkinshin: And it's even better to write something like `Contains('.')`

tests/BenchmarkDotNet.Tests/Reports/ScaledPrecisionTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                if (row[scaledIndex].Split('.').Count() > 0)
+                {
+                    // check precision of scaled column to be 2 or 3 decimal places
```
AndreyAkinshin: Please, introduce a method with a meaningful name instead of this comment.

PR #592, opened 2017-11-24
Removed a xunit workaround because an issue has been fixed

AlekseiKudelia:
I removed a xunit workaround. An issue has been fixed already.
https://github.com/xunit/xunit/issues/876 and https://github.com/xunit/xunit/pull/1007


AndreyAkinshin:
@AlekseiKudelia, thanks!

PR #595, opened 2017-11-27
Migrate all project to new project system.

mfilippov:


AndreyAkinshin:
LGTM.

adamsitnik:
@mfilippov looks good to me!  Could you just add information to our [building](https://github.com/dotnet/BenchmarkDotNet/blob/master/docs/guide/Contributing/Building.md) doc which version of VS/Rider is required to be able to build it?

PR #597, opened 2017-11-29
Fix typo in WithCustomBuildConfiguration API

benjamin-hodgson:
Technically renaming a parameter is a breaking change. I think in practice it's unlikely that any users of that API specified `buildConfiguraiton` using a named parameter, though.


adamsitnik:
@benjamin-hodgson thanks!

Issue #582, opened 2017-11-16
Print amount of logical and physical core


AndreyAkinshin:
Right now (v0.10.10) we only print the amount of logical cores (via `Environment.ProcessorCount`). It would be nice to also print the amount of physical cores and amount of physical processors. Since there are no managed API for it, we need 3 different implementations (for Windows, Linux, and macOS).

AndreyAkinshin:
We can use this for Windows: https://stackoverflow.com/a/2670568/184842

morgan-kn:
@AndreyAkinshin could you please assign me on this issue?

AndreyAkinshin:
@morgan-kn, I added you to the collaborators list. Please, accept the invitation, then I will be able to assign you.

AndreyAkinshin:
The feature is implemented (PR #607)

PR #607, opened 2017-12-18
Print amount of logical and physical core #582

morgan-kn:


AndreyAkinshin:
@morgan-kn, thanks for the PR! Looks good. I will review it and check on different hardware on Thursday.

src/BenchmarkDotNet.Core/Environments/HostEnvironmentInfo.cs
```diff
@@ -71,7 +75,9 @@ protected HostEnvironmentInfo()
             BenchmarkDotNetVersion = GetBenchmarkDotNetVersion();
             OsVersion = new Lazy<string>(RuntimeInformation.GetOsVersion);
             ProcessorName = new Lazy<string>(RuntimeInformation.GetProcessorName);
-            ProcessorCount = Environment.ProcessorCount;
+            LogicalCoresCount = Environment.ProcessorCount;
+            PhysicalCoresCount = new Lazy<int?>(RuntimeInformation.GetPhysicalCoresCount);
+            PhysicalProcessorsCount = new Lazy<int?>(RuntimeInformation.GetPhysicalProcessorsCount);
```
AndreyAkinshin: I think it would be better to use names like `LogicalCoreCount`, `PhysicalCoreCount`, `PhysicalProcessorCount` (without `s`).

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -136,14 +136,72 @@ internal static string GetProcessorName()
...
+                    // ignored
+                }
+            }
+#endif
+
+            if (IsLinux())
+                return ExternalToolsHelper.ProcCpuInfo.Value.PhysicalCoresCount;
+            
+            if (IsMacOSX())
+                return  int.TryParse(ExternalToolsHelper.Sysctl.Value.GetValueOrDefault("hw.physicalcpu"), out int hwPhysicalcpu) ? hwPhysicalcpu : (int?) null;
```
AndreyAkinshin:
```
return  int.TryParse
      ^^
```
double space

AndreyAkinshin:
It's great that now we have detailed information about CPU, but the main problem now is huge info line in the summary:
```ini
Processor=Intel Core i7-4790K CPU 4.00GHz (Haswell), LogicalCoresCount=8, PhysicalCoresCount=4, PhysicalProcessorsCount=1
```
When you share such line with others on GitHub, some important info is hidden. People have to scroll the summary to the right and it's not a good thing. It would be great to keep the total length of this line < 90 characters (I don't want to split it and introduce another line).
Probably, we can write something like:
```ini
Processor=Intel Core i7-4790K CPU 4.00GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
```
or even:
```ini
Intel Core i7-4790K CPU 4.00GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
```
(`Processor=` can be omitted because it's a pretty obvious part)
I'm not sure about perfect option here, but it should be short.

src/BenchmarkDotNet.Core/Environments/HostEnvironmentInfo.cs
```diff
@@ -85,7 +91,7 @@ protected HostEnvironmentInfo()
         public override IEnumerable<string> ToFormattedString()
         {
             yield return $"{BenchmarkDotNetCaption}=v{BenchmarkDotNetVersion}, OS={OsVersion.Value}";
-            yield return $"Processor={ProcessorName.Value}, ProcessorCount={ProcessorCount}";
+            yield return $"Processor={ProcessorName.Value}, LogicalCoresCount={LogicalCoresCount}, PhysicalCoresCount={PhysicalCoresCount.Value?.ToString() ?? RuntimeInformation.Unknown}, PhysicalProcessorsCount={PhysicalProcessorsCount.Value?.ToString() ?? RuntimeInformation.Unknown}";
```
AndreyAkinshin: This line is huge. Could we introduce additional variables? (Especially for things like `PhysicalCoresCount.Value?.ToString() ?? RuntimeInformation.Unknown`, it's not a good idea to use complicated expression inside very long interpolation string).

src/BenchmarkDotNet.Core/Helpers/ExternalToolsHelper.cs
```diff
@@ -17,7 +17,7 @@ public static class ExternalToolsHelper
         /// Output of the `cat /proc/info` command.
         /// Linux only.
         /// </summary>
-        public static readonly Lazy<Dictionary<string, string>> ProcCpuInfo = LazyDic(RuntimeInformation.IsLinux, "cat", "/proc/cpuinfo", ':');
+        public static readonly Lazy<ProcCpuInfoParser> ProcCpuInfo = LazyProcCpuInfo(RuntimeInformation.IsLinux, "cat", "/proc/cpuinfo");
```
AndreyAkinshin: It's a good idea to replace `Dictionary<string, string>` by `ProcCpuInfoParser`!

src/BenchmarkDotNet.Core/Helpers/ProcCpuInfoParser.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var list = Regex.Split(content ?? "", "\r?\n");
+            foreach (string line in list)
+                if (line.IndexOf(separator) != -1)
+                {
+                    var lineParts = line.Split(separator);
+                    if (lineParts.Length >= 2)
+                    {
+                        if (units == null)
+                            items.Add(units = new Dictionary<string, string>());
```
AndreyAkinshin: I don't like tricks like `items.Add(units = new Dictionary<string, string>())`. It looks nice and smart, but it usually leads to bugs when somebody else tries to modify such code. Also, this method looks complicated. Could we simplify it? For example, we can introduce `[CanBeNull] Dictionary<string, string> ParseSection` method, it should make this code more readable.

src/BenchmarkDotNet.Core/Portability/RuntimeInformation.cs
```diff
@@ -136,14 +136,72 @@ internal static string GetProcessorName()
...
 
             if (IsMacOSX())
                 return ProcessorBrandStringHelper.Prettify(ExternalToolsHelper.Sysctl.Value.GetValueOrDefault("machdep.cpu.brand_string") ?? "");
 
             return Unknown;
         }
 
+        internal static int? GetPhysicalCoresCount()
+        {
+#if !CORE
```
AndreyAkinshin: What about "Windows + .NET Core" case?

tests/BenchmarkDotNet.Tests/Portability/ProcCpuInfoParserTests.cs
```diff
@@ -0,0 +1,436 @@
+﻿using BenchmarkDotNet.Helpers;
+using Xunit;
+
+namespace BenchmarkDotNet.Tests.Portability
+{
+    public class ProcCpuInfoParserTests
+    {
+        [Fact]
+        public void TestEmpty()
```
AndreyAkinshin: In other test classes, we *usually* put `Test` at the end of a test method (e.g. `EmptyTest` instead of `TestEmpty`). Please, follow the convention.

tests/BenchmarkDotNet.Tests/Portability/ProcCpuInfoParserTests.cs
```diff
@@ -0,0 +1,436 @@
...
+        {
+            var parser = new ProcCpuInfoParser(string.Empty);
+            Assert.Equal(null, parser.PhysicalProcessorsCount);
+            Assert.Equal(null, parser.PhysicalCoresCount);
+            Assert.Equal(null, parser.ProcessorName);
+        }
+
+        [Fact]
+        public void TestBad()
```
AndreyAkinshin: I think it would be better to use `Malformed` instead of `Bad`.

adamsitnik:
> @adamsitnik, what do you think?

I like this idea and I agree with your commend (let's keep the summary small to make it easy to understand it and not miss anything important)

src/BenchmarkDotNet.Core/Helpers/ProcCpuInfoParser.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public string ProcessorName { get; }
+
+        [NotNull]
+        private static List<Dictionary<string, string>> Parse([CanBeNull] string content, char separator)
+        {
+            var items = new List<Dictionary<string, string>>();
+            Dictionary<string, string> units = null;
+
+            var list = Regex.Split(content ?? "", "\r?\n");
```
tmds: nit: `/proc/cpuinfo` doesn't have '\r' chars

tests/BenchmarkDotNet.Tests/Portability/Cpu/CpuInfoFormatterTests.cs
```diff
@@ -0,0 +1,39 @@
...
+using ApprovalTests.Reporters;
+using BenchmarkDotNet.Portability.Cpu;
+using BenchmarkDotNet.Tests.Mocks;
+using Xunit;
+
+namespace BenchmarkDotNet.Tests.Portability.Cpu
+{
+    [UseReporter(typeof(XUnit2Reporter))]
+    [UseApprovalSubdirectory("ApprovedFiles")]
+    public class CpuInfoFormatterTests
```
AndreyAkinshin: If you have more than one approval test class, you should mark all of them with `[Collection("ApprovalTests")]`, it prevents running them in parallel. We have a problem with `NamerFactory.AdditionalInformation`: it's a static singleton. If you run each test individually, everything is OK. In the case of multithreading execution, the `AdditionalInformation` value from one test can be used in another.
See also: https://xunit.github.io/docs/running-tests-in-parallel.html

tests/BenchmarkDotNet.Tests/Portability/Cpu/WmicCpuInfoParserTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            Assert.Equal("Intel(R) Xeon(R) CPU E5-2630 v3 @ 2.40GHz", parser.ProcessorName);
+            Assert.Equal(2, parser.PhysicalProcessorCount);
+            Assert.Equal(16, parser.PhysicalCoreCount);
+            Assert.Equal(32, parser.LogicalCoreCount);
+        }
+
+        [Fact]
+        public void RealOneProcessorFourCoresTest()
+        {
+            #region cpuinfo
```
AndreyAkinshin: Why do we need `#region` here? Could we remove it?

morgan-kn: Fixed

tests/BenchmarkDotNet.Tests/Portability/Cpu/CpuInfoFormatterTests.cs
morgan-kn: FIxed

tests/BenchmarkDotNet.Tests/Exporters/ApprovalTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if CLASSIC || NETCOREAPP2_0
+﻿#if CLASSIC
```
AndreyAkinshin: Probably it's a merge bug. We need `#if CLASSIC || NETCOREAPP2_0` here (now we can run ApprovalTests on Linux and macOS on .NET Core 2.0)

src/BenchmarkDotNet.Core/Portability/Cpu/MOSCpuInfoLoader.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Management;
+using BenchmarkDotNet.Environments;
+
+namespace BenchmarkDotNet.Portability.Cpu
+{
+    public class MOSCpuInfoLoader : ICpuInfo
+    {
+        public MOSCpuInfoLoader()
```
AndreyAkinshin: Could we hide the constructor and make it private?

In BenchmarkDotNet, we typically do something like this:
```cs
internal class MosCpuInfoLoader : ICpuInfo
{
    internal static readonly Lazy<ICpuInfo> Instance = new Lazy<ICpuInfo>(() => new MosCpuInfoLoader());
        
    private MosCpuInfoLoader()
    {
        // ...
```

It solves the following problems:
1. We try to ensure that we run such "heavy" logic only once.
2. RuntimeInformation logic shouldn't allocate additional memory after warmup (it can affect MemoryDiagnoser)
3. With the private constructor and internal singleton, a user can't accidentally invoke it and get an exception on a wrong OS.
4. Users don't need *public* OS-specific classes; they should get all information via "recommended" API. All magic should be hidden inside.
5. (A note): we typically use `Mos` notation instead of `MOS` for three-letter abbreviations.

P.S. Please, make the changes for other `ICpuInfo` implementations as well.

tests/BenchmarkDotNet.Tests/Portability/Cpu/CpuInfoFormatterTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace BenchmarkDotNet.Tests.Portability.Cpu
+{   
+    [Collection("ApprovalTests")]
+    [UseReporter(typeof(XUnit2Reporter))]
+    [UseApprovalSubdirectory("ApprovedFiles")]
+    public class CpuInfoFormatterTests
+    {
+        [Fact]
+        public void FormatTest()
```
AndreyAkinshin: Another ApprovalTests "feature" we should mark this tests with `[MethodImpl(MethodImplOptions.NoInlining)]` =)
See https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/build/0.10.11.386-aynykfpx/tests

src/BenchmarkDotNet.Core/Portability/Cpu/SysctlCpuInfoParser.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Collections.Generic;
+using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Helpers;
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Portability.Cpu
+{
+    public class SysctlCpuInfoParser : ICpuInfo
```
AndreyAkinshin: For `ProcCpuInfoParser`, `WmicCpuInfoParser`, `SysctlCpuInfoParser`, it's better to split parsing logic and results.
I suggest writing something like this:
```cs
internal static class SysctlCpuInfoParser
{
    [NotNull]
    internal static ICpuInfo Parse([CanBeNull] string content)
    {
        var sysctl = SectionsHelper.ParseSection(content, ':');
        var processorName = sysctl.GetValueOrDefault("machdep.cpu.brand_string");
        var physicalProcessorCount = GetPositiveIntValue(sysctl, "hw.packages");
        var physicalCoreCount = GetPositiveIntValue(sysctl, "hw.logicalcpu");
        var logicalCoreCount = GetPositiveIntValue(sysctl, "hw.physicalcpu");
        return new CpuInfo(processorName, physicalProcessorCount, physicalCoreCount, logicalCoreCount);
    }


    [CanBeNull]
    private static int? GetPositiveIntValue([NotNull] Dictionary<string, string> sysctl, string keyName)
    {
        if (sysctl.TryGetValue(keyName, out string value) &&
            int.TryParse(value, out int result) &&
            result > 0)
            return result;
        return null;
    }
}

public class CpuInfo : ICpuInfo
{
    public string ProcessorName { get; }
    public int? PhysicalProcessorCount { get; }
    public int? PhysicalCoreCount { get; }
    public int? LogicalCoreCount { get; }

    public CpuInfo(string processorName, int? physicalProcessorCount, int? physicalCoreCount, int? logicalCoreCount)
    {
        ProcessorName = processorName;
        PhysicalProcessorCount = physicalProcessorCount;
        PhysicalCoreCount = physicalCoreCount;
        LogicalCoreCount = logicalCoreCount;
    }
}
```

AndreyAkinshin:
@morgan-kn, great job, everything looks great! I will merge it as soon as you finish the requested fixed (which are minor).

src/BenchmarkDotNet.Core/Portability/Cpu/SysctlCpuInfoParser.cs
morgan-kn: @AndreyAkinshin ok, thanks

src/BenchmarkDotNet.Core/Portability/Cpu/MOSCpuInfoLoader.cs
morgan-kn: @AndreyAkinshin How do you think MosCpuInfoLoader has to implement ICpuInfo or it has to have method Load() which returns CpuInfo (from the next comment)?

AndreyAkinshin: @morgan-kn, I think, we can write something like this (it should be consistent):

```cs
internal static class MosCpuInfo
{
    internal static Lazy<ICpuInfo> Instance = new Lazy<ICpuInfo>(Load);

    private static ICpuInfo Load()
    {
        var processorModelNames = new HashSet<string>();
        uint physicalCoreCount = 0;
        uint logicalCoreCount = 0;
        int processorsCount = 0;

        var mosProcessor = new ManagementObjectSearcher("SELECT * FROM Win32_Processor");
        foreach (var moProcessor in mosProcessor.Get().Cast<ManagementObject>())
        {
            var name = moProcessor[WmicCpuInfoKeyNames.Name]?.ToString();
            if (!string.IsNullOrEmpty(name))
            {
                processorModelNames.Add(name);
                processorsCount++;
                physicalCoreCount += (uint) moProcessor[WmicCpuInfoKeyNames.NumberOfCores];
                logicalCoreCount += (uint) moProcessor[WmicCpuInfoKeyNames.NumberOfLogicalProcessors];
            }
        }

        return new CpuInfo(
            processorModelNames.Count > 0 ? string.Join(", ", processorModelNames) : null,
            processorsCount > 0 ? processorsCount : (int?) null,
            physicalCoreCount > 0 ? (int?) physicalCoreCount : null,
            logicalCoreCount > 0 ? (int?) logicalCoreCount : null);
    }
}
```

morgan-kn: @AndreyAkinshin  Yes, I thought about it, but is that ok that field Instance in MosCpuInfo class will have another type but not MosCpuInfo?

AndreyAkinshin: @morgan-kn, ok, what do you think about such naming:
```
internal static class MosCpuInfoProvider
{
    internal static Lazy<ICpuInfo> MosCpuInfo = new Lazy<ICpuInfo>(Load);
```

morgan-kn: @AndreyAkinshin ok, sounds good, thanks

AndreyAkinshin:
@morgan-kn, right now we have the following properties in `HostEnvironmentInfo`:
```cs
public Lazy<CpuInfo> CpuInfo { get; protected set; }
public int LogicalCoreCount { get; protected set; }
```
However, `CpuInfo` also contains the `LogicalCoreCount` property. Am I right that `HostEnvironmentInfo.LogicalCoreCount` can be removed?

morgan-kn:
@AndreyAkinshin, sorry for the late reply, I missed the question. Yes, HostEnvironmentInfo.LogicalCoreCount can be removed. I can make this fix today or tomorrow.

AndreyAkinshin:
>  I can make this fix today or tomorrow.

Great! Waiting for the PR.

Issue #614, opened 2018-01-04
Build fails with "'Microsoft.NETCore.App', version '1.1.2' was not found" probably due to 1.1.4 runtime not being available


nietras:
At the end of running `build.ps1` it fails with:
```
Running .NET Core 1.1.2 tests for framework netcoreapp1.1...
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '1.1.2' was not found.
  - Check application dependencies and target a framework version installed at:
      \
  - Alternatively, install the framework version '1.1.2'.
An error occurred when executing task 'FastTests'.

----------------------------------------
Teardown
----------------------------------------
Executing custom teardown action...
Error: One or more errors occurred.
        .NET Core CLI: Process returned an error (exit code -2147450749).
```
Note that in the beginning the `1.1.4` runtime fails downloadbut is
```
dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Runtime/1.1.4/dotnet-runtime-1.1.4-win-x86.zip
dotnet-install: Cannot download: https://dotnetcli.azureedge.net/dotnet/Runtime/1.1.4/dotnet-runtime-1.1.4-win-x86.zip
```
Not sure why... but going to the URL:
```
This XML file does not appear to have any style information associated with it. The document tree is shown below.
<Error>
<Code>BlobNotFound</Code>
<Message>
The specified blob does not exist. RequestId:0e9eef4c-001e-0094-474a-85c805000000 Time:2018-01-04T10:54:37.9967890Z
</Message>
</Error>
```

adamsitnik:
@nietras thanks for reporting the problem

@Ky7m could you take a look? I wonder how our CI works without it?

Ky7m:
Sure. I'll take care! 
@nietras can you give more details, please?

1. Do you run this in your local environment? OS: Windows?
2. In your logs you see the following output:
```
dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Runtime/1.1.4/dotnet-runtime-1.1.4-win-x86.zip
dotnet-install: Cannot download: https://dotnetcli.azureedge.net/dotnet/Runtime/1.1.4/dotnet-runtime-1.1.4-win-x86.zip
```
but it's expected and next line should say that dotnet-install switched to use legacy link
```
[...]

Issue #612, opened 2018-01-01
Disassembler DisassembleMethod fails with `Object reference not set to an instance of an object.`,


nietras:
Running a benchmark with disassembler turned on for something like:
```
 [DisassemblyDiagnoser(printAsm: true, printSource: true, recursiveDepth: 7)]
    [SimpleJob(RunStrategy.Monitoring, launchCount: 1, warmupCount: 2, targetCount: 11)]
    //[RyuJitX64Job()]
    public class RandomSort
    {
        const int Length = 2000000;
        static readonly int[] _random = CreateRandomArray<int>(Length, i => i);
        int[] _work = new int[Length];

        [IterationSetup]
        public void IterationSetup()
        {
            //Console.WriteLine(nameof(IterationSetup));
            Array.Copy(_random, _work, Length);
        }

        [Benchmark(Baseline = true)]
        public void ArraySort()
        {
            // CLR code
            Array.Sort(_work);
        }

        [Benchmark]
        public void SpanSort()
        {
            new Span<int>(_work).Sort();
        }
}
```

after settings `recursiveDepth: 7` this fails with:

```
Failed to disassemble with following exception:
Object reference not set to an instance of an object.
   at Microsoft.Diagnostics.RuntimeExt.ClrSourceExtensions.FindNearestLine(PdbFunction function, Int32 ilOffset)
   at BenchmarkDotNet.Disassembler.Program.<GetSource>d__9.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at BenchmarkDotNet.Disassembler.Program.DisassembleMethod(MethodInfo methodInfo, State state, Settings settings)
   at BenchmarkDotNet.Disassembler.Program.Disassemble(Settings settings, ClrRuntime runtime, State state)
   at BenchmarkDotNet.Disassembler.Program.Handle(Settings settings)
   at BenchmarkDotNet.Disassembler.Program.Main(String[] args)
```

nietras:
cc: @adamsitnik

nietras:
On that note is there any way to get dissassembly for runtime code? I.e. code I do not have the source for?

adamsitnik:
@nietras the workaround: show only asm, no source code:

```cs
 [DisassemblyDiagnoser(printAsm: true, printSource: false, recursiveDepth: 7)]
```

[...]

PR #615, opened 2018-01-04
Quick fix Disassembler.Program.GetMethod when more than one method found just return null

nietras:
Quick fix for issue reported later in https://github.com/dotnet/BenchmarkDotNet/issues/612

Had build issues also https://github.com/dotnet/BenchmarkDotNet/issues/614

Note this does simply ignores the case when more than one method found... didn't know how to add a test or debug this, but simply added this and now BDN at least doesn't crash when there is this issue.

### DIFF
--- a/build/pack.bat
+++ b/build/pack.bat
@@ -1,0 +1,7 @@
+dotnet pack .\src\BenchmarkDotNet.Core\BenchmarkDotNet.Core.csproj -c Release
+dotnet pack .\src\BenchmarkDotNet.Toolchains.Roslyn\BenchmarkDotNet.Toolchains.Roslyn.csproj -c Release
+dotnet pack .\src\BenchmarkDotNet\BenchmarkDotNet.csproj -c Release
+dotnet pack .\src\BenchmarkDotNet.Diagnostics.Windows\BenchmarkDotNet.Diagnostics.Windows.csproj  -c Release
+rmdir artifacts /s /q
+mkdir artifacts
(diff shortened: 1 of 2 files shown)


nietras:
also added a `pack.bat` file for convenience, perhaps call this from `build-and-pack.cmd` to avoid redundancy.

nietras:
For reference and to be able to test this more easily for myself and possible others are here the packages built with this change:

[artifacts.zip](https://github.com/dotnet/BenchmarkDotNet/files/1603242/artifacts.zip)

nietras:
It would probably be better to fix the `CecilNameToClrmdName` method which doesn't seem to handle a lot of types i.e. generic methods etc. Refactoring code and adding tests for these would be good too 😄

adamsitnik:
@nietras thanks for the fix!

I tried to reproduce the problem on my own. It turned out that in this case we have 3 different methods with same full signature: "System.IntPtr.op_Explicit(IntPtr)"

The problem is that they return different types (int, long and void*) but the return type is not a part of the full signature in ClrMD. AFAIK there is no way to get the method return type with the available APIs. I am going to ping @leculver and ask if it is somehow possible.

nietras:
@adamsitnik great! Yeah I thought it might have something to do with explicit conversion operators... is this part of the nightly build then?

adamsitnik:
>  is this part of the nightly build then?

@nietras yes, every time we merge a PR/push to master we release new NuGet package to our CI feed.

PR #625, opened 2018-01-15
Fix xunit warnings connected to usage of Assert.Equal() to check for Null

Ky7m:
Replace Assert.Equal(null,...) with Assert.Null(...) to simplify asserts and fix warnings from xunit analyzers.


AndreyAkinshin:
@Ky7m, thanks!

Issue #541, opened 2017-09-07
Mono Support for DisassemblyDiagnoser


AndreyAkinshin:
Consider the following benchmarks:
```cs
 [DisassemblyDiagnoser(printAsm: true, printSource: true)] // !!! use the new diagnoser!!
    [Config(typeof(Config))]
    public class Simple
    {
        private class Config : ManualConfig
        {
            public Config()
            {
                Add(Job.Dry.With(Runtime.Mono));
            }
        }
        int[] field = Enumerable.Range(0, 100).ToArray();

        [Benchmark]
        public int SumLocal()
        {
            var local = field; // we use local variable that points to the field

            int sum = 0;
            for (int i = 0; i < local.Length; i++)
                sum += local[i];

            return sum;
        }

        [Benchmark]
        public int SumField()
        {
            int sum = 0;
            for (int i = 0; i < field.Length; i++)
                sum += field[i];

            return sum;
        }
    }
```
Here are my disasm results on Linux and Windows:
![simple-linux](https://user-images.githubusercontent.com/2259237/30168252-84fa238a-93f1-11e7-959c-fb62cbd89760.png)
![simple-windows](https://user-images.githubusercontent.com/2259237/30168256-88f36ff0-93f1-11e7-8a97-5c55ad05df3e.png)
It doesn't look like a correct results.

adamsitnik:
It's a bug, however I am afraid that I will not have any free time soon to fix it (I have 3 conferences next week). I am going to mark it as up-for-grabs.

adamsitnik:
How to fix the problem:

1. Compile the code provided by @AndreyAkinshin 
2. Run `mono -v -v -v -v --compile $namespace.$typeName:$methodName $pathToExe`
3. Take it's output, add it as a new test to [this](https://github.com/dotnet/BenchmarkDotNet/blob/c6bbda4e1eb69218e7981d8aa2b0d59344c5ebaf/tests/BenchmarkDotNet.Tests/MonoDisassemblyOutputParserTests.cs#L8) class with tests
4. Modify our [parser](https://github.com/dotnet/BenchmarkDotNet/blob/c6bbda4e1eb69218e7981d8aa2b0d59344c5ebaf/src/BenchmarkDotNet.Core/Diagnosers/MonoDisassembler.cs#L45) to get the tests passing

AndreyAkinshin:
I will do it, thanks for the guide.

AndreyAkinshin:
[...]

PR #637, opened 2018-01-30
Mono Support for DisassemblyDiagnoser #541

morgan-kn:


AndreyAkinshin:
Thanks, @morgan-kn!

src/BenchmarkDotNet.Core/Diagnosers/CopiedDataContracts.cs
```diff
@@ -53,6 +53,8 @@ public class DisassembledMethod
         public string Problem { get; set; }
 
         public Map[] Maps { get; set; }
+
+        public string CommandLine { get; set; }
```
adamsitnik: the file is called `CopiedDataContracts` and it starts with `// keep it in sync with src\BenchmarkDotNet.Disassembler.x64\DataContracts.cs!`

could you please add the same field to the `src\BenchmarkDotNet.Disassembler.x64\DataContracts.cs` file?

src/BenchmarkDotNet.Core/Running/BenchmarkRunnerCore.cs
```diff
@@ -366,7 +366,7 @@ private static BenchmarkReport RunCore(Benchmark benchmark, ILogger logger, Read
             {
                 logger.WriteLineInfo("// Run, Diagnostic [SeparateLogic]");
 
-                separateLogicCompositeDiagnoser.Handle(HostSignal.AfterAll, new DiagnoserActionParameters(null, benchmark, config));
+                separateLogicCompositeDiagnoser.Handle(HostSignal.SeparateLogic, new DiagnoserActionParameters(null, benchmark, config));
```
adamsitnik: @morgan-kn very good catch! you have fixed my bug 💯

AndreyAkinshin:
> Could we also give our users a warning if this tool is not present?

Absolutely! In fact, it's almost ready, I just want to check the detection logic on a few additional machines. It will be a part of the v0.10.13 release.

adamsitnik:
@AndreyAkinshin awesome!

src/BenchmarkDotNet.Core/Diagnosers/CopiedDataContracts.cs
morgan-kn: @adamsitnik should I fill this field? If so, I can suggest something like CommandLine = Environment.CommandLine in both DisassembleMethod and CreateEmpty.

src/BenchmarkDotNet.Core/Running/BenchmarkRunnerCore.cs
morgan-kn: @adamsitnik thanks =)

src/BenchmarkDotNet.Core/Diagnosers/CopiedDataContracts.cs
adamsitnik: @morgan-kn no, just add it to the class. 

Some background: Windows disassembler must not have any dependencies to BDN, but it serializes the result to file. BDN has to be able to deserialize it, and also have no dependencies to the disassembler. Hence these two projects must duplicate same data contract, and we need this lame, manual files sync ;/

morgan-kn: @adamsitnik thank you =)

AndreyAkinshin: @adamsitnik, why do we have to sync these files manually?
In the `BenchmarkDotNet.Disassembler.x86` we use linked files:
```xml
<Compile Include="..\BenchmarkDotNet.Disassembler.x64\DataContracts.cs"
         Link="DataContracts.cs" />
```
What do you think about using the same trick for `BenchmarkDotNet.Core`?

adamsitnik: @AndreyAkinshin I remember that I have tried this approach, but for some reason, it did not work. I am not sure what was the problem that I hit back then ;/ It's most probably worth to give it a try once again

PR #639, opened 2018-02-01
Portability.Cpu tests improvements

morgan-kn:
For CpuInfoParserTests: remove #region directive, test data was moved to test files.


AndreyAkinshin:
@morgan-kn, thanks!

PR #657, opened 2018-02-22
Fixed code block formatting

svick:
The old version shows fine on GitHub, but not on the website:

I *think* this change will fix it, but I haven't verified that.


adamsitnik:
@svick thanks for improving our docs!

Issue #658, opened 2018-02-22
[Params] for enums should include all values by default


svick:
I think that with `[Params]` properties whose type is some `enum`, a fairly common situation is that I want all values of that `enum` to be included. It would be nice if doing this didn't require listing all those values in the attribute. I.e., instead of having to write:

```c#
[Params(CurrentCulture, CurrentCultureIgnoreCase, InvariantCulture, InvariantCultureIgnoreCase, Ordinal, OrdinalIgnoreCase)]
public StringComparison StringComparison { get; set; }
```

I could write just:

```c#
[Params]
public StringComparison StringComparison { get; set; }
```

Do you think this would be a useful feature? Would this be considered a breaking change? (If yes, something like `[EnumParams]` could be added instead.)

adamsitnik:
I like this idea, but I am not sure how we make it more intuitive. Our users are used to providing all values.

Maybe something like:

```cs
[AllEnumValues]
public StringComparison StringComparison { get; set; }
```

@svick what do you think?

svick:
@adamsitnik Yeah, I think that's reasonable (and it's what I meant by `[EnumParams]`).

Regarding naming, I think that since it's a variation on `Params` (just like `ParamsSource`), I think the name should include "Params". But I'm not sure what exactly would be the right name: `EnumParams` is too cryptic and something like `ParamsAllEnumValues` is too long.

adamsitnik:
@svick so maybe `EnumAllParam` ?

Naming is hard ;p

gsomix:
Hey, @svick, @adamsitnik, @AndreyAkinshin, do you mind if I reopen #664 with additional changes (tests, validator)?

svick:
@gsomix Feel free to go ahead. I don't know when I would have the time to do it myself.

PR #664, opened 2018-02-24
Added [ParamsAllValues]

svick:
Closes https://github.com/dotnet/BenchmarkDotNet/issues/658.

The name is `ParamsAllValues`, because it can get all values not just for `enum`, but also for other types with tiny number of values (see the documentation).

### DIFF
--- a/docs/guide/Advanced/Params.md
+++ b/docs/guide/Advanced/Params.md
@@ -118,3 +118,49 @@ public class IntroIParam
 | Benchmark | (100,20) | 120.4 ms | 0.0843 ms | 0.0788 ms |
 | Benchmark | (200,10) | 210.4 ms | 0.0892 ms | 0.0834 ms |
 | Benchmark | (200,20) | 220.4 ms | 0.0949 ms | 0.0887 ms |
+
+# ParamsAllValues
+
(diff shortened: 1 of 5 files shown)


AndreyAkinshin:
Hey, @svick, thanks for the PR!

Could you replace integration tests with regular unit tests?
Our integration tests take too much time, we use it only in cases when we want to check that we can build and run a benchmark against specific toolchains.
In you case, you just want to check that `ParamsAllValues` was parsed correctly. You don't have to actually run a benchmark for this.

I guess that the best way to test this feature is approval tests (see `MarkdownExporterApprovalTests` as an example). It's still not a perfect solution, but it's the simplest one for now (and we can run such tests almost instantly.

AndreyAkinshin:
I think that we also should implement a validator (e.g., see `BaselineValidator`) instead of throwing `InvalidOperationException`. This approach allows printing a nice user-friendly message which describes the problem. All validation errors can be collected via special API (without try-catch blocks) and it doesn't break all execution process for all benchmarks, it just disables a benchmark with invalid config.

AndreyAkinshin:
@svick, I close this PR since we don't have any updates since February. Feel free to open another one in case if you resolve the review comments.

PR #675, opened 2018-03-06
Fix IDE0034: 'default' expression can be simplified

shoelzer:
This was suggested by VS2017 Code Analysis.


AndreyAkinshin:
@shoelzer thanks!

PR #686, opened 2018-03-14
Update Overview.md

AlexGhiondea:
The links pointed to .htm files instead of the available .md files.

### DIFF
--- a/docs/guide/Overview.md
+++ b/docs/guide/Overview.md
@@ -131,7 +131,7 @@ public class Md5VsSha256
     }
 ```
 
-Read more:  [Jobs](Configs/Jobs.htm), [Configs](Configs/Configs.htm)
+Read more:  [Jobs](Configs/Jobs.md), [Configs](Configs/Configs.md)
 
 
 ## Columns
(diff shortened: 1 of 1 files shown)


AlexGhiondea:
@adamsitnik how can I restart the CI for this?

adamsitnik:
@AndreyAkinshin why do we link to `.htm` files instead of the `.md`? To get our docs working? Can the doc compiler replace the `.md` links with `.html` links?

AndreyAkinshin:
@AlexGhiondea, @adamsitnik 
Currently, we use [DocNet](https://github.com/FransBouma/DocNet) which doesn't transform `.md` files to `.htm` automatically. Thus, we have to keep `.htm` links (for now).

Probably, we can find a workaround for it, but I don't think that we should spend time on it.
To be honest, I want to migrate to [DocFX](https://github.com/dotnet/docfx) eventually. This project is more popular, it has more features, and it's already used for the most of the .NET Foundation projects.

AndreyAkinshin:
@adamsitnik, I created a separated issue for it: https://github.com/dotnet/BenchmarkDotNet/issues/710

PR #688, opened 2018-03-15
.NET Standard 2.0

adamsitnik:
I would like to open a discussion with this PR. I do not insist to merge it or anything like that.

I was supposed to move all our attributes to a new lib, BenchmarkDotNet.Annotations and make it target .NET Standard 1.0.

I started doing this and realized that this is not trivial. Why? Because I would have to move everything the attributes use to new dll. IConfig, Benchmark and all other basic types we use everywhere.

I am a lazy person and I decided to stop. 

I had another idea: what if I just port everything to .NET Standard 2.0? BDN is a  benchmarking tool, so I am not sure if we really need to target 1.0 (BDN is not a dependency of any other lib)

Advantages:

1. We can now simply target .NET Standard 2.0 for new runtimes and .NET 4.6 for those of our users who don't want to use .NET Standard and are using full .NET Framework
2. A lot of ugly `#if` defines got removed
3. I merged `BenchmarkDotNet.Toolchains.Roslyn` back to `BenchmarkDotNet.Core`, so we have fewer packages
4. I ported `BenchmarkDotNet.Diagnostics.Windows` to .NET Standard 2.0

I could even merge BenchmarkDotNet.Core to BenchmarkDotNet. So we would have **one** package

**Disadvantages**: I had to drop .NET Core 1.1 support. But I wonder if any of our users still use it.

@AndreyAkinshin @ig-sinicyn  @Ky7m  what do you guys think?


AndreyAkinshin:
To be honest, I don't like the idea of dropping .NET Core 1.1 support.
According to [.NET Support policy](https://www.microsoft.com/net/support/policy), the End of Support Date for .NET Core 1.1 is June 27, 2019.

One year ago we started trying to make our core library runtime-independent (https://github.com/dotnet/BenchmarkDotNet/pull/440), it was a cool idea. It would be great if we can keep our annotation package runtime-agnostic (e.g., .NETStandard1.0) and introduce an interface (let's call it `IRuntime`) for runtime-specific logic. An implementation of this interface can be dynamically loaded. It can look like follows:

* BenchmarkDotNet.Annotations (.NET Standard 1.0): Runtime-agnostic public API for attributes and configs + `IRuntime` definition (without logic which actually runs benchmarks)
* `BenchmarkDotNet.Runtime.NETCore11`, `BenchmarkDotNet.Runtime.NETCore20`, `BenchmarkDotNet.Runtime.NETFramework46`, and so on: `IRuntime` implementation for different runtimes (without ugly `#ifdef`)
* `BenchmarkDotNet.Engine`: a package with logic which actually runs benchmarks without dependencies on runtime packages
* `BenchmarkDotNet` which reference all other packages.

Advantages:
* A lot of ugly `#if` defines got removed
* .NETCore1.1 support (and probably NETCore1.0 support (with some limitations))
* Such system looks good and it's pretty pluggable and extendable.

Disadvantages:
* I understand that such refactoring is not trivial, it's not easy to refactor the existed code base in such way.

adamsitnik:
I remember the idea of having runtime-independent BDN because I was the person who implemented it. It took me a lot of time (40+ hours) and I did not like the final result for one reason: I had to load the runtime-dependent things in a dynamic way. It was very fragile: even a small difference could lead to a crash (missing dll/wrong name/some dependency problem). The thing that I am most afraid is that some other tool will not copy all required dependencies and we can't load the runtime and we can't even run a single benchmark.

Another problem with the idea of `IRuntime` is that 99% of our code is not runtime specific, and the 1% which is, is scattered around all types. So there is no single type that needs it but there are 10 types with just some `#if` defines (please take a look at the files I have modified). And the `#if` statements are usually for something that should be part of .NET Standard 1.0, but for some reason was not and was added to .NET Standard 2.0. This is why the .NET Standard 2.0 code is so simple: everything we need is already there.

I don't like the idea of dropping .NET  Core 1.1 either, but I believe that nobody is using 1.1 for performance-sensitive code. Why do I think so? People who use .NET Core are early adopters. When 2.0 came out and they could easily port from 1.1 to 2.0 and gain huge perf boost and a lot of missing APIs they just did. In my previous job, I was developing a huge trading system and it took me only a day to port it from 1.1 to 2.0. It was a free lunch.

Also, most of our users are early adopters and they are always interested in running latest .NET Core.

I see it this way: if we move to .NET Standard 2.0 everything is plain simple and is always going to work as expected. If we want to target .NET Standard 1.0 and keep the support of .NET Core 1.1 we increase the complexity for very small gain. I just don't think it's worth the struggle.

Ky7m:
In general, I assume BDN is more a tool than nuget package-helper for your business logic in a project (e.g., like Json.NET) and it is not required to support minimal version of .NET Standard. 
On the other hand, in case of dropping .NET Core 1.1 compatibility end users won't be able to run the same tests against different runtimes, correct? If so, I think this only one huge stopper.

AndreyAkinshin:
@adamsitnik, OK, you convinced me. Let's try to drop .NET Core 1.1. support. In case of the negative feedback by the community, we always can try to support it again. =)

A few things that you should keep in mind:
* I still think that it's a good idea to introduce an interface for `RuntimeInformation`. The main reason is the ability to mock it in the tests. We have a lot of logic which is executed only in a specific environment. It would be nice if we can cover this logic in all possible environments. (e.g. I want to run .NETFramework specific tests from .NETCore+macOS environment; here I'm talking about functional tests instead of integration tests which actually run the benchmarks and take a huge amount of time).
* `BenchmarkDotNet v0.10.14` shouldn't include any breaking changes (because it contains two important features: arguments and histograms). Please, use the `master` branch only for small changes or hotfixes.
* `BenchmarkDotNet v0.11.0` will include many major changes. I guess it makes sense to create a separate branch for it.

Ky7m:
Love it!!!! With this change a build process can be simplified a lot :)

adamsitnik:
> OK, you convinced me.

@AndreyAkinshin great!

> I still think that it's a good idea to introduce an interface for RuntimeInformation

I agree, but it is used in many, many places. We would need to make it an interface and create it before running the benchmarks and pass from the top to the bottom to be able to use it everywhere. It's a refactoring challenge :D

> Please, use the master branch only for small changes or hotfixes.

ok, so I will create `0.10.11` branch and push it there.

AndreyAkinshin:
> I agree, but it is used in many, many places. We would need to make it an interface and create it before running the benchmarks and pass from the top to the bottom to be able to use it everywhere. It's a refactoring challenge :D

Since, we need it only in tests, we can try something like that:

```cs
public class Runtime : IRuntime
{
    public static private IRuntime instance;

    public static public IRuntime Intsnace => instance;

    // For tests only
    internal static void ChangeImplementation(IRuntime newRuntime)
    {
        instance = newRuntime;
    }
}
// ...
public class UnitTests
{
    public UnitTests()
    {
        Runtime.ChangeImplementation(MockRuntime.Default);
    }
}
```
It shouldn't be a huge refactoring, we have to only replace `RuntimeInformation` by `Runtime.Instance`. What do you think?

adamsitnik:
> It shouldn't be a huge refactoring, we have to only replace RuntimeInformation by Runtime.Instance. What do you think?

that makes thing simpler ;) I can add it to my PR

adamsitnik:
I am closing this PR as I have created a new branch "v11". I will be using the new branch to implement some more features for the 0.11 release

AndreyAkinshin:
@adamsitnik, Ok. I have a few more bug fixes that I want to include in v0.10.14. Nothing major, it should be easy to merge or cherry-pick it into "v11". I hope v0.10.14 will be released soon, so we will be able to switch master to "v11". In case of any critical bugs, we can also release v0.10.15 which will not include any breaking changes.

adamsitnik:
@AndreyAkinshin I like your plan! I will create issues for the things that I want to implement for 0.11.0

PR #693, opened 2018-03-18
Jit runtime validation

morgan-kn:


AndreyAkinshin:
@adamsitnik, I was involved in the design of this feature. The main reason why do wee need analysers here is following: we don't always know in advance what kind of jobs are valid. For example, it's almost impossible to say can we use RyuJIT or not (it depends on the current installed version of .NET Framework + Registry keys).

I agree that it would be nice to have a validator which prevents running obviously invalid jobs, but we still need an analyzer for the final verification.

AndreyAkinshin:
@morgan-kn, `CharactersAsParamsAreSupported` is red on Travis (see https://travis-ci.org/dotnet/BenchmarkDotNet/jobs/356756856), `CheckWindows(runtime: Mono, jit: LegacyJit, platform: X86, exptectedText: ...` is red on AppVeyor (https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/build/0.10.13.481-fvoncjsb/tests)
Could you please take a look?

AndreyAkinshin:
@morgan-kn, in last 4 commits, your email is `morgan_kn@skbkontur.ru` (instead of `morgan.usu@gmail.com`) which is not linked to your GitHub profile. Could you please add the new email to your profile or rewrite the history with the old email (I want to get a nice commit history).

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -214,12 +214,12 @@ internal static string GetRuntimeVersion()
...
         internal static Runtime GetCurrentRuntime()
         {
+            if (IsMono)
+                return Runtime.Mono;
             if (IsFullFramework)
                 return Runtime.Clr;
             if (IsNetCore)
                 return Runtime.Core;
-            if (IsMono)
-                return Runtime.Mono;
```
AndreyAkinshin: Why do we change the order of conditions here? If we need a specific order, it would be nice to have a comment which explains the requirements.

tests/BenchmarkDotNet.IntegrationTests/JitRuntimeValidationTest.cs
```diff
@@ -0,0 +1,97 @@
...
+        private const string ToolchainSupportsOnlyRyuJit = "Currently dotnet cli toolchain supports only RyuJit";
+
+        public JitRuntimeValidationTest(ITestOutputHelper outputHelper)
+        {
+            Output = outputHelper;
+        }
+
+        public static TheoryData<Runtime, Jit, Platform, string> DataForWindows = new TheoryData<Runtime, Jit, Platform, string>
+        {
+            // our CI would need to have Mono installed..
```
AndreyAkinshin: @Ky7m do you know, is it possible to use the latest Mono version in our appveyor build?

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
morgan-kn: @AndreyAkinshin I did it on the base of the order in RuntimeInformation.GetRuntimeVersion(). When I ran my tests on Mono, RuntimeInformation.GetCurrentRuntime() detected runtime as CLR.

src/BenchmarkDotNet/Configs/DefaultConfig.cs
```diff
@@ -45,7 +45,7 @@ public IEnumerable<IAnalyser> GetAnalysers()
             yield return OutliersAnalyser.Default;
             yield return MinIterationTimeAnalyser.Default;
             yield return IterationSetupCleanupAnalyser.Default;
-            yield return MultimodalDistributionAnalyzer.Default;
+            yield return MultimodalDistributionAnalyzer.Default;            yield return RuntimeErrorAnalyser.Default;
```
AndreyAkinshin: You missed a new line here.

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
AndreyAkinshin: In this case, we definitely need a comment.

tests/BenchmarkDotNet.IntegrationTests/JitRuntimeValidationTest.cs
Ky7m: Appveyor has outdated version and doesn't have an option to configure a required version. We can add an additional step in appveyor yml config to install mono (via chocolatey), so it's possible. Let me play around and create a separate PR. Does it make sense?

src/BenchmarkDotNet/Running/BenchmarkRunner.cs
```diff
@@ -369,6 +369,19 @@ private static BenchmarkReport RunCore(Benchmark benchmark, BenchmarkId benchmar
 
                 executeResults.Add(executeResult);
 
+                var errors = executeResults.SelectMany(r => r.Data)
+                    .Union(executeResults.SelectMany(r => r.ExtraOutput))
+                    .Where(line => line.Contains(ValidationHelper.ConsoleErrorPrefix))
```
adamsitnik: nit: if this is a prefix, then maybe `StartsWith` would be better than `Contains`?

src/BenchmarkDotNet/Environments/Jit.cs
```diff
@@ -2,6 +2,12 @@
 {
     public enum Jit
     {
+        /// <summary>
+        /// Undefined
+        /// <remarks>By default</remarks>
+        /// </summary>
+        Undefined,
```
adamsitnik: If user runs some benchmarks with following config:

```cs
DefaultConfig.Instance
	.With(Job.Default)
	.With(Job.RyuJitX64)
```

What is going to be printed in the `Jit` column for the default job? Is it going to be `Undefined` or `Default`?

@AndreyAkinshin  Maybe we should call this value a `Default` to avoid possible confusion? (users will ask what is Undefined Jit?)

morgan-kn: @adamsitnik 
If you do not set Jit in Benchmark configuration then benchmark will be ran with LegacyJIT (because it is the first element in Jit enum). 
In my branch I added the new enum element to avoid such behavior. If I run benchmark with your config I get such summary:

// * Summary *

BenchmarkDotNet=v0.10.14.20180523-develop, OS=Windows 10.0.17134
Intel Core i7-4710MQ CPU 2.50GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
Frequency=2435775 Hz, Resolution=410.5470 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3101.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3101.0
  RyuJitX64  : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3101.0

Jit=RyuJit  Platform=X64  

 Method |       Job |      Mean |     Error |    StdDev |
------- |---------- |----------:|----------:|----------:|
   Test |   Default | 0.0020 ns | 0.0039 ns | 0.0035 ns |
   Test | RyuJitX64 | 0.0000 ns | 0.0000 ns | 0.0000 ns |

adamsitnik: @morgan-kn thanks for taking care of this! I remember that we used  to display `LegacyJit` in some places because of the default enum value ;)

morgan-kn: @adamsitnik I'm going to do post review changes. Should I rename Jit.Undefined to Jit.Default? I think it's a good idea. =)

adamsitnik: @morgan-kn I think the same, please change it. 👍

Issue #689, opened 2018-03-16
DisassemblyDiagnoser for Mono does not work on Windows and Linux


dvorn:
MonoDisassembler.OutputParser.Parse seems to be looking for the string "(__TEXT,__text) section" to find the beginning of disassembled code ([see](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Diagnosers/MonoDisassembler.cs#L56)).  However this string is seemingly produced by the otool on Mac.  On Windows and Linux, the objdump tool is used and the assembly in the log is prepended  by the string "Disassembly of section .text:".

Thus, OutputParser does not find the beginning of assembly and then crashes with NullReferenceException.

AndreyAkinshin:
@morgan-kn, please take a look.

AndreyAkinshin:
@dvorn, thanks for the report! The fix will be available in v0.10.14.

PR #694, opened 2018-03-18
Fix 689

morgan-kn:


AndreyAkinshin:
@morgan-kn, thanks, everything works great on Windows and Linux.
My only concern is about the commit history. Here are your two commits:
```
$ git log --pretty=format:"%an <%ae>: %s" -n2 morgan-kn/Fix689
Irina Ananyeva <--global>: MonoDisassembler improvements
Irina Ananyeva <--global>: Fix 689
```
As you can see, your current email is `--global`. GitHub fails to match you and your commits because of this. Could you please amend these commits with correct email?
Also, it would be nice to improve the commit message and replace "Fix 689" by "Fix #689" (it automatically close the fixed issue after the merge, see https://help.github.com/articles/closing-issues-using-keywords/). It would be even better to provide short summary about changes (e.g. "ProcessHelper: handle null Data values in OutputDataReceived, fix #689").

PR #709, opened 2018-04-01
Added an analyser that will award you with achievements.

exyi:
The analyser will award you with achievements to motivate you to benchmark more and make the code faster. It remembers some stats about every benchmark, so it can spot that you've made it run faster/with less alocations. And, as you'd expect from achievement, it will not give you the same one multiple times.

image

The engine is also extensible, so you can implement new achievement pack as a library :)

To motivate dev further to make the core faster, we could also include  some minigames or HackerNews/reddit/twitter client that will unlock new features for these achievements. And it will stop the game when the benchmark finishes, so you'll **have to** run the benchmark again and again to complete the level. In a year or two, running .NET code would be faster that Rust or whatever.

### DIFF
--- a/src/BenchmarkDotNet.Core/Achievements/AchievementAnalyser.cs
+++ b/src/BenchmarkDotNet.Core/Achievements/AchievementAnalyser.cs
@@ -2,17 +2,8 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection.Metadata.Ecma335;
 using BenchmarkDotNet.Analysers;
-using BenchmarkDotNet.Columns;
-using BenchmarkDotNet.Diagnosers;
-using BenchmarkDotNet.Engines;
-using BenchmarkDotNet.Exporters;
-using BenchmarkDotNet.Extensions;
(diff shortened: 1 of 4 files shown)


AndreyAkinshin:
Hey @exyi!
Gamification with a set of achievements is a nice idea. However, I don't think that it should be included in the library by default. We are not trying to motivate people to make code faster. BenchmarkDotNet is a tool for performance investigations, it's not a game. Also, we are trying to keep the console output as small as possible so people are able to focus on really important data. We already have too many information on the output and it's not a good idea to add some additional lines which people don't really need.

Feel free to create own NuGet package (e.g. `Achievements.BenchmarkDotNet` or `Gamification.BenchmarkDotNet`). In this case, people who want this feature will be able to install it. People who don't need this feature will be able to continue performance investigations in the "distraction free mode."
Let me know if you miss any API which you need for building own package.

exyi:
Spoiler: 
image

And you can see from the path, I don't actually miss API for this ;)

AndreyAkinshin:
@exyi, sorry, didn't look at the date. I reviewed this PR on April 3 in my time zone.

Issue #701, opened 2018-03-22
Implement BenchView exporter


adamsitnik:
BenchView is an internal MS tool for reporting benchmark results. We are planning to open source it soon. Its uploader requires a file of certain format to work (It's a CSV so there will be no strong dependency to any db or anything like that).

The goal is to use BDN to run the benchmarks and export them to BenchView.

PR #725, opened 2018-04-20
Extend exported json file with FullName using xunit naming convention for integration purpose

adamsitnik:
this PR provides an initial implementation of the exporter for BenchView integration. #701

edit: it's not a dedicated exporter, I just extended the JsonExporter with new field

@jorive I need to figure out how to set you as an reviewer of this PR


jorive:
@adamsitnik LGTM modulo generating better names for generics, or asserting if two different benchmarks have the same name? thoughts?

jorive:
LGTM

adamsitnik:
> I would not use the BenchView name on the source code

@jorive only the PR title and branch name contain the "BenchView" keyword

PR #728, opened 2018-04-22
BenchmarkReport Exporter

Rizzen:
I found the TODO in BenchmarkRunner.
`` // TODO: make exporter``
So I created new Exporter and moved logic into it.
Does it make sense?


adamsitnik:
@Rizzen thank you! --todo;

PR #737, opened 2018-05-02
Return value validator

### LINKED ISSUE #136: [Feature request] Fastcheck for correctness of benchmark implementations?

ltrzesniewski:
This PR basically fixes #136. The use case is described there. I got tired of writing additional validation logic every time I wrote a benchmark.

It adds a `ReturnValueValidator` class which checks that benchmark methods return the same values for a given parameter set. Void methods are allowed.

It also adds `ExecutionValidatorAttribute` and `ReturnValueValidatorAttribute` for convenience.


adamsitnik:
@ltrzesniewski could you also add tests for some collections (ie: array, dictionary) and custom types implementing `IEqutable`?

adamsitnik:
@ltrzesniewski really good job! 👍 thanks for doing this!

ltrzesniewski:
Thanks! 👍
I totally forgot about collection types, these won't compare as equal right now. I'll fix that.

ltrzesniewski:
@adamsitnik I added a custom equality comparison logic which should be quite similar in behavior to what unit test framework equality assertions usually do, but without going as far. I preferred to optimize for simplicity. 😄 

Also, I noticed that `ExecutionValidator` always printed the generic `TargetInvocationException` message, so I fixed that as well.

PR #743, opened 2018-05-04
Improve .NET Framework version detection

epeshk:
1) Add .NET Framework 4.7.2 version constant
2) Check for Reference Assemblies directory existence only for Toolchain select. This directory exists only when Developer Pack is installed. I think it would be better to display the version of framework (runtime, not sdk) in benchmarks output
3) Remove hardcoded Program Files directory path. It fix version detection on 32bit systems.
4) Refactor: move framework version detection from CsProjClassicNetToolchain, get rid of code clones

### DIFF
--- a/src/BenchmarkDotNet/Helpers/FrameworkVersionHelper.cs
+++ b/src/BenchmarkDotNet/Helpers/FrameworkVersionHelper.cs
@@ -53,7 +53,14 @@ namespace BenchmarkDotNet.Helpers
         
         // Reference Assemblies exists when Developer Pack is installed
         private static bool IsDeveloperPackInstalled(string version) => Directory.Exists(Path.Combine(
-            Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86), @"Reference Assemblies\Microsoft\Framework\.NETFramework", 'v' + version));
+            GetProgramFilesX86DirectoryPath(), @"Reference Assemblies\Microsoft\Framework\.NETFramework", 'v' + version));
 
+        private static string GetProgramFilesX86DirectoryPath()
+        {
+            var folder = Environment.Is64BitOperatingSystem
+                ? Environment.SpecialFolder.ProgramFilesX86
+                : Environment.SpecialFolder.ProgramFiles;
+            return Environment.GetFolderPath(folder);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


epeshk:
close and re-open for trigger travis-ci build

src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
```diff
@@ -75,44 +87,10 @@ private static IToolchain GetCurrentVersion()
...
-        [NotNull]
-        internal static string GetCurrentNetFrameworkVersion()
-        {
-            var toolchain = GetCurrentVersionBasedOnWindowsRegistry() as CsProjClassicNetToolchain;
-            if (toolchain == null)
-                return "?";
-            string version = toolchain.targetFrameworkMoniker.Replace("net", "");
-            return string.Join(".", version.ToCharArray());
+            
+            // this logic is put to a separate method to avoid any assembly loading issues on non Windows systems
```
AndreyAkinshin: 👍

AndreyAkinshin:
@epeshk thanks!

src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
adamsitnik: 👍 great!

src/BenchmarkDotNet/Helpers/FrameworkVersionHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+                       .FirstOrDefault(v => releaseNumber >= v.minReleaseNumber && IsDeveloperPackInstalled(v.version))
+                       .version;
+        }
+        
+        // Reference Assemblies exists when Developer Pack is installed
+        private static bool IsDeveloperPackInstalled(string version) => Directory.Exists(Path.Combine(
+            ProgramFilesX86DirectoryPath, @"Reference Assemblies\Microsoft\Framework\.NETFramework", 'v' + version));
+
+        private static readonly string ProgramFilesX86DirectoryPath = Environment.GetFolderPath(
+            Environment.Is64BitOperatingSystem
```
adamsitnik: 💯 thanks for removing the magic strings! the solution now is very clear and easy to understand. + it handles 32bit OS too ;)

src/BenchmarkDotNet/Helpers/FrameworkVersionHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+                if (ndpKey == null)
+                    return null;
+                return Convert.ToInt32(ndpKey.GetValue("Release"));
+            }
+        }
+        
+        [NotNull]
+        internal static string GetCurrentNetFrameworkVersion()
+        {
+            if (!(GetReleaseNumberFromWindowsRegistry() is int releaseNumber))
```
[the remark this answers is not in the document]
epeshk: Oh! Don't notice it, thanks!
https://github.com/dotnet/BenchmarkDotNet/pull/746

Issue #741, opened 2018-05-04
Give users nice warning when T in generic benchmark is not public


adamsitnik:
We now print an ugly error, instead of that the users should get a nice warning from a validator.

Repro:

```cs
public class Bench<T>  where T : new()
{
    [Benchmark] T New() => new T();
}

private class NotPublic { }

BenchmarkRunner.Run<Bench<NotPublic>>():
```

```
04740317-cbb6-4b4f-8ea5-770120ef36c5.notcs(1220,50): error CS0122: 'Request' is inaccessible due to its protection level [C:\Projects\corefxlab\tests\Benchmarks\bin\Release\netcoreapp2.1\04740317-cbb6-4b4f-8ea5-770120ef36c5\BenchmarkDotNet.Autogenerated.csproj]
04740317-cbb6-4b4f-8ea5-770120ef36c5.notcs(655,50): error CS0122: 'Request' is inaccessible due to its protection level [C:\Projects\corefxlab\tests\Benchmarks\bin\Release\netcoreapp2.1\04740317-cbb6-4b4f-8ea5-770120ef36c5\BenchmarkDotNet.Autogenerated.csproj]
04740317-cbb6-4b4f-8ea5-770120ef36c5.notcs(90,50): error CS0122: 'Request' is inaccessible due to its protection level [C:\Projects\corefxlab\tests\Benchmarks\bin\Release\netcoreapp2.1\04740317-cbb6-4b4f-8ea5-770120ef36c5\BenchmarkDotNet.Autogenerated.csproj]
    0 Warning(s)
    3 Error(s)
```

Rizzen:
Solution introduced in https://github.com/dotnet/BenchmarkDotNet/pull/744

PR #744, opened 2018-05-07
BenchmarkClass Validator

Rizzen:
Fixes https://github.com/dotnet/BenchmarkDotNet/issues/741
Made BenchmarkClassValidator that checks generic parameters for public or nested public accessibility (in that case benchmark works fine too). Giving user-friendly warning in case of error.


src/BenchmarkDotNet/Validators/BenchmarkClassValidator.cs
```diff
@@ -0,0 +1,30 @@
...
+using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Running;
+
+namespace BenchmarkDotNet.Validators
+{
+    public class BenchmarkClassValidator: IValidator
+    {
+        public static readonly IValidator Default = new BenchmarkClassValidator();
+        
+        public bool TreatsWarningsAsErrors => false;
```
adamsitnik: this should return true => if `T` is private we can't benchmark it, so it's an error (not warning)

src/BenchmarkDotNet/Validators/BenchmarkClassValidator.cs
```diff
@@ -0,0 +1,30 @@
...
+            var notValid = validationParameters.Benchmarks.Where(x => x.Target.Type.IsGenericType)
+                                                          .Where(HasPrivateGenericArguments).ToList();
+            foreach (var benchmark in notValid)
+            {
+                yield return new ValidationError(true, $"Generic class {benchmark.Target.Type.GetDisplayName()} has non public generic argument(s)");
+            }
+        }
+
+        private bool HasPrivateGenericArguments(Benchmark benchmark) => benchmark.Target.Type.GetGenericArguments().Any(a => !a.IsPublic
+                                                                                                                          && !a.IsNestedPublic);
```
adamsitnik: could you please use `||` instead of `&&` in the condition?

src/BenchmarkDotNet/Validators/BenchmarkClassValidator.cs
```diff
@@ -0,0 +1,30 @@
...
+        public static readonly IValidator Default = new BenchmarkClassValidator();
+        
+        public bool TreatsWarningsAsErrors => false;
+
+        private BenchmarkClassValidator() { }
+        
+        public IEnumerable<ValidationError> Validate(ValidationParameters validationParameters)
+        {
+            var notValid = validationParameters.Benchmarks.Where(x => x.Target.Type.IsGenericType)
+                                                          .Where(HasPrivateGenericArguments).ToList();
```
adamsitnik: nit: `.ToList()` is not mandatory here, you can just yield directly from the `foreach` loop or return and `IEnumberable`

```cs
=> validationParameters.Benchmarks
       .Where(x => x.Target.Type.IsGenericType && HasPrivateGenericArguments(x.Target.Type))
       .Select(benchmark => new ValidationError(true, $"Generic class {benchmark.Target.Type.GetDisplayName()} has non public generic argument(s)"));
```

src/BenchmarkDotNet/Validators/BenchmarkClassValidator.cs
Rizzen: I guess i cant, cause public nested class has IsPublic=false and public non-nested has IsNested=false, so in case of OR condition we have (!a.IsPublic || !a.IsNestedPublic) = true for both public cases and for private case too - thats wrong.

adamsitnik: @Rizzen my bad, sorry! and thanks for the explanation! 

ps. this is a perfect case for unit test ;)

Rizzen: No, I think i forgot about De Morgan's laws

Rizzen: // Restore deleted comment
I guess i cant, cause public nested class has IsPublic=false and public non-nested has IsNestedPublic=false, so in case of OR condition we have (!a.IsPublic || !a.IsNestedPublic) = true for both public cases and for private case too -thats wrong

Rizzen: @adamsitnik What do you think about two ways?

adamsitnik: @Rizzen I like `Any(a => !(a.IsPublic || a.IsNestedPublic))` but I really think that we should have unit tests for this. You should be able to write something similar to [these](https://github.com/dotnet/BenchmarkDotNet/blob/master/tests/BenchmarkDotNet.Tests/Validators/ExecutionValidatorTests.cs) tests for all 4 test cases (public/not public * nested/not nested)

Rizzen:
Made edits and wrote some tests

PR #754, opened 2018-05-14
generate IParams for users in smart way

adamsitnik:
This change generates the right `IParam` for the users for any complex types returned by `[ArgumentsSource]` and `[ParamsSource]` 

I was tired of writing another IParam when I realized that I don't need to do that.

Idea: 

1. generate code which is going to call the `[Source]` getter/method 
2. it always returns `IEnumerable` so we can call `ToArray()` on it
3. use array indexer to get the argument

Examples and auto-generated code:

```cs
[ParamsSource(nameof(DictionaryAsParam))]
public Dictionary<int, string> DictionaryParamInstance;

public IEnumerable<object> DictionaryAsParam => new object[] { new Dictionary<int, string>() { { 1234, "it's an instance getter" } } };

generates:
Runnable_0 instance = new Runnable_0(); 
instance.DictionaryParamInstance= (System.Collections.Generic.Dictionary<System.Int32, System.String>)instance.DictionaryAsParam().ToArray()[0];;
```

```cs
[Benchmark]
[ArgumentsSource(nameof(NonPrimitive))]
public void Simple(SomeClass someClass, SomeStruct someStruct)

public IEnumerable<object[]> NonPrimitive()
{
    yield return new object[] { new SomeClass(Enumerable.Range(0, 100).ToArray()), new SomeStruct(100) };
    yield return new object[] { new SomeClass(Enumerable.Range(0, 1000).ToArray()), new SomeStruct(1000) };
}

__argField0 = (BenchmarkDotNet.IntegrationTests.WithNonPrimitiveArgumentsSource.SomeClass)NonPrimitive().ToArray()[0][0];;
__argField1 = (BenchmarkDotNet.IntegrationTests.WithNonPrimitiveArgumentsSource.SomeStruct)NonPrimitive().ToArray()[0][1];;
```

### DIFF
--- a/docs/guide/Advanced/Arguments.md
+++ b/docs/guide/Advanced/Arguments.md
@@ -46,7 +46,7 @@ public class IntroArguments
 
 In case you want to use a lot of values, you should use `[ArgumentsSource]`.
 
-You can mark one or several fields or properties in your class by the `ArgumentsSource` attribute. In this attribute, you have to specify the name of public method/property which is going to provide the values (something that implements `IEnumerable`). The source must be within benchmarked type! If the values are complex types (not compile-time constants) you should implement `IParam` interface. See [more](Params.md#IParam).
+You can mark one or several fields or properties in your class by the `ArgumentsSource` attribute. In this attribute, you have to specify the name of public method/property which is going to provide the values (something that implements `IEnumerable`). The source must be within benchmarked type! 
 
 ```cs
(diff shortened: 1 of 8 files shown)


adamsitnik:
@AndreyAkinshin please take a look (it's 3:30 am here and I am falling asleep and I need really good review)

AndreyAkinshin:
LGTM

PR #757, opened 2018-05-16
Add SummaryOrderPolicy.Defined to return benchmarks as instantiated

### LINKED ISSUE #720: Add SummaryOrderPolicy.Defined
Suggested by @stephentoub

> I've got this in my config:

```cs
        Add(Job.Default.With(Platform.X64).With(CsProjClassicNetToolchain.Current.Value));
        Add(Job.Default.With(Platform.X64).With(CsProjCoreToolchain.NetCoreApp20));
        Add(Job.Default.With(Platform.X64).With(CsProjCoreToolchain.NetCoreApp21)); 
```

> I wanted the results to always be shown in that order, but there doesn't seem to be a SummaryOrderPolicy that enables that... I was expecting a SummaryOrderPolicy.Defined, or something similar.  To achieve it I had to create my own:

```cs
    private sealed class OriginalOrderProvider : DefaultOrderProvider
    {
        public override IEnumerable<Benchmark> GetExecutionOrder(Benchmark[] benchmarks) => benchmarks;
    } 
```

afmorris:
Fixes #720 

This is a pretty naive implementation of the `SummaryOrderPolicy.Defined`, just returning the incoming benchmarks without doing any sorting against the collection.

An alternative here would be to update the `BenchmarkComparer` to accept a `SummaryOrderPolicy` parameter, and to not do any comparison in that case, but I opted to not change the constructor of `BenchmarkComparer` in this case. Let me know if that's more in line with what you guys were thinking.

### DIFF
--- a/src/BenchmarkDotNet/Order/DefaultOrderProvider.cs
+++ b/src/BenchmarkDotNet/Order/DefaultOrderProvider.cs
@@ -32,12 +32,12 @@ namespace BenchmarkDotNet.Order
 
         public virtual IEnumerable<Benchmark> GetExecutionOrder(Benchmark[] benchmarks)
         {
-            var list = benchmarks.ToList();
             if (summaryOrderPolicy == SummaryOrderPolicy.Defined)
             {
                 return benchmarks;
             }
-
+            
+            var list = benchmarks.ToList();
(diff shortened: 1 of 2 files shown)


adamsitnik:
@afmorris thank you!

Issue #745, opened 2018-05-08
It should be possible to specify the generic arguments by using attributes


adamsitnik:
Let's consider

```cs
public class Bench<T>
{
   [Benchmark] public T Create() => Activator.CreateInstance<T>();
}
```

As of today, to run it for given `T` the users needs to define `T` when running the benchmarks from `BenchmarkRunner` or `BenchmarkSwitcher` level. Example: `BenchmarkRunner.Run<Bench<int>>()`;

It would be nice if this could be defined via attributes, like:

```cs
[GenericBenchmark(typeof(int))]
[GenericBenchmark(typeof(char))]
public class Bench<T>
{
   [Benchmark] public T Create() => Activator.CreateInstance<T>();
}
```

/cc @ahsonkhan

Rizzen:
Possible solution introduced in https://github.com/dotnet/BenchmarkDotNet/pull/758

Also, there is a question - should BenchmarkDotNet allows users to use non-single generic parameter classes as benchmarks?
E.g:
```C#
[GenericBenchmark(typeof(int), typeof(char))]
[GenericBenchmark(typeof(char), typeof(string))]
public class GenericBenchmarks<T, K>
{
    [Benchmark] public T CreateT() => Activator.CreateInstance<T>();

    [Benchmark] public K CreateK() => Activator.CreateInstance<K>();
}
```

adamsitnik:
@Rizzen I changed the naming from `GenericBenchmark` to `GenericTypeArguments` to make it more intuitive.

PR #758, opened 2018-05-18
Generic Benchmark Attribute

Rizzen:
Allows to define generic parameter of benchmark class via attribute.
Solves https://github.com/dotnet/BenchmarkDotNet/issues/745
Example:

```C#
[GenericBenchmark(typeof(int))]
[GenericBenchmark(typeof(char))]
public class GenericBenchmarks<T>
{
    [Benchmark] public T Create() => Activator.CreateInstance<T>();
}
```

### DIFF
--- a/samples/BenchmarkDotNet.Samples/Other/GenericBenchmarks.cs
+++ b/samples/BenchmarkDotNet.Samples/Other/GenericBenchmarks.cs
@@ -1,0 +1,13 @@
+﻿using System;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Attributes.Jobs;
+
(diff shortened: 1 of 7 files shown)


Rizzen:
Also made some generic class constraint checks. Invalid classes justs skips
E.g.
```C#
[GenericBenchmark(typeof(int))]
[GenericBenchmark(typeof(char))]
public class GenericBenchmarks<T> where T: class
{
    [Benchmark] public T Create() => Activator.CreateInstance<T>();
}
```
will be skipped with warning.

Rizzen:
Added multiple argument support? made some refactoring and wrote tests.

adamsitnik:
@Rizzen once again big thanks!

PR #761, opened 2018-05-23
stop the ETW session on Ctrl+C + restore console colors ;), fixes #729

### LINKED ISSUE #729: Handle Ctrl+C/Break
1. end ETW session
2. restore console settings

/cc @jorive

adamsitnik:
/cc @jorive

PS. Testing this is a PITA

### DIFF
--- a/src/BenchmarkDotNet.Diagnostics.Windows/EtwDiagnoser.cs
+++ b/src/BenchmarkDotNet.Diagnostics.Windows/EtwDiagnoser.cs
@@ -40,6 +40,8 @@ namespace BenchmarkDotNet.Diagnostics.Windows
 
             Session = CreateSession(parameters.Benchmark);
 
+            Console.CancelKeyPress += OnConsoleCancelKeyPress;
+
             EnableProvider();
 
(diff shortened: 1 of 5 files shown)


src/BenchmarkDotNet.Diagnostics.Windows/EtwDiagnoser.cs
```diff
@@ -76,6 +78,8 @@ protected void Stop()
             WaitForDelayedEvents();
 
             Session.Dispose();
+
+            Console.CancelKeyPress -= OnConsoleCancelKeyPress;
```
jorive: >CancelKeyPress [](start = 20, length = 14)

This event will only handle Ctrl-C and Ctrl-Break signals. It will not dispose the session if the cmd window is closed.

adamsitnik:
@jorive I added the "close console window" case

jorive:
LGTM

---
In reply to: [391471661](https://github.com/dotnet/BenchmarkDotNet/pull/761#issuecomment-391471661) [](ancestors = 391471661)

PR #766, opened 2018-05-28
Introduce OutlierMode

AndreyAkinshin:


AndreyAkinshin:
@adamsitnik, please review, but don't merge (I'm going to update the documentation as soon as you approve the naming).

src/BenchmarkDotNet/Mathematics/OutlierMode.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace BenchmarkDotNet.Mathematics
+{
+    public enum OutlierMode
+    {
+        None, OnlyUpper, OnlyLower, All
```
adamsitnik: I would add some xml docs here (some users might not understand what is Upper and Lower in this context)

src/BenchmarkDotNet/Engines/EngineResolver.cs
```diff
@@ -24,16 +25,16 @@ private EngineResolver()
...
             Register(AccuracyMode.EvaluateOverheadCharacteristic, () => true);
-            Register(AccuracyMode.RemoveOutliersCharacteristic, job =>
+            Register(AccuracyMode.OutlierModeCharacteristic, job =>
             {
                 var strategy = job.ResolveValue(RunMode.RunStrategyCharacteristic, this);
                 switch (strategy)
                 {
                     case RunStrategy.Throughput:
-                        return true;
+                        return OutlierMode.OnlyUpper;
```
adamsitnik: @AndyAyersMS this change implements your request to remove the outliers that are above the upper fence (the ones below lower fence are now going to remain)

@AndreyAkinshin 👍

src/BenchmarkDotNet/Jobs/JobExtensions.cs
```diff
@@ -52,7 +52,7 @@ public static class JobExtensions
         public static Job WithMinIterationTime(this Job job, TimeInterval value) => job.WithCore(j => j.Accuracy.MinIterationTime = value);
         public static Job WithMinInvokeCount(this Job job, int value) => job.WithCore(j => j.Accuracy.MinInvokeCount = value);
         public static Job WithEvaluateOverhead(this Job job, bool value) => job.WithCore(j => j.Accuracy.EvaluateOverhead = value);
-        public static Job WithRemoveOutliers(this Job job, bool value) => job.WithCore(j => j.Accuracy.RemoveOutliers = value);
```
adamsitnik: @AndreyAkinshin I think that we could keep `WithRemoveOutliers` for backward compatibility and implement it to keep the old behavior. Just to have fewer breaking changes.

```cs
[Obsolete("Please use the new WithOutlierMode instead")]
public static Job WithRemoveOutliers(this Job job, bool value) 
    => job.WithOutlierMode(value ? OutlierMode.All : OutlierMode.None);
```

AndreyAkinshin: @adamsitnik fixed.

src/BenchmarkDotNet/Mathematics/OutlierMode.cs
AndreyAkinshin: Done.

src/BenchmarkDotNet/Jobs/JobExtensions.cs
```diff
@@ -52,7 +52,9 @@ public static class JobExtensions
...
+        public static Job WithOutlierMode(this Job job, OutlierMode value) => job.WithCore(j => j.Accuracy.OutlierMode = value);
+        [Obsolete("Please use the new WithOutlierMode instead")]
+        public static Job WithRemoveOutliers(this Job job, bool value) => job.WithCore(j => j.Accuracy.OutlierMode = OutlierMode.OnlyUpper);
```
adamsitnik: @AndreyAkinshin you forgot the case where `value == false` (users don't want to remove the outliers)

AndreyAkinshin: Fixed.

PR #760, opened 2018-05-22
don't execute long operations more than once per iteration

adamsitnik:
Background: because of the default unroll factor = 16, every iteration was invoking benchmark at least 16 times. 

So if the benchmark took 1s, then:

1. jitting took 16x1s
2. warmup took 6 (MinIterationCount for EngineWarmupStage)  x 16 x 1s = 96 s
3. every target iteration was taking 16s

**Note:** In the default mode, this was described in our docs and was possible to change with custom config. We just did not have the time to fix it.

This PR introduces following changes:

1. EngineFactory is supposed to create an engine which is "ready to run", which means it's supposed to call `GlobalSetup`. (hence the method name change `Create` => `CreateReadyToRun`)
2. Engine implements `IDisposable`, which calls `GlobalCleanup`
3. When the strategy requires us to jit the code first (default mode) and to run the pilot (no explicit value for invocationCount), the engine factory does jitting and **also checks how long does it take to run the benchmark once** (unrol factor=invocation count=1).
 a. If it took longer than IterationTime (0.5s by default), the engine is created for job with one invocation per iteration
 b. otherwise, old rules applies
4. Strategies which don't require jitting don't follow this path (by design)
5. I had to duplicate both the C# and IL we generate to have possibility to choose the mode and keep the Engine immutable. It does not look as bad as  I expected it to look ;)
6. Because the changes applies mostly to `EngineFactory`, I was able to write some nice unit tests (not integration tests)

```cs
public class Sleeep
{
    [Benchmark] public void Sleep1s() => Thread.Sleep(TimeSpan.FromSeconds(1));
}
```

Before: Total time: 00:06:16 (376.01 sec)
After:   Total time: 00:00:27 (27.88 sec)

This fixes #736, makes #730 easy and contributes to #550.

@AndreyAkinshin please take a look ;)
@krzysztofcwalina this fixes the issue you have faced in ML.NET

PR #771, opened 2018-06-01
have two main actions: with unroll and without, for no unroll icrease the step by 1 in pilot (not *2)

adamsitnik:
I am currently working on porting CoreCLR to BenchmarkDotNet.
CoreCLR contains a lot of "benchmark game" benchmarks. These are long-running benchmarks.

After #760 the time it takes to run the benchmarks has greatly improved (ML.NET benchmarks from 20 to 2 minutes). However, I have found two extra scenarios for improvement:

 1. The time to execute benchmark is slighly less than IterationTime. Let's say it's 0.4s when the IterationTime is 0.5s. In that case we invoke it 16 times per iteration (default unroll factor value). It's too much, we should invoke it just once.
2. The time to execute benchmark is just **few** times less than IterationTime. Let's say it's 0.1s when the IterationTime is 0.5s. In that case we invoke it 16 times per iteration (default unroll factor value). It's too much, we should invoke it few times to fit within IterationTime .

My changes are quite simple:

1. If the time to execute benchmark is slighly less than IterationTime, run it once per iteration. No Pilot needed.
2. If the time to execute benchmark is just **few** times less than IterationTime, start the Pilot with MinInvokeCount = 2 (we know 1 is not enough), and increment it with every pilot iteration (not double, it's typically too much for these benchmarks).

Example:

```cs
[Benchmark]
public void Sleep100() => Thread.Sleep(100);
```

```log
IdleJitting  1: 1 op, 803135.91 ns, 803.1359 us/op
MainJitting  1: 1 op, 100786001.20 ns, 100.7860 ms/op

Pilot        1: 2 op, 200567013.38 ns, 100.2835 ms/op
Pilot        2: 3 op, 301861923.92 ns, 100.6206 ms/op
Pilot        3: 4 op, 401721624.60 ns, 100.4304 ms/op
Pilot        4: 5 op, 503124019.07 ns, 100.6248 ms/op

MainWarmup   1: 5 op, 503016250.03 ns, 100.6033 ms/op
```

### DIFF
--- a/src/BenchmarkDotNet/Engines/EnginePilotStage.cs
+++ b/src/BenchmarkDotNet/Engines/EnginePilotStage.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Horology;
 using BenchmarkDotNet.Jobs;
 
@@ -67,7 +68,10 @@ namespace BenchmarkDotNet.Engines
                 if (invokeCount >= MaxInvokeCount)
                     break;
 
-                invokeCount *= 2;
+                if (unrollFactor == 1 && invokeCount < EnvResolver.DefaultUnrollFactorForThroughput)
+                    invokeCount += 1;
+                else
+                    invokeCount *= 2;
             }
             WriteLine();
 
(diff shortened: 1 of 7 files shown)


src/BenchmarkDotNet/Engines/EnginePilotStage.cs
```diff
@@ -67,7 +68,10 @@ private long RunAuto()
                 if (invokeCount >= MaxInvokeCount)
                     break;
 
-                invokeCount *= 2;
+                if (unrollFactor == 1 && invokeCount < EnvResolver.DefaultUnrollFactorForThroughput)
```
adamsitnik: @AndreyAkinshin the `invokeCount < EnvResolver.DefaultUnrollFactorForThroughput` check is for the benchmarks, where the first call is super expensive and every next can be very fast (like initiating the cache and later just reading from it)

So we are going to increment it until we reach `EnvResolver.DefaultUnrollFactorForThroughput` and after it just double it.

PR #777, opened 2018-06-04
Trying to fix the build

adamsitnik:


adamsitnik:
[Explanation](https://github.com/dotnet/sdk/issues/1615#issuecomment-332952774):

> By default, if you don't explicitly specify an sdk version in a global.json, the SDK used will be the highest version installed across:

> The 'sdk' folder next to the dotnet.exe being executed
> The %PROGRAMFILES%\dotnet\sdk folder
> You can opt out of this behavior by setting

> $env:DOTNET_MULTILEVEL_LOOKUP = '0'

PR #783, opened 2018-06-07
BenchmarkDotNet.Samples refactoring, fixes #583

### LINKED ISSUE #583: BenchmarkDotNet.Samples refactoring
In the early versions of BenchmarkDotNet, the Samples project had two purposes:
1. It was our set of manual integration tests (we didn't have real integration tests)
2. It contained a lot of "cool" benchmarks.

Right now we have good integration tests and the community have a lot of new awesome BenchmarkDotNet-based benchmarks. The most precious piece of BenchmarkDotNet.Samples is the Intro folder. My suggestion:

1. All Intro samples which can be useful for users as a demonstration of our API (like `IntroParamsSource` or `IntroRankColumn`) should be well-documented and should be a part of the official documentation. Right now we are copying parts of these samples into documentation manually; we should do it automatically.
2. If we still have some interesting samples which should be executed manually (they can be useful for diagnostics), but we don't want to move it into IntegrationTests projects (because they take too much time), we should move it into a separate project (like `ManualIntegrationTests`). An example: `IntroEmpty`.
3. All other samples which are not useful for users or BenchmarkDotNet developers (like `Framework_SelectVsConvertAll` or `IL_Loops`) should be removed from the source code. These samples were useful in the past, but they are not useful now.

AndreyAkinshin:
* All non-Intro samples are removed from the repo
* Added the "Samples" section in the documentation
* Intro*-samples have corresponding articles in this docs section. The source code is included as-is with the help of the `[!code-csharp[IntroParams.cs](../../../samples/BenchmarkDotNet.Samples/IntroParams.cs)]` DocFX syntax. Thus, the samples are completely synchronized, we will not have outdated samples in the documentation.
* Samples can be included in other documentation sections via `[!include[IntroParams](../samples/IntroParams.md)]`.
* We have a new article which covers parameterization: it's just a compilation of 5 samples. You can find a preview here: https://andreyakinshin.github.io/bdn-docs-wip/articles/features/parameterization.html
* It's recommended to use `[`[Params]`](xref:BenchmarkDotNet.Attributes.ParamsAttribute)` in docs, so we get clickable links to the API.
* Currently, only parametrization is covered by the new kind of samples. Other samples are (temporary) in the `WithoutDocs` samples.


samples/BenchmarkDotNet.Samples/Other/GenericBenchmarks.cs
```diff
@@ -1,13 +0,0 @@
-﻿using System;
-using BenchmarkDotNet.Attributes;
-using BenchmarkDotNet.Attributes.Jobs;
-
-namespace BenchmarkDotNet.Samples.Other
```
adamsitnik: this file should be part of intro

AndreyAkinshin: Fixed.

PR #785, opened 2018-06-08
Improve filtering from console args

adamsitnik:
As of today, if the user wants to filter the benchmarks by namespace/class/category, the user needs to provide the filter, and then for the benchmark switcher put `*` for the filter to work...

This change: we filter in once place, if the user provides filter from console args the benchmark switcher will not print the benchmarks, just run them.

I need that because we have MANY benchmarks in CoreCLR + CoreFX and I need to make the filtering simple.


AndreyAkinshin:
LGTM. Could you also update docs (`docs/articles/configs/filters.md`) and add a few examples?

adamsitnik:
@AndreyAkinshin I added the docs and fixed `--join` (sth has broke it along the way)

AndreyAkinshin:
The `CheckCore` test doesn't work anymore.

docs/articles/samples/IntroCategories.md
```diff
@@ -0,0 +1,23 @@
...
+uid: BenchmarkDotNet.Samples.IntroCategories
+---
+
+## Sample: IntroCategories
+
+Combined together with `[BenchmarkCategory]` attribute, you can group the benchmarks into categories and filter them by categories.
+
+### Source code
+
+[!code-csharp[IntroFilters.cs](../../../samples/BenchmarkDotNet.Samples/IntroCategories.cs)]
```
AndreyAkinshin: I guess it should be IntroCategories instead of IntroFilters.

docs/articles/configs/filters.md
```diff
@@ -8,82 +8,37 @@ name: Filters
...
-    [Benchmark]
-    [BenchmarkCategory("B", "1")]
-    public void B1() => Thread.Sleep(10); // Will be benchmarked
-    
-    [Benchmark]
-    [BenchmarkCategory("B", "2")]
-    public void B2() => Thread.Sleep(10);
-}
-```
+[!include[IntroFilters](../samples/IntroCategories.md)]
```
AndreyAkinshin: Here we also need `IntroCategories`

adamsitnik:
@AndreyAkinshin thanks for the review!

PR #790, opened 2018-06-11
add link to inprocesstoolchain

IanKemp:


AndreyAkinshin:
@IanKemp, thanks for the PR. Please, use `uid`s for such links. It's better because `uid` is a persistent identifier, the link will continue to work even if the relative path of the file is changed.

See also:
https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#define-uid
https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#different-syntax-of-cross-reference

PR #796, opened 2018-06-13
docs: multiversion combobox

AndreyAkinshin:
Preview: https://andreyakinshin.github.io/bdn-docs-wip/
image

Benefits:
* We can keep the old version of documentation on `https:/benchmarkdotnet.org/vNUMBER/`
* We can deploy documentation for the master branch on `https://benchmarkdotnet.org/`
* The implementation of this combobox is pretty simple and doesn't change the template.

Current problems:
* We still need a set of scripts for build & deploy, but it sounds solvable.


AndreyAkinshin:

PR #799, opened 2018-06-17
Cpu info improvement

Rizzen:
In order to improve BDN with constant folding analyzer we need to detect CPU speed.
So, I guess we can improve CpuInfo class to do it. For now it is PoC for Windows and Linux (cause I haven't Mac atm).

- On Windows still easy to use ```wmic``` but with additional ```CurrentClockSpeed``` parameter

- On Linux it's a bit difficult - ```cat /proc/cpuinfo``` output contains only current CPU Frequency. We can use ```lscpu | grep MHz``` that returns three parameters of CPU - min, max and current CPU Frequency. So i thought that we can measure actual CPU speed with something like dummy method, that loads CPU and  runs ```lscpu | grep MHz``` during dummy work ([example](https://gist.github.com/Rizzen/1255ceb685702f29a25c11388ba96b0b)). But values of that measure have a huge distribution (i had from 1300 to 3100 MHz). Thus I guess we can use ```lscpu | grep max MHz```.

- On Mac I hope we can use ```sysctl hw.cpufrequency``` but it has to be checked.


src/BenchmarkDotNet/Portability/Cpu/CpuInfo.cs
```diff
@@ -6,13 +6,15 @@ public class CpuInfo
         public int? PhysicalProcessorCount { get; }
         public int? PhysicalCoreCount { get; }
         public int? LogicalCoreCount { get; }
+        public double? CurrentClockSpeed { get; } //MHz
```
AndreyAkinshin: It will be better to use `Frequency` here instead of the `//MHz` comment.

src/BenchmarkDotNet/Portability/Cpu/CpuInfoFormatter.cs
```diff
@@ -33,7 +33,8 @@ private static string Format(string processorName, int? physicalProcessorCount,
                 parts.Add("1 physical core");
             if (physicalCoreCount > 1)
                 parts.Add($"{physicalCoreCount} physical cores");
-
+            if (processorFreq > 0)
+                parts.Add($", processor frequency = {processorFreq} MHz");
```
AndreyAkinshin: ProcessorBrandString usually contains the information about processor frequency. We want to keep this section as small as possible, so there is no need to duplicate this info.

src/BenchmarkDotNet/Portability/Cpu/ProcCpuInfoProvider.cs
```diff
@@ -18,9 +20,20 @@ private static CpuInfo Load()
...
             return null;
         }
+        
+        private static string GetCpuSpeed()
+        {
+            var output = ProcessHelper.RunAndReadOutput("/bin/bash","-c \"lscpu | grep \"max MHz\"\"")?
+                                      .Split('\n').First()
+                                      .Split(' ').Last();
+            
+            return $"\ncpu freq\t:{output}";
```
AndreyAkinshin: We should use `ProcCpuInfoKeyNames.CpuFreq` here.

AndreyAkinshin:
> On Mac I hope we can use sysctl hw.cpufrequency but it has to be checked.

On `macOS High Sierra 10.13.5 (17F77) / Intel Core i7-4870HQ CPU 2.50GHz (Haswell)`, I have the following output:
```
$ sysctl hw.cpufrequency
hw.cpufrequency: 2500000000
```

src/BenchmarkDotNet/Portability/Cpu/CpuInfoFormatter.cs
AndreyAkinshin: If we have a difference between frequency in the brand string and the actual max frequency, probably it makes sense to patch the branch string and get something like `Intel(R) Core(TM) i5-4308U CPU @ 2.80GHz (Max: 3.30GHz)`.
What do you think?

AndreyAkinshin: Hmm... It seems that `wmic CurrentClockSpeed` and `sysctl hw.cpufrequency` equal to the frequency value in the brand string. Let's call it "nominal frequency".
I don't know how to get the nominal frequency on Linux. `lscpu` is a good way to get the minimum and maximum frequency. Maybe it makes sense to introduce separate properties for such values.

src/BenchmarkDotNet/Helpers/ProcessHelper.cs
```diff
@@ -38,6 +38,7 @@ internal static string RunAndReadOutput(string fileName, string arguments = "")
                 process.WaitForExit();
                 return output;
             }
+
```
AndreyAkinshin: We don't need empty line here.

AndreyAkinshin:
@Rizzen, I like this approach. It's definitely good to know advanced information about CPU frequency.

src/BenchmarkDotNet/Portability/Cpu/CpuInfoFormatter.cs
Rizzen: I think that the way to patch the brand string is awesome idea.

On my machine brand string doesn't match CurrentClockSpeed value
On ```wmic cpu get Name, CurrentClockSpeed /Format:List``` I have:
```
CurrentClockSpeed=3701
Name=Intel(R) Core(TM) i5-2500 CPU @ 3.30GHz
```
Of Linux - what if we measure ```lscpu```  with example from https://gist.github.com/Rizzen/1255ceb685702f29a25c11388ba96b0b and take max value? 
Anyway I discovered that max value measured under load is close to Max value from ```lscpu```  so I think we can consider max lscpu value as Frequency.

AndreyAkinshin: On the `CpuInfo` level, it makes sense to introduce 3 properties: `NominalFrequency`, `MaxFrequency`, and `MinFrequency`. Next, we can provide `EffectiveFrequency` API which returns a proper value for our analyzers.
By the way, on macOS, we can get all 3 values:
```
$ sysctl -a | grep cpufreq
hw.cpufrequency: 2500000000
hw.cpufrequency_min: 2500000000
hw.cpufrequency_max: 2500000000
```

Rizzen:
@AndreyAkinshin Big thanks for your help! So I implemented Nominal, Max and Min frequencies for all platforms. Also I implemented experimental EffectiveCpuFrequency API for Linux (for Windows and Mac we just used values that we already have - there is no other way).

Rizzen:
I made some more investigations with LinuxEffectiveCpuProvider. So, there is a weird results.
I have ```Intel Core i5-2520M 2.50Ghz```. With that there was min measured result ~800 MHz, max ~3100 MHz, average was ~3000 MHz.
Thus, I hope on Linux we can parse frequency from brand string and use it in case when measured effective frequency is lower.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
```diff
@@ -8,6 +8,9 @@
     <PackageId>BenchmarkDotNet</PackageId>
     <RootNamespace>BenchmarkDotNet</RootNamespace>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <Optimize>true</Optimize>
+  </PropertyGroup>
```
AndreyAkinshin: Please remove these lines

src/BenchmarkDotNet/Environments/ProcessorBrandStringHelper.cs
```diff
@@ -10,13 +10,18 @@ public static class ProcessorBrandStringHelper
         /// Transform a processor brand string to a nice form for summary.
         /// </summary>
         /// <param name="processorName">Original processor brand string</param>
+        /// <param name="frequency">processor actual frequency</param>
         /// <returns>Prettified version</returns>
         [NotNull]
-        public static string Prettify([NotNull] string processorName)
+        public static string Prettify([NotNull] string processorName, double? frequency = null)
```
AndreyAkinshin: It should be renamed to maxFrequency; `Frequency?` should be used instead of `double?`

src/BenchmarkDotNet/Portability/Cpu/CpuInfo.cs
```diff
@@ -6,13 +6,29 @@ public class CpuInfo
         public int? PhysicalProcessorCount { get; }
         public int? PhysicalCoreCount { get; }
         public int? LogicalCoreCount { get; }
+        public double? NominalFrequency { get; }
+        public double? MinFrequency { get; }
+        public double? MaxFrequency { get; }
+        public double? EffectiveFrequency { get; }
```
AndreyAkinshin: `Frequency` should be used everywhere instead of `double`

src/BenchmarkDotNet/Portability/EffectiveFrequency/MacEffectiveCpuProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+    internal class MacEffectiveCpuProvider
+    {
+        internal static readonly Lazy<CpuInfo> MacEffectiveCpuInfo = new Lazy<CpuInfo>(GetInfo);
+        
+        private static CpuInfo GetInfo()
+        {
+            
+            if (RuntimeInformation.IsMacOSX())
+            {
+                string content = ProcessHelper.RunAndReadOutput("sysctl", "-a");
```
AndreyAkinshin: We already call `sysctl -a` in `SysctlCpuInfoProvider`. Please, reuse the result instead of another call.

AndreyAkinshin:
@Rizzen, sorry for the late reply, I was on a business trip.

> Thus, I hope on Linux we can parse frequency from brand string and use it in case when measured effective frequency is lower.

It's a pretty good idea. I guess it will be enough for our analyzers. I suggest to do the following:

* Let's keep only `MinFrequency`, `MaxFrequency`, `NominalFrequency` (without "EffectiveFrequency"); the type of all properties should be `Frequency`
* `NominalFrequency` should use `sysctl` on macOS, `wmic` on Windows, and the cpu brand string on Linux
* `NominalFrequency` will be used in our analyzers. It's still good to know the min/max values, but I don't know how we can use it for now.
* If `MaxFrequency` is known and there is a difference between `NominalFrequency` and `MaxFrequency` (we have this situation on Linux), the cpu brand string should be patched (already implemented).

What do you think?

Rizzen:
@AndreyAkinshin I think that is pretty nice solution, cause it looks like that is only way to get reproducable nominal CPU frequency on Linux. Also I totally forgot about Frequency, thanks for remind!

src/BenchmarkDotNet/Environments/ProcessorBrandStringHelper.cs
```diff
@@ -37,12 +38,11 @@ public static string Prettify([NotNull] string processorName, double? frequency
...
         /// </summary>
         /// <param name="frequency"></param>
-        private static string GetBrandStyledActualFrequency(double? frequency)
+        private static string GetBrandStyledActualFrequency(Frequency? frequency)
         {
             if (frequency == null)
                 return null;
-            var a = frequency / 1000;
-            return $"{a.Value.ToString("N2").Replace(',','.')}GHz";
+            return $"{frequency.Value.ToGHz().ToString("N2").Replace(',','.')}GHz";
```
AndreyAkinshin: It will be better to use `ToStr()` (which use correct culture) instead of ugly `.Replace(',','.')`

src/BenchmarkDotNet/Portability/Cpu/ProcCpuInfoParser.cs
```diff
@@ -61,8 +61,14 @@ internal static CpuInfo ParseOutput([CanBeNull] string content)
         private static double ParseFrequencyFromBrandString(string brandString)
```
AndreyAkinshin: This method should return `Frequency`. Also, it would be nice to make it internal and write a few unit tests.

Rizzen: So should CpuInfo ctor has nominalFrequency, minFrequency and maxFrequency arguments of type Frequency?

AndreyAkinshin: Yes. We should use `Frequency` everywhere when it's possible.

src/BenchmarkDotNet/Portability/Cpu/ProcCpuInfoParser.cs
```diff
@@ -13,7 +16,12 @@ internal static CpuInfo ParseOutput([CanBeNull] string content)
             var logicalCores = SectionsHelper.ParseSections(content, ':');
             var processorModelNames = new HashSet<string>();
             var processorsToPhysicalCoreCount = new Dictionary<string, int>();
+            
             var logicalCoreCount = 0;
+            var nominalFrequency = (Frequency)0d;
+            var minFrequency = (Frequency)0d;
+            var maxFrequency = (Frequency)0d;
```
AndreyAkinshin: The statement looks ugly to me. Let's introduce
```cs
public static readonly Frequency Zero = new Frequency(0);
```
in `Frequency.cs` and use `Frequency.Zero` here.

src/BenchmarkDotNet/Portability/Cpu/ProcCpuInfoParser.cs
```diff
@@ -25,15 +33,46 @@ internal static CpuInfo ParseOutput([CanBeNull] string content)
                 if (logicalCore.TryGetValue(ProcCpuInfoKeyNames.ModelName, out string modelName))
                 {
                     processorModelNames.Add(modelName);
+                    nominalFrequency = ParseFrequencyFromBrandString(modelName);
                     logicalCoreCount++;
                 }
+                
+                if (logicalCore.TryGetValue(ProcCpuInfoKeyNames.MinFrequency, out string minCpuFreqValue)
+                    && double.TryParse(minCpuFreqValue.Replace(',','.'), out double minCpuFreq))
```
AndreyAkinshin: `double.TryParse` uses the current culture by default. Thus, this construction will not work in some environments; we should always use `HostEnvironmentInfo.MainCultureInfo` for each parsing operation. To avoid copy-pasting, I suggest introducing additional methods in `Frequency.cs`:
```cs
public static bool TryParse(string s, FrequencyUnit unit, out Frequency freq)
{
    var success = double.TryParse(s, NumberStyles.Any, HostEnvironmentInfo.MainCultureInfo, out double result);
    freq = new Frequency(result, unit);
    return success;
}

public static bool TryParseHz(string s, out Frequency freq) => TryParse(s, FrequencyUnit.Hz, out freq);
public static bool TryParseKHz(string s, out Frequency freq) => TryParse(s, FrequencyUnit.KHz, out freq);
public static bool TryParseMHz(string s, out Frequency freq) => TryParse(s, FrequencyUnit.MHz, out freq);
public static bool TryParseGHz(string s, out Frequency freq) => TryParse(s, FrequencyUnit.GHz, out freq);
```
Then, we can use `Frequency.TryParseMHz` everywhere.

src/BenchmarkDotNet/Environments/ProcessorBrandStringHelper.cs
```diff
@@ -10,13 +12,18 @@ public static class ProcessorBrandStringHelper
         /// Transform a processor brand string to a nice form for summary.
         /// </summary>
         /// <param name="processorName">Original processor brand string</param>
+        /// <param name="frequency">processor actual frequency</param>
         /// <returns>Prettified version</returns>
         [NotNull]
-        public static string Prettify([NotNull] string processorName)
+        public static string Prettify([NotNull] string processorName, Frequency? frequency = null)
```
AndreyAkinshin: It should be renamed to `maxFrequency`

src/BenchmarkDotNet/Portability/Cpu/ProcCpuInfoProvider.cs
```diff
@@ -18,9 +19,28 @@ private static CpuInfo Load()
...
+                                      .Split('\n')
+                                      .SelectMany(x => x.Split(':'))
+                                      .ToArray();
+            
+            if (output == null || output.Length < 6)
+                return null;
+                
+            double.TryParse(output[1].Trim(), out double currentValue);
+            double.TryParse(output[3].Trim().Replace(',', '.'), out double maxValue);
+            double.TryParse(output[5].Trim().Replace(',', '.'), out double minValue);
```
AndreyAkinshin: It would be nice to add a comment with an example of the output. It will help to understand what kind of data do we parse.

AndreyAkinshin:
@Rizzen as soon as you fix my last remarks, the PR will be ready for merging.

AndreyAkinshin:
@Rizzen, thanks!

PR #800, opened 2018-06-17
job Mutators

### LINKED ISSUE #713: How to specify the invocation/launch count per type when using Config for multiple runtimes?
@vitek-karas asked me this question today and honestly, I am not sure if we have a good solution for that case.

Background: The user has a few types with benchmarks and wants to run all of them for different runtimes **and** have a different invocation/iteration count per type.

The problem: when I set a `SimpleJob` attribute per type and I configure custom jobs in the config class I pass to runner, the settings don't get merged, but instead the jobs gets added:

```cs
class Program
{
    static void Main(string[] args) =>
        BenchmarkSwitcher.FromAssembly(typeof(Program).GetTypeInfo().Assembly).Run(args,
            DefaultConfig.Instance.With(Job.Clr).With(Job.Core));
}

[SimpleJob(runStrategy: RunStrategy.Monitoring, warmupCount: 1, targetCount: 10, invocationCount: 1)]
public class First
{
    [Benchmark] public void Nothing() { }
}

[SimpleJob(runStrategy: RunStrategy.Monitoring, warmupCount: 1, targetCount: 1, invocationCount: 1)]
public class Second
{
    [Benchmark] public void Nothing() { }
}
```

|  Method |     Job | Runtime | RunStrategy | TargetCount | UnrollFactor | WarmupCount |          Mean |     Error |    StdDev |
|-------- |-------- |-------- |------------ |------------ |------------- |------------ |--------------:|----------:|----------:|
| Nothing | Default |     Clr |  Monitoring |           1 |            1 |           1 | 3,421.2400 ns |        NA | 0.0000 ns |
| Nothing |     Clr |     Clr |     Default |     Default |           16 |     Default |     0.9211 ns | 0.0075 ns | 0.0063 ns |
| Nothing |    Core |    Core |     Default |     Default |           16 |     Default |     0.0000 ns | 0.0000 ns | 0.0000 ns |

I got it working by having a dedicated config type per type with benchmarks, but this idea sucks because it requires too much code.

```cs
class Program
{
    {
        BenchmarkSwitcher.FromAssembly(typeof(Program).GetTypeInfo().Assembly).Run(args);
    }
}

public class MyConfig : ManualConfig
{
    public MyConfig()
    {
        Add(GetPerTypeSettings(Job.Clr));
        Add(GetPerTypeSettings(Job.Core));
    }

    protected virtual Job GetPerTypeSettings(Job runtimeJob) => runtimeJob;
}

    
public class MyConfigFirst : MyConfig
{
    protected override Job GetPerTypeSettings(Job runtimeJob) => runtimeJob.With(RunStrategy.Monitoring).WithInvocationCount(1).WithTargetCount(1);
}

[Config(typeof(MyConfigFirst))]
public class First
{
    [Benchmark] public void Nothing() { }
}

public class MyConfigSecond : MyConfig
{
    protected override Job GetPerTypeSettings(Job runtimeJob) => runtimeJob.With(RunStrategy.Monitoring).WithInvocationCount(1).WithTargetCount(10);
}

[Config(typeof(MyConfigSecond))]
public class Second
{
    [Benchmark] public void Nothing() { }
}
```

``` ini

BenchmarkDotNet=v0.10.13.20180409-develop, OS=Windows 10.0.16299.309 (1709/FallCreatorsUpdate/Redstone3)
Frequency=3507503 Hz, Resolution=285.1031 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2633.0
  Job-LKIZOH : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2633.0
  Job-BINUMK : .NET Core 1.1.7 (CoreCLR 4.6.26201.01, CoreFX 4.6.24705.01), 64bit RyuJIT

InvocationCount=1  RunStrategy=Monitoring  TargetCount=1  

```
|  Method | Runtime |     Mean | Error |
|-------- |-------- |---------:|------:|
| Nothing |     Clr | 827.1 us |    NA |
| Nothing |    Core | 290.2 us |    NA |

@AndreyAkinshin @ig-sinicyn do you guys have a better idea how to get it work with less code? Or maybe I will have to change the way we merge configs? Sth like: if two different jobs configure different settings merge them into one instead of adding all of them. ?

adamsitnik:
As of today, whenever we define few jobs, they are being added, not merged.

Example:
2 jobs for Runtimes: Core and Mono
1 job with IterationCount = 20

So far the result was 3 jobs: one for Core, one for Mono, and one for the current runtime with IterationCount = 20.

This PR allows telling what is mutator (and should be merged). I also provided few attributes, which I need on my own.

@AndreyAkinshin could you take a look? I will add the docs tomorrow.

@jorive My motivation is that I want to have single config for CoreCLR and CoreFX benchmarks, but make some exceptions per some types where the settings are not enough.

fixes #713


AndreyAkinshin:
I like it! `Meta.IsMutator` looks really nice.

My only concern is about the Attribute Style API. Attributes is a convenient way to introduce new jobs. The new attributes mutate existed jobs. Will it be obvious? Maybe we have to add a note about mutators on the first line of xml docs for such attributes.

adamsitnik:
> My only concern is about the Attribute Style API. Attributes is a convenient way to introduce new jobs. The new attributes mutate existed jobs. Will it be obvious?

@AndreyAkinshin I was also thinking about this. I would say that what we have now is what users expect naturally: the old job attribues always add a new job (it was expected behaviour), the new attributes cover settings that should be "merged".

```cs
[ClrJob, CoreJob]
[GcServer(true)]
public class MyBenchmarkClass
```

Is going to be merged to a config with two jobs:

* CoreJob with `GcServer=true`
* ClrJob with `GcServer=true`

I decided not to use `Mutator` in the attributes names just to keep it simple and intuitive. I also added `AllowMultiple = false` so it's now impossible to do:

```cs
[GcServer(true)]
[GcServer(false)] // compilation error
public class MyBenchmarkClass
```

I also added missing xml docs to the code, updated the docs and added some predefined attributes for the things I believe people are going to use.

adamsitnik:
@AndreyAkinshin build is red because dotnet cli can't be downloaded from Azure Blob Storage, I have reported this problem internally in MS

AndreyAkinshin:
@adamsitnik, the appveyor build is red because of the following problem: https://github.com/dotnet/BenchmarkDotNet/issues/801

Issue #804, opened 2018-06-25
What is the point of BuildScriptFilePath ?


Shazwazza:
I'm in the process of attempting to implement #290 and noticed that the Generators create a `.bat` build script which seems quite handy, but then the Builders don't actually use the build script generated.

For example the `DotNetCliBuilder` runs it's own `restore` and `build` commands and doesn't use the generated build script to do the building.  The Roslyn Builder outputs a log entry for the `BuildScriptFilePath` but then doesn't use it at all.

This might be intentional, so sorry if I'm being naive (the solution is quite a learning curve :)

adamsitnik:
Hi @Shazwazza 

It's a very good question!

BDN used to generate **and run** this script. At some point of time, we moved from doing this to calling Roslyn API directly. The script was preserved for situations when we fail to build the benchmark project, apply `[KeepBenchmarkFiles]` attribute and run it manually to reproduce the problem. So we keep it only for eventual troubleshooting.

You can ignore it if you are developing a custom toolchain.

Shazwazza:
Ah, ok thanks :) I might add some code docs to those classes in my PR.

Issue #290, opened 2016-10-23
Question: Any official way to benchmark same method between different assembly versions?


Shazwazza:
Hi,

I can't seem to find anything specifically about what I'd like to do, but if this has already been discussed some other place please let me know :)

Otherwise, what I'd like to do is benchmark the same method of an assembly but with different assembly versions. I know I can accomplish this in various ways and use reflection to load in the type/method based on different assembly files but am wondering if there is any official way to do this?

adamsitnik:
H! @Shazwazza ! 

Currenly we don't support it, but I am planning to add such feature. 

My initial goal was to enable comparison for for assemblies that are available on the nuget. Sample scenario: what's the difference between current implementation and latest version available on nuget? 

What would be your scenario?

Shazwazza:
yes that is exactly what I was planning on implementing :) 

I haven't written anything quite yet but wanted to try to get something going as early as next week. The idea that I had was something along the lines of:

* During the benchmark initialization, have some code config to point to either a referenced assembly or a nuget alias + version
* Use Nuget API (or whatever) to go fetch the configured Nuget assembly versions from whatever Nuget sources are configured
* Then load in the class/method from either the referenced assembly or the Nuget fetched assembly
* Run the benchmark

adamsitnik:
One question: are you using .csprojs or .xprojs? I have an idea how dotnet cli could do the nuget restore for us. 

If you are using csprojs then I think that the best thing to do would be to extend our RoslynToolchain. [Here](https://github.com/PerfDotNet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Toolchains.Roslyn/Toolchains/Classic/RoslynBuilder.cs#L32) you can see how we build the code with Roslyn. I think that the solution would be to extend our Job class with list of assembiles/nugets that would be used by the [RoslynGenerator](https://github.com/PerfDotNet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Toolchains.Roslyn/Toolchains/Classic/RoslynGenerator.cs#L48)

then it would a matter of executing all benchmarks in two configurations (by using two separate Jobs): without custom assemblies and with custom assemblies.

Shazwazza:
For the current project I'm working with it would be csproj. Thanks for the pointers, I'll fork the project and have a look and hopefully i can find some time :)

GeertvanHorrik:
I am currently working on this for my own purposes. I will push this to the (open source) repository soon:

[...]

Issue #305, opened 2016-11-23
Summary Save/Load/Combine


AndreyAkinshin:
Now BenchmarkDotNet allows to create a nice set of jobs for different environments. However, sometimes it's impossible to get all desired result with a single run. Examples:
* Comparing performance before and after refactoring
* Comparing performance between different version of a NuGet package (probably will be implemented in the future, but it's not available for now)
* Comparing performance between different OS (Windows/Linux/MacOS)

So, I suggest to add `Save`/`Load`/`Combine` methods which will help to solve this problems. An example:
```cs
// Part 1: WIndows
var summary = BenchmarkRunner.Run<MyBench>();
summary.Save("windows.json");
// Part 2: Linux
var summary = BenchmarkRunner.Run<MyBench>();
summary.Save("linux.json")
// Part 3: Common summary table
var summaryWindows = Summary.Load("windows.json");
var summaryLinux   = Summary.Load("linux.json");
var summary = Summary.Combine(summaryWindows, summaryLinux);
```

Some additional thoughts / points for discussions.
* Is it ok to save summary in the json format?
* Probably it would be nice to have an API for adding additional columns to the current summary like `var newSummary = mySummary.WithAdditionalColumn("VersionOfMyLibrary", "0.10.0")`.

@adamsitnik, @mattwarren, @ig-sinicyn, @terrajobst, what do you think?

ig-sinicyn:
@AndreyAkinshin, please, not JSON. There's standard format for storing and exchanging the data - XML.
It has good conventions for value escaping and date representation. It has strong specification without silly logical errors such as undetermined behavior for duplicate keys.
There's API for it out of the box, no additional dependencies required.

Why JSON?
About API: why not expose it as an `Exporter`?

Oh, and now goes the worst part: how the `Combine()` should work at all? Averaging? No way.

### Case 1.
Imagine two benchmark methods, `A` and `B`
```
On PC1:
`A` - 1000ns (Vector.HardwareAccelerated` is false); `B` - 100 ns

On PC2
`A` - 150ns (Vector.HardwareAccelerated` is true); `B` - 300 ns
```
now what?
The bad news: we have no need to imagine it. It's a real case we were faced a year ago.

### Case 2.
```
On PC1:
`A` - 200ns

On PC2
`B` - 150ns
```
how should these be merged at all? There's no common baseline for it so we have no idea what the actual ratio is. It can be like
```
On PC2
`B` - 150ns; A - 50ns
[...]

PR #805, opened 2018-06-25
WIP for supporting testing between Nuget dependencies

Shazwazza:
This is a WIP to resolve the task https://github.com/dotnet/BenchmarkDotNet/issues/290

I've got this working with the `CsProjCoreToolchain` and have added a sample showing the syntax and how it works. During development i noticed that the `BuildScriptFilePath` doesn't ever seem to be used so raised a question/issue about that here https://github.com/dotnet/BenchmarkDotNet/issues/804

What I need help understanding at this stage is how to get this working with the RoslynToolchain. I can see that the Roslyn Generator calls into `csc` and there's a `/references` switch so I assume I'd need to find a way to go fetch the nuget packages and get these wired up but I'm not sure if there's some sort of automated way like with the dotnet cli that can be used.  Any ideas on this one?

Please provide any feedback you might have. Hopefully i haven't done something too odd, there's a bit of a learning curve involved in this codebase :)

### DIFF
--- a/samples/BenchmarkDotNet.Samples/IntroNuget.cs
+++ b/samples/BenchmarkDotNet.Samples/IntroNuget.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using BenchmarkDotNet.Analysers;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Exporters.Csv;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Toolchains.CsProj;
+
+namespace BenchmarkDotNet.Samples
+
+    [Config(typeof(Config))]
+    public class IntroNuget
+    {
+        private class Config : ManualConfig
+        {
+            public Config()
+            {
+                //TODO: So far only implemented with any toolchain using DotNetCliBuilder
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "11.0.1").WithId("11.0.1"));
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "10.0.3").WithId("10.0.3"));
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "9.0.1").WithId("9.0.1"));
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "8.0.1").WithId("8.0.1"));
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "7.0.1").WithId("7.0.1"));
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "6.0.1").WithId("6.0.1"));
+                Add(DefaultConfig.Instance.GetColumnProviders().ToArray());
+                Add(DefaultConfig.Instance.GetLoggers().ToArray());
+                Add(CsvExporter.Default);
+            }
+        }
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            var jsonConvertMethod = Assembly.Load("Newtonsoft.Json").GetType("Newtonsoft.Json.JsonConvert")
+                .GetMethods().Where(x => x.Name == "SerializeObject" && x.GetParameters().Length == 1)
+                .First();
+
+            Serializer = s => (string)jsonConvertMethod.Invoke(null, new object[] { s });
+        }
+
+        public Func<object, string> Serializer { get; private set; }
+
+        [Benchmark]
+        public void SerializeAnonymousObject() => Serializer(new { hello = "world", price = 1.99, now = DateTime.UtcNow });
+    }
+}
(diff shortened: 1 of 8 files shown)


samples/BenchmarkDotNet.Samples/IntroNuget.cs
```diff
@@ -0,0 +1,52 @@
...
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Exporters.Csv;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Toolchains.CsProj;
+
+namespace BenchmarkDotNet.Samples
+
```
adamsitnik: missing '{' after the namespace (`IntroNuget.cs(15,34): error CS1514: {`)

samples/BenchmarkDotNet.Samples/IntroNuget.cs
```diff
@@ -0,0 +1,52 @@
...
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "7.0.1").WithId("7.0.1"));
+                Add(Job.MediumRun.With(CsProjCoreToolchain.Current.Value).WithNuget("Newtonsoft.Json", "6.0.1").WithId("6.0.1"));
+                Add(DefaultConfig.Instance.GetColumnProviders().ToArray());
+                Add(DefaultConfig.Instance.GetLoggers().ToArray());
+                Add(CsvExporter.Default);
+            }
+        }
+
+        [GlobalSetup]
+        public void Setup()
```
adamsitnik: If the project that defines benchmarks (let's call it A) references `Newtonsoft.Json 9.0.1` and we are going to generate two projects:

1. 'Newtonsoft.Json 11.0.1' (B)
2. 'Newtonsoft.Json 6.0.1' (C)

Then which version of 'Newtonsoft.Json' is going to be restored and loaded during the execution? Will it be 11.0.1  for B and 9.0.1 for C? Is this why you are using reflection here?

adamsitnik:
@Shazwazza big thanks for doing this!

samples/BenchmarkDotNet.Samples/IntroNuget.cs
Shazwazza: That's the million dollar question :) I'm using reflection in this test for 2 reasons: 1) so we don't have to reference this assembly in this project and 2) precisely to avoid this issue

I've been working to get Benchmark.Net to run benchmarks between many of our products' versions and have struggled with this very question. In an ideal world, you could do this:

* Project "A" (the benchmark project) references the minimum version of the nuget dependency that is being tested in the list of Jobs so that reflection doesn't have to be used to call the APIs and it builds
* The list of Jobs specifies the exact version that will be tested and that is what will be resolved from Nuget
* So long as the API signatures haven't changed between all of these versions and the assembly isn't strong-named (or there's no major version changes) then it should 'just work'

For simple projects/tests the above should be fine but of course ideal worlds don't always exist. In our case there's quite a bit of method signature changes between versions so this solution isn't going to work.  I've had a couple of ideas (some perhaps zany!): 

1. Have 'baseline' projects, one for each minimum nuget version to be tested for a particular API surface area. Then define the Jobs in these projects that reference the higher nuget version dependencies that you know will work. Then you could call these command line apps in sequence  and combine the csv results either manually or potentially with some sort of custom `IExporter`. Otherwise, you could have a parent command line app that calls these in sequence and if we could serialize/deserialize the `Summary` then we could combine them and send them to an exporter (i.e. #305 ) ... This is a pretty round about way to get this done though!
1. Have 'baseline' nuget packages that include the benchmarks to run then in your Jobs, you would specify the baseline nuget package which by default references the minimum nuget version to be tested for a given API surface area, and also include the specific Nuget version of the dependency to be tested. There's a problem with this approach though because there's no way to convert this to `BenchmarkRunInfo` that the `BenchmarkRunner` needs. Something would have to be created/modified to be able to resolve the `BenchmarkRunInfo.Type` and `BenchmarkRunInfo.BenchmarkCase` for each Job run and the runner would somehow still need to know about which cases it will be running.
1. Create a new `BenchmarkRunner.RunProjects` methods which would be similar to above but instead of building nuget packages containing the benchmarks, you could point the `BenchmarkRunner` at a list of csproj files containing benchmarks. This also suffers from the same problem listed above though.
1. Utilize something like the `RunUrl` (maybe `RunFiles`) where you point the `BenchmarkRunner` to a list of files, these files also contain metadata in a header (like yaml or something) which is parsed out and could contain assembly/nuget references. The `RunFiles` method has a parameter for the user to specify which benchmarks to run by name. Then Benchmark.Net builds the first file, finds the specified benchmarks by name which it uses as it's `BenchmarkRunInfo`, it executes that benchmark, then moves on to the next file, builds it, verifies it, runs it, and so on, all in the while collection a list of `Summary` objects which it can then combine at the end with `Summary.Join`

Sorry for very long message, but been meaning to write this all down somewhere! Let me know what you think or any ideas you've got.

adamsitnik:
@Shazwazza please excuse me for a late reply

On Monday we should release 0.11.0, then I take a week off and after that, I will most probably merge your PR to a separate branch and test it.

Rbn3D:
@Shazwazza Does this PR allow benchmark between two versions of the same project, but one as project reference (source) and the other as nuget?

This is my exact use case.

I'm trying to test performance optimizations i'm doing for a existing nuget project, and the ideal way to go would be to benchmark the nuget version agaisnt the source version directly.

This way I can add the benchmark project to the repo, and the performance of the changes I'm doing could be measured easily, without even having to commmit, or publish to nuget. Just like running a test suite :)

Any ideas?

Thanks in advance.

Shazwazza:
@Rbn3D This PR indeed would let you benchmark between different versions of your product but it's not really intended to work with a reference and also Nugets, it works only with Nuget references but this will still work for you since i'm sure you can make a Nuget package of your current version.

The way it would need to work is that your benchmark project would reference your minimum version of your product that has the same API surface area so you can write your benchmarks using those APIs, then you would create each benchmark to run against specific nuget versions which would load in the newer versions of your product and execute the newer versions of those same APIs. See here for more details: https://github.com/dotnet/BenchmarkDotNet/pull/805#discussion_r197746744

Shazwazza:
@adamsitnik Hi adam, just bumping this one :) I realize this is out of date with latest changes so happy to try to get this updated, but seeing as though I'd love a hacktoberfest t-shirt, could i close this PR and submit a new one? ;)

adamsitnik:
@Shazwazza sure thing!

PR #824, opened 2018-07-09
Use 3rd party lib for console args parsing + support globs for filtering

### LINKED ISSUE #795: Broken BenchmarkSwitcher
Repro:
```
dotnet BenchmarkDotNet/samples/BenchmarkDotNet.Samples/bin/Release/netcoreapp2.1/BenchmarkDotNet.Samples.dll IntroBasic
```
BenchmarkSwitcher doesn't parse `IntroBasic` from arguments and prints the list of available benchmarks.

adamsitnik:
Today I tried to fix #795 and when I realized how hard it's going to be to add it to our custom hand-written console args parsing I decided to rewrite it and reuse some 3rd party library to make my life easier.

I also added glob support as suggested by @jorive

Now we can do something like:

`--job dry --filter *` - will run all benchmarks once
`--runtimes clr core mono` - will run selected benchmarks for Core, CLR and Mono using Default job
`--job short --filter System.Linq*ToArray` - will use Job.Short to run all benchmarks from System.Linq namespace with method name which ends with `ToArray`

I also removed few things which I believe nobody is using but are adding extra complexity.


adamsitnik:
I have fixed the tests and added the docs

AndreyAkinshin:
@adamsitnik, I don't like an idea to reference `CommandLineParser` in the `BenchmarkDotNet` package. It can introduce problems for people who already use a specific version of `CommandLineParser` and want to use `BenchmarkDotNet`. Could we introduce an additional package (e.g., `BenchmarkDotNet.CommandLine`) and move all command line-related logic into this package. People who want to run benchmarks via command line will be able to do it with this package. People who don't need this feature will be able to use `BenchmarkDotNet` without additional dependency.

src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+using BenchmarkDotNet.Mathematics;
+using CommandLine;
+using CommandLine.Text;
+
+namespace BenchmarkDotNet.ConsoleArguments
+{
+    public class CommandLineOptions
+    {
+        [Option('j', "job", Required = false, Default = "Default", HelpText = "Dry/Short/Medium/Long or Default")]
+        public string BaseJob { get; set; }
```
jorive: >string [](start = 15, length = 6)

Here I would use an `enum` (`JobType`?), then Default = JobType.Default.
The same for the options below.

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
```diff
@@ -50,73 +49,41 @@ public static BenchmarkSwitcher FromAssemblyAndTypes(Assembly assembly, Type[] t
         /// <summary>
         /// Run all available benchmarks.
         /// </summary>
-        public IEnumerable<Summary> RunAll() => Run(new[] { "*" });
+        public IEnumerable<Summary> RunAll() => Run(new[] { "--filter", ".*" });
```
jorive: >.* [](start = 73, length = 2)

We are asking users for glob, but expect regex wildcard. We should be consistent here.

adamsitnik: @jorive I changed regex  into real glob

PR #825, opened 2018-07-12
Accuracy based job attributes

Zhentar:
The documentation (rightly) encourages encourages trusting the automatic determination of iteration count to achieve the desired accuracy. But this advice is undermined by the attributes to achieve faster testing runs - using hardcoded iteration counts. I think these attributes effectively address that, trusting the algorithm and simply informing it of relaxed accuracy requirements.

I'd also like to add a `DiagnosersOnly` job/attribute, but I'm not sure how to best address the varying iteration requirements (e.g. the disassembly diagnoser only ever needs one run, memory diagnoser benefits from multiples. PMC benefits from more warmups); I would appreciate any input on that

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,0 @@
-# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
-
-# All files
-[*]
-indent_style = space
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+
+# All files
+[*]
+indent_style = space
--- a/src/BenchmarkDotNet/Attributes/Jobs/ModerateAccuracyJobAttribute.cs
+++ b/src/BenchmarkDotNet/Attributes/Jobs/ModerateAccuracyJobAttribute.cs
@@ -1,0 +1,11 @@
+﻿using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Attributes
+{
+    public class ModerateAccuracyAttribute : JobConfigBaseAttribute
+    {
+        public ModerateAccuracyAttribute() : base(Job.ModerateAccuracy)
+        {
+        }
+    }
+}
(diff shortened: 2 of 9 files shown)


.editorconfig
```diff
@@ -0,0 +1,5 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
```
AndreyAkinshin: .editorconfig is not related to the accuracy based job attributes. Please, remove it from this pull request.

src/BenchmarkDotNet/Attributes/Jobs/ModerateAccuracyJobAttribute.cs
```diff
@@ -0,0 +1,11 @@
+﻿using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Attributes
+{
+    public class ModerateAccuracyAttribute : JobConfigBaseAttribute
```
AndreyAkinshin: All attributes which are not mutators should have the "Job" suffix.

AndreyAkinshin:
@Zhentar thanks for the PR, this idea looks interesting.
We need an example in our documentation which explains these attributes and shows a minimal case when these attributes can be useful.

Zhentar:
Oops, forgot to comment here - I've made the requested changes. 

If the documentation comment was directed at me, I'll need some guidance on how one goes about making such updates.

AndreyAkinshin:
@Zhentar, you should add an Intro example in the `BenchmarkDotNet.Samples` project. This example should show a case when these attributes can be useful. Please, make it as short as possible. Next, you should add a markdown file with the same name in `docs/articles/samples` which contains explanations for these attributes. Use the same docs format which we use in other samples.

AndreyAkinshin:
@Zhentar any progress with this PR? It doesn't make sense to add new accuracy modes without examples which demonstrate that they are really useful.

AndreyAkinshin:
@Zhentar we don't have any updates in this PR since July, so I close this PR. Feel free to open another one If you come up with good examples of introduced accuracy modes.

Issue #828, opened 2018-07-15
Benchmark Reporting - Deadlock/Getting Stuck Reading From Process.StandardOutput (NetCore 2.1)


houseofcat:
#### Update (7/15/2018 @ 7:25 PM EST)
Decoupled my Benchmark project from a local project and instead added reference to the NuGet it corresponds to.

In addition to that, I installed 2.302 SDK (`NetCore 2.1.2`).

Standard Sleep benchmark is working again.

Moving to BenchmarkDotNet nightlies (BDN v0.10.14.685, Core v0.10.14.508).

#### Update (7/15/2018 @ 8:07 PM EST)
Nothing seems to work.

Thread stuck here:
![image](https://user-images.githubusercontent.com/1414138/42739871-1b3c0d3e-886d-11e8-8251-f98258663e0b.png)

Found the exception in background:
![image](https://user-images.githubusercontent.com/1414138/42739968-4fa063bc-886e-11e8-9c86-8285b17b5b94.png)

Not sure how I can help but the file it's stuck on seems to be the .log.

I allowed continuation found the code frozen in one spot - continuously frozen in one spot:
![image](https://user-images.githubusercontent.com/1414138/42740075-e8a4afe0-886f-11e8-9178-7e21727d28f9.png)

```
Not Flagged	>	40620	1	Main Thread	Main Thread  

System.Private.CoreLib.dll!System.IO.FileStream.ReadFileNative    

[Managed to Native Transition]   

System.Private.CoreLib.dll!System.IO.FileStream.ReadFileNative(Microsoft.Win32.SafeHandles.SafeFileHandle handle, System.Span<byte> bytes, System.Threading.NativeOverlapped* overlapped, out int errorCode) Line 1190  
System.Private.CoreLib.dll!System.IO.FileStream.ReadNative(System.Span<byte> buffer) Line 507   
System.Private.CoreLib.dll!System.IO.FileStream.ReadSpan(System.Span<byte> destination) Line 433   
System.Private.CoreLib.dll!System.IO.FileStream.Read(byte[] array, int offset, int count) Line 302   
System.Private.CoreLib.dll!System.IO.StreamReader.ReadBuffer() Line 637  
System.Private.CoreLib.dll!System.IO.StreamReader.ReadLine() Line 812  
BenchmarkDotNet.dll!BenchmarkDotNet.Loggers.SynchronousProcessOutputLoggerWithDiagnoser.ProcessInput()  
[...]

PR #830, opened 2018-07-16
Read StandardOutput in a smart way to avoid infinite loops

thyams:
Potential fix for SynchronousProcessOutputLoggerWithDiagnoser infinite loop.
Related to my issue in #828 . Detected an abnormal streamread lock/loop. After some investigation, it looks to be potentially an infinite loop at the StreamReader.ReadLine(). It is capable of returning `string.Empty` and `null` as valid values.  The loop proceeds to continue when `string.IsNullOrEmpty`.

If EndOfStream is never achieved/reached, then a break or timeout will need to be added.

Ran the unit test that calls Execute and found that it froze during WaitForExit. I saw the comment should we add a Timeout and proceeded to do so.

Added timeout in Execute, detected freeze during UnitTest.
`ToolchainTests.CustomToolchainsAreSupported()`

* The infinite loop won't be fixed if the stream never ends.
* The infinite loop won't be fixed if the StreamReader can't read the underlying file (OS lock contention).
   * Wanted to have added a 5 second timeout to read the line to assist with this issue, but the underlying stream doesn't support timeouts.

The Execute timeout probably needs to be reviewed and debated about, I have chosen an arbitrary 30 seconds on a WaitForExit().

Commit 2/3, attempted to add timeouts then undo them.

Commit 4 attempted enhancement utilizing StreamReader peek() to see if it can bypass a ReadLine timeout.

Similar Stackoverflow Examples:
https://stackoverflow.com/questions/35318588/streamreader-readline-will-hang-in-an-infinite-loop
https://stackoverflow.com/questions/286533/filestream-streamreader-problem-in-c-sharp
https://stackoverflow.com/questions/13705209/check-if-a-streamreader-can-read-another-line

### DIFF
--- a/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
+++ b/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
@@ -38,22 +38,26 @@ namespace BenchmarkDotNet.Loggers
         internal void ProcessInput()
         {
             string line;
-            while ((line = process.StandardOutput.ReadLine()) != null)
+
+            // ReadLine() can return string.Empty/null as values.
+            while (!process.StandardOutput.EndOfStream)
             {
-                logger.WriteLine(LogKind.Default, line);
+                line = process.StandardOutput.ReadLine();
 
-                if (string.IsNullOrEmpty(line))
-                    continue;
-
-                if (!line.StartsWith("//"))
-                    LinesWithResults.Add(line);
-                else if (Engine.Signals.TryGetSignal(line, out var signal))
+                if (!string.IsNullOrEmpty(line))
                 {
-                    diagnoser?.Handle(signal, diagnoserActionParameters);
-                    process.StandardInput.WriteLine(Engine.Signals.Acknowledgment);
+                    logger.WriteLine(LogKind.Default, line);
+
+                    if (!line.StartsWith("//"))
+                    { LinesWithResults.Add(line); }
+                    else if (Engine.Signals.TryGetSignal(line, out var signal))
+                    {
+                        diagnoser?.Handle(signal, diagnoserActionParameters);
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
```diff
@@ -38,22 +38,30 @@ public SynchronousProcessOutputLoggerWithDiagnoser(ILogger logger, Process proce
...
+                            diagnoser?.Handle(signal, diagnoserActionParameters);
+                            process.StandardInput.WriteLine(Engine.Signals.Acknowledgment);
+                        }
+                        else
+                        { LinesWithExtraOutput.Add(line); }
+                    }
                 }
-                else
-                    LinesWithExtraOutput.Add(line);
+                else { break; } // No more 'characters' can be read.
```
adamsitnik: could you please remove the `else { break; }`  and move the condition to while loop?

`while (!process.StandardOutput.EndOfStream && process.StandardOutput.Peek() > 0)`

src/BenchmarkDotNet/Toolchains/Executor.cs
```diff
@@ -69,7 +69,11 @@ private ExecuteResult Execute(Process process, BenchmarkCase benchmarkCase, Sync
 
             loggerWithDiagnoser.ProcessInput();
 
-            process.WaitForExit(); // should we add timeout here?
+            // Timeout Added. Freeze detected.
+            if (!process.WaitForExit(30000))
```
adamsitnik: 30000 ms is 30 s which is definitely too little for majority of the benchmarks we run.

Please revert this change and create a new issue where we can discuss what the default timeout should be. We should also make it configurable

thyams:
@adamsitnik changes made, off to work for me!

adamsitnik:
@thyams once again big thanks!

From time to time our Travis MacOs build hangs and timeouts. I hope that this change is going to fix this issue too!

PR #854, opened 2018-08-16
Exclude Directory.Build.props/targets from generated csproj files

agocke:
Unless explicitly excluded, MSBuild will walk up the directory tree looking for
Directory.Build.props and Directory.Build.targets files. This is problematic for
one of the most common uses of Directory.Build.* files: redirecting output
directories. The rest of the benchmarking infrastructure has strong expectations
of where files will be produced, so things will often go wrong if they are redirected.


adamsitnik:
Thanks!

Issue #859, opened 2018-08-22
Strange max frequency values on Windows


AndreyAkinshin:
In v0.11.0, we introduced advanced `CpuInfo` which contains the maximum CPU frequency.
On Windows, this value is based on the `wmic` output (`CurrentClockSpeed`). It turned out that sometimes these values look pretty strange: the evaluated frequency can be lower than the brand string frequency. Sometimes the difference between brand string and evaluated frequencies is about 0.01GHz.
The examples are listed below.

---

https://github.com/in-async/WeightedSamplingWithoutReplacement/blob/511652d58450e308be11378fe95cef55c984d2b6/README.md
```
Intel Core i5-3570K CPU 3.40GHz (Max: 3.30GHz) (Ivy Bridge), 1 CPU, 4 logical and 4 physical cores
```

---

https://github.com/silver-xu/Span-Memory-Benchmark/blob/d574522be1077e0c058afc9c711bc7970da7bb6f/SpanMemoryBenchmark/bin/Release/netcoreapp2.1/BenchmarkDotNet.Artifacts/results/SpanSample.Benchmarks-report-github.md
```
Intel Core i7-6650U CPU 2.20GHz (Max: 2.21GHz) (Skylake), 1 CPU, 4 logical and 2 physical cores
```

---

https://github.com/evilguest/linq2d/blob/9eaac5164bc9ffd4b160aaab93b9cde77b0b871c/README.md
```
Intel Core i7-6600U CPU 2.60GHz (Max: 2.61GHz) (Skylake), 1 CPU, 4 logical and 2 physical cores
```

---

https://github.com/ryzngard/Chipotle.CSV/blob/71d39352b02aea990834c2e2ae541ff6c5f45bfb/BenchmarkDotNet.Artifacts/Combined.md
```
Intel Core i7-8650U CPU 1.90GHz (Max: 1.91GHz) (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
```

---

https://github.com/JohanLarsson/BenchmarkBox/blob/d90f963540248f7b4071ea01bc52597d7743f815/BenchmarkBox/FindVsFirstOrDefault.md
```
Intel Xeon CPU E5-2637 v4 3.50GHz (Max: 3.49GHz), 2 CPU, 16 logical and 8 physical cores
```

---

https://github.com/dhhoang/IonDotnet/blob/87053280a9328d901cae2ca7f3aa4eb5f49aed37/README.md
```
[...]

PR #860, opened 2018-08-23
Fix strange CPU Frequency values

Rizzen:
For https://github.com/dotnet/BenchmarkDotNet/issues/859


src/BenchmarkDotNet/Portability/Cpu/MosCpuInfoProvider.cs
```diff
@@ -33,7 +33,7 @@ private static CpuInfo Load()
                     processorsCount++;
                     physicalCoreCount += (uint) moProcessor[WmicCpuInfoKeyNames.NumberOfCores];
                     logicalCoreCount += (uint) moProcessor[WmicCpuInfoKeyNames.NumberOfLogicalProcessors];
-                    nominalClockSpeed = maxClockSpeed = minClockSpeed = (uint) moProcessor[WmicCpuInfoKeyNames.CurrentClockSpeed];
+                    nominalClockSpeed = maxClockSpeed = minClockSpeed = (uint) moProcessor[WmicCpuInfoKeyNames.MaxClockSpeed];
```
AndreyAkinshin: I think it would be better to set only `maxClockSpeed`.
`nominalClockSpeed` can be parsed from the CPU brand string.
It also makes sense to don't set `minClockSpeed` on Windows (because we don't know how to get the right value).

Issue #865, opened 2018-08-30
Truncated parameter strings used as group causes problems


marklam:
If a benchmark has parameters which are long strings, they get truncated to 
(_first five chars_)…(_last five chars_) [_length_]

If two parameters are the same length and have the same first/last five characters, they benchmarks get grouped together. 
If one of the methods is marked as a baseline, BenchmarkDotNet complains that the group has two baselines.

```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using System.Threading;

namespace RunnerCS
{
    public class ThingsToBenchmark
    {
        [Params("12345thisisaLongparameter54321", 
                "12345thisisaRongparameter54321")]
        public string Parameter { get; set; }

        [Benchmark(Baseline = true)]
        public void Unmodified()
        {
            Thread.Sleep(100);
        }

        [Benchmark]
        public void Faster()
        {
            Thread.Sleep(90);
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            var summary = BenchmarkRunner.Run<ThingsToBenchmark>();
        }
    }
}
```
produces `Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Parameter=12345(...)54321 [30]] in class ThingsToBenchmark has 2`

adamsitnik:
@marklam big thanks for the very good bug report!

PR #866, opened 2018-08-31
Fix behavior of Baseline property of Benchmark attribute in integration with Params Attribute

Caballero77:
For issue https://github.com/dotnet/BenchmarkDotNet/issues/865


Caballero77:
Hello @adamsitnik 

Thank you very much for review. I have fixed your remark.

PR #878, opened 2018-09-17
EtwProfiler Diagnoser

adamsitnik:
This PR introduces a new Diagnoser that will allow MS to switch from xunit-performance to BenchmarkDotNet.

What it does: it allows us to also collect profiles using ETW. It can be done as a separate run (the default setting, the profiler overhead is not included in the benchmark results), or together with benchmarking (the overhead is included, but also in the Overhead stage to minimize it).

The end result is an `.etl` file that we can open with some profiler like PerfView. The trace file contains not only traces but also CLR events, which can be used by the CLR Team to investigate possible regressions. I made sure that it works for both .NET and .NET Core and than 0 stacks are broken (it was not trivial)

As for now it works only on Windows, but in the future I could also make it work with other OSes using EventPipe mechanism from CoreCLR (I would love to do that, but I currently don't have time for it ).

@AndreyAkinshin I tried to keep the number of changes in BenchmarkDotNet.dll as small as possible, most of the things landed in our Diagnostics package. The new things are `Metrics`, which are reported by Diagnosers and exported together with the measurements.

/cc @jorive @danmosemsft


adamsitnik:
Sample FlameGraph from PerfView:
rangetoarray flamegraph1

AndreyAkinshin:
@adamsitnik, the changes look great! Give me a few days to check it in different environments. Meanwhile, could you please add docs about new features?

src/BenchmarkDotNet/Engines/IterationStage.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace BenchmarkDotNet.Engines
 {
-    public enum IterationStage
+    public enum IterationStage : int
```
jorive: Should it be called `TaskName`?

src/BenchmarkDotNet/Engines/IterationMode.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace BenchmarkDotNet.Engines
 {
-    public enum IterationMode
+    public enum IterationMode : int
```
jorive: Should it be used as the `OpcodeName` with a start/stop?

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            return this;
+        }
+    }
+
+    internal class KernelSession : Session
+    {
+        public KernelSession(DiagnoserActionParameters details, int bufferSizeInMb) : base(KernelTraceEventParser.KernelSessionName, details, bufferSizeInMb) { }
+        
+        protected override string FileExtension => ".kernel.etl";
```
jorive: I think that on Windows 8+, you do not need to generate two etl files.

adamsitnik:
@jorive I have divided the single generic "iteration event" into separate events (WorkloadJitting, Pilot, Actual etc) as you suggested. Now when opening the trace file with WPA it's much easier to filter the time:

image

Also I made the `EtwProfiler` very configurable, user can provide all of the settings for the Kernel keywords and user providers we enable. Also after changing the intervals to those recommended the size of a trace file has dropped significantly (-30%) but the accuracy of the PMCs remained the same (because BDN runs the benchmarks MANY times ;) )

adamsitnik:
@AndreyAkinshin now I am going to add the docs ;p

docs/articles/configs/diagnosers.md
```diff
@@ -12,18 +12,20 @@ The current Diagnosers are:
...
   [![NuGet](https://img.shields.io/nuget/v/BenchmarkDotNet.svg)](https://www.nuget.org/packages/BenchmarkDotNet.Diagnostics.Windows/).
   Please see Adam Sitnik's [blog post](http://adamsitnik.com/Hardware-Counters-Diagnoser/) for all the details.
 - Disassembly Diagnoser.
   It allows you to disassemble the benchmarked code to asm, IL and C#/F#.
   Please see Adam Sitnik's [blog post](http://adamsitnik.com/Disassembly-Diagnoser/) for all the details.
+- ETW Profiler (`EtwProfiler`).
+  It allows you to not only benchmark, but also profile the code. It's using TraceEvent, which internally uses ETW and exports all the information to a trace file. The trace file contains all of the stack traces captured by the profiler, PDBs to resolve symbols for both native and managed code and captured GC, JIT and CLR events. Please use one of the free tools: PerfView or Windows Performance Analyzer to analyze and visualize the data from trace file. You can find this diagnoser in a separate package with diagnosers for Windows (`BenchmarkDotNet.Diagnostics.Windows`): [![NuGet](https://img.shields.io/nuget/v/BenchmarkDotNet.svg)](https://www.nuget.org/packages/BenchmarkDotNet.Diagnostics.Windows/)
```
AndreyAkinshin: I guess we need an Intro* sample which demonstrates how we can get some useful information with the help of `EtwProfiler`. It may include a few PerfView screenshots (e.g., FlameGraph).
Also, it would be nice to show how `EtwProfiler` can be configured.

adamsitnik:
@AndreyAkinshin I added the docs. The gif files were heavy so I just referenced the ones from my blog. I hope it's ok

docs/articles/features/etwprofiler.md
```diff
@@ -0,0 +1,94 @@
...
+
+`EtwProfiler` uses `TraceEvent` library which internally uses Event Tracing for Windows (ETW) to capture stack traces and important .NET Runtime events.
+
+Before the process with benchmarked code is started, EtwProfiler starts User and Kernel ETW sessions. Every session writes data to it's own file and captures different data. User session listens for the .NET Runtime events (GC, JIT etc) while the Kernel session gets CPU stacks and Hardware Counter events. After this, the process with benchmarked code is started. During the benchmark execution all the data is captured and written to a trace file. Moreover, BenchmarkDotNet Engine emits it's own events to be able to differentiate jitting, warmup, pilot and actual workload when analyzing the trace file. When the benchmarking is over, both sessions are closed and the two trace files are merged into one.
+
+## Limitations
+
+What we have today comes with following limitations:
+
+* EtwProfiler works only on Windows (one day I might implement similar thing for Unix using EventPipe)
```
AndreyAkinshin: We should avoid phrases with "I" in the official documentation.

adamsitnik:
I corrected the docs. I am going to use it for a few days and if I don't find any new bugs I am going to squash it into master.

docs/articles/features/etwprofiler.md
```diff
@@ -0,0 +1,94 @@
...
+
+1. Open Events window
+2. Put "WorkloadActual" in the Filter box and hit enter.
+3. Press control or shift and choose the Start and Stop events from the left panel. Hit enter.
+4. Choose iteration that you want to investigate (events are sorted by time).
+5. Select two or more cells from the "Time MSec" column.
+6. Right click, choose "Open Cpu Stacks".
+7. Choose the process with benchmarks, right-click, choose "Drill Into"
+
+![](https://adamsitnik.com/images/etwprofiler/perfview_events.gif)
```
jorive: Should this image be checked in in this repo instead?

PR #886, opened 2018-10-01
Enabled GcModeTests.CanEnableServerGcMode

dlemstra:
In #221 there was a request to investigate why CanEnableServerGcMode failed on AppVeyor. I enabled this and ran a build on my own AppVeyor account to check if this still happened and it looks like this is fixed.

This PR enables the unit test and that would mean we could close issue #221.


dlemstra:
The travis build has failed but that looks unrelated. It also failed for some other pull requests. The appveyor build has passed now.

adamsitnik:
@dlemstra thank you!

PR #887, opened 2018-10-01
Dependencies update

adamsitnik:
I am currently trying to add the dependency to BenchmarkDotNet in CoreFX.

But MSBuild complains about some of our old dependencies.. So I will be updating this PR until I can get both builds green ;)


AndreyAkinshin:
@adamsitnik, do we really have to put dll files in `src/BenchmarkDotNet/Disassemblers/net46/*`?

adamsitnik:
@AndreyAkinshin honestly I don't remember. I will remove them and see if all builds passes on all OSes

adamsitnik:
@AndreyAkinshin everything works fine, good catch!

PR #888, opened 2018-10-01
Fix duplicate example in RunStrategy guide

stevedesmond-ca:
The `RunStrategy` guide currently shows 2 identical `ColdStart` examples. I'm assuming we want to show `Monitoring` as the second one, instead.


AndreyAkinshin:
@stevedesmond-ca, thanks!

PR #890, opened 2018-10-02
RPlotExporter: use https to download dependencies

xavero:
Using a secure https connection should be preferred over http.

Download over http can be blocked by some corporate firewalls/proxy.


adamsitnik:
@xavero thank you!

Issue #521, opened 2017-08-07
Support async Setup/Cleanup


AndreyAkinshin:

AndreyAkinshin:
Related issue: https://github.com/dotnet/BenchmarkDotNet/issues/517

dlemstra:
@adamsitnik I already started working on the async `GlobalCleanup` part because the `GlobalSetup` will be merged soon. Is there a reason why `ExecutionValidatorBase` does not check the `GlobalCleanup`?

adamsitnik:
@dlemstra I have merged #892

>  Is there a reason why ExecutionValidatorBase does not check the GlobalCleanup?

no, it must be a bug. It would be great if you could fix it

dlemstra:
Will try to put that in a new PR then.

PR #892, opened 2018-10-03
Added support for async GlobalSetup.

dlemstra:
This PR adds support for an `async` `GlobalSetup` as requested in #521. I am also planning to add support for `async` `GlobalCleanup.` I can add this to the same PR if that is preferred or I could open a separate PR for that.

### DIFF
--- a/src/BenchmarkDotNet/Code/DeclarationsProvider.cs
+++ b/src/BenchmarkDotNet/Code/DeclarationsProvider.cs
@@ -70,12 +70,16 @@ namespace BenchmarkDotNet.Code
                 return EmptyAction;
             }
 
-            if (method.ReturnType != typeof(Task))
+            if (method.ReturnType == typeof(Task) ||
+                method.ReturnType == typeof(ValueTask) ||
+                (method.ReturnType.IsGenericType &&
+                    (method.ReturnType.GetGenericTypeDefinition() == typeof(Task<>) ||
+                     method.ReturnType.GetGenericTypeDefinition() == typeof(ValueTask<>))))
             {
-                return method.Name;
+                return $"() => {method.Name}().GetAwaiter().GetResult()";
             }
 
-            return $"() => {method.Name}().GetAwaiter().GetResult()";
+            return method.Name;
         }
     }
 
--- a/src/BenchmarkDotNet/Code/DeclarationsProvider.cs
+++ b/src/BenchmarkDotNet/Code/DeclarationsProvider.cs
@@ -28,7 +28,7 @@ namespace BenchmarkDotNet.Code
 
         public string WorkloadTypeName => Descriptor.Type.GetCorrectCSharpTypeName();
 
-        public string GlobalSetupMethodName => Descriptor.GlobalSetupMethod?.Name ?? EmptyAction;
+        public string GlobalSetupMethodName => GetMethodName(Descriptor.GlobalSetupMethod);
 
         public string GlobalCleanupMethodName => Descriptor.GlobalCleanupMethod?.Name ?? EmptyAction;
 
@@ -62,6 +62,21 @@ namespace BenchmarkDotNet.Code
             => Descriptor.WorkloadMethod.GetParameters()
                 .Where(arg => !string.IsNullOrEmpty(arg.ParameterType.Namespace))
                 .Select(arg => $"using {arg.ParameterType.Namespace};");
+
+        private string GetMethodName(MethodInfo method)
+        {
+            if (method == null)
+            {
+                return EmptyAction;
+            }
+
+            if (method.ReturnType != typeof(Task))
+            {
+                return method.Name;
+            }
+
+            return $"() => {method.Name}().GetAwaiter().GetResult()";
+        }
     }
 
     internal class VoidDeclarationsProvider : DeclarationsProvider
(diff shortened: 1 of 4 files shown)


adamsitnik:
Hi @dlemstra 

Big thanks for the PR! However it's not complete yet.

BDN generates and compiles the code for every benchmark.

[Here](https://github.com/dotnet/BenchmarkDotNet/blob/382a4af4f1a1d057077c69eb37026545910b966e/src/BenchmarkDotNet/Templates/BenchmarkType.txt#L60) is the template for a C# file which is filled with value [here](https://github.com/dotnet/BenchmarkDotNet/blob/29471e8a6a800bf389590b9ac76df6c1845b2f2f/src/BenchmarkDotNet/Code/CodeGenerator.cs#L58)

So as of today we have sth like:

`private Action globalSetupAction = $nameOfTheSetupMethod$`

which should be changed to sth like:

`private Action globalSetupAction = () => $nameOfTheSetupMethod$.GetAwaiter().GetResult()`

for async methods. We generate similar code for async benchmarks [link](https://github.com/dotnet/BenchmarkDotNet/blob/29471e8a6a800bf389590b9ac76df6c1845b2f2f/src/BenchmarkDotNet/Code/DeclarationsProvider.cs#L150)

dlemstra:
@adamsitnik I am not sure if I understand completely what you want me to change. I just pushed a patch to overload the `GlobalSetupMethodName` inside `TaskDeclarationsProvider`. But I don't know how to test my changes.

adamsitnik:
Hi @dlemstra 

The `TaskDeclarationsProvider` is used when the benchmark method returns a Task, not when Setup or Cleanup method returns a Task.

My suggestion: please move the logic from `TaskDeclarationsProvider` to base class. If the setup method is async, use the awaiter. If it's not, keep it the old way.

You wrote a really nice test for the `ExecutionValidator`. This validator is used on demand when our users want to check if it's possible to run the benchmark to tell early if something is wrong. 

What would be the best: get familiar with `AllSetupAndCleanupTest` and write two new test cases:

1. Async benchmark and async setup
2. Sync benchmark, async setup

`AllSetupAndCleanupTest` is an integration test which actually runs the benchmark. Last week I added some [docs](https://github.com/dotnet/BenchmarkDotNet/blob/master/docs/articles/guides/troubleshooting.md) about Debugging Benchmarks. You can read this doc and use sth like the code below to get a better understanding of what is generated and executed by BenchmarkDotNet:

```cs
[GlobalSetup]
public async Task Setup()
{
    System.Diagnostics.Debugger.Launch();
}
```

adamsitnik:
@dlemstra and once again big thanks for doing this!

src/BenchmarkDotNet/Code/DeclarationsProvider.cs
```diff
@@ -62,6 +62,21 @@ public IEnumerable<string> ArgumentsNamespaces
...
                 .Select(arg => $"using {arg.ParameterType.Namespace};");
+
+        private string GetMethodName(MethodInfo method)
+        {
+            if (method == null)
+            {
+                return EmptyAction;
+            }
+
+            if (method.ReturnType != typeof(Task))
```
adamsitnik: an asynchronous Setup method should be returning a `Task`. However knowing our users some of them will also return a `Task<T>` or even a `ValueTask<T>`.

Could you please handle these two cases as well?

```cs
if (method.ReturnType.GetTypeInfo().IsGenericType
    && (method.ReturnType.GetTypeInfo().GetGenericTypeDefinition() == typeof(Task<>)
        || method.ReturnType.GetTypeInfo().GetGenericTypeDefinition() == typeof(ValueTask<>)))
```

tests/BenchmarkDotNet.IntegrationTests/AllSetupAndCleanupTest.cs
```diff
@@ -62,6 +63,42 @@ public void AllSetupAndCleanupMethodRunsTest()
...
 
+        [Fact]
+        public void AllSetupAndCleanupMethodRunsAsyncTest()
+        {
+            var logger = new OutputLogger(Output);
+            var miniJob = Job.Default.With(RunStrategy.Monitoring).WithWarmupCount(2).WithIterationCount(3).WithInvocationCount(1).WithUnrollFactor(1).WithId("MiniJob");
+            var config = CreateSimpleConfig(logger, miniJob);
+
+            CanExecute<AllSetupAndCleanupAttributeBenchmarksAsync>(config);
+            Output.WriteLine(OutputDelimeter);
```
adamsitnik: I understand that re-writing this to `Output` is for debugging purposes?

Should we keep it in the code?

dlemstra: I just copy pasted the previous test and only changed the class that was used. No idea what that was added but we can remove it without any issues.

src/BenchmarkDotNet/Code/DeclarationsProvider.cs
dlemstra: Also added support for Task<T>, ValueTask and ValueTask<T>.

adamsitnik:
@dlemstra LGTM!

Due to an incident with their services, (https://blog.github.com/2018-10-21-october21-incident-report/) Github temporarily paused all webhooks events, which will result in PR triggers not firing.

As soon as webhooks are back and the build passes I am going to merge it.

adamsitnik:
@dlemstra thanks! great job!

PR #894, opened 2018-10-03
Fix GlobalCleanupAttributeTest.GlobalCleanupMethodRunsTest

dlemstra:
This PR fixes corrects the tests because this test will also pass when
```C#
[Benchmark]
public void Benchmark()
{
    Console.WriteLine(BenchmarkCalled);
}
```

is changed to

```C#
[Benchmark]
public void Benchmark()
{
    //Console.WriteLine(BenchmarkCalled);
}
```

This is happening because there is a check if log contains `GlobalCleanupCalled` instead `BenchmarkCalled`. When the log only contains `GlobalCleanupCalled` this will happen:

```C#
string log = logger.GetLog();
Assert.Contains(GlobalCleanupCalled + System.Environment.NewLine, log);
// True
Assert.True(
    log.IndexOf(GlobalCleanupCalled + System.Environment.NewLine) >
    log.IndexOf(BenchmarkCalled + System.Environment.NewLine));
// 0 > -1 == true
```

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/GlobalCleanupAttributeTest.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/GlobalCleanupAttributeTest.cs
@@ -24,8 +24,8 @@ namespace BenchmarkDotNet.IntegrationTests
             string log = logger.GetLog();
             Assert.Contains(BenchmarkCalled + System.Environment.NewLine, log);
             Assert.True(
-                log.IndexOf(GlobalCleanupCalled + System.Environment.NewLine) >
-                log.IndexOf(BenchmarkCalled + System.Environment.NewLine));
+                log.IndexOf(BenchmarkCalled + System.Environment.NewLine) <
+                log.IndexOf(GlobalCleanupCalled + System.Environment.NewLine));
         }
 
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
@dlemstra, thank you!

dlemstra:
@AndreyAkinshin What is going on with the tests on travis? Would you be interested in giving Azure DevOps Pipelines a try? I don't mind setting that up in another PR if you would like to have this added.

AndreyAkinshin:
> What is going on with the tests on travis?

It seems that the OS image preparation takes too much time on Linux. Our Linux build is flaky because Travis often kills it because of the 50 min timeout. We don't have this problem on macOS (build takes about 30 min).
@Ky7m, is it possible to do something with it?

> Would you be interested in giving Azure DevOps Pipelines a try? I don't mind setting that up in another PR if you would like to have this added.

It would be great!

dlemstra:
@AndreyAkinshin Will give it a try this weekend.

dlemstra:
@ky7m I could implement that if you dont mind me doing that

Ky7m:
@dlemstra it would be awesome, just let me know if you need any help from my side.

Issue #722, opened 2018-04-18
Add scaled column for Allocated Memory


adamsitnik:
Suggested by @stephentoub

> There's also no equivalent for the memory columns that I could find, which makes things inconsistent in the results, since some of these benchmarks are more about memory than they are about throughput.

Side note: The column should be visible when one of the benchmarks is marked as baseline

gsomix:
I'm on it.

adamsitnik:
@gsomix could you wait until I merge #878 ? I should merge it in 24h. It contains some things that would break your PR for sure (it's touching the columns for diagnosers)

gsomix:
@adamsitnik OK, I'm just reading code for now. :)

gsomix:
@adamsitnik I'm looking for some guidance for this issue. There are classes for baselines (e.g. `BaselineCustomColumn`) that tightly coupled with `Statistics`. We don't have any statistics for memory usage, only raw numbers of allocated bytes or gc operations. Should I somehow reuse and refactor exist baseline columns, or just to create new one (like `MetricScaledColumn`)?

adamsitnik:
@gsomix I think that you could detect if any of the Benchmarks have a basline, if so MemoryDiagnoser could return one more metric - scaled allocated bytes

Kavignon:
I’ll take this if that’s ok!

PR #902, opened 2018-10-08
More command line args

adamsitnik:
Recently I am using BDN on a large scale and in some scenarios I can't recompile the code, I need to be able to specify most of the settings from the command line. This PR adds most of the things to supported command line arguments.

## Runtimes

The `--runtimes` or just `-r` allows you to run the benchmarks for selected Runtimes. Available options are: Mono, CoreRT, net46, net461, net462, net47, net471, net472, netcoreapp2.0, netcoreapp2.1, netcoreapp2.2, netcoreapp3.0.

Example: run the benchmarks for .NET 4.7.2 and .NET Core 2.1:

```log
dotnet run -c Release -- --runtimes net472 netcoreapp2.1
```

## Number of invocations and iterations

* `--launchCount` - how many times we should launch process with target benchmark. The default is 1.
* `--warmupCount` - how many warmup iterations should be performed. If you set it, the minWarmupCount and maxWarmupCount are ignored. By default calculated by the heuristic.
* `--minWarmupCount` - minimum count of warmup iterations that should be performed. The default is 6.
* `--maxWarmupCount` - maximum count of warmup iterations that should be performed. The default is 50.
* `--iterationTime` - desired time of execution of an iteration. Used by Pilot stage to estimate the number of invocations per iteration. 500ms by default.
* `--iterationCount` - how many target iterations should be performed. By default calculated by the heuristic.
* `--minIterationCount` - minimum number of iterations to run. The default is 15.
* `--maxIterationCount` - maximum number of iterations to run. The default is 100.
* `--invocationCount` - invocation count in a single iteration. By default calculated by the heuristic.
* `--unrollFactor` - how many times the benchmark method will be invoked per one iteration of a generated loop. 16 by default
* `--runOncePerIteration` - run the benchmark exactly once per iteration. False by default.

Example: run single warmup iteration, from 9 to 12 actual workload iterations.

```log
dotnet run -c Release -- --warmupCount 1 --minIterationCount 9 --maxIterationCount 12
```

## Specifying custom default settings for console argument parser

If you want to have a possibility to specify custom default Job settings programmatically and optionally overwrite it with console line arguments, then you should create a global config with single job marked as `.AsDefault` and pass it to `BenchmarkSwitcher` together with the console line arguments.

Example: run single warmup iteration by default.

```cs
    => BenchmarkSwitcher
        .FromAssembly(typeof(Program).Assembly)
        .Run(args, GetGlobalConfig());

static IConfig GetGlobalConfig()
    => DefaultConfig.Instance
        .With(Job.Default
            .WithWarmupCount(1)
            .AsDefault()); // the KEY to get it working
```

Now, the default settings are: `WarmupCount=1` but you might still overwrite it from console args like in the example below: 

```log
dotnet run -c Release -- --warmupCount 2
```

/cc @AndreyAkinshin @jorive

### DIFF
--- a/docs/articles/guides/console-args.md
+++ b/docs/articles/guides/console-args.md
@@ -43,12 +43,12 @@ You can also filter the benchmarks by categories:
 
 ## Runtimes
 
-The `--runtimes` or just `-r` allows you to run the benchmarks for selected Runtimes. Available options are: Clr, Mono, Core and CoreRT.
+The `--runtimes` or just `-r` allows you to run the benchmarks for selected Runtimes. Available options are: Mono, CoreRT, net46, net461, net462, net47, net471, net472, netcoreapp2.0, netcoreapp2.1, netcoreapp2.2, netcoreapp3.0.
 
-Example: run the benchmarks for .NET and .NET Core:
+Example: run the benchmarks for .NET 4.7.2 and .NET Core 2.1:
 
 ```log
-dotnet run -c Release -- --runtimes clr core
+dotnet run -c Release -- --runtimes net472 netcoreapp2.1
 ```
 
 ## More
@@ -61,4 +61,7 @@ dotnet run -c Release -- --runtimes clr core
 * `--affinity` Affinity mask to set for the benchmark process
 * `--allStats` (Default: false) Displays all statistics (min, max & more)
 * `--attribute` Run all methods with given attribute (applied to class or method)
+* `--monoPath` custom Path for Mono
+* `--cliPath` custom Path for dotnet cli
+* `--coreRt` path to ILCompiler for CoreRT
 
--- a/docs/articles/guides/console-args.md
+++ b/docs/articles/guides/console-args.md
@@ -51,6 +51,26 @@ Example: run the benchmarks for .NET 4.7.2 and .NET Core 2.1:
 dotnet run -c Release -- --runtimes net472 netcoreapp2.1
 ```
 
+## Number of invocations and iterations
+
+* `--launchCount` - how many times we should launch process with target benchmark. The default is 1.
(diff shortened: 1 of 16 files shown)


adamsitnik:
@AndreyAkinshin  I am going to merge it now to be able to merge #878 and unblock #722 and few other things that I am working on and test it all together so we can release 0.11.2.

docs/articles/guides/console-args.md
```diff
@@ -43,12 +43,57 @@ You can also filter the benchmarks by categories:
 
 ## Runtimes
 
-The `--runtimes` or just `-r` allows you to run the benchmarks for selected Runtimes. Available options are: Clr, Mono, Core and CoreRT.
+The `--runtimes` or just `-r` allows you to run the benchmarks for selected Runtimes. Available options are: Mono, CoreRT, net46, net461, net462, net47, net471, net472, netcoreapp2.0, netcoreapp2.1, netcoreapp2.2, netcoreapp3.0.
```
AndreyAkinshin: Do we really need `net46`, `net461`, `net47`, `net471`, `net472`?
When we are talking about a runtime for a benchmark, we typically want to run it on .NET Framework. Since it's impossible to install several versions of the .NET Framework side-by-side, we want to run it on the currently installed version of .NET Framework. However, we don't have this option anymore: we have to specify a particular version of the .NET Framework each time.

adamsitnik: @AndreyAkinshin I think that we can have both, I added `core` and `clr` back in https://github.com/dotnet/BenchmarkDotNet/commit/e7d4d96113e822393383afde8a2b83208f9861c0

PR coming on Monday ;)

AndreyAkinshin: Awesome!

Issue #447, opened 2017-05-11
Implement ColoredLogger for LinqPad


AndreyAkinshin:
Original discussion: https://github.com/dotnet/BenchmarkDotNet/issues/445#issuecomment-300695708

bgrainger:
As an avid BDN and LINQPad user, I'm interested in implementing this (for Hacktoberfest). My understanding of the requirements is:

* Add the code from the linked comment
  * Use Reflection to call `Util.WithStyle` to avoid taking a dependency
  * Detect LINQPad Dark Mode and use alternate colour scheme
* Detect whether running under LINQPad (either using `AppDomain.CurrentDomain.FriendlyName` or assembly location)
* Automatically add a `LinqPadLogger` to the default configuration if running under LINQPad? (Not sure about this)

Out-of-scope (unless easy):

* Add a `ToDump()` method to dump the Summary in a formatted way

Without objection, I'll start working on this.

AndreyAkinshin:
Hey @bgrainger! The plan looks good to me.

bgrainger:
#903 implemented `LinqPadLogger`; this issue should now be closed.

PR #903, opened 2018-10-10
Add LINQPad logging

bgrainger:
Adds a logger that sends coloured output to LINQPad, fixing #447.

This new logger is enabled by default (in `DefaultConfig`) if the benchmark detects that it's being run under LINQPad. This makes it really easy to use, but I wasn't sure if this "magic" was appropriate; let me know if you'd like it to be done a different way.

### DIFF
--- a/src/BenchmarkDotNet/Configs/DefaultConfig.cs
+++ b/src/BenchmarkDotNet/Configs/DefaultConfig.cs
@@ -37,7 +37,10 @@ namespace BenchmarkDotNet.Configs
 
         public IEnumerable<ILogger> GetLoggers()
         {
-            yield return ConsoleLogger.Default;
+            if (LinqPadLogger.IsAvailable)
+                yield return LinqPadLogger.Instance;
+            else
+                yield return ConsoleLogger.Default;
         }
 
         public IEnumerable<IAnalyser> GetAnalysers()
--- a/src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
+++ b/src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
@@ -1,0 +1,95 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace BenchmarkDotNet.Loggers
+{
+    public sealed class LinqPadLogger : ILogger
+    {
+        private const string DefaultColor = "";
+
+        // The Util.WithStyle method from LINQPad
+        private readonly MethodInfo withStyle;
+
+        private readonly IReadOnlyDictionary<LogKind, string> colorScheme;
+
+        public static readonly Lazy<LinqPadLogger> lazyInstance = new Lazy<LinqPadLogger>(() =>
+        {
+            // Detect if being run from LINQPad; see https://github.com/dotnet/BenchmarkDotNet/issues/445#issuecomment-300723741
+            MethodInfo withStyle = null;
+            if (AppDomain.CurrentDomain.FriendlyName.StartsWith("LINQPad Query Server", StringComparison.OrdinalIgnoreCase))
+            {
+                try
+                {
+                    // Use reflection to avoid taking a dependency on the LINQPad assembly
+                    var util = Type.GetType("LINQPad.Util, LINQPad, Version=1.0.0.0, Culture=neutral, PublicKeyToken=21353812cd2a2db5", throwOnError: true);
+                    withStyle = util.GetMethod("WithStyle", BindingFlags.Static | BindingFlags.Public);
+                }
+                catch (Exception)
+                {
+                }
+            }
+
+            if (withStyle != null)
+            {
+                var isDarkTheme = (bool) (withStyle.DeclaringType.GetProperty("IsDarkThemeEnabled", BindingFlags.Static | BindingFlags.Public)?.GetValue(null) ?? false);
+                var colorScheme = isDarkTheme ? CreateDarkScheme() : CreateLightScheme();
+                return new LinqPadLogger(withStyle, colorScheme);
+            }
+
+            return null;
+        });
+
+        public static bool IsAvailable => Instance != null;
+
+        public static ILogger Instance => lazyInstance.Value;
+
+        private LinqPadLogger(MethodInfo withStyle, IReadOnlyDictionary<LogKind, string> colorScheme)
+        {
+            this.withStyle = withStyle;
+            this.colorScheme = colorScheme;
+        }
+
+        public void Write(LogKind logKind, string text) => Write(logKind, Console.Write, text);
+
+        public void WriteLine() => Console.WriteLine();
+
+        public void WriteLine(LogKind logKind, string text) => Write(logKind, Console.WriteLine, text);
+
+        private void Write(LogKind logKind, Action<object> write, string text) =>
+            write(WithStyle(text, "color:" + GetColor(logKind)));
+
+        private object WithStyle(object data, string htmlStyle) =>
+            withStyle.Invoke(null, new[] { data, htmlStyle });
+
+        private string GetColor(LogKind logKind) =>
+            colorScheme.TryGetValue(logKind, out var color) ? color : DefaultColor;
+
+        // Converted from ConsoleLogger.CreateColorfulScheme using https://stackoverflow.com/a/28211539
+        private static IReadOnlyDictionary<LogKind, string> CreateDarkScheme() =>
+            new Dictionary<LogKind, string>
+            {
+                { LogKind.Default, "#C0C0C0" },
+                { LogKind.Help, "#008000" },
+                { LogKind.Header, "#FF00FF" },
+                { LogKind.Result, "#008080" },
+                { LogKind.Statistic, "#00FFFF" },
+                { LogKind.Info, "#808000" },
+                { LogKind.Error, "#FF0000" },
+                { LogKind.Hint, "#008080" }
+            };
+
+        private static IReadOnlyDictionary<LogKind, string> CreateLightScheme() =>
+            new Dictionary<LogKind, string>
+            {
+                { LogKind.Default, "#404040" },
+                { LogKind.Help, "#008000" },
+                { LogKind.Header, "#800080" },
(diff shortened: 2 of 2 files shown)


src/BenchmarkDotNet/Configs/DefaultConfig.cs
```diff
@@ -37,7 +37,10 @@ public IEnumerable<IExporter> GetExporters()
 
         public IEnumerable<ILogger> GetLoggers()
         {
-            yield return ConsoleLogger.Default;
+            if (AppDomain.CurrentDomain.FriendlyName.StartsWith("LINQPad Query Server", StringComparison.OrdinalIgnoreCase))
```
AndreyAkinshin: It would be better to have all LINQ-specific logic in a single class. I suggest hiding this check inside LinqPadLogger. We can introduce a public static field `LinqPadLogger.IsAvaliable`.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Reflection;
+
+namespace BenchmarkDotNet.Loggers
+{
+    public sealed class LinqPadLogger : ILogger
+    {
+        private const string DefaultColor = "";
+
+        // Use Reflection to avoid taking a dependency on the LINQPad assembly
+        static readonly Type linqPadUtil = Type.GetType("LINQPad.Util, LINQPad, Version=1.0.0.0, Culture=neutral, PublicKeyToken=21353812cd2a2db5", throwOnError: true);
```
AndreyAkinshin: Why do you use `throwOnError: true` here?

bgrainger: `Type.GetType(string)` will simply return `null` if the type isn't found, which will immediately cause a `NullReferenceException` in the static constructor. This will throw a more informative `TypeLoadException`.

bgrainger: (And I use a named parameter because I personally dislike "naked" Boolean parameter values; this is more self-documenting.)

src/BenchmarkDotNet/Configs/DefaultConfig.cs
bgrainger: Great suggestion; fixed.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
AndreyAkinshin: Thus, it will throw an exception once the static field is evaluated. It can be a problem in non-LINQPAD-environment if somebody calls some LinqPadLogger API.
It would be better to have a static instance which is initialized in the static constructor and correctly handle all the exception. This static instance should be initialized only in the LINQPAD environment (IsAvailable=true).

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public static bool IsAvailable => AppDomain.CurrentDomain.FriendlyName.StartsWith("LINQPad Query Server", StringComparison.OrdinalIgnoreCase);
+
+        public static ILogger Default { get; } = new LinqPadLogger();
+
+        private readonly IReadOnlyDictionary<LogKind, string> colorScheme;
+
+        public LinqPadLogger()
+        {
+            var isDarkTheme = (bool) linqPadUtil.GetProperty("IsDarkThemeEnabled", BindingFlags.Static | BindingFlags.Public).GetValue(null);
```
AndreyAkinshin: Did you check how it works in old LINQPad versions which don't have this property? This case should be handled correctly without any additional exceptions.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    public sealed class LinqPadLogger : ILogger
+    {
+        private const string DefaultColor = "";
+
+        // Use Reflection to avoid taking a dependency on the LINQPad assembly
+        static readonly Type linqPadUtil = Type.GetType("LINQPad.Util, LINQPad, Version=1.0.0.0, Culture=neutral, PublicKeyToken=21353812cd2a2db5", throwOnError: true);
+        private static readonly MethodInfo withStyle = linqPadUtil.GetMethod("WithStyle", BindingFlags.Static | BindingFlags.Public);
+
+        public static bool IsAvailable => AppDomain.CurrentDomain.FriendlyName.StartsWith("LINQPad Query Server", StringComparison.OrdinalIgnoreCase);
```
AndreyAkinshin: Could you please make it lazy?
In BenchmarkDotNet, we are trying to evaluate all the heavy calls only once (on the first usage). It helps to avoid undesired allocations and pauses at random moments.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace BenchmarkDotNet.Loggers
+{
+    public sealed class LinqPadLogger : ILogger
+    {
+        private const string DefaultColor = "";
```
AndreyAkinshin: It would be better to use an existing color which looks readable in both themes.
It can protect us from a situation when we add new LogKind and forget to update LINQPad color schemes.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
bgrainger: Yes, I just realised this.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
bgrainger: This works as expected by default; LINQPad sets the default colour to black (in light mode) and white (in dark mode), so everything is readable.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
bgrainger: Will fix this to handle `GetProperty` returning `null` (i.e., property not found).

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
bgrainger: Moved Reflection out of the static constructor and into a `Lazy<bool>` that makes `IsAvailable = true` if all prerequisites are met. 

Reflection will only occur during `DefaultConfig.GetLoggers` if the `AppDomain.CurrentDomain.FriendlyName` test passes, which seems OK to me; let me know if that's not acceptable.

src/BenchmarkDotNet/Loggers/LinqPadLogger.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        private static readonly Lazy<bool> isAvailable = new Lazy<bool>(() =>
+        {
+            if (AppDomain.CurrentDomain.FriendlyName.StartsWith("LINQPad Query Server", StringComparison.OrdinalIgnoreCase))
+            {
+                try
+                {
+                    // Use Reflection to avoid taking a dependency on the LINQPad assembly
+                    linqPadUtil = Type.GetType("LINQPad.Util, LINQPad, Version=1.0.0.0, Culture=neutral, PublicKeyToken=21353812cd2a2db5", throwOnError: true);
+                    withStyle = linqPadUtil.GetMethod("WithStyle", BindingFlags.Static | BindingFlags.Public);
```
bgrainger: I'm not happy with a Lazy value provider having the side-effect of setting static variables; will rework.

bgrainger: Implemented this change (using static field initialisation for laziness).

AndreyAkinshin:
Now it looks much better. However, I still have some additional suggestions. Currently, it's possible to use `LinqPadLogger` without checking `LinqPadLogger.Instance.IsAvailable`. It may produce some strange bugs. Also, we still need lazy initialization for this instance. I guess, we can rewrite it in the following form:

```
public static bool IsAvailable => Instance.Value != null;
public static LinqPadLogger Instance => Instance.Value
private static readonly Lazy<LinqPadLogger> lazyInstance = ...
```
In this case, it will be impossible to use invalid LinqPadLogger instance (we will get `NullReferenceException` in this case). Also, the instance will be initialized only after the first `IsAvaliable` or `Instance` call.
What do you think?

bgrainger:
I wasn't sure whether using `Lazy<T>` was a significant improvement over a static field, since `LinqPadLogger.IsAvailable` was being accessed from `DefaultConfig`, which would force the initialisation either way. However, I do agree that it's a problem that a semi-functional instance (that just uses `Console.Write`) can be used even if `IsAvailable` is false. I'll make the changes you suggest.

Thanks for your patient assistance in helping me rework this!

AndreyAkinshin:
Awesome, thanks! It works great in both themes.

> I wasn't sure whether using Lazy<T> was a significant improvement over a static field,

Now `LinqPadLogger` is a part of public API and we don't know how it will be used. By default, we are trying to make all "heavy" calls lazy. It may look unnecessary in some specific cases, but we already had an unexpected infrastructure impact on benchmarks in the past. Now we are trying to apply the lazy pattern everywhere: it helps to minimize side-effects from the infrastructure.

---

P.S.
By default, LINQPad uses a non-monospaced font for the results. The summary table looks pretty ugly in this mode. In my local setup, I use CSS from [here](https://forum.linqpad.net/discussion/1150/how-do-i-get-monospaced-results), but it would be nice to enable a monospaced font for `LinqPadLogger` by default. What do you think?

bgrainger:
Yes, using a monospaced font (for all output) seems like it would make it much more readable. (Not everything _needs_ to be monospaced but it would probably look very odd to have a mix of monospaced and proportional fonts.) 

This seems like a simple change; I can open a PR.

Issue #731, opened 2018-04-26
Add constant folding analyser


adamsitnik:
If the result == 0 then most probably the benchmark was returning a const value. We should warn the users about it.

Example:

```cs
[Config(typeof(JitsConfig))]
public class MathBenchmarks
{
    private class JitsConfig : ManualConfig
    {
        public JitsConfig()
        {
            Add(Job.Default.With(Jit.LegacyJit).With(Platform.X86).WithId("Legacy x86"));
            Add(Job.Default.With(Jit.LegacyJit).With(Platform.X64).WithId("Legacy x64"));
            Add(Job.Default.With(Jit.RyuJit).With(Platform.X64).WithId("Ryu x64"));
        }
    }

    [Benchmark]
    public double Sqrt14() 
        => Math.Sqrt(1) + Math.Sqrt(2) + Math.Sqrt(3) + Math.Sqrt(4) +
           Math.Sqrt(5) + Math.Sqrt(6) + Math.Sqrt(7) + Math.Sqrt(8) +
           Math.Sqrt(9) + Math.Sqrt(10) + Math.Sqrt(11) + Math.Sqrt(12) +
           Math.Sqrt(13) + Math.Sqrt(14);
}
```

![image](https://user-images.githubusercontent.com/6011991/39320000-a69de034-4937-11e8-891a-ba2381482dc6.png)

Rizzen:
I'll take a look.

adamsitnik:
Fixed in #906

Issue #601, opened 2017-12-02
Surprising results


JohanLarsson:
```cs
public class UnitVectorBenchmarks
{
    [Benchmark(Baseline = true)]
    public double CreateUnitVectorReturnX()
    {
        return new UnitVector(1, 2).X;
    }

    [Benchmark]
    public double Return1()
    {
        return 1;
    }

    public struct UnitVector
    {
        public readonly double X;
        public readonly double Y;

        public UnitVector(double x, double y)
        {
            X = x;
            Y = y;
        }
    }
}
```

Result in:
```
 |                                 Method |       Mean |     Error |    StdDev | Scaled | ScaledSD | Allocated |
 |--------------------------------------- |-----------:|----------:|----------:|-------:|---------:|----------:|
 |                CreateUnitVectorReturnX |  0.0088 ns | 0.0128 ns | 0.0236 ns |      ? |        ? |       0 B |
 |                                Return1 |  0.1331 ns | 0.0262 ns | 0.0644 ns |      ? |        ? |       0 B |
```

Not sure if I'm dumbing something but it does not look right.

adamsitnik:
Hi @JohanLarsson 

Thanks for reporting the problem.

If you use our `[DisassemblyDiagnoser]` you can see that `new UnitVector(1, 2).X;` gets recognized by JIT as a constant.

```
00007ffe`1c444a90 ConsoleApp15.UnitVectorBenchmarks.CreateUnitVectorReturnX()
00007ffe`1c444a90 c4e17b100507000000 vmovsd xmm0,qword ptr [00007ffe`1c444aa0]

Method got inlined
ConsoleApp15.UnitVectorBenchmarks+UnitVector..ctor(Double, Double)
```

```
00007ffe`1c464a90 ConsoleApp15.UnitVectorBenchmarks.Return1()
00007ffe`1c464a90 c4e17b100507000000 vmovsd xmm0,qword ptr [00007ffe`1c464aa0]
```

If you apply `[MethodImpl(MethodImplOptions.NoOptimization)]` attribute to method `CreateUnitVectorReturnX()` the assembly is:

```
00007ffe`1c474a90 ConsoleApp15.UnitVectorBenchmarks.CreateUnitVectorReturnX()
00007ffe`1c474a9a 33c9            xor     ecx,ecx
00007ffe`1c474a9c 488d45f0        lea     rax,[rbp-10h]
[...]

PR #906, opened 2018-10-11
Zero measurement analyser

Rizzen:
Second attempt to solve https://github.com/dotnet/BenchmarkDotNet/issues/731 and https://github.com/dotnet/BenchmarkDotNet/issues/601
We detecting CPU max frequency, then calculates resolution (time for 1 cycle). Then we check distribution of benchmark result against half of 1 CPU cycle (to avoid false positive results due to some noise when benchmark takes around 1 CPU cycle) with statistical test.
Then, if result is Zero Measurement, we warns user about it.
@AndreyAkinshin what do you think about this approach?


src/BenchmarkDotNet/Analysers/ZeroMeasurementAnalyser.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        private static readonly TimeInterval FallbackCPUResolutionValue = TimeInterval.FromNanoseconds(0.2d);
+        
+        private ZeroMeasurementAnalyser() { }
+
+        protected override IEnumerable<Conclusion> AnalyseReport(BenchmarkReport report, Summary summary)
+        {
+            var currentFrequency = summary.HostEnvironmentInfo.CpuInfo.Value.MaxFrequency;
+            if (!currentFrequency.HasValue || currentFrequency <= 0)
+                currentFrequency = Frequency.FromGHz(1 / FallbackCPUResolutionValue.Nanoseconds);
```
AndreyAkinshin: It's better to use `TimeInterval.ToFrequency()`

src/BenchmarkDotNet/Analysers/ZeroMeasurementHelper.cs
```diff
@@ -0,0 +1,17 @@
...
+    public class ZeroMeasurementHelper
+    {
+        /// <summary>
+        /// Checks distribution against Zero Measurement hypothesis
+        /// </summary>
+        /// <returns>True if measurement is ZeroMeasurement</returns>
+        public static bool CheckZeroMeasurement(double[] results, double threshold)
+        {
+            var sample = new [] { threshold, threshold };
+            return !WelchTest.Instance.IsGreater(results, sample).NullHypothesisIsRejected;
```
AndreyAkinshin: If we have `Overhead` iterations, it will be much better to compare `Workload` and `Overhead` samples.
If we don't have `Overhead` iterations, we should perform an honest statistical test against fixed value instead of emulating it via the `{ threshold, threshold }` sample.

Rizzen: Welch test doesn't allow check against single value.

AndreyAkinshin: We should implement a one sample t-test. (If you want, I can add the corresponding API for you).

Rizzen: Thank you, but I guess I can handle it.

Rizzen:
Added test against overhead, when it possible. Added one sample t-test, to analyse distribution when no overhead.

AndreyAkinshin:
@Rizzen, now it looks much better.
We also need unit tests for the new statistical test in `BenchmarkDotNet.Tests.Mathematics.StatisticalTesting`. You can check the output in R (`t.test(x, mu = mu, alternative = "greater")` should produce the same result)

src/BenchmarkDotNet/Mathematics/StatisticalTesting/OneSampleTTest.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            double mean = sample.Mean;
+            double stdDev = sample.StandardDeviation;
+            double n = sample.N;
+            double df = n - 1;
+            
+            threshold = threshold ?? RelativeThreshold.Default;
+
+            double t = (mean - value) / 
+                       (stdDev * Math.Sqrt(n));
```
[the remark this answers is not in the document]
Rizzen: Yeah, of course, thank you!

Rizzen:
Fixed remarks and added statistical tests.
@AndreyAkinshin Thank you for your help!

AndreyAkinshin:
Currently, we have the following "feature": conclusions with the same message are merged into a single conclusion. Originally, it was implemented for analyzers like `EnvironmentAnalyser` (it doesn't make sense to print a message like `use RELEASE instead of DEBUG` for each benchmark). Unfortunately, because of this "feature", `ZeroMeasurementAnalyser` prints a warning only for the first "empty" methods (even if we have several "empty" methods).
The conclusion system should be improved. For example, we can introduce `bool Mergeable` property for `Conclusion`. If `Conclusion.Mergable == false`, `Conclusion.Equals` should  use the `Report` value.

Rizzen:
Ok, I'll do that.

Rizzen:
@AndreyAkinshin Added Mergeable parameter to Conclusion and Report usage in Conclusion.Equals.

src/BenchmarkDotNet/Analysers/Conclusion.cs
```diff
@@ -11,36 +11,45 @@ public sealed class Conclusion : IEquatable<Conclusion>
...
-            => new Conclusion(analyserId, ConclusionKind.Warning, message, report);
+        public static Conclusion CreateWarning(string analyserId, string message, [CanBeNull] BenchmarkReport report = null, bool mergeable = true) 
+            => new Conclusion(analyserId, ConclusionKind.Warning, message, report, mergeable);
 
-        public static Conclusion CreateError(string analyserId, string message, [CanBeNull] BenchmarkReport report = null) 
-            => new Conclusion(analyserId, ConclusionKind.Error, message, report);
+        public static Conclusion CreateError(string analyserId, string message, [CanBeNull] BenchmarkReport report = null, bool mergeable = true) 
+            => new Conclusion(analyserId, ConclusionKind.Error, message, report, mergeable);
 
         public bool Equals(Conclusion other)
```
AndreyAkinshin: `GetHashCode` should be also updated because the `Distinct()` call use a hash set internally. In the current implementation, BenchmarkDotNet still (randomly) merges some ZeroMeasurement-conclusions.

src/BenchmarkDotNet/Analysers/Conclusion.cs
```diff
@@ -11,36 +11,45 @@ public sealed class Conclusion : IEquatable<Conclusion>
...
         {
             if (ReferenceEquals(null, other))
                 return false;
             if (ReferenceEquals(this, other))
                 return true;
+            if (!Mergeable)
```
AndreyAkinshin: `other.Mergeable` should also be checked to make `Equals` symmetric.

tests/BenchmarkDotNet.Tests/Analysers/ZeroMeasurementHelperTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        [InlineData(0.27049, 0.27410, 0.27433, 0.27468, 0.27476, 0.27591, 0.27644, 0.27704, 0.27724, 0.27766, 0.27792, 0.27878, 0.28025)]
+        public void CheckZeroMeasurement_OneSample_Around_One_CPU_Cycle_Method(params double[] distribution)
+        {
+            var zeroMeasurement = ZeroMeasurementHelper.CheckZeroMeasurementOneSample(distribution, ThresholdMock);
+            Assert.False(zeroMeasurement);
+        }
+        
+        [Theory]
+        [InlineData(0.2703, 0.2701, 0.2703, 0.2701, 0.2703, 0.2701, 0.2703, 0.2701, 0.2703, 0.2701, 0.2703, 0.2701)]
+        public void CheckZeroMeasurement_OneSample_Exactly_One_CPU_Cycle_Method(params double[] distribution)
```
AndreyAkinshin: You can remove `CheckZeroMeasurement` from the title because we have this information in the class name. The current test name is too huge.

src/BenchmarkDotNet/Analysers/Conclusion.cs
Rizzen: Thank you! Totally forgot about hashcode.

adamsitnik:
@Rizzen great job!!!

PR #908, opened 2018-10-13
Added [ParamsAllValues]

### LINKED ISSUE #658: [Params] for enums should include all values by default

gsomix:
Closes #658 

- [x] Rebase #664 
- [x] Add approval tests for the feature
- [x] Implement validators
- [x] Add lost documentation


gsomix:
@adamsitnik @AndreyAkinshin Could you take a look? Thanks!

AndreyAkinshin:
@adamsitnik could you also review this PR?

gsomix:
@adamsitnik Done! I have also refactored `if-else` to pattern matching. Hope now it's way more cleaner.

adamsitnik:
@gsomix great job! thank you!

Issue #463, opened 2017-06-08
Review interface IConfig


Backs:
Look at defenition of IConfig:

    public interface IConfig
    {
        IEnumerable<IColumnProvider> GetColumnProviders();
        IEnumerable<IExporter> GetExporters();
        IEnumerable<ILogger> GetLoggers();
        IEnumerable<IDiagnoser> GetDiagnosers();
        IEnumerable<IAnalyser> GetAnalysers();
        IEnumerable<Job> GetJobs();
        IEnumerable<IValidator> GetValidators();
        IEnumerable<HardwareCounter> GetHardwareCounters();
        IEnumerable<IFilter> GetFilters();        

        IOrderProvider GetOrderProvider();
        ISummaryStyle GetSummaryStyle();

        ConfigUnionRule UnionRule { get; }

        /// <summary>
        /// determines if all auto-generated files should be kept or removed after running benchmarks
        /// </summary>
        bool KeepBenchmarkFiles { get; }
    }

As we see, many methods return IEnumerable<T>. In some cases, I see, it's better to return IReadOnlyCollection to prevent excessive ToList or ToArray calls.
Examples:

https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Running/BenchmarkConverter.cs#L36

`var rawJobs = config?.GetJobs().ToArray() ?? Array.Empty<Job>();`
Collection is better.

https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Running/BenchmarkConverter.cs#L49

`var filters = config.GetFilters().ToList();`
Can be replaced with Collection.

https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Configs/ConfigExtensions.cs#L15
Alse, I think `CompositeLogger`, `CompositeExporter` and others should have constructors that accept IReadOnlyCollection.

https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Diagnostics.Windows/PmcDiagnoser.cs#L166
`var stats = new PmcStats(parameters.Config.GetHardwareCounters().ToArray());`
`PmcStats` constructor accepts IReadOnlyCollection.

Internal implementation of ManualConfig is based on List<T>, so it can be simple replaced with:
`public IReadOnlyCollection<ILogger> GetLoggers() => loggers.AsReadOnly();`

[...]

PR #912, opened 2018-10-15
Duplicates handling for IConfig = big refactor, fixes #360

### LINKED ISSUE #360: [Suggestion] Duplicates handling for IConfig

adamsitnik:
This is a big refactor to solve #360 requested by @ig-sinicyn a long time ago.

My motivation was a few minor bugs and logic for building the final config spread amongst the entire library.

The idea:

1. Make `ManualConfig` dummy. Just a bag of settings with no logic
2. Make all `CompositeXYZ` dummy. They just get something and use it
3. Put all of the logic for building a FinalConfig to one place.
4. Add tests
5. Make the `FinalConfig` immutable
6. If you want to create `CompositeXYZ` only `FinalConfig` can do that for you


src/BenchmarkDotNet/Configs/FinalConfig.cs
```diff
@@ -0,0 +1,103 @@
...
+using BenchmarkDotNet.Order;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Validators;
+using JetBrains.Annotations;
+using RunMode = BenchmarkDotNet.Diagnosers.RunMode;
+
+namespace BenchmarkDotNet.Configs
+{
+    public class FinalConfig : IConfig
```
AndreyAkinshin: I don't like the class name. Maybe it should be `ImmutableConfig`?

AndreyAkinshin:
@adamsitnik, in general, the PR looks great. My only concern is about the name of `FinalConfig`. Maybe it should be `ImmutableConfig` or `SealedConfig`?

adamsitnik:
@AndreyAkinshin heh, I also was not sure about the name ;)

Internally we have few configs (global/from args/from attributes). Before we run the benchmarks, we merge all of these configs into one. And this is the config that we run the benchmarks against.

This is why I have called it `Final`. Initially, I have called it `FinalImmutableConfig` but it was too long.

I am not myself a big fan of the `FinalConfig` name, but what I want to express is that this is the config that we run the benchmarks against, no more merging or adding new things. Do you think that `ImmutableConfig` is a better name for it? I am totally open to renaming it or other suggestions.

@ig-sinicyn maybe you have some ideas?

AndreyAkinshin:
I think, `ImmutableConfig` is better because it describes the main "feature" of this config. The class name shouldn't express how we are going to use it in a particular use case, we have field/property names for that.

adamsitnik:
@AndreyAkinshin I have solved the merge conflicts and finished the refactor. I think it's ready for a review

This PR fixes #938, #360 and #463

PR #913, opened 2018-10-16
.NET Core Toolchains improvements

adamsitnik:
This PR  brings the following changes:

0. I have changed the way we build auto-generated program. There were few edge cases reported by the users (/cc @eerhardt) where `dotnet restore --no-dependencies` and `dotnet build --no-restore --no-dependencies` were causing some serious trouble like failing to build or running the benchmark against old code (the dependency was not fully rebuilt because user was doing `dotnet run -f X -- -runtime Y` where `X != Y`). Now we perform a full restore and build and if something does not work, because the file is in use or sth like that we fall back to the optimized path (`--no-dependencies`). From the perf perspective it should not be a problem because we build a single project for all benchmarks (we used to buid one per every benchmark).
1. I added back `core` and `clr` as default runtime monikers for the `--runtime` option (./cc @AndreyAkinshin )
2. I made it possible to specify custom `dotnet restore --packages` location by adding `--packages` argument. It's required for CoreCLR/FX and some other repos that don't restore to the global cli folder, but to repo/packages (/cc @jorive)
3. I improved our troubleshooting scripts, now they contain exactly the same thing our toolchains do, including a call to custom dotnet cli with full path
4. I did a refactor of our code that uses `dotnet cli`. I think it's much cleaner now!


docs/articles/guides/console-args.md
```diff
@@ -43,14 +43,33 @@ You can also filter the benchmarks by categories:
...
+Example: run the benchmarks for .NET Core 3.0 and latest .NET SDK installed on your PC:
+
+```log
+dotnet run -c Release -f netcoreapp3.0 -- --runtimes clr core
+```
+
+But same command executed with `-f netcoreapp2.0` is going to run the benchmarks for .NET Core 2.0:
+
+```log
+dotnet run -c Release -f netcoreapp3.0 -- --runtimes clr core
```
AndreyAkinshin: I guess, it should be `dotnet run -c Release -f netcoreapp2.0 -- --runtimes clr core`

adamsitnik: it should! good catch!

adamsitnik:
@AndreyAkinshin  Thank you for the code review!

PR #916, opened 2018-10-18
Update console-args.md - add information about `--list` option

wojtpl2:
Add information about `--list` option in console-args.md file.
I forgot about it ;)


adamsitnik:
could you also provide an example with `--filter`? It's a cool feature, it would be great if our users know about it.

wojtpl2:
I've created whole section for this option, but I'm not sure if this is the best solution;)

adamsitnik:
thanks!

Issue #875, opened 2018-09-10
[DevOps] Run builds on Linux, macOS, and Windows using Azure Pipelines


Ky7m:
https://azure.microsoft.com/en-us/blog/announcing-azure-pipelines-with-unlimited-ci-cd-minutes-for-open-source/

Ky7m:
feel free to assign it to me, I am returning to the normal work mode:)

Ky7m:
@AndreyAkinshin, can we create Azure DevOps project and setup extension https://github.com/marketplace/azure-pipelines ?
Just let me know if you need any guidance for these steps

AndreyAkinshin:
@Ky7m it seems that I do not have permission to modify this app on the dotnet organization.
@jongalloway could you help us?

Ky7m:
There are some failed tests with the following exception:

> Wrong CoreFx version: was 4.6.26629.01, should be 4.6.26531.02

Any ideas, @adamsitnik?

adamsitnik:
@Ky7m the `CustomCoreClrToolchain` is heavily dependent on the `dotnet cli` behavior which is following: you can restore custom `Runtime/NetCoreApp` package if it's newer than the one in the `dotnet cli` sdk folder.

Did you update `dotnet cli`? The solution would be to bump the version numbers in these tests.

Ky7m:
Thank you @adamsitnik , will take a look.
By the way, there is a view of tests https://dev.azure.com/ky7m/BenchmarkDotNet/_build/results?buildId=15&view=ms.vss-test-web.test-result-details (looks better in comparison with Travis CI)

adamsitnik:
Currently this test:

https://github.com/dotnet/BenchmarkDotNet/blob/1e62355f209a25c7a33f9ab7e7e03b0afe7d851f/tests/BenchmarkDotNet.IntegrationTests/CustomCoreClrToolchainTests.cs#L18-L33

Is using: https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.win-x64.Microsoft.Private.CoreFx.NETCoreApp/4.5.0-rtm-26531-02

Could you try to change `4.5.0-rtm-26531-02` to `4.5.0-servicing-26703-08` and `4.6.26531.02` to `4.6.26703.08` ?

Ky7m:
@adamsitnik, I reverted back version of dotnet cli and it solved the issue. Thanks!

AndreyAkinshin:
[...]

PR #917, opened 2018-10-19
Add Azure Pipelines support

Ky7m:
1. Add integration with Azure Pipelines (I had to disable several tests, it is temporarily)
2. Remove limit for build time on Travis
3. Upgrade xunit packages


azure-pipelines.Ubuntu.yml
```diff
@@ -0,0 +1,15 @@
+trigger:
+  batch: true
+  branches:
+    # blacklist
```
adamsitnik: nit: this comment is not required, `excluded` is kind of obvious and enough

azure-pipelines.Windows.yml
```diff
@@ -0,0 +1,16 @@
...
+
+name: $(Year:yyyy).$(Month).$(DayOfMonth)$(Rev:.r)
+
+jobs:
+- template: build/azure-pipelines.job.template.yml
+  parameters:
+    name: Windows
+    vmImage: 'vs2017-win2016'
+    scriptFileName: .\build.ps1
+    timeoutInMinutes: 300
```
adamsitnik: why the `timeoutInMinutes` is 300 for Windows when the default value in the template is 120?

Ky7m: I need to remove this, but I am still investigating where deadlock is. Usually, it takes 40 minutes to complete build with all test suites on Windows, and in some cases 3 hours isn't enough to finish the build. Only on Windows. macOS and Linux builds are stable.

Ky7m:
@adamsitnik, my pleasure. CoreRT tests are back, pre-requisites were required for Ubuntu agent only.

adamsitnik:
> CoreRT tests are back, pre-requisites were required for Ubuntu agent only

@Ky7m awesome! do we have any blockers now except of the `CustomCoreClrToolchainTests`? Please let me know if you need some help.

adamsitnik:
@Ky7m if you wonder why the builds are not triggered https://blog.github.com/2018-10-21-october21-incident-report/

adamsitnik:
> not clear behavior of test SingleBenchmarkCanBeExecutedForMultipleRuntimes

it's weird, it was rather always stable. The goal of this test is to run the same benchmark for .NET Core and CLR. Were you able to reproduce it locally?

Ky7m:
> it's weird, it was rather always stable. The goal of this test is to run the same benchmark for .NET Core and CLR. Were you able to reproduce it locally?

I guess it is related to fact that tests are run in parallel and it might lead to such behavior. I have a few ideas on how to check it.

Ky7m:
I see that SingleBenchmarkCanBeExecutedForMultipleRuntimes takes more than 30 minutes.

Ky7m:
@adamsitnik, I am done with changes and disabled xunit feature to avoid parallelization of test runs.

Ky7m:
@adamsitnik let me know if I can try to fix CustomCoreClrToolchainTests.

Ky7m:
Closes #875.

AndreyAkinshin:
@Ky7m @adamsitnik can we merge this PR? Or do we have any additional issues that should be resolved first?

adamsitnik:
@Ky7m sorry for the delay!

AndreyAkinshin:
@Ky7m I don't see any builds here: https://dev.azure.com/dotnet/BenchmarkDotNet/_build
Should we do any additional steps to make it work?

Ky7m:
@AndreyAkinshin it was configured, need to check if triggers work. Run manually https://dev.azure.com/dotnet/BenchmarkDotNet/_build

PR #920, opened 2018-10-21
OCD Whitespace and tabs cleanup

dlemstra:
While working on another PR I noticed that this file could use some whitespace cleanup.


adamsitnik:
@dlemstra thanks!

PR #922, opened 2018-10-22
Enables benchmarking betweeen different Nuget packages

Shazwazza:
referenced issue: https://github.com/dotnet/BenchmarkDotNet/issues/290

Some notes on this PR:

* Original PR was here https://github.com/dotnet/BenchmarkDotNet/pull/805 but the code became out-dated with recent changes to the benchmark.net project. There are some interesting notes about running benchmarks between Nuget versions https://github.com/dotnet/BenchmarkDotNet/pull/805#discussion_r197746744 (will summarize more below)
* This only supports running with the `CsProjCoreToolchain`, I've added a couple `TODO` notes in the `Roslyn/Generator.cs` since I think it might be possible to use the roslyn toolchain for this but I think will require a bunch of manual nuget.exe work (?)
* Have added an `IntroNuget` sample. As per the notes mentioned above, this uses Reflection in the sample so that we don't need a direct reference to the Newtonsoft.Json, however if you think it would be more clear as far as sample code goes, I can add a direct reference to the minimum version tested of Newtsonsoft.Json to demonstrate how a benchmark project should be setup to test between different Nuget versions... just let me know.

### How to benchmark between Nuget versions

Ideally this is how it is done:

* The benchmark project references the minimum version of the nuget dependency that is being tested in the list of Jobs so that reflection doesn't have to be used to call the APIs and it builds
* The list of Jobs specifies the exact version that will be tested and that is what will be resolved from Nuget
* So long as the API signatures haven't changed between all of these versions then it should 'just work'

When each benchmark is run, it will add the correct nuget package dependency and build the project with that specific version and run the benchmarks against it.

This scenario will not work if API signatures have changed between versions. In those cases, different benchmark projects would need to be setup for the same API surface area and then the resulting reports could be combined.


src/BenchmarkDotNet/Jobs/JobExtensions.cs
```diff
@@ -184,6 +184,31 @@ public static class JobExtensions
...
         public static Job With(this Job job, IReadOnlyList<Argument> arguments) => job.WithCore(j => j.Infrastructure.Arguments = arguments);
+        
+        /// <summary>
+        /// Runs the job with a specific Nuget dependency which will be resolved during the Job build process
+        /// </summary>
+        /// <param name="job"></param>
+        /// <param name="packageName">The Nuget package name</param>
+        /// <param name="packageVersion">The Nuget package version</param>
+        /// <returns></returns>
+        public static Job WithNuget(this Job job, string packageName, string packageVersion) => job.WithCore(j => j.Infrastructure.NugetReferences = new List<NugetReference> { new NugetReference(packageName, packageVersion) });
```
adamsitnik: the current value of `j.Infrastructure.NugetReferences`  is replaced with the new package, not added to the list.

```
job
   .WithNuget("A". "1.0.0");
   .WithNuget("B". "1.0.0");
```

would result only with a list of single element (B)

src/BenchmarkDotNet/Jobs/NugetReference.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class NugetReference
+    {
+        public NugetReference(string packageName, string packageVersion)
+        {
+            if (string.IsNullOrWhiteSpace(packageName))
+                throw new ArgumentException("message", nameof(packageName));
+
+            PackageName = packageName;
+            PackageVersion = packageVersion;
+            if (!string.IsNullOrWhiteSpace(PackageVersion) && !Version.TryParse(PackageVersion, out var version))
```
adamsitnik: nit: consistency: could you do both checks and then assign the properties?

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -126,5 +153,15 @@ private static string GetCustomMsBuildArguments(BenchmarkCase benchmarkCase, IRe
 
             return string.Join(" ", msBuildArguments.Select(arg => arg.TextRepresentation));
         }
+
+        private static IEnumerable<string> GetNugetAddPackageCommands(BenchmarkCase benchmarkCase, IResolver resolver)
+        {
+            if (!benchmarkCase.Job.HasValue(InfrastructureMode.NugetReferencesCharacteristic))
+                return Enumerable.Empty<string>();
+
+            var nugetRefs = benchmarkCase.Job.ResolveValue(InfrastructureMode.NugetReferencesCharacteristic, resolver).OfType<NugetReference>();
```
adamsitnik: is the `OfType<NugetReference>()` required here?

src/BenchmarkDotNet/Toolchains/Roslyn/Generator.cs
```diff
@@ -27,6 +27,7 @@ protected override string[] GetArtifactsToCleanup(ArtifactsPaths artifactsPaths)
 
         protected override void GenerateBuildScript(BuildPartition buildPartition, ArtifactsPaths artifactsPaths)
         {
+            //TODO: For nuget refs i think we need to use the MsBuildArgument/Argument to tell msbuild to fetch nuget refs?
```
adamsitnik: I think that we can just add a validation rule to the RoslynToolchain and say that this feature is not supported there

https://github.com/dotnet/BenchmarkDotNet/blob/a971a435ce6e6ca25d246e5e2cd56c5b2cf4739d/src/BenchmarkDotNet/Toolchains/Roslyn/RoslynToolchain.cs#L23-L44

src/BenchmarkDotNet/Toolchains/Roslyn/Generator.cs
```diff
@@ -47,6 +48,7 @@ protected override void GenerateBuildScript(BuildPartition buildPartition, Artif
                 prefix + string.Join(" ", list));
         }
 
+        //TODO: We need to get nuget dependencies in here somehow
```
adamsitnik: same as above

samples/BenchmarkDotNet.Samples/IntroNuget.cs
```diff
@@ -0,0 +1,52 @@
...
+using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Exporters.Csv;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Toolchains.CsProj;
+
+namespace BenchmarkDotNet.Samples
+{
+    [Config(typeof(Config))]
+    public class IntroNuget
```
adamsitnik: I think that this example is too advanced for the Intros.

Maybe you could just add a comment saying that user can reference only newer version of packages that are used?

I think that you should also add a dependency to `Newtonsoft.Json` and call the method directly, without reflection

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
Shazwazza: Nope, i think that's copy/paste fail :)

Shazwazza:
@adamsitnik I've pushed a few more commits to address your review including a few tests, let me know if there's anything i've missed

adamsitnik:
@Shazwazza thank you! great job!

Shazwazza:
Whoohoo! awesome :) 🎉

Shazwazza:
Hi @adamsitnik , I was wondering if there's already a way in BDN to specify nuget package sources? Reason I'm asking is because perhaps the https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Jobs/NugetReference.cs can be extended to have a 'Source' property which is then used when calling the `add package` cmd? Would be happy to make that change if it makes sense just let me know. Cheers!

adamsitnik:
> I was wondering if there's already a way in BDN to specify nuget package sources?

the only way is to use a `nuget.config` file (as for every msbuild project)

if this is not enough for your scenario, a PR would be very welcomed as usual

thanks!

PR #923, opened 2018-10-23
async GlobalCleanup support

dlemstra:
The `GlobalCleanup` attribute is not checked inside the `ExecutionValidatorBase` class. This patch adds support for that. 

The check inside the `Validate` method checks the count of the errors before and after executing `ExecuteBenchmarks`. And other option would be having `ExecuteBenchmarks` return a boolean but that would be a breaking change.

### DIFF
--- a/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
+++ b/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
@@ -44,7 +44,14 @@ namespace BenchmarkDotNet.Validators
                     continue;
                 }
 
+                int errorCount = errors.Count;
+
                 ExecuteBenchmarks(benchmarkTypeInstance, typeGroup, errors);
+
+                if (errorCount == errors.Count)
+                {
+                    TryToCallGlobalCleanup(benchmarkTypeInstance, errors);
+                }
             }
 
             return errors;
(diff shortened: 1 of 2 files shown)


tests/BenchmarkDotNet.Tests/Validators/ExecutionValidatorTests.cs
```diff
@@ -108,6 +151,39 @@ public override void GlobalSetup()
...
+
+            Assert.True(OverridesGlobalCleanup.WasCalled);
+            Assert.Empty(validationErrors);
+        }
+
+        public class BaseClassWithThrowingGlobalCleanup
+        {
+            [GlobalCleanup]
+            public virtual void GlobalCleanup()
+            {
```
adamsitnik: nit: good place to use `=>` and save some space

tests/BenchmarkDotNet.Tests/Validators/ExecutionValidatorTests.cs
```diff
@@ -108,6 +151,39 @@ public override void GlobalSetup()
...
+            public void NonThrowing() { }
+        }
+
+        public class OverridesGlobalCleanup : BaseClassWithThrowingGlobalCleanup
+        {
+            public static bool WasCalled;
+
+            [GlobalCleanup]
+            public override void GlobalCleanup()
+            {
```
adamsitnik: nit: good place to use `=>` and save some space

src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
```diff
@@ -44,7 +44,14 @@ public IEnumerable<ValidationError> Validate(ValidationParameters validationPara
                     continue;
                 }
 
+                int errorCount = errors.Count;
+
                 ExecuteBenchmarks(benchmarkTypeInstance, typeGroup, errors);
+
+                if (errorCount == errors.Count)
```
adamsitnik: I think that we should try to execute the cleanup no matter if there were any exceptions thrown when executing the benchmark.

Example:
Setup - open db connection
Execute - single sql query
Cleanup - close the db connection (mandatory)

dlemstra: Good point, will make the change.

adamsitnik:
@dlemstra thank you!

PR #893, opened 2018-10-03
Custom environment info

gsomix:
### Progress
- [x] Add a property for custom environment lines in `IConfig`
- [x] Add API in `ManualConfig` which allows adding new custom lines
- [x] Implement the parsing of the `CustomEnvironmentInfo` attribute in `BenchmarkConverter`
- [x] Implement processing of `void -> string` static methods
- [x] Implement processing of `void -> IEnumerable<string>` static methods
- [x] Introduce API in Summary which returns the "classic" part of the environment plus custom lines
- [x] Use the new Summary API in all exporters which print this information
- [x] Implement error handling for methods' signatures mismatch
- [x] Write tests.
- [x] Describe the new feature in the documentation and add an Intro example

Closes #384

PR #924, opened 2018-10-23
[WIP] Second attempt on custom environment info

### LINKED ISSUE #384: Print Vector<byte>.Count as part of machine info
Helps to know this when benchmarking SIMD operations

gsomix:
Closes #384 

- [x] Add `[CustomEnvironmentInfo]` attribute
- [x] Implement attribute parsing
- [x] Add invocation of custom info method to benchmark's template and code generator
- [ ] Summary
- [ ] Validator, tests, docs

### DIFF
--- a/src/BenchmarkDotNet/Code/CodeGenerator.cs
+++ b/src/BenchmarkDotNet/Code/CodeGenerator.cs
@@ -72,7 +72,9 @@ namespace BenchmarkDotNet.Code
                     .Replace("$MeasureGcStats$", buildInfo.Config.HasMemoryDiagnoser() ? "true" : "false")
                     .Replace("$Encoding$", buildInfo.Config.Encoding.ToTemplateString())
                     .Replace("$DisassemblerEntryMethodName$", DisassemblerConstants.DisassemblerEntryMethodName)
-                    .Replace("$WorkloadMethodCall$", provider.GetWorkloadMethodCall(passArguments)).ToString();
+                    .Replace("$WorkloadMethodCall$", provider.GetWorkloadMethodCall(passArguments))
+                    .Replace("$CustomEnvInfo$", GetCustomEnvInfo(provider))
+                    .ToString();
 
                 benchmarkTypeCode = Unroll(benchmarkTypeCode, benchmark.Job.ResolveValue(RunMode.UnrollFactorCharacteristic, EnvironmentResolver.Instance));
 
@@ -93,6 +95,24 @@ namespace BenchmarkDotNet.Code
                 .ToString();
 
             return benchmarkProgramContent;
+        }
+
+        private static string GetCustomEnvInfo(DeclarationsProvider provider)
+        {
+            if (provider.CustomEnvInfoMethodName is null)
+            {
+                return "Array.Empty<string>()";
+            }
+
+            switch (provider.CustomEnvInfoMethodReturnType)
+            {
+                case Type t when t == typeof(string):
+                    return $"new string[] {{ {provider.CustomEnvInfoMethodName}() }}";
+                case Type t when typeof(IEnumerable<string>).IsAssignableFrom(t):
+                    return $"{provider.CustomEnvInfoMethodName}()";
+                default:
+                    return "Array.Empty<string>()";
+            }
         }
 
         private static void AddNonEmptyUnique(HashSet<string> items, string value)
--- a/src/BenchmarkDotNet/Code/CodeGenerator.cs
+++ b/src/BenchmarkDotNet/Code/CodeGenerator.cs
@@ -101,7 +101,7 @@ namespace BenchmarkDotNet.Code
         {
             if (provider.CustomEnvInfoMethodName is null)
             {
-                return "Array.Empty<string>()";
+                return "System.Array.Empty<string>()";
             }
 
             switch (provider.CustomEnvInfoMethodReturnType)
--- a/src/BenchmarkDotNet/Code/CodeGenerator.cs
+++ b/src/BenchmarkDotNet/Code/CodeGenerator.cs
@@ -111,7 +111,7 @@ namespace BenchmarkDotNet.Code
                 case Type t when typeof(IEnumerable<string>).IsAssignableFrom(t):
                     return $"{provider.CustomEnvInfoMethodName}()";
                 default:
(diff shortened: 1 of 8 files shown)


src/BenchmarkDotNet/Code/CodeGenerator.cs
```diff
@@ -95,6 +97,24 @@ internal static string Generate(BuildPartition buildPartition)
             return benchmarkProgramContent;
         }
 
+        private static string GetCustomEnvInfo(DeclarationsProvider provider)
+        {
+            if (provider.CustomEnvInfoMethodName is null)
+            {
+                return "Array.Empty<string>()";
```
adamsitnik: Could you please use the full name? 

```suggestion
                return "System.Array.Empty<string>()";
```

src/BenchmarkDotNet/Code/CodeGenerator.cs
```diff
@@ -95,6 +97,24 @@ internal static string Generate(BuildPartition buildPartition)
...
+            }
+
+            switch (provider.CustomEnvInfoMethodReturnType)
+            {
+                case Type t when t == typeof(string):
+                    return $"new string[] {{ {provider.CustomEnvInfoMethodName}() }}";
+                case Type t when typeof(IEnumerable<string>).IsAssignableFrom(t):
+                    return $"{provider.CustomEnvInfoMethodName}()";
+                default:
+                    return "Array.Empty<string>()";
```
adamsitnik: Could you please use the full name?

```suggestion
                    return "System.Array.Empty<string>()";
```

src/BenchmarkDotNet/Code/CodeGenerator.cs
gsomix: OK, let's try this suggestons feature. :)

gsomix:
@adamsitnik  I'm trying to figure out where to put environment info in the report. There are 3 possible solutions:

- add info to summary. I don't like this solution because summary should be clear and concise.
- print info in detailed reports for every job. It's better, but harder to discover.
- introduce new conclusion for environment info

What do you think? Thanks!

adamsitnik:
@gsomix maybe a new column? @AndreyAkinshin what do you think?

gsomix:
@adamsitnik New column makes sense only for short info, e.g. `Vector<byte>.Count`. I think we can do that and just drop support of `IEnumerable<string>` version at all.

adamsitnik:
@gsomix good idea! if users want more info, they can always concatenate the strings

gsomix:
Guys, sorry for the delya. I'm going back to work.

roji:
@gsomix any plans to continue work on this?

gsomix:
@roji Yeah, sorry. I'm going to take a fresh look at PR this weekend.

gsomix:
@roji Sorry, I cannot find a time for it, I'm really busy at work last days. I'm going to close it, but feel free rebase it and push to merge.

roji:
Thanks for your message @gsomix, I'll try to pick this up in the coming days.

MarekM25:
@adamsitnik, @gsomix 
I would like to verify my understanding of this issue. Please, correct me if I am wrong. So, the expected summary is:
``` ini

BenchmarkDotNet=v0.11.5.20191006-develop, OS=Windows 10.0.17763.775 (1809/October2018Update/Redstone5)
Intel Core i5-8250U CPU 1.60GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
  [Host] : .NET Framework 4.7.2 (4.7.3460.0), X64 LegacyJIT DEBUG  [AttachedDebugger]
  Dry    : .NET Framework 4.7.2 (4.7.3460.0), X64 RyuJIT

Job=Dry  IterationCount=1  LaunchCount=1  
RunStrategy=ColdStart  UnrollFactor=1  WarmupCount=1  

```
|           Method |     Mean | Error | Custom environment info |
|----------------- |---------:|------:|------------------------:|
|            Sleep | 24.51 ms |    NA | Vector<byte>.Count = 32 |
| Thread.Sleep(10) | 16.62 ms |    NA | Vector<byte>.Count = 32 |

I plan to add it to a summary by modifying _MethodsToBenchmarksWithFullConfig_ in _BenchmarkConverter_ class. There I will add a new column to config by executing customEnvInfoMethods added by @gsomix .

gsomix:
@MarekM25 @roji Sorry for the late answer. It's a difficult question. Indeed the environment variables cannot vary according to benchmark methods, but consider benchmarking against different Jobs. Personally I'd prefer to place that information into dedicated column, because along with grouping by methods it's more descriptive way than putting additional lines to summary header. But I'm not sure.

@adamsitnik @AndreyAkinshin What do you think guys?

MarekM25:
@roji @gsomix 
I also prefer the first idea. For me, it is more natural and elegant. For example:
``` ini

BenchmarkDotNet=v0.11.5.20191006-develop, OS=Windows 10.0.17763.775 (1809/October2018Update/Redstone5)
Intel Core i5-8250U CPU 1.60GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
  [Host] : .NET Framework 4.7.2 (4.7.3460.0), X64 LegacyJIT DEBUG  [AttachedDebugger]
  Dry    : .NET Framework 4.7.2 (4.7.3460.0), X64 RyuJIT

Job=Dry  IterationCount=1  LaunchCount=1  
RunStrategy=ColdStart  UnrollFactor=1  WarmupCount=1  
Custom environment info: Vector<byte>.Count = 32
```
If it is about implementation, I propose to use static HostEnvironmentInfo class and there add a string property for CustomEnvInfo. We will have one custom info for all executed benchmark classes.

gsomix:
@MarekM25 I'm not sure that's good idea, because we might want to actually invoke methods to retrieve information about environment from generated code. So it may be different from host environment variables and also may vary according to different Runtimes/Jobs.

Please read this comment from Adam in first PR for this issue.
https://github.com/dotnet/BenchmarkDotNet/pull/893#pullrequestreview-164559842

Thanks for working on it!

MarekM25:
@gsomix 
Do you have some implementation tips for it? I see that you did it in generated code, but how to do that in summary? I mean not by adding new column but below environment info with vary for different Runtimes/Jobs.

PR #926, opened 2018-10-25
Added support for async GlobalCleanup.

dlemstra:
This PR adds support for async GlobalCleanup to the `DeclarationsProvider` and that is the last step that needs to be done to complete #521.


adamsitnik:
@dlemstra thank you!

PR #932, opened 2018-10-28
Partition benchmark run info based on added nuget packages

### LINKED ISSUE #931: Same NuGet version used when benchmarking different packages
As [discussed in Gitter](https://gitter.im/dotnet/BenchmarkDotNet?at=5bd48f01c7bf7c36627bed54), my attempt to benchmark two different NuGet versions of FakeItEasy seem to be running the benchmarks against a single version. I'm using BenchmarkDotNet version 0.11.1.821 from the nightlies, and started from the example configuration added in #922. 

I quote the Gitter conversation:

> Hmmm. As I continue working with this, I think it's doing the wrong thing. I don't know if I'm doing something wrong or if there's a bug. I think we're always running the code from one NuGet version. I have a benchmark that's dramatically faster with my 4.10.0… version than the 4.9.1, but if I run the two jobs, I get numbers in line with 4.10 only:

``` ini

BenchmarkDotNet=v0.11.1.821-nightly, OS=Windows 10.0.17134.345 (1803/April2018Update/Redstone4)
Intel Core i7-2720QM CPU 2.20GHz (Sandy Bridge), 1 CPU, 8 logical and 4 physical cores
Frequency=2143568 Hz, Resolution=466.5119 ns, Timer=TSC
.NET Core SDK=2.1.402
  [Host] : .NET Core 2.1.4 (CoreCLR 4.6.26814.03, CoreFX 4.6.26814.02), 64bit RyuJIT
  4.9.1  : .NET Core 2.1.4 (CoreCLR 4.6.26814.03, CoreFX 4.6.26814.02), 64bit RyuJIT

Job=4.9.1  NugetReferences=FakeItEasy 4.9.1  Toolchain=.NET Core 2.1  
IterationCount=1  LaunchCount=1  WarmupCount=10  

```
|  Method |     Mean | Error |
|-------- |---------:|------:|
| Literal | 135.1 us |    NA |

------

``` ini

BenchmarkDotNet=v0.11.1.821-nightly, OS=Windows 10.0.17134.345 (1803/April2018Update/Redstone4)
Intel Core i7-2720QM CPU 2.20GHz (Sandy Bridge), 1 CPU, 8 logical and 4 physical cores
Frequency=2143568 Hz, Resolution=466.5119 ns, Timer=TSC
.NET Core SDK=2.1.402
  [Host]                      : .NET Core 2.1.4 (CoreCLR 4.6.26814.03, CoreFX 4.6.26814.02), 64bit RyuJIT
  4.10.0-make-servicelocat001 : .NET Core 2.1.4 (CoreCLR 4.6.26814.03, CoreFX 4.6.26814.02), 64bit RyuJIT

Job=4.10.0-make-servicelocat001  NugetReferences=FakeItEasy 4.10.0-make-servicelocat001  Toolchain=.NET Core 2.1  
IterationCount=1  LaunchCount=1  WarmupCount=10  

```
|  Method |     Mean | Error |
|-------- |---------:|------:|
| Literal | 15.23 us |    NA |

------

``` ini

BenchmarkDotNet=v0.11.1.821-nightly, OS=Windows 10.0.17134.345 (1803/April2018Update/Redstone4)
Intel Core i7-2720QM CPU 2.20GHz (Sandy Bridge), 1 CPU, 8 logical and 4 physical cores
Frequency=2143568 Hz, Resolution=466.5119 ns, Timer=TSC
.NET Core SDK=2.1.402
  [Host]                      : .NET Core 2.1.4 (CoreCLR 4.6.26814.03, CoreFX 4.6.26814.02), 64bit RyuJIT
  4.10.0-make-servicelocat001 : .NET Core 2.1.4 (CoreCLR 4.6.26814.03, CoreFX 4.6.26814.02), 64bit RyuJIT
  4.9.1                       : .NET Core 2.1.4 (CoreCLR 4.6.26814.03, CoreFX 4.6.26814.02), 64bit RyuJIT

Toolchain=.NET Core 2.1  IterationCount=1  LaunchCount=1  
WarmupCount=10  

```
|  Method |                         Job |                        NugetReferences |     Mean | Error |
|-------- |---------------------------- |--------------------------------------- |---------:|------:|
| Literal | 4.10.0-make-servicelocat001 | FakeItEasy 4.10.0-make-servicelocat001 | 15.24 us |    NA |
| Literal |                       4.9.1 |                       FakeItEasy 4.9.1 | 15.31 us |    NA |

My benchmark project references 4.9.1, and build my configuration looks like this:

```c#
public Config(params string[] versions)
{
    // Specify jobs with different versions of the same Nuget package to benchmark.
    // The Nuget versions referenced on these jobs must be greater or equal to the 
    // same Nuget version referenced in this benchmark project.
    foreach (var version in versions)
    {
        Add(Job.MediumRun
            .WithIterationCount(1)
            .WithLaunchCount(1)
            .With(CsProjCoreToolchain.Current.Value)
            .WithNuget("FakeItEasy", version)
            .WithId(version));
    }
    Add(DefaultConfig.Instance.GetColumnProviders().ToArray());
    Add(DefaultConfig.Instance.GetLoggers().ToArray());
    Add(CsvExporter.Default);
    Add(MarkdownExporter.GitHub);
}
```

the 4.10.0-make-servicelocat001 package is not an official one, but existed in a local NuGet source I had on my computer.

[FakeItEasy.Benchmark.zip](https://github.com/dotnet/BenchmarkDotNet/files/2522032/FakeItEasy.Benchmark.zip) contains the source I used to reproduce, including the output I got when I ran `benchmark.ps1`, which attempts to benchmark 4.9.1 alone, 4.10.0-make-servicelocat001 alone, and the two together. The last attempt gives the suspect results.
I included `nuget/4.10.0-make-servicelocat001.nupkg` in case someone wants to reproduce.

In the .log file for the combined run, I see start dotnet add package FakeItEasy -v 4.10.0-make-servicelocat001 only. No similar line for 4.9.1, even when I change my .csproj to reference something older, like 4.8.0.

I'll try to debug myself, but am occupied for the rest of today. With any luck, I'm doing something silly.

blairconrad:
Fixes #931. Related to #922.

I tried to follow the local coding style, but my have failed. In particular, I saw no consistent method for calculating hashes. I'm happy to make any changes to conform.

The partitioner needs an implementation where package references are considered different if either name or version differ, so I amended `Equals` and `GetHashCode`. then `HashSet` wouldn't recognize duplicate packages, so I replaced it with a custom collection. I had it throw when the nuget reference's package name is a duplicate, since configuring different package versions of the same package for one job is an error, and it seems better to fail fast rather than run with perhaps an unintended version.

Thanks for considering this. Again, I'm happy to make any changes at all.

### DIFF
--- a/src/BenchmarkDotNet/Jobs/NugetReferenceList.cs
+++ b/src/BenchmarkDotNet/Jobs/NugetReferenceList.cs
@@ -1,0 +1,85 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace BenchmarkDotNet.Jobs
+{
+    /// <summary>
+    /// An ordered list of Nuget references. Does not allow duplicate references with the same PackageName.
+    /// </summary>
+    internal class NugetReferenceList : IReadOnlyCollection<NugetReference>
+    {
+        private readonly List<NugetReference> references = new List<NugetReference>();
+
+        public NugetReferenceList(IReadOnlyCollection<NugetReference> readOnlyCollection)
+        {
+            foreach (var nugetReference in readOnlyCollection)
+            {
+                Add(nugetReference);
+            }
+        }
+
+        public int Count => references.Count;
+
+        public IEnumerator<NugetReference> GetEnumerator() => references.GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public void Add(NugetReference reference)
+        {
+            if (reference == null)
+                throw new ArgumentNullException(nameof(reference));
+
+            var insertionIndex = references.BinarySearch(reference, PackageNameComparer.Default);
+            if (0 <= insertionIndex && insertionIndex < Count)
+                throw new ArgumentException($"Nuget package {reference.PackageName} was already added", nameof(reference));
+
+            references.Insert(~insertionIndex, reference);
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj))
+                return false;
+            if (ReferenceEquals(this, obj))
+                return true;
+            if (obj.GetType() != GetType())
+                return false;
+
+            var otherList = (NugetReferenceList)obj;
+            if (Count != otherList.Count)
+                return false;
+
+            for (int i = 0; i < Count; i++)
+            {
+                if (!references[i].Equals(otherList.references[i]))
+                    return false;
+            }
+
+            return true;
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                int hashCode = 0;
+                foreach (var nugetReference in references)
+                {
+                    hashCode = hashCode * 397 + nugetReference.GetHashCode();
+                }
+
+                return hashCode;
+            }
+        }
+
+        private class PackageNameComparer : IComparer<NugetReference>
+        {
+            private PackageNameComparer() { }
+
+            public static PackageNameComparer Default { get; } = new PackageNameComparer();
+
+            public int Compare(NugetReference x, NugetReference y) => Comparer<string>.Default.Compare(x.PackageName, y.PackageName);
+        }
+    }
+}
--- a/src/BenchmarkDotNet/Jobs/NugetReferenceList.cs
+++ b/src/BenchmarkDotNet/Jobs/NugetReferenceList.cs
@@ -7,9 +7,13 @@ namespace BenchmarkDotNet.Jobs
     /// <summary>
     /// An ordered list of Nuget references. Does not allow duplicate references with the same PackageName.
     /// </summary>
-    internal class NugetReferenceList : IReadOnlyCollection<NugetReference>
+    public class NugetReferenceList : IReadOnlyCollection<NugetReference>
     {
         private readonly List<NugetReference> references = new List<NugetReference>();
+
+        public NugetReferenceList()
+        {
+        }
 
         public NugetReferenceList(IReadOnlyCollection<NugetReference> readOnlyCollection)
         {
--- a/src/BenchmarkDotNet/Running/BenchmarkPartitioner.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkPartitioner.cs
@@ -45,6 +45,8 @@ namespace BenchmarkDotNet.Running
                     return false;
                 if (AreDifferent(jobX.Infrastructure.Arguments, jobY.Infrastructure.Arguments)) // arguments can be anything (Mono runtime settings or MsBuild parameters)
                     return false;
+                if (AreDifferent(jobX.Infrastructure.NugetReferences, jobY.Infrastructure.NugetReferences))
+                    return false;
                 if (!jobX.Environment.Gc.Equals(jobY.Environment.Gc)) // GC settings are per .config/.csproj
                     return false;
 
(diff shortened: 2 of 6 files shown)


AndreyAkinshin:
@Shazwazza, @adamsitnik could you please review this PR?

src/BenchmarkDotNet/Running/BenchmarkPartitioner.cs
```diff
@@ -78,6 +80,8 @@ public int GetHashCode(BenchmarkCase obj)
                     hashCode ^= job.Infrastructure.BuildConfiguration.GetHashCode();
                 if (job.Infrastructure.Arguments != null && job.Infrastructure.Arguments.Any())
                     hashCode ^= job.Infrastructure.Arguments.GetHashCode();
+                if (job.Infrastructure.NugetReferences != null)
+                    hashCode ^= job.Infrastructure.GetHashCode();
```
adamsitnik:
```suggestion
                    hashCode ^= job.Infrastructure.NugetReferences.GetHashCode();
```

src/BenchmarkDotNet/Jobs/NugetReferenceList.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace BenchmarkDotNet.Jobs
+{
+    /// <summary>
+    /// An ordered list of Nuget references. Does not allow duplicate references with the same PackageName.
+    /// </summary>
+    internal class NugetReferenceList : IReadOnlyCollection<NugetReference>
```
adamsitnik: hmm, maybe we should give up on the idea of `NugetReference` and use `Dictionary<string, string>` instead?

blairconrad:
> The only thing I would like you to change before I merge this fix is:
> 
> ```diff
> - hashCode ^= job.Infrastructure.GetHashCode();
> + hashCode ^= job.Infrastructure.NugetReferences.GetHashCode();
> ```

Of course. So glad you caught this. Copy/paste error. I will definitely fix. Do you prefer an amended commit or a `fixup!` to be later squashed?

I would also be happy to explore the use of a `Dictionary` (I'm assuming package name is the key and the version the value) for the nuget packages, if that's your preference. Although I'm not sure it would solve all our problems, at least without also adding

1.  an up-front check to ensure we don't replace an existing entry with a new one that has a different version, which is currently encapsulated in `NugetReferenceList`, and
2. custom equality-checking and hashcode-generation, which is currently encapsulated in `NugetReferenceList`.

but if you prefer that code to be separate from the container, I can make the change.

blairconrad:
Whoops. Builds failed because I'd left `NugetReferenceList` as internal. I had an unpushed change in my local repo that resolves this, but will wait on it before hearing your feelings on amendments vs. fixup commits (and also wondering if my `Dictionary` comments will turn you pack onto the `NugetRefernceList`: if not, the problem goes away anyhow).

adamsitnik:
You don’t need to squash, I can do it from github UI level when merging the PR. When it comes to using the dictionary we can leave it for later. I will do some PoC myself to validate this idea.

Shazwazza:
I didn't realize changes were needed for the `BenchmarkPartitioner`, wasn't code that i had actually looked at, clearly should have tested further ;) And looks like a problem snuck in there with my later changes to the HashSet of NugetReference. Good to see there wasn't a ton of code that needed changing. 

Only nitpick might be that if we keep the NugetReferenceList maybe we should make it implement IReadOnlyList instead of IReadOnlyCollection since it seems that most other collections are exposed as this and means in the test https://github.com/dotnet/BenchmarkDotNet/pull/922/files#diff-048cab01ff351e9c6602501a273532d5R480 that ElementAt doesn't need to be used an instead an indexer can be used. I ended up using IReadOnlyCollection only due to the HashSet usage since it doesn't implement IReadOnlyList

blairconrad:
I see that AppVeyor failed, but do not see an actual failure. Does the build time out after an hour? I can't imagine what in my change would blow the time up, especially considering that the Travis build was as fast as usual.

adamsitnik:
@blairconrad thank you! the CI failure seems to be unrelated, don't worry

blairconrad:
Yay! Thanks for the reviews and merge, folks.

PR #940, opened 2018-10-31
Add support for mono AOT pass

alexanderkyte:


alexanderkyte:
Fixes #867

alexanderkyte:
@adamsitnik any stylistic changes you'd like to see here?

alexanderkyte:
@AndreyAkinshin seems like they may be interested in this as well

AndreyAkinshin:
@alexanderkyte thanks for the PR and sorry for the late reply.
Could you add a few usage examples of this feature in the documentation?

src/BenchmarkDotNet/Toolchains/Roslyn/Builder.cs
```diff
@@ -11,9 +12,72 @@
...
+            try {
+                using (var process = new Process { StartInfo = CreateMonoAot(monoPath, dllPath) })
+                {
+                    logger.WriteLineInfo("// Execute: " + process.StartInfo.FileName + " " + process.StartInfo.Arguments);
+                    ConsoleHandler.Instance.SetProcess(process);
+                    process.Start();
+                    process.WaitForExit(); // should we add timeout here?
+
+                    if (process.ExitCode != 0)
+                        logger.WriteLineError("Attempt to AOT failed: " + process.StartInfo.FileName + " " + process.StartInfo.Arguments);
```
adamsitnik: **important**: in this case you should also:

1. log the output so the user knows why it has failed
2. return `BuildResult.Failure` instead of `result` from base builder which is success in this case

src/BenchmarkDotNet/Toolchains/Roslyn/Builder.cs
```diff
@@ -11,9 +12,72 @@
...
+            };
+            start.FileName = monoPath;
+            start.Arguments = String.Format("{0} {1}", mono.AotArgs, absFilePath);
+            if (mono.MonoBclPath != null)
+                start.EnvironmentVariables ["MONO_PATH"] = mono.MonoBclPath;
+
+            return start;
+        }
+
+        public MonoAotBuilder (MonoRuntime monoReference)
```
adamsitnik: please move the ctor to the top of the file, also it's a good place to use expression `=>`

src/BenchmarkDotNet/Toolchains/Roslyn/RoslynToolchain.cs
```diff
@@ -14,6 +15,12 @@ public class RoslynToolchain : Toolchain
     {
         public static IToolchain Instance = new RoslynToolchain();
 
+        [PublicAPI]
+        public RoslynToolchain (MonoRuntime mono) : base("Roslyn", new Generator(), new MonoAotBuilder(mono), new Executor())
```
adamsitnik: nit: spacing
important: name

```suggestion
        public RoslynToolchain(MonoRuntime mono) : base("MonoAOT", new Generator(), new MonoAotBuilder(mono), new Executor())
```

src/BenchmarkDotNet/Toolchains/Roslyn/RoslynToolchain.cs
```diff
@@ -14,6 +15,12 @@ public class RoslynToolchain : Toolchain
...
+        {
+
```
adamsitnik: nit: empty line

```suggestion
```

adamsitnik:
@alexanderkyte @EgorBo guys any chances that you could finish this PR?

alexanderkyte:
Definitely :-)

Sorry for the delay, I've got some time allocated for this toward the end of this sprint. I'll try to get the cosmetic changes done sooner than later.

src/BenchmarkDotNet/Toolchains/Roslyn/Builder.cs
alexanderkyte: 1. The output logging request here confuses me. The `CreateMonoAot` function sets `RedirectStandardOutput = true`, so we should be seeing the output from this command in the stdout of the BDN runner. I'll add the error code though.

2. Will do

alexanderkyte:
@adamsitnik I've made the requested changes to this PR. 

Do you have any tests on CI that currently use mono (ie: is this change being tested outside of mono's usage of the library?).

Issue #787, opened 2018-06-09
Grand renaming


AndreyAkinshin:
Since, we are going to introduce a lot of breaking changes in v0.11.0, it would be nice to rename a few things as well. =)
Current naming scheme has an important problem: we use the same generic words like "Benchmark", "Main", "Target" in different places for different purposes.
Thus, you can't say what "Target" means without a context.
This grand renaming has two main goals:

* Use one term for one purpose. We shouldn't reuse words for different things.
* Avoid "generic" words with unclear meaning if possible.

Here is my list of renaming:

### IterationMode

I suggest to split `IterationMode` to `IterationMode` and `IterationStage` as follows:

- [x] `Iteration`
  * `Mode`: `Idle`, `Workload`
  * `Stage`: `Jitting`, `Pilot`, `Warmup`, `General`, `Result`

Thus, we can use the `Workload` term for our the benchmark which we want to benchmark (no more `Main`, `Target`, `Method`, etc.)

### Benchmark

- [x] `Benchmark` -> `BenchmarkCase` (solving a conflict between `class Benchmark` and `class BenchmarkAttribute` which is used as `[Benchmark]`)
- [x] `Benchmark.Target` -> `BenchmarkCase.Descriptor`

Descriptor should include the following properties:

- [x] `Type`
- [x] `GlobalSetupMethod`, `GlobalCleanupMethod`, `IterationSetupMethod`, `IterationCleanupMethod`
- [x] `WorkloadMethod`

### Other renames

- [x] ~`Analyser` -> `Inspector` (solving a conflict between American "Analyser" and British "Analyzer")~,
  `Conclusion` -> `Inspection` (consistency)
- [x] `OrderProvider` -> `Orderer` (prefer one-word term)
- [x] ~`SummaryStyle` -> `Styler` (prefer one-word term)~
- [x] `IsBaseline` -> `Baseline` (consistency)
- [x] Move `EnvironmentVariables` from `InfrastructureMode` to `EnvMode`
- [x] `General` -> `Actual`

AndreyAkinshin:
@adamsitnik, what do you think?

adamsitnik:
> I suggest to split IterationMode to IterationMode and IterationStage as follows:

The only suggestion I have now is to rename `Idle` to `Overhead` 

> Benchmark -> BenchmarkCase
> Benchmark.Target -> BenchmarkCase.Descriptor
> OrderProvider -> Orderer

👍 

[...]

PR #941, opened 2018-11-01
Fix example code (isBaseline -> baseline)

PathogenDavid:
#787 changed the parameter name on job attributes, but the example code on the landing page still used the old name. I grepped the docs folder and this appears to be the only example that had this issue.


AndreyAkinshin:
@PathogenDavid, thanks!

PR #944, opened 2018-11-02
Fixed typo in IntroTagColumn sample

ahmedalejo:
the sample read:
```csharp
Add(new TagColumn("Foo or Bar", name => name.Substring(0, 3)));
```
instead of
```csharp
Add(new TagColumn("Kind", name => name.Substring(0, 3)));
```


AndreyAkinshin:
@ahmedalejo thanks!

Issue #939, opened 2018-10-31
We need an option to stop running when the first benchmark fails.


jorive:
When running benchmarks in the lab, it is ideal to fail fast and avoid using resources unnecessarily. When a benchmark fails we would like to acknowledge the developers as soon as possible.
Currently, the harness keeps executing all benchmarks to completion. Is there a way to tell BDN to fail fast?

adamsitnik:
I think that it would be nice to add such feature. 

It could be added to `IConfig` as `StopOnFirstError` with default value = false and exposed via console line arguments to override.

@wojtpl2 maybe you would like to give it a try?

wojtpl2:
Ok next week

jorive:
@adamsitnik What is the reasoning for `default value = false`?

adamsitnik:
@jorive to don't change the current default behavior (breaking change). I like the current one: last night I started the benchmark run for all the benchmarks we have and today in the morning I had the results for few thousands of them. 8 failed had no results but I had the errors in the logs. If it would stop on the first error I would have to fix the problem and rerun all the benchmarks again. With current behavior, I just fixed and re-run 8 of them.

adamsitnik:
Fixed by #947

PR #947, opened 2018-11-06
Add option to stop running when the first benchmark fails

wojtpl2:
Fixes for #939
I'm not sure all of my changes, especially adding property and changing constructor in BenchmarkReport.


wojtpl2:
@adamsitnik I've also added `--stopOnFirstError` command line argument.

AndreyAkinshin:
Could you please also review this PR?

Issue #948, opened 2018-11-07
BenchmarkDotNet.Mathematics.StatisticalTesting.MannWhitneyTest.PValueForSmallN(int n, int m, double u)


ArheusLis:
If I use only ClrJob & CoreJob - all fine.
But then I tried MonoJob or CoreRtJob after all test pass I see: System.IndexOutOfRangeException: 'Index was outside the bounds of the array.' in BenchmarkDotNet.Mathematics.StatisticalTesting.MannWhitneyTest.PValueForSmallN(int n, int m, double u) .
May be I'm using something wrong?

Bechmark is simple:
[Benchmark]
        public void ERange()
        {
            for (int i = 0; i < N; i++)
            {
                Enumerable.Range(0, N);
            }
        }

and 
    [CoreRtJob][CoreJob] [ClrJob][MonoJob("mono", @"C:\Program Files\Mono\bin\mono.exe")]

ArheusLis:
На 0.11 проблема не воспроизводится.
На 0.12 - регулярно.

wojtpl2:
I can't reproduce this error. What value of N do you use? I checked 1, 10, 1000 and I didn't get this error. 

I checked that MannWhitneyTest.cs file haven't been changing since 28/9/2018.

I created branche _issue_948_ in my fork https://github.com/wojtpl2/BenchmarkDotNet/tree/issue_948 and I added your simple benchmark https://github.com/wojtpl2/BenchmarkDotNet/commit/cbc0713430ace399174370219e2f64fc64460bd6. Can you check if I'm doing everything right?

wojtpl2:
Problem is in https://github.com/dotnet/BenchmarkDotNet/commit/05cc8d15ef88e382bbb1827d766d7275c3e42abd commit. I'll try to understend it and fix. It was added in 0.11.2.

wojtpl2:
I have stack trace from debug:
```
   at BenchmarkDotNet.Mathematics.StatisticalTesting.MannWhitneyTest.PValueForSmallN(Int32 n, Int32 m, Double u) in C:\Work\BenchmarkDotNet\src\BenchmarkDotNet\Mathematics\StatisticalTesting\MannWhitneyTest.cs:line 26

```

wojtpl2:
I've created PR #951 with failing test.

PR #951, opened 2018-11-09
Add failing test for #948

wojtpl2:
@AndreyAkinshin I caught parameters that reproduce error. 
Exception is throwing from this line: https://github.com/dotnet/BenchmarkDotNet/blob/05cc8d15ef88e382bbb1827d766d7275c3e42abd/src/BenchmarkDotNet/Mathematics/StatisticalTesting/MannWhitneyTest.cs#L23-L24
when `i=15, j=1, k=1` and `w = new long[20, 15, 266]`


AndreyAkinshin:
@wojtpl2 thanks!

Issue #68, opened 2016-01-14
Power management


omariom:
My ultrabook's CPU frequency goes up and down accommodating to the load.
That's because it is in Balanced power mode. To keep it upmost I set it to High performance manually.

It would be convenient if BenchmarkDotNet  did that for us.

This is what I use in my microbechmarking tool for LinqPad.
I found it in Vance Morrison's MeasureIt.

``` C#
using System;
using System.Runtime.InteropServices;
using System.Diagnostics;

/// <summary>
/// PowerManagement allows you to access the funtionality of the Control Panel -> Power Options
/// dialog in windows.  (Currently we only use VISTA APIs). 
/// </summary>
public static unsafe class PowerManagment
{
    public static Guid HighPerformance = new Guid(0x8c5e7fda, 0xe8bf, 0x4a96, 0x9a, 0x85, 0xa6, 0xe2, 0x3a, 0x8c, 0x63, 0x5c);
    public static Guid Balenced        = new Guid(0x381b4222, 0xf694, 0x41f0, 0x96, 0x85, 0xff, 0x5b, 0xb2, 0x60, 0xdf, 0x2e);
    public static Guid PowerSaver      = new Guid(0xa1841308, 0x3541, 0x4fab, 0xbc, 0x81, 0xf7, 0x15, 0x56, 0xf2, 0x0b, 0x4a);

    public static Guid CurrentPolicy
    {
        get
        {
            Guid* retPolicy = null;
            Guid ret = Guid.Empty;
            try
            {
                int callRet = PowerGetActiveScheme(IntPtr.Zero, ref retPolicy);
                if (callRet == 0)
                {
                    ret = *retPolicy;
                    Marshal.FreeHGlobal((IntPtr)retPolicy);
                }
            }
            catch (Exception) { }
            return ret;
        }
    }
    public static bool Set(Guid newPolicy)
    {
        try
        {
            return PowerSetActiveScheme(IntPtr.Zero, ref newPolicy) == 0;
        }
        catch (Exception) { }
        return false;
    }

    #region private 
        [DllImport("powrprof.dll")]
    private static extern int PowerGetActiveScheme(IntPtr ReservedZero, ref Guid* policyGuidRet);

    [DllImport("powrprof.dll")]
    private static extern int PowerSetActiveScheme(IntPtr ReservedZero, ref Guid policyGuid);
    #endregion
}
```

AndreyAkinshin:
[...]

PR #952, opened 2018-11-10
Implemented power-management, add docs (#68)

MarekM25:


AndreyAkinshin:
@MarekM25 thanks for the PR.

docs/articles/configs/powerplans.md
```diff
@@ -0,0 +1,19 @@
+---
+uid: docs.powerplans
+name: Power Plans
+---
+
+# Power Plans
+
+This description concerns only version v0.11.2 and above of BenchmarkDotNet. In the previous versions, benchmarks in the power plans.
```
AndreyAkinshin: We don't need a note about the BenchmarkDotNet version. The documentation is always relevant only for the latest version of BenchmarkDotNet.

samples/BenchmarkDotNet.Samples/IntroPowerPlan.cs
```diff
@@ -0,0 +1,12 @@
+﻿using BenchmarkDotNet.Attributes;
+
+namespace BenchmarkDotNet.Samples
+{
+    [HighPerformancePowerPlan(false)]
+    public class IntroPowerPlan
```
AndreyAkinshin: In the IntroSample, we need two jobs (enabled and disabled high performance power plan) to show the difference between them. An example of the summary table should be included in `IntroPowerPlan.md`.

src/BenchmarkDotNet/Attributes/HighPerformancePowerPlanAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BenchmarkDotNet.Attributes
+{
+    /// <summary>
+    /// determines if BenchmarDotNet changes power plan to High Performance
+    /// </summary>
+    [PublicAPI]
+    public class HighPerformancePowerPlanAttribute : Attribute, IConfigSource
```
AndreyAkinshin: It makes sense to add `[AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = true)]` here. It will allow comparing enabled/disabled configurations.

src/BenchmarkDotNet/Running/PowerManagementApplier.cs
```diff
@@ -0,0 +1,70 @@
...
+using BenchmarkDotNet.Helpers;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Portability;
+
+namespace BenchmarkDotNet.Running
+{
+    internal class PowerManagementApplier
+    {
+        private Guid? _userCurrentPowerPlan;
+        private bool _powerPlanChanged = false;
```
AndreyAkinshin: We don't use underscores as private field prefixes. Please, remove it.

tests/BenchmarkDotNet.IntegrationTests/PowerManagementApplierTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    public class PowerManagementApplierTests : BenchmarkTestExecutor
+    {
+        public const string PowerSaverGuid = "e22fd527-0c09-43ad-83d6-ba300affc27d";
+
+        public PowerManagementApplierTests(ITestOutputHelper output) : base(output) { }
+
+        [Fact]
+        public void TestSettingAndRevertingBackGuid()
+        {
+            if (RuntimeInformation.IsWindows())
```
AndreyAkinshin: Please, use `FactWindowsOnlyAttribute` instead of such checks.

src/BenchmarkDotNet/Running/PowerManagementApplier.cs
```diff
@@ -0,0 +1,70 @@
...
+        private bool _isInitialized = false;
+
+        internal void ApplyPerformancePlan(ILogger logger, bool highPerformancePowerPlan)
+        {
+            if (RuntimeInformation.IsWindows())
+            {
+                if (highPerformancePowerPlan && _powerPlanChanged == false)
```
AndreyAkinshin: Could we move the Apply logic from constructor to a separate method?

AndreyAkinshin:
@MarekM25 the build is restarted.

> To be sure I thought about only enabling/disabling the high-performance plan. From my perspective, allowing multiple attributes with the current implementation is not enough (of course, I might be wrong). However, it is a remarkably great idea to do it, and I suggest doing more here. I mean not only analyzing disabled and enabled a high-performance plan but making it possible to compare any power plan for a user. For example, we could compare HighPerformance, Balanced, PowerSaver. What is more, I consider allowing a user to both use power plan enum and provide it by GUID string. The first way contains popular power plan GUIDs such as I mentioned previously. The second approach is essential because companies could have own, custom power plans for any device. This feature allows targeting benchmarks when we could expect what power plan is used by an end user. We could benchmark different power plans if we know that our software will be run on a server, a personal laptop or mobile phone. I could do it willingly as my next issue or by continuing actual:) What do you think?

It looks pretty interesting. However, let's do the minimal implementation first. Currently, `HighPerformancePowerPlan` is a property of a config. That's why you had to duplicate your benchmark in the sample:

```cs
    [HighPerformancePowerPlan(false)]
    public class IntroPowerPlanDisabled
    {
        [Benchmark]
        public int SplitJoin()
            => string.Join(",", new string[1000]).Split(',').Length;
    }
    // By default benchmark.net uses high-performance power plan.
    // There is no need to set it to true explicitly
    [HighPerformancePowerPlan(true)]
    public class IntroPowerPlanEnabled
    {
        [Benchmark]
        public int SplitJoin()
            => string.Join(",", new string[1000]).Split(',').Length;
    }
```

The PowerPlan should be a property of a job (like `Platform`, `Jit`, and so on). In this case, users will be able to apply different power plans to the same benchmark class without code duplication.

MarekM25:
Hi @AndreyAkinshin ,
I finished working on this PR. Could you have a look? What do you think?
The build failed, but the same test locally and before squashing commits works fine. Is it possible that the test is unstable?

AndreyAkinshin:
@MarekM25 I will review this PR on the next week. Meanwhile, could you please resolve conflicts with the master branch?

MarekM25:
> 
> 
> @MarekM25 I will review this PR on the next week. Meanwhile, could you please resolve conflicts with the master branch?

Thx, done ;)

AndreyAkinshin:
@MarekM25 sorry for such a delay, I had to verify this branch on different laptops. It works great! Thanks for your contribution!

MarekM25:
> @MarekM25 sorry for such a delay, I had to verify this branch on different laptops. It works great! Thanks for your contribution!

Thx, it was pleasure for me!

Issue #834, opened 2018-07-19
Better list of suggested benchmarks for wrong filter


AndreyAkinshin:
If we make a typo in a benchmark filter expression, BenchmarkDotNet will print the first 40 known benchmark names (see https://github.com/dotnet/BenchmarkDotNet/issues/829). It will be better to show names which are similar to the filter value (in the case of a typo, it will be easy to fix it). We can use [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) for this feature.

adamsitnik:
Fixed by #957

PR #957, opened 2018-11-14
Better list of suggested benchmarks for wrong filter #834

morgan-kn:
Better list of suggested benchmarks for the wrong filter #834


AndreyAkinshin:
@morgan-kn, thanks for the PR!

src/BenchmarkDotNet/ConsoleArguments/MisspellingsFinder.cs
```diff
@@ -0,0 +1,66 @@
...
+        public MisspellingsFinder(IReadOnlyList<Type> benchmarks)
+        {
+            foreach (var benchmark in benchmarks)
+            {
+                benchmarkNames.AddRange(benchmark.GetMethods()
+                    .Where(methodInfo => methodInfo.HasAttribute<BenchmarkAttribute>())
+                    .Select(methodInfo => $"{benchmark.FullName}.{methodInfo.Name}"));
+            }
+
+            benchmarkNames.AddRange(benchmarks.Select(bn => bn.FullName).Distinct());
```
AndreyAkinshin: It would be also nice to add benchmark type names (without namespaces).

src/BenchmarkDotNet/ConsoleArguments/MisspellingsFinder.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public string[] Find([NotNull] string userInput)
+        {
+            if (userInput == null) 
+                throw new ArgumentNullException(nameof(userInput));
+            return benchmarkNames.Where(name => GetLevenshteinDistance(userInput, name) <= PossibleMisspellingCount).ToArray();
+        }
+
+        private static int GetLevenshteinDistance(string string1, string string2)
```
AndreyAkinshin: Could you write a few unit tests for `GetLevenshteinDistance`? (just for pure string input without the reflection logic).

src/BenchmarkDotNet/ConsoleArguments/MisspellingsFinder.cs
```diff
@@ -0,0 +1,36 @@
...
+                    .Where(methodInfo => methodInfo.HasAttribute<BenchmarkAttribute>())
+                    .Select(methodInfo => $"{benchmark.FullName}.{methodInfo.Name}"));
+            }
+
+            benchmarkNames.AddRange(benchmarks.Select(bn => bn.FullName).Distinct());
+            benchmarkNames.AddRange(benchmarks.Select(bn => bn.Namespace).Distinct());
+            benchmarkNames.AddRange(benchmarks.Select(bn => bn.Name).Distinct());
+        }
+
+        public string[] Find([NotNull] string userInput)
```
AndreyAkinshin: Since the updated implementation actually sorts benchmarks and move most relevant results to the top, I'm not sure that `MisspellingsFinder.Find` is the best name for this API. Could we come up with better name?

morgan-kn: @AndreyAkinshin is CorrectionsSuggester ok?

Issue #955, opened 2018-11-13
How to dynamically load BenchmarkDotNet.Diagnostics.Windows without NuGet hacks


adamsitnik:
We have `BenchmarkDotNet.Diagnostics.Windows.dll` and `BenchmarkDotNet.dll`.  `BenchmarkDotNet.Diagnostics.Windows.dll` has dependency to `BenchmarkDotNet.dll`, not the opposite.

When the user specifies `--profiler ETW` we parse command line arguments in `BenchmarkDotNet.dll` and we try to load `BenchmarkDotNet.Diagnostics.Windows.dll` in a dynamic way using `Assembly.Load` and others:

https://github.com/dotnet/BenchmarkDotNet/blob/60ea1705d7d8e31eb1292e5bf785f818b9cbd0a2/src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs#L97-L112

@eerhardt how can we load it without `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>` in NuGet friendly way?

eerhardt:
This is exactly the scenario that [`DependencyContext`](https://github.com/dotnet/core-setup/blob/master/src/managed/Microsoft.Extensions.DependencyModel/DependencyContext.cs) is meant to solve.

You can find all the dependencies of the current application by looking in `DependencyContext.Default.RuntimeLibraries`.

Note this typically only works on .NET Core projects, which automatically create `.deps.json` files. It will work on .NET Framework projects, if the application is built to produce a `.deps.json` file (not the default on .NET Framework).

wojtpl2:
I did this issue. I'll push PR in no time.

wojtpl2:
@adamsitnik Do you have any other task that I can take care of? I can finish #912 if you want. Of course, I'll keep your history of changes. Or I can take something else.

PR #958, opened 2018-11-14
Use DependencyContext to load diagnostics assembly

wojtpl2:
Fix for #955

### DIFF
--- a/src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
+++ b/src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
@@ -6,7 +6,6 @@ using System.Threading;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Portability;
-using Microsoft.Extensions.DependencyModel;
 
 namespace BenchmarkDotNet.Diagnosers
 {
@@ -59,16 +58,16 @@ namespace BenchmarkDotNet.Diagnosers
         {
             try
             {
-                var benchmarkDotNetCoreAssembly = typeof(DefaultConfig).GetTypeInfo().Assembly;
+                var benchmarkDotNetAssembly = typeof(DefaultConfig).GetTypeInfo().Assembly;
 
-                var diagnosticsAssembly = LoadDiagnosticsAssembly(benchmarkDotNetCoreAssembly);
+                var diagnosticsAssembly = Assembly.Load(new AssemblyName(DiagnosticAssemblyName));
 
-                if (diagnosticsAssembly.GetName().Version != benchmarkDotNetCoreAssembly.GetName().Version)
+                if (diagnosticsAssembly.GetName().Version != benchmarkDotNetAssembly.GetName().Version)
                 {
                     string errorMsg =
                         $"Unable to load: {DiagnosticAssemblyFileName} version {diagnosticsAssembly.GetName().Version}" +
                         Environment.NewLine +
-                        $"Does not match: {Path.GetFileName(benchmarkDotNetCoreAssembly.Location)} version {benchmarkDotNetCoreAssembly.GetName().Version}";
+                        $"Does not match: {Path.GetFileName(benchmarkDotNetAssembly.Location)} version {benchmarkDotNetAssembly.GetName().Version}";
                     ConsoleLogger.Default.WriteLineError(errorMsg);
                 }
                 else
@@ -93,30 +92,6 @@ namespace BenchmarkDotNet.Diagnosers
                 MemoryDiagnoser.Default,
                 DisassemblyDiagnoser.Create(new DisassemblyDiagnoserConfig())
             };
-        }
-
-        private static Assembly LoadDiagnosticsAssembly(Assembly benchmarkDotNetCoreAssembly)
-        {
-            // We couldn't use Assembly.GetEntryAssembly()?.GetReferencedAssemblies() because
-            // it not enough to just install NuGet to be "referenced", the project has to consume the dll for real to be on the referenced assembly list
-            // So we use DependencyContext that reads *.deps.json file that contains all referenced assembly. 
-            var inCompileLibraries = DependencyContext.Default.CompileLibraries.Any(l => l.Name.Equals(DiagnosticAssemblyName, StringComparison.OrdinalIgnoreCase));
-            var inRuntimeLibraries = DependencyContext.Default.RuntimeLibraries.Any(l => l.Name.Equals(DiagnosticAssemblyName, StringComparison.OrdinalIgnoreCase));
-
-            if (inCompileLibraries || inRuntimeLibraries)
-            {
-                return Assembly.Load(new AssemblyName(DiagnosticAssemblyName));
-            }
-
-            // we use the location of BenchmarkDotNet.dll, because it should be in the same folder
-            string directoryName = new FileInfo(benchmarkDotNetCoreAssembly.Location).DirectoryName
-                ?? throw new DirectoryNotFoundException(benchmarkDotNetCoreAssembly.Location);
-            string diagnosticAssemblyBinPath = Path.Combine(directoryName, DiagnosticAssemblyFileName);
-            if (File.Exists(diagnosticAssemblyBinPath))
-                return Assembly.LoadFile(diagnosticAssemblyBinPath);
-
-            // Assembly.LoadFrom(fileName) searches in current directory, not bin, but it's our last chance
-            return Assembly.LoadFrom(DiagnosticAssemblyFileName);
         }
 
         private static IDiagnoser CreateDiagnoser(Assembly loadedAssembly, string typeName)
--- a/src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
+++ b/src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Portability;
+using Microsoft.Extensions.DependencyModel;
 
 namespace BenchmarkDotNet.Diagnosers
 {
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
```diff
@@ -96,10 +97,16 @@ private static IDiagnoser[] LoadClassic()
 
         private static Assembly LoadDiagnosticsAssembly(Assembly benchmarkDotNetCoreAssembly)
         {
+            // We couldn't use Assembly.GetEntryAssembly()?.GetReferencedAssemblies() because
```
adamsitnik: I think that we should keep the old way (to make it work for old .NET frameworks) and extend it with the new one.

So it would be:

```cs
var referencedAssemblyName = Assembly.GetEntryAssembly()?.GetReferencedAssemblies().SingleOrDefault(name => name.Name == DiagnosticAssemblyName);	            if (referencedAssemblyName != default(AssemblyName))
    return Assembly.Load(referencedAssemblyName);

var inCompileLibraries = DependencyContext.Default.CompileLibraries.Any(l => l.Name.Equals(DiagnosticAssemblyName, StringComparison.OrdinalIgnoreCase));
var inRuntimeLibraries = DependencyContext.Default.RuntimeLibraries.Any(l => l.Name.Equals(DiagnosticAssemblyName, StringComparison.OrdinalIgnoreCase));

if (inCompileLibraries || inRuntimeLibraries)
    return Assembly.Load(new AssemblyName(DiagnosticAssemblyName));
```

src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
```diff
@@ -96,10 +97,16 @@ private static IDiagnoser[] LoadClassic()
...
-            var referencedAssemblyName = Assembly.GetEntryAssembly()?.GetReferencedAssemblies().SingleOrDefault(name => name.Name == DiagnosticAssemblyName);
-            if (referencedAssemblyName != default(AssemblyName))
-                return Assembly.Load(referencedAssemblyName);
+            // So we use DependencyContext that reads *.deps.json file that contains all referenced assembly. 
+            var inCompileLibraries = DependencyContext.Default.CompileLibraries.Any(l => l.Name.Equals(DiagnosticAssemblyName, StringComparison.OrdinalIgnoreCase));
+            var inRuntimeLibraries = DependencyContext.Default.RuntimeLibraries.Any(l => l.Name.Equals(DiagnosticAssemblyName, StringComparison.OrdinalIgnoreCase));
+
+            if (inCompileLibraries || inRuntimeLibraries)
+            {
+                return Assembly.Load(new AssemblyName(DiagnosticAssemblyName));
```
adamsitnik: I would assume that the `DependencyContext` exposes some method to load the library without providing the name. Here you just check that it's on the list and if it is you load it by name. How `Assembly.Load` will know about the `.deps` file here?

@eerhardt how should we use `DependencyContext` to load an assembly?

wojtpl2: I found this way there https://github.com/dotnet/corefx/issues/11639#issuecomment-300565009.
I created project on the side and this solution works.

I don't know how `Assembly.Load` works. I'll try to check.

src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
wojtpl2: I'll check old .Net frameworks path.

eerhardt: Agreed. `DependencyContext` doesn't work "by default" on .NET Framework. DependencyContext is mostly just for the .NET Core case.

You should also check for `DependencyContext.Default == null`, which will be the case when a `.deps.json` file isn't present.

src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
```diff
@@ -96,10 +97,16 @@ private static IDiagnoser[] LoadClassic()
 
         private static Assembly LoadDiagnosticsAssembly(Assembly benchmarkDotNetCoreAssembly)
         {
+            // We couldn't use Assembly.GetEntryAssembly()?.GetReferencedAssemblies() because
             // it not enough to just install NuGet to be "referenced", the project has to consume the dll for real to be on the referenced assembly list
-            var referencedAssemblyName = Assembly.GetEntryAssembly()?.GetReferencedAssemblies().SingleOrDefault(name => name.Name == DiagnosticAssemblyName);
-            if (referencedAssemblyName != default(AssemblyName))
-                return Assembly.Load(referencedAssemblyName);
+            // So we use DependencyContext that reads *.deps.json file that contains all referenced assembly. 
+            var inCompileLibraries = DependencyContext.Default.CompileLibraries.Any(l => l.Name.Equals(DiagnosticAssemblyName, StringComparison.OrdinalIgnoreCase));
```
eerhardt: You shouldn't check the `CompileLibraries` at all. These are not appropriate for runtime scenarios, and are mainly there to support things like ASP.NET MVC Razor, which compiles code at runtime using Roslyn.

src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
eerhardt: `DependencyContext` allows you to get the information about a dependency, but it doesn't have assembly loading capabilities. You still use the normal `Assembly.Load` APIs to load the assembly.

The reason I suggested using `DependencyContext` is because I assumed the main thing you need/want here is you want to get the AssemblyVersion of the referenced `BenchmarkDotNet.Diagnostics.Windows.dll`, right?  (That information can be found [on this API](https://github.com/dotnet/core-setup/blob/bb31dc3b52e38ca4e679f741cd0263dc239b926f/src/managed/Microsoft.Extensions.DependencyModel/RuntimeFile.cs#L24)).

If it works to just `Assembly.Load(new AssemblyName(DiagnosticAssemblyName));`, then you don't need the `DependencyContext` nor the bit about `Assembly.GetEntryAssembly()?.GetReferencedAssemblies()` at all.  Just call `Assembly.Load` passing in the .dll name.

wojtpl2:
@eerhardt @adamsitnik Big thanks for this review. It is amazing how much I can learn in this project. I'll fix this PR..

src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs
wojtpl2: Calling `Assembly.Load` is enough for .net core and full framework. 
For full framework, it works even if instead of referencing a library, I copy the dll to the bin directory.

wojtpl2:
As @eerhardt wrote, BDN doesn't need DependencyContext at all. I tested this solution for .net core and full framework.

adamsitnik:
@wojtpl2, unfortunately, it needs it. The difference is when you work with our local project built on your PC (when everything is in bin) vs BDN installed via NuGet (when BDN is in NuGet folder sth like C:\Users\adsitnik\.nuget\packages)

The problem that we try to solve is the following:

1. User installs `BenchmarkDotNet.Diagnostics.Windows` NuGet package but does not consume any of the types defined in this package.
2. User uses `BenchmarkSwitcher` and passes args to it
3. User provides `--profiler ETW` command line argument, the way we use `Assembly.Load` today fails from loading `BenchmarkDotNet.Diagnostics.Windows` because it's not in the bin folder. Here is where we need to use `DependencyContext`

Repro:

```log
dotnet run -c Release -- --filter * --profiler ETW
```

```cs
class Program
{
}

public class Sample
{
    [Benchmark] public void TheMethod() { }  
}
```

Reproducing this issue is not easy. The best way to do it know to me:

1. Use dotnet publish and publish a BenchmarkDotNet NuGet packages
2. Copy the packages to a folder
3. Create new console app (dotnet new console)
4. Create new nuget.config file (dotnet new nugetconfig), add the folder where you store the packages as a feed to the list in this nuget.config file
5. Install BDN from the local feed
6. Run the sample I provided above

Once you are able to repro the problem locally and provide a fix you need to remove the local NuGet from `C:\Users\$username\.nuget\packages` and reinstall the new version. If you try to install a NuGet package with same version, but different content NuGet client won't update the things stored in C:\Users\$username\.nuget\packages

wojtpl2:
I've reproducing it in different way:
I have a project:
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>net4.7.2;netcoreapp2.1</TargetFrameworks>
  </PropertyGroup>

  <ItemGroup>
    <!-- <PackageReference Include="BenchmarkDotNet" Version="0.11.2" />  -->
    <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.11.2" />
    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.26" />
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="..\src\BenchmarkDotNet\BenchmarkDotNet.csproj" />
  </ItemGroup>

</Project>

```

and program.cs without consuming type from `BenchmarkDotNet.Diagnostics.Windows`:
```c#
using System;
using System.Security.Cryptography;

namespace ConsoleApp1
{
   
    public class Md5VsSha256
    {
        private const int N = 10000;
        private readonly byte[] data;

        private readonly SHA256 sha256 = SHA256.Create();
        private readonly MD5 md5 = MD5.Create();

        public Md5VsSha256()
        {
            data = new byte[N];
            new Random(42).NextBytes(data);
        }

        [Benchmark]
        public byte[] Sha256() => sha256.ComputeHash(data);

        [Benchmark]
        public byte[] Md5() => md5.ComputeHash(data);
    }
    class Program
    {
        {
            var switcher = BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);
        }
    }
}
```

After build my release folder looks like:
image

I was running test by:
```
dotnet ConsoleApp1.dll --filter *Md5* --profiler ETW
```

Before my changes I got exception. But of course I can try your path. I'll let you know.

wojtpl2:
I check your path, and I have to say that it works. `DependencyContext` doesn't load assambly. It only read `*.deps.json` file. 

@eerhardt explained exactly why it works.:
> `DependencyContext` allows you to get the information about a dependency, but it doesn't have assembly loading capabilities. You still use the normal `Assembly.Load` APIs to load the assembly.
> 
> The reason I suggested using `DependencyContext` is because I assumed the main thing you need/want here is you want to get the AssemblyVersion of the referenced `BenchmarkDotNet.Diagnostics.Windows.dll`, right? (That information can be found [on this API](https://github.com/dotnet/core-setup/blob/bb31dc3b52e38ca4e679f741cd0263dc239b926f/src/managed/Microsoft.Extensions.DependencyModel/RuntimeFile.cs#L24)).
> 
> If it works to just `Assembly.Load(new AssemblyName(DiagnosticAssemblyName));`, then you don't need the `DependencyContext` nor the bit about `Assembly.GetEntryAssembly()?.GetReferencedAssemblies()` at all. Just call `Assembly.Load` passing in the .dll name.

wojtpl2:
Without changes I got exception:
```
c:\Work\Test_955
λ dotnet run -c Release -- --filter * --profiler ETW
Error loading BenchmarkDotNet.Diagnostics.Windows.dll: FileNotFoundException - Could not load file or assembly 'c:\Work\Test_955\BenchmarkDotNet.Diagnostics.Windows.dll'. The system cannot find the file specified.
// Validating benchmarks:
Unable to resolve IProfiler diagnoser using dynamic assembly loading.
            Please make sure that you have installed the latest BenchmarkDotNet.Diagnostics.Windows package.
If you are using `dotnet build` you also need to consume one of its public types to make sure that MSBuild copies it to the output directory. The alternative is to use `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>` in your project file.
```
After changes:
```
c:\Work\Test_955
dotnet run -c Release -- --filter * --profiler ETW
// Validating benchmarks:
// ***** BenchmarkRunner: Start   *****
```

adamsitnik:
@wojtpl2 Great!

btw I am an idiot. There is much simpler way of testing it: just create new project, install BDN packages and copy-paste the code for loading the dlls to the new app and run it.

wojtpl2:
You are not an idiot ;) Thanks to you I started my journey with open source. 
btw. what task can I take? Maybe I can help you in #912? It is interesting me. Or I can do something else.

eerhardt:
Thanks, @wojtpl2, for the great work here! Fixing this issue will help quite a few customers, especially ML.NET, where this issue was [spawned from](https://github.com/dotnet/machinelearning/pull/1599/files#r232782005).

adamsitnik:
@wojtpl2 I just tested it myself and everything works fine. Once again thanks for help!

PR #960, opened 2018-11-15
Expose StatisticalTestColumn via command line arguments

adamsitnik:
@AndreyAkinshin I took the Statistical Tests you implemented and created a simple analyzer on top of them. It helps me a lot for new .NET framework releases.


src/BenchmarkDotNet/Analysers/RegressionAnalyser.cs
```diff
@@ -0,0 +1,48 @@
...
+                    continue;
+
+                var baseline = benchmarks.Single(benchmark => benchmark.Job.Meta.Baseline);
+                var diff = benchmarks.Single(benchmark => !benchmark.Job.Meta.Baseline);
+                
+                var x = summary[baseline].ResultStatistics.GetOriginalValues().ToArray();
+                var y = summary[diff].ResultStatistics.GetOriginalValues().ToArray();
+
+                var mannWhitneyTestResult = StatisticalTestHelper.CalculateTost(MannWhitneyTest.Instance, x, y, Threshold);
+                var welchTestResult = StatisticalTestHelper.CalculateTost(WelchTest.Instance, x, y, Threshold);
```
AndreyAkinshin: I don't think that it's a good idea to use Welch's t-test in general case because it's work fine only for distributions which are close to normal. I suggest using only Mann-Whitney.

src/BenchmarkDotNet/Analysers/RegressionAnalyser.cs
```diff
@@ -0,0 +1,48 @@
...
+
+                var conclusions = new[] { welchTestResult.Conclusion, mannWhitneyTestResult.Conclusion };
+                
+                if (conclusions.All(conclusion => conclusion == EquivalenceTestConclusion.Same) || conclusions.All(conclusion => conclusion == EquivalenceTestConclusion.Base))
```
AndreyAkinshin: Do we really need a conclusion for _each_ pair of benchmarks? If we really need it, it's better to introduce additional column. If we care only about actual regressions, it makes sense to print only "Slower" results.

tests/BenchmarkDotNet.Tests/Analysers/RegressionAnalyserTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            var conclusion = sut.Analyse(summary).Single();
+            
+            Assert.Equal(ConclusionKind.Hint, conclusion.Kind);
+            Assert.Contains("Same", conclusion.Message);
+        }
+        
+        [Fact]
+        public void RegressionsAreDetected()
+        {
+            var @base = new[] { 10.0, 10.01, 10.02, 10.0, 10.03, 10.02, 9.99, 9.98, 10.0, 10.02 };
```
AndreyAkinshin: It's better to use `baseline` instead of `@base`.

src/BenchmarkDotNet/Analysers/RegressionAnalyser.cs
```diff
@@ -0,0 +1,48 @@
...
+        public override string Id => "Regression";
+
+        private Threshold Threshold { get; }
+
+        protected override IEnumerable<Conclusion> AnalyseSummary(Summary summary)
+        {
+            foreach (var sameArguments in summary.BenchmarksCases.GroupBy(benchmark => (benchmark.Descriptor, benchmark.Parameters)))
+            {
+                var benchmarks = sameArguments.ToArray();
+                if (benchmarks.Length != 2)
```
AndreyAkinshin: Do we really need this limitation? What if I want to compare the baseline against several other jobs?

adamsitnik:
@AndreyAkinshin please take one more look. I decided to remove the RegressionAnalyser and just reuse existing StatisticalTestColumn. I added some tests, removed one bug (same input for Welch was causing stack overflow exception), exposed it via console line arugments and added some docs for the args.

adamsitnik:
Sample:

```cs
class Program
{
}

public class MySample
{
    [Benchmark]
    public void Sleep()
    {
#if NETFRAMEWORK
        Thread.Sleep(50);
#elif NETCOREAPP2_0
        Thread.Sleep(45);
#elif NETCOREAPP2_1
        Thread.Sleep(55);
#endif
    }
    
    [Benchmark]
    public void Same()
    {
        Thread.Sleep(50);
    }
}
```

```log
dotnet run -c Release -f netcoreapp2.1 --filter *MySample* --runtimes net46 netcoreapp2.0 netcoreapp2.1 --statisticalTest MannWhitney --thresholdUnit Ratio --thresholdValue 0.01
```

| Method | Runtime |     Toolchain |     Mean |     Error |    StdDev | Ratio | MannWhitney(1%) |
|------- |-------- |-------------- |---------:|----------:|----------:|------:|---------------- |
|  Sleep |     Clr |         net46 | 50.51 ms | 0.1833 ms | 0.1714 ms |  1.00 |            Base |
|  Sleep |    Core | netcoreapp2.0 | 45.53 ms | 0.1262 ms | 0.1181 ms |  0.90 |          Faster |
|  Sleep |    Core | netcoreapp2.1 | 55.50 ms | 0.1217 ms | 0.1138 ms |  1.10 |          Slower |
|        |         |               |          |           |           |       |                 |
|   Same |     Clr |         net46 | 50.47 ms | 0.1795 ms | 0.1679 ms |  1.00 |            Base |
|   Same |    Core | netcoreapp2.0 | 50.55 ms | 0.1873 ms | 0.1752 ms |  1.00 |            Same |
|   Same |    Core | netcoreapp2.1 | 50.55 ms | 0.2162 ms | 0.2022 ms |  1.00 |            Same |

AndreyAkinshin:
I don't like that we force users to specify too many parameters.
1. We should use default values for thresholds (users should be able to skip these parameters)
2. `--thresholdUnit Ratio --thresholdValue 0.01` looks too clumsy. From the practical point of view, it will be better to write something like `--threshold 1%` or `--threshold 5ms` (and parse it).
3. We shouldn't force users to type "MannWhitney" each time, it should be the implicit default value.

adamsitnik:
@AndreyAkinshin thanks for the feedback! fixed!

as soon as the tests pass you can squash it and release 0.11.3

AndreyAkinshin:
@adamsitnik, now it looks much better, but I'm still not happy about the API.
It's not obvious that `--threshold 5%` activates the statistical test. Also, it's not obvious how to run the statistical test with the default threshold.

adamsitnik:
> Also, it's not obvious how to run the statistical test with the default threshold.

I would prefer not to create any default threshold. What should the value be? 1%? 2%? 5%? 10%? IMHO it's better to leave it up to the users to decide.

> It's not obvious that --threshold 5% activates the statistical test. 

So maybe I should just call it `--statisticalTest`  instead and always use MannWhitney if it's better?

So it would be just  `--statisticalTest 5%`  to turn it on?

adamsitnik:
The test  failure is not related to my last commit, it must have been a .NET Framework change/update on AppVeyor server. I can't repo it locally.

@AndreyAkinshin I am going to merge it now to be able to push release notes

PR #966, opened 2018-11-18
Fix typos in ConfigParser and CommandLineOptions

morgan-kn:


adamsitnik:
@morgan-kn thank you!

PR #969, opened 2018-11-21
Disable batch mode and explicitly enable build trigger for master branch

Ky7m:
It seems to me that we have to define branch for master builds explicitly.


adamsitnik:
@Ky7m thanks!

Issue #553, opened 2017-09-23
Suggestion: New BDN run mode that causes cache misses between iterations


damageboy:
I got this "idea" from https://probablydance.com/2017/02/26/i-wrote-the-fastest-hashtable/

The basic premise of what @skarupke did is to get more realistic benchmark results by creating many thousands of identical test objects (dictionaries in that case) and call the benchmarked code on different instances on every invocation.

In this way every known cache has its worst performance exhibited (L1..3/TLB) because no cache-locality is enforced. A nifty aspect of this approach is that is required no privileged code to run in order to invalidate the caches, but it rather overloads them.

I think this is an interesting concept to bring in the BDN framework, as it allows users to get a better sense of how their code will behave in the real world where its not being run in a tight loop and everything comfortably fits into L3/L2/L1 caches.

As a framework feature I think this would require a decent size of work:
1. Somehow track all the *amount* of memory allocation performed by the tested code, in order to aggregate it into "bytes allocated per tested object"
2. allocate as many object that are required to saturate L1/L2/L3/TLB caches given the previously collected aggregation
3. run each test iteration on a different instance in sequence
4. Profit!

adamsitnik:
@damageboy thanks for the idea, I really like it! The referenced blog post itself is really awesome!

@AndreyAkinshin what do you think? 

I just wonder what would be the  best way of invalidating the cache without any manged allocations (to keep the rest of our code like MemoryDiagnoser working)

btw today it could be achieved by using the `[InterationSetup]` attribute ([docs](http://benchmarkdotnet.org/Advanced/SetupAndCleanup.htm#iterationsetup))

AndreyAkinshin:
I also like this idea. I don't think that we have to introduce another run strategy, but we can introduce another bool characteristic (like `ColdCpuCache` or `InvalidateCpuCahe`). It doesn't make sense for throughput mode and microbenchmarks, but it can be very useful for ColdStart/Monitoring + macrobenchmarks.
And we don't need managed allocations for this feature, it will be enough to preallocate an array (e.g. same size as L3), and just "touch" all the elements several times (in fact, we don't need *all* elements, it will be enough to "touch" one byte from each 64B cache line).
Another tricky thing here is that each physical core typically has own L1/L2 cache. So, *probably* it's a good idea to create several threads with specified processor affinity, each thread will invalidate own caches.

damageboy:
[...]

PR #971, opened 2018-11-22
Cache clearing strategies - Issue #553

wojtpl2:
PR for #553 
@AndreyAkinshin I'm not sure of this solution. I've changed `Allocations` strategy. Now this strategy should work good for `Job` with and without `--affinity` parameter.

#### None
```
dotnet BenchmarkDotNet.Samples.dll --filter *IntroColdCpuCache*
```
```
              Method |     Mean |    Error |   StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
-------------------- |---------:|---------:|---------:|---------:|---------:|---------:|---------:|---------:|
 AllocationsStrategy | 701.3 us | 5.492 us | 5.137 us | 697.2 us | 699.5 us | 705.2 us |   0.6767 |    2.385 |
```
#### Native
```
dotnet BenchmarkDotNet.Samples.dll --filter *IntroColdCpuCache* --CacheClearingStrategy Native
```
```
              Method |     Mean |    Error |   StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
-------------------- |---------:|---------:|---------:|---------:|---------:|---------:|---------:|---------:|
 AllocationsStrategy | 697.4 us | 1.767 us | 1.379 us | 696.4 us | 697.1 us | 698.6 us |   0.4028 |    1.650 |
```
#### Allocations
```
dotnet BenchmarkDotNet.Samples.dll --filter *IntroColdCpuCache* --CacheClearingStrategy Allocations
```
```
              Method |     Mean |    Error |   StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
-------------------- |---------:|---------:|---------:|---------:|---------:|---------:|---------:|---------:|
 AllocationsStrategy | 700.8 us | 3.367 us | 3.149 us | 698.3 us | 700.1 us | 703.0 us |   0.1887 |    1.833 |
```

## Affinity 1
Result for affinity set to 1.
#### None
```
dotnet BenchmarkDotNet.Samples.dll --filter *IntroColdCpuCache* --affinity 1
```
```
              Method |     Mean |    Error |   StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
-------------------- |---------:|---------:|---------:|---------:|---------:|---------:|---------:|---------:|
 AllocationsStrategy | 714.3 us | 3.988 us | 3.731 us | 711.5 us | 713.6 us | 716.8 us |   0.3048 |    2.087 |
```
#### Native 
```
dotnet BenchmarkDotNet.Samples.dll --filter *IntroColdCpuCache* --CacheClearingStrategy Native --affinity 1
```
```
              Method |     Mean |    Error |   StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
-------------------- |---------:|---------:|---------:|---------:|---------:|---------:|---------:|---------:|
 AllocationsStrategy | 709.1 us | 2.537 us | 2.249 us | 707.2 us | 708.7 us | 711.1 us |  -0.0152 |    1.580 |
```
#### Allocations
```
dotnet BenchmarkDotNet.Samples.dll --filter *IntroColdCpuCache* --CacheClearingStrategy Allocations --affinity 1
```
```
              Method |     Mean |    Error |   StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
-------------------- |---------:|---------:|---------:|---------:|---------:|---------:|---------:|---------:|
 AllocationsStrategy | 708.7 us | 3.439 us | 3.217 us | 707.1 us | 709.0 us | 711.1 us |  -0.2776 |    1.867 |
```


src/BenchmarkDotNet/Engines/CacheClearingStrategies/AllocationsCacheClearingStrategy.cs
```diff
@@ -0,0 +1,166 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.ConstrainedExecution;
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Threading;
+using Microsoft.Win32.SafeHandles;
+
+namespace BenchmarkDotNet.Engines.CacheClearingStrategies
+{
+    internal class AllocationsCacheClearingStrategy : ICacheClearingStrategy
```
adamsitnik: I would rename this type to mention OS, sth like AllocationsCacheClearingStrategy**ForWindows**

src/BenchmarkDotNet/Engines/CacheClearingStrategies/AllocationsCacheClearingStrategy.cs
```diff
@@ -0,0 +1,166 @@
...
+                foreach (int affinity in affinities)
+                {
+                    var thread = new Thread(() => AllocateMemory(affinity)) { IsBackground = true, };
+                    thread.Start();
+
+                    threads[index++] = thread;
+                }
+            }
+
+            //Wait for all thread
```
adamsitnik: nit: java style comment (obvious comment)

```suggestion
```

src/BenchmarkDotNet/Engines/CacheClearingStrategies/AllocationsCacheClearingStrategy.cs
```diff
@@ -0,0 +1,166 @@
...
+        [ReliabilityContract(Consistency.WillNotCorruptState, Cer.Success), DllImport("kernel32.dll", CharSet = CharSet.Auto, SetLastError = true, ExactSpelling = true)]
+        private static extern bool CloseHandle(IntPtr handle);
+
+        [DllImport("kernel32")]
+        private static extern int GetCurrentThreadId();
+
+        [DllImport("kernel32.dll", CharSet = CharSet.Auto, SetLastError = true)]
+        private static extern SafeThreadHandle OpenThread(int access, bool inherit, int threadId);
+    }
+}
```
adamsitnik: could you please:

1. move `extern`s to the top of the file to make it clear that this type is using WinAPI
2. order the private methods by the importance? I would like to see the methods that do the most important things at the top, the helpers like `GetAffinitiesForSelectedProcessors` at the bottom. Just to get good overview what this code is doing quickly
3. add `<summary>` comments to the methods that perform actual cache cleanup? just two sentences what these methods do and why
4. in the comments add a link to the website/book which describeds how to do it (I assume you have used one)

src/BenchmarkDotNet/Engines/CacheClearingStrategies/CacheClearingStrategiesFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+                case CacheClearingStrategy.Allocations:
+                    return new AllocationsCacheClearingStrategy();
+                case CacheClearingStrategy.Native:
+                    return new NativeCacheClearingStrategy();
+                default:
+                    throw new NotSupportedException($"Not supported cache clearing strategy: {cacheClearingStrategy}.");
+            }
+        }
+    }
+}
```
adamsitnik: personally, I don't like the idea of mixing if and switch like that, what do you say about:

```cs
if (!RuntimeInformation.IsWindows() || cacheClearingStrategy == CacheClearingStrategy.None)
    return null;
else if (cacheClearingStrategy == CacheClearingStrategy.Allocations)
    return new AllocationsCacheClearingStrategy();
else if (cacheClearingStrategy == CacheClearingStrategy.Native:
    return new NativeCacheClearingStrategy();
else
    throw new NotSupportedException($"Not supported cache clearing strategy: {cacheClearingStrategy}.");
```

moreover, since you have introduced `ICacheClearingStrategy` you could go fully "object oriented" and instead of returning `null` use "null object pattern"

src/BenchmarkDotNet/Engines/Engine.cs
```diff
@@ -43,13 +44,14 @@ public class Engine : IEngine
...
         private readonly bool includeMemoryStats;
+        private readonly ICacheClearingStrategy cacheClearingStrategy;
 
         internal Engine(
             IHost host,
             IResolver resolver,
             Action dummy1Action, Action dummy2Action, Action dummy3Action, Action<long> overheadAction, Action<long> workloadAction, Job targetJob,
             Action globalSetupAction, Action globalCleanupAction, Action iterationSetupAction, Action iterationCleanupAction, long operationsPerInvoke,
-            bool includeMemoryStats, Encoding encoding, string benchmarkName)
+            bool includeMemoryStats, Encoding encoding, string benchmarkName, CacheClearingStrategy cacheClearingStrategy)
```
adamsitnik: is something can be created using `targetJob.ResolveValue` you should not add a new parameter to the ctor and  edit the codegen but rather resolve the value as we do for example for `Clock = targetJob.ResolveValue(InfrastructureMode.ClockCharacteristic, Resolver);`

src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
```diff
@@ -145,7 +145,10 @@ public class CommandLineOptions
         
         [Option("statisticalTest", Required = false, HelpText = "Threshold for Mann–Whitney U Test. Examples: 5%, 10ms, 100ns, 1s")]
         public string StatisticalTestThreshold { get; set; }
-        
+
+        [Option("cacheClearingStrategy", Required = false, Default = CacheClearingStrategy.None, HelpText = "Cache clearing strategy.")]
```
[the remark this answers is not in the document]
adamsitnik: And please add a validation rule and return error when somebody tries it on non supported OS

src/BenchmarkDotNet/Engines/Engine.cs
```diff
@@ -138,6 +141,8 @@ public Measurement RunIteration(IterationData data)
             if(!isOverhead)
                 IterationSetupAction();
 
+            ClearCache();
```
adamsitnik: please add a comment explaining why it's before  `GcCollect` (I assume that because one of the strategies can allocate managed memory?)

src/BenchmarkDotNet/Engines/Engine.cs
```diff
@@ -163,6 +168,17 @@ public Measurement RunIteration(IterationData data)
...
+        {
+            IntPtr? affinity = null;
+            if (TargetJob.Job.Environment.HasValue(EnvironmentMode.AffinityCharacteristic))
+            {
+                affinity = TargetJob.Job.Environment.Affinity;
+            }
+
+            cacheClearingStrategy?.ClearCache(affinity);
+        }
+
```
adamsitnik: can be simplified to:

```cs
private void ClearCache()
    => TargetJob.Job.Environment.HasValue(EnvironmentMode.AffinityCharacteristic)
        ? cacheClearingStrategy?.ClearCache(TargetJob.Job.Environment.Affinity)
        : cacheClearingStrategy?.ClearCache(null);

```

src/BenchmarkDotNet/Configs/IConfig.cs
```diff
@@ -55,5 +56,7 @@ public interface IConfig
         IEnumerable<BenchmarkLogicalGroupRule> GetLogicalGroupRules();
 
         bool StopOnFirstError { get; }
+
+        CacheClearingStrategy CacheClearingStrategy { get; }
```
adamsitnik: important:: I think that this setting is very similar to Affinity and it would be nice to compare the results for the same benchmark with two different jobs.

So please move it from `IConfig` to `CacheClearingStrategyCharacteristic` in `EnvironmentMode`

https://github.com/dotnet/BenchmarkDotNet/blob/a76f438a81e3e0235a2e9b1249a5280038189689/src/BenchmarkDotNet/Jobs/EnvironmentMode.cs#L14

also, please don't forget to register a default value for it in the resolver

wojtpl2:
@adamsitnik Thanks for review. I'll fix it. But I have question. Have you tested it or just read it?

Now I see that when I run 
```ini
dotnet BenchmarkDotNet.Samples.dll --filter *IntroColdCpuCache* --CacheClearingStrategy Allocations --affinity 1
```

I don't need to use [SetProcessorAffinity](https://github.com/dotnet/BenchmarkDotNet/pull/971/files#diff-d03f5523c2651e80c1e0654b4067bf05R91 ) on the threads. The process has affinity set to 1 and all threads are run on the same processor. But I don't yet if it is the law or just accident. If it is normal behevior, it could work on linux with `--affinity X`. (where X equals 1, 2, 4, 8, 16 ..)

adamsitnik:
@wojtpl2 so far I did not test it myself

You can add an integration test for that and see which OS is going to scream ;)

src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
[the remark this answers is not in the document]
wojtpl2: Or maybe "cleaningCache". "cache" is too general. You may need it in the future for something else?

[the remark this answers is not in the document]
adamsitnik: @wojtpl2 sounds good to me! 👍

src/BenchmarkDotNet/Engines/Engine.cs
wojtpl2: I wasn't aware of it. I wanted to do it after `IterationSetupAction()`. I'll add message.

wojtpl2:
@adamsitnik I made `CacheClearingStrategyCharacterisitcs` and now I can compere results for different strategies.
```ini
   Method | Affinity | CacheClearingStrategy |     Mean |     Error |    StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
--------- |--------- |---------------------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|---------:|
 ArraySum | 00000001 |           Allocations | 632.4 us |  5.174 us |  4.039 us | 630.0 us | 631.7 us | 633.4 us |   0.7022 |    2.408 |
 ArraySum | 00000001 |                Native | 635.0 us |  4.159 us |  3.687 us | 632.0 us | 634.9 us | 637.4 us |   0.3465 |    1.710 |
 ArraySum | 00000001 |                  None | 636.3 us |  6.696 us |  5.591 us | 631.3 us | 634.7 us | 642.5 us |   0.2689 |    1.476 |
 ArraySum | 11111111 |           Allocations | 658.9 us | 13.938 us | 27.186 us | 637.0 us | 650.4 us | 672.1 us |   0.9307 |    2.965 |
 ArraySum | 11111111 |                Native | 638.6 us |  8.604 us |  7.627 us | 633.6 us | 636.9 us | 643.2 us |   0.6121 |    2.443 |
 ArraySum | 11111111 |                  None | 639.0 us | 11.414 us | 10.118 us | 633.0 us | 638.2 us | 641.6 us |   1.0159 |    3.990 |
```

I don't know where I should keep validation of CacheClearingStrategyCharacterisitcs.
I didn't make any integration tests and unit test. Should I add tests that work for all systems with
```c#
if (!RuntimeInformation.IsWindows())
```

src/BenchmarkDotNet/Engines/CacheClearingStrategies/NativeCacheClearingStrategy.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace BenchmarkDotNet.Engines.CacheClearingStrategies
+{
+    internal class NativeCacheClearingStrategy : ICacheClearingStrategy
+    {
+        [DllImport("kernel32.dll")]
+        private static extern bool FlushInstructionCache(IntPtr hProcess, IntPtr lpBaseAddress, UIntPtr dwSize);
```
AndreyAkinshin: Am I understand correctly that `FlushInstructionCache` flushes only instruction cache instead of CPU L1/L2/L3 chaches? In this case, we don't need this kind of strategy.
P.S. See https://blogs.msdn.microsoft.com/oldnewthing/20031208-00/?p=41583

samples/BenchmarkDotNet.Samples/IntroColdCpuCache.cs
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+
+        private readonly int[] array = Enumerable.Range(1, 1024 * 1024).ToArray();
+
+        [Benchmark]
+        public int ArraySum()
+        {
+            int result = 0;
+            for (int i = 0; i < array.Length; i++)
```
AndreyAkinshin: It's better to use `i += 16` instead of `i++` (to hit each 64B cache line only once).

src/BenchmarkDotNet/Engines/Engine.cs
```diff
@@ -138,6 +146,9 @@ public Measurement RunIteration(IterationData data)
             if(!isOverhead)
                 IterationSetupAction();
 
+            // this line should be before GcCollect because some strategies can allocate managed memory
+            cacheClearingStrategy?.ClearCache(affinity);
```
AndreyAkinshin: It doesn't make sense to run it for all kind of iterations (including Jitting; Overhead). Only workload iterations should be affected.

src/BenchmarkDotNet/Engines/CacheClearingStrategies/MemoryAllocator.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public void AllocateMemory()
+        {
+            const int howManyMb = 14;
+            const int howManyPass = 10;
+            const int sizeOfOneArray = 64;
+
+            for (int i = 0; i < (howManyMb * 1024 * 1024 / sizeOfOneArray) * howManyPass; i++)
+            {
+                var tmpArray = new int[sizeOfOneArray];
```
AndreyAkinshin: Do we really have to allocate memory before each iteration? I guess it will be enough to preallocate a huge array in advance and reuse it.

wojtpl2: @AndreyAkinshin Thanks for review! What do you mean by "preallocate a huge array"? Should I create a private array:
```c#
private readonly int[] memory = new int[howManyMb* 1024 * 1024]
```
And then use it instead of allocate a new one?

src/BenchmarkDotNet/Engines/Engine.cs
wojtpl2: I've changed it to:
```c#
            if (data.IterationMode == IterationMode.Workload && data.IterationStage == IterationStage.Actual)
            {
                // this line should be before GcCollect because some strategies can allocate managed memory
                cacheClearingStrategy?.ClearCache(affinity);
            }
```

src/BenchmarkDotNet/Engines/CacheClearingStrategies/MemoryAllocator.cs
AndreyAkinshin: Yes, we should create this array only once.

wojtpl2: I've pushed changes.

src/BenchmarkDotNet/Engines/CacheClearingStrategies/MemoryAllocator.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Linq;
+using System.Runtime.CompilerServices;
+
+namespace BenchmarkDotNet.Engines.CacheClearingStrategies
+{
+    internal class MemoryAllocator : IMemoryAllocator
+    {
```
AndreyAkinshin: Since it doesn't allocate memory anymore, it should be renamed.

src/BenchmarkDotNet/Engines/CacheClearingStrategies/MemoryAllocator.cs
```diff
@@ -0,0 +1,26 @@
...
+        private const int HowManyMb = 14;
+        private const int SizeOfArray = HowManyMb * 1024 * 1024;
+        private readonly int[] memory = Enumerable.Range(1, SizeOfArray).ToArray();
```
AndreyAkinshin: We don't need LINQ here; it's enough to write `new int[SizeOfArray]`.

src/BenchmarkDotNet/Engines/CacheClearingStrategies/MemoryAllocator.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public void AllocateMemory()
```
AndreyAkinshin: The loop should be repeated several times to clear the cache.

src/BenchmarkDotNet/Engines/CacheClearingStrategies/MemoryAllocator.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            for (int i = 0; i < SizeOfArray; i++)
```
AndreyAkinshin: We can use `i += 16` instead of `i++`, it will work faster with the same effect.

src/BenchmarkDotNet/Engines/CacheClearingStrategies/MemoryAllocator.cs
```diff
@@ -0,0 +1,26 @@
...
+            {
+                Consumer(memory[i]);
```
AndreyAkinshin: Let's try to do `memory[i]++` (we will get read/write operatins instead of a single read operation).

src/BenchmarkDotNet/Engines/Engine.cs
```diff
@@ -135,9 +143,15 @@ public Measurement RunIteration(IterationData data)
             bool isOverhead = data.IterationMode == IterationMode.Overhead;
             var action = isOverhead ? OverheadAction : WorkloadAction;
 
-            if(!isOverhead)
+            if (!isOverhead)
                 IterationSetupAction();
 
+            if (data.IterationMode == IterationMode.Workload && data.IterationStage == IterationStage.Actual)
```
AndreyAkinshin: It's better to use `IsWorkload()` extension method here (and introduce a similar method for the Actual Stage).

wojtpl2:
After all changes you can see difference in result:

```ini
   Method | Affinity | CacheClearingStrategy |     Mean |     Error |    StdDev |       Q1 |   Median |       Q3 | Skewness | Kurtosis |
--------- |--------- |---------------------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|---------:|
 ArraySum | 00000001 |           Allocations | 4.065 ms | 0.1101 ms | 0.1224 ms | 3.985 ms | 4.005 ms | 4.132 ms |   1.1943 |    2.884 |
 ArraySum | 00000001 |                  None | 3.978 ms | 0.0319 ms | 0.0299 ms | 3.950 ms | 3.982 ms | 3.999 ms |  -0.1554 |    1.800 |
 ArraySum | 11111111 |           Allocations | 4.096 ms | 0.0794 ms | 0.1087 ms | 4.029 ms | 4.067 ms | 4.110 ms |   0.9610 |    3.009 |
 ArraySum | 11111111 |                  None | 4.044 ms | 0.0503 ms | 0.0471 ms | 4.032 ms | 4.057 ms | 4.078 ms |  -1.1839 |    3.260 |
```

@AndreyAkinshin I used IsWorkload method, but it could be breaking changes. This method was in `BenchmarkDotNet.Reports.MeasurementExtensions`. I needed this method for `BenchmarkDotNet.Engines.IterationData`. To achieve it I introduced IIterationInfo and made extension method to this interface, but namespace was changed. If you want, I can revert this changes.

src/BenchmarkDotNet/Engines/Engine.cs
wojtpl2: I've used this method, but please see the comment below.

src/BenchmarkDotNet/Engines/CacheClearingStrategies/CacheMemoryCleaner.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace BenchmarkDotNet.Engines.CacheClearingStrategies
+{
+    /// <summary>
+    /// This class clean CPU cache. To archive it,
```
adamsitnik: nit: the comment is not complete?

src/BenchmarkDotNet/Engines/IterationInfoExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Engines
+{
+    public static class IterationInfoExtensions
+    {
+        [PublicAPI] public static bool IsOverhead(this IIterationInfo iterationInfo) => iterationInfo.IterationMode == IterationMode.Overhead;
```
adamsitnik: when calling an interface method on a struct we introduce boxing

to avoid boxing, we need to make this method generic (however I am not 100% sure if an extension method can be generic):

```cs
public static bool IsOverhead<T>(this T iterationInfo) where T : IIterationInfo => iterationInfo.IterationMode == IterationMode.Overhead;
```

we try to avoid boxing at all cost in Engine, mostly to make sure that Engine does not allocate any managed memory which could spoil the results

more: https://adamsitnik.com/Value-Types-vs-Reference-Types/#boxing

AndreyAkinshin:
@wojtpl2 I made a lot of experiments on different hardware, but I can't get a statistically significant difference between cache clearing strategies. =( Currently, I don't know what should we improve to resolve this problem. We need a proper IntroSample which shows the `CachEClearingStrategy` effect before we merge this branch.

samples/BenchmarkDotNet.Samples/IntroColdCpuCache.cs
```diff
@@ -0,0 +1,45 @@
...
+        private class CompereCacheClearingStrategies : ManualConfig
+        {
+            public CompereCacheClearingStrategies()
+            {
+                Add(Job.Default.WithCacheClearingStrategy(CacheClearingStrategy.Allocations));
+                Add(Job.Default.WithCacheClearingStrategy(CacheClearingStrategy.None));
+                Add(Job.Default.WithAffinity((IntPtr) 1).WithCacheClearingStrategy(CacheClearingStrategy.Allocations));
+                Add(Job.Default.WithAffinity((IntPtr) 1).WithCacheClearingStrategy(CacheClearingStrategy.None));
+
+                Add(StatisticColumn.Q1,
```
AndreyAkinshin: We don't actually need Q1/Median/Q3/Skewness/Kurtosis in this sample. Instead of this, we need StatisticalTestColumn here.

src/BenchmarkDotNet/Engines/IterationInfoExtensions.cs
wojtpl2: Yes Extension method can be generic. Before changes: 
image
After changes:
image

wojtpl2: Thanks for this suggestion!

src/BenchmarkDotNet/Engines/CacheClearingStrategies/AllocationsCacheClearingStrategyForWindows.cs
```diff
@@ -0,0 +1,121 @@
...
+        [SuppressUnmanagedCodeSecurity]
+        private class SafeThreadHandle : SafeHandleZeroOrMinusOneIsInvalid
+        {
+            public SafeThreadHandle() : base(true) { }
+
+            protected override bool ReleaseHandle() => CloseHandle(handle);
+        }
+
+        [ReliabilityContract(Consistency.WillNotCorruptState, Cer.Success)]
+        [DllImport("kernel32.dll", CharSet = CharSet.Auto, SetLastError = true, ExactSpelling = true)]
```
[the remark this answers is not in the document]
wojtpl2: In this place we can't use `ProcessThread.ProcessorAffinity` because I create normal `Thread`. `Thread` doesn't have `ProcessorAffinity` option.

samples/BenchmarkDotNet.Samples/IntroColdCpuCache.cs
wojtpl2: I've added: `[StatisticalTestColumn(StatisticalTestKind.MannWhitney, ThresholdUnit.Microseconds, 0.2, true)]`. I'm not sure if it is good.

wojtpl2:
@AndreyAkinshin I can't prepare `IntroSample` that shows significant difference between cache clearing strategies. 

I added `Add(HardwareCounter.CacheMisses);` and this column shows that `CacheClearingStrategy.Allocations` works but it does not affect the test time. 

```ini
                         Method | Affinity | CacheClearingStrategy |       Mean |     Error |    StdDev | Ratio | MannWhitney(200ns)/p-values | RatioSD | CacheMisses/Op |
------------------------------- |--------- |---------------------- |-----------:|----------:|----------:|------:|---------------------------- |--------:|---------------:|
                       ArraySum |     0001 |           Allocations |   3.938 ms | 0.0086 ms | 0.0072 ms |  1.01 |         Same: 0.1511|0.8600 |    0.02 |        165,186 |
                       ArraySum |     0001 |                  None |   3.946 ms | 0.0072 ms | 0.0064 ms |  1.01 |       Slower: 0.0345|0.9690 |    0.02 |        132,259 |
                       ArraySum |     1111 |           Allocations |   3.874 ms | 0.0570 ms | 0.0533 ms |  0.99 |       Faster: 0.9615|0.0424 |    0.02 |        191,387 |
                       ArraySum |     1111 |                  None |   3.920 ms | 0.0762 ms | 0.0676 ms |  1.00 |                   Base: ?|? |    0.00 |        122,825 |
                                |          |                       |            |           |           |       |                             |         |                |
                      Accessing |     0001 |           Allocations |   5.837 ms | 0.0242 ms | 0.0214 ms |  0.99 |       Faster: 0.9997|0.0006 |    0.01 |        138,068 |
                      Accessing |     0001 |                  None |   5.972 ms | 0.0223 ms | 0.0186 ms |  1.01 |       Slower: 0.0003|0.9997 |    0.01 |        118,619 |
                      Accessing |     1111 |           Allocations |   5.822 ms | 0.0954 ms | 0.0892 ms |  0.99 |       Faster: 0.9922|0.0089 |    0.02 |        156,769 |
                      Accessing |     1111 |                  None |   5.903 ms | 0.0774 ms | 0.0686 ms |  1.00 |                   Base: ?|? |    0.00 |        114,303 |
                                |          |                       |            |           |           |       |                             |         |                |
    ArrayOfArrayAccessingByRows |     0001 |           Allocations |  36.021 ms | 0.0700 ms | 0.0621 ms |  0.99 |       Faster: 1.0000|0.0000 |    0.01 |        293,092 |
    ArrayOfArrayAccessingByRows |     0001 |                  None |  36.048 ms | 0.0613 ms | 0.0543 ms |  0.99 |       Faster: 0.9997|0.0004 |    0.01 |         60,698 |
    ArrayOfArrayAccessingByRows |     1111 |           Allocations |  36.368 ms | 0.2496 ms | 0.2335 ms |  1.00 |         Same: 0.2305|0.7819 |    0.01 |        488,099 |
    ArrayOfArrayAccessingByRows |     1111 |                  None |  36.296 ms | 0.2096 ms | 0.1960 ms |  1.00 |                   Base: ?|? |    0.00 |         60,508 |
                                |          |                       |            |           |           |       |                             |         |                |
 ArrayOfArrayAccessingByColumns |     0001 |           Allocations | 268.189 ms | 0.7375 ms | 0.6158 ms |  1.01 |       Slower: 0.0000|1.0000 |    0.01 |      3,934,660 |
 ArrayOfArrayAccessingByColumns |     0001 |                  None | 299.346 ms | 0.7359 ms | 0.6523 ms |  1.13 |       Slower: 0.0000|1.0000 |    0.01 |      2,062,363 |
 ArrayOfArrayAccessingByColumns |     1111 |           Allocations | 266.720 ms | 2.4447 ms | 1.9087 ms |  1.01 |       Slower: 0.0434|0.9616 |    0.01 |      5,903,726 |
 ArrayOfArrayAccessingByColumns |     1111 |                  None | 264.835 ms | 2.6314 ms | 2.1973 ms |  1.00 |                   Base: ?|? |    0.00 |      2,363,422 |

```

I was trying find similar solution for c++ but I couldn't find one. There is intresting thing: https://stackoverflow.com/questions/48527189/is-there-a-way-to-flush-the-entire-cpu-cache-related-to-a-program 

I'm afride that we will not able to prepare the right test. And this is blind alley.

wojtpl2:
How I run the test:
```
cd c:\Work\BenchmarkDotNet\samples\BenchmarkDotNet.Samples

dotnet run -c Release -f netcoreapp2.1 -- --filter *IntroColdCpuCache* 
```

adamsitnik:
> I'm afride that we will not able to prepare the right test. And this is blind alley.

Could you try with smaller IterationTime? maybe `--iterationTime 100` ? I think that we most probably perform too many invocations per iteration to make the difference visible

wojtpl2:
@adamsitnik `--iterationTime 100` doesn't help on my computer. But I have one more idea.

wojtpl2:
I've done many experiments and I think that clearing the cache does not work at all.

I've tryed different aproach here https://github.com/wojtpl2/BenchmarkDotNet/tree/issue_553_version_2 
Instead of use `Thread`, I made `BencharkDotNet.MemoryAllocator.dll` application that allocates memory. 
`AllocationsCacheClearingStrategyForWindows` doesn't do any allocation. It just run new `Process` (`dotnet BencharkDotNet.MemoryAllocator.dll`). 

I got result:
```
   Method | Affinity | CacheClearingStrategy |     Mean |     Error |    StdDev | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op | CacheMisses/Op |
--------- |--------- |---------------------- |---------:|----------:|----------:|------:|--------:|------------:|------------:|------------:|--------------------:|---------------:|
 ArraySum |     0001 |           Allocations | 4.051 ms | 2.9497 ms | 0.1617 ms |  0.97 |    0.05 |           - |           - |           - |                   - |        101,872 |
 ArraySum |     0001 |                  None | 4.187 ms | 0.1578 ms | 0.0087 ms |  1.00 |    0.02 |           - |           - |           - |                   - |         90,553 |
 ArraySum |     1111 |           Allocations | 4.042 ms | 2.0394 ms | 0.1118 ms |  0.96 |    0.04 |           - |           - |           - |                   - |        100,659 |
 ArraySum |     1111 |                  None | 4.194 ms | 1.3753 ms | 0.0754 ms |  1.00 |    0.00 |           - |           - |           - |                   - |         85,740 |
```
It even works faster with CacheClearingStrategy 😞

wojtpl2:
> 3\. run each test iteration on a different instance in sequence

I forgot to do it 😅 It was describe in #553. It will be the last chance.

PR #979, opened 2018-11-29
ARM support

adamsitnik:
@AndyAyersMS used BDN to run benchmarks for ARM and failed.

BDN was generating wrong target architecture based on IntPtr.Size, which returns 8 for both x64 and ARM64. The  solution is to use `System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture` instead.


adamsitnik:
Azure DevOps builds are green, I am merging this PR to unblock some folks.

AndreyAkinshin:
👍

Issue #975, opened 2018-11-26
Sample benchmark using CoreJob fails to run in LINQPad 5 (ClrJob runs fine)


Turnerj:
I'm pretty sure I'm doing something wrong but can't work out what. Using `ClrJob`, the benchmark runs flawlessly. Using `CoreJob`, I get errors like "Generate Exception: Unable to find LINQPadQuery in {Path}..."

I thought it might have been around the .Net SDK version so I downloaded the latest and installed it - still having the exact same error.

One thing I will note with my LINQPad configuration is that I created (and gave permission for) the "BenchmarkDotNet.Artifacts" folder and files to be written to "C:\Program Files (x86)\LINQPad5\BenchmarkDotNet.Artifacts". I tried with various other paths from auto-generated ones to just using the system temp path, they all come back with this issue for `CoreJob`.

LINQPad: v5.36.03 (Optimisations enabled for build)
BDN: v0.11.3 (default Nuget feed)
Code:
```csharp
void Main()
{
	var summary = BenchmarkRunner.Run<Md5VsSha256>();
}

// Define other methods and classes here
[CoreJob]
public class Md5VsSha256
{
	private SHA256 sha256 = SHA256.Create();
	private MD5 md5 = MD5.Create();
	private byte[] data;

	[Params(1000, 10000)]
	public int N;

	[GlobalSetup]
	public void Setup()
	{
		data = new byte[N];
		new Random(42).NextBytes(data);
	}

	[Benchmark]
	public byte[] Sha256() => sha256.ComputeHash(data);

	[Benchmark]
	public byte[] Md5() => md5.ComputeHash(data);
}
```

Full Output:
```
// Validating benchmarks:

Assembly LINQPadQuery, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null is located in temp. If you are running benchmarks from xUnit you need to disable shadow copy. It's not supported by design.

// ***** BenchmarkRunner: Start   *****

// ***** Building 1 exe(s) in Parallel: Start   *****

Unable to find .sln file. Will use current directory C:\Program Files (x86)\LINQPad5 to search for project file. If you don't use .sln file on purpose it should not be a problem.

// ***** Done, took 00:00:00 (0.02 sec)   *****

// Found benchmarks:

//   Md5VsSha256.Sha256: Core(Runtime=Core) [N=1000]

//   Md5VsSha256.Md5: Core(Runtime=Core) [N=1000]

[...]

PR #980, opened 2018-12-01
Detect .NET Core benchmark failures from LINQPad

Turnerj:
When running a .NET Core benchmark job from LINQPad, you will encounter an error like the following:
> // Generate Exception: Unable to find LINQPadQuery in C:\Program Files (x86)\LINQPad5 and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj

As noted in [this comment](https://github.com/dotnet/BenchmarkDotNet/issues/975#issuecomment-443297722), this is because to run a .NET Core job requires the benchmarks to be in .NET Core (LINQPad does not provide this).

This change detects that a .NET Core benchmark job was run from LINQPad and provides an additional error message stating this fact.

Detection of LINQPad is the same implementation used in [`JitOptimizationsValidator`](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Validators/JitOptimizationsValidator.cs) which uses the assembly name.

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkRunner.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunner.cs
@@ -197,7 +197,19 @@ namespace BenchmarkDotNet.Running
                     reports.Add(new BenchmarkReport(false, benchmark, buildResult, buildResult, default, default, default, default));
 
                     if (buildResult.GenerateException != null)
+                    {
                         logger.WriteLineError($"// Generate Exception: {buildResult.GenerateException.Message}");
+
+                        if (benchmark.Job.Environment == EnvironmentMode.Core)
+                        {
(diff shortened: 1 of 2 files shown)


adamsitnik:
@Turnerj big thanks for the PR!

In BDN we have the concept of toolchain, which builds the benchmarks but also validates them before that.

Could you please move this logic to [CsProjCoreToolchain.IsSupported](https://github.com/dotnet/BenchmarkDotNet/blob/6fb83099144462986bf9b9455c16bfd4a128e01e/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs#L63)? It will help us to prevent the error

Turnerj:
No problem - that's a much better spot to put it. I'll update my PR shortly :)

adamsitnik:
@Turnerj thank you!

Issue #786, opened 2018-06-09
An easy way to check for statistically significant difference between benchmarks


AndreyAkinshin:
Currently, we have optional `WelchTTestPValueColumn` which help you to verify that there is a statistically significant difference between benchmarks. However, it doesn't work great with default run strategy because this strategy typically doesn't perform enough iterations. Users have to manually choose a satisfactory amount of iterations. Thus, it's possible to do such checks, but the user experience is not good enough. We can do the following:

* Introduce additional property in `AccuracyMode`. Let's call it `StopСriterion` (let me know if you have better ideas about naming). It will contain logic which should decide when do we have enough iterations.
* Currently, we have hardcoded logic inside `EngineTargetStage`. Let's move it to a class called `StdErrStopCriterion`.
* We can introduce `WelchStopCriterion` which will do additional iterations until we sure that it's enough for the Welch's Two Sample t-test. (Bonus: users will be able to write own criterion)
* `StopCriterion` should be able to affect `IOrderProvider.GetExecutionOrder` and ask to run baseline benchmarks first.
* `EngineTargetStage.RunAuto` should get additional information about benchmarks like `IsBaseline` value. The non-baseline benchmarks should get all measurements from the baseline benchmark in the corresponded group.

Original request: https://twitter.com/AnthonyLloyd123/status/1005388154046644230

AndreyAkinshin:
Another [idea](https://twitter.com/AnthonyLloyd123/status/1005408508978847744) by Anthony Lloyd:

> Perhaps as a first step a warning if the values are too close to be reliable to a given confidence. Smart welch can make a run very long if tests are close.

caaavik-msft:
I've been looking into this particular issue recently and the way I see it there are two use cases here to consider:

1. Comparing two benchmark cases
2. Comparing a benchmark case against historical data collected previously

The first option would be if you have the ability to test the implementations side by side, but the second option may be needed if you don't have an easy way to test the baseline again (e.g. running in CI against latest commit but comparing to previous versions). 

[...]

PR #984, opened 2018-12-08
Introduce StoppingCriteria

AndreyAkinshin:
In this PR, we introduce StoppingCriteria with a few implementations: `FixedStoppingCriteria` and `AutoWarmupStoppingCriteria`.
It's the first step of a big Engine refactoring.
This refactoring has the following benefits:

1. **Better code isolation.**
   Currently, the logic of stopping criteria and the logic of benchmark execution are mixed inside the same classes.
   It's not always easy to find a problem in the stopping criteria.
2. **Better test coverage.**
   Currently, it's pretty hard to cover stopping logic by tests.
3. **Better engine customization.**
   In the future, users will be able to easily define own stopping criteria if they don't like defaults.
4. **Better diagnostics.**
   Now we know the reason why do we stop iterations; this information can be shown in logs.
   Also, we have a set of warnings about invalid configs.
5. **A possibility to implement "advanced" stopping criteria.**
   For example, #786 requires a stopping criteria which knows about baseline benchmarks iterations.
   In the current infrastructure, it's pretty hard to introduce this feature without major changes.


src/BenchmarkDotNet/Engines/StoppingCriteria/AutoWarmupStoppingCriteria.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        private readonly int minIterationCount;
+        private readonly int maxIterationCount;
+        private readonly int minFluctuationCount;
+
+        private readonly string maxIterationMessage;
+        private readonly string minFluctuationMessage;
+
+        /// <summary>
+        /// The idea of the implementation is simple: if measurements monotonously decrease or increase, the steady state is not achieved;
```
adamsitnik: @AndreyAkinshin  I love the fact that you added those comments! 🥇

src/BenchmarkDotNet/Engines/StoppingCriteria/AutoWarmupStoppingCriteria.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// If the required number of fluctuations is not achieved but the number of iterations equal to <paramref name="maxIterationCount"/>,
+        /// we should stop the iterations.
+        /// </param>
+        public AutoWarmupStoppingCriteria(int minIterationCount, int maxIterationCount, int minFluctuationCount = DefaultMinFluctuationCount)
+        {
+            this.minIterationCount = minIterationCount;
+            this.maxIterationCount = maxIterationCount;
+            this.minFluctuationCount = minFluctuationCount;
+
+            maxIterationMessage = $"The maximum amount of iteration ({maxIterationCount}) is achieved";
```
adamsitnik: nit: personally I would create those messages when we need them, not up-front.

src/BenchmarkDotNet/Engines/StoppingCriteria/AutoWarmupStoppingCriteria.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        protected override string GetTitle() => $"{nameof(AutoWarmupStoppingCriteria)}(" +
+                                                $"{nameof(minIterationCount)}={minIterationCount}, " +
+                                                $"{nameof(maxIterationCount)}={maxIterationCount}, " +
+                                                $"{nameof(minFluctuationCount)}={minFluctuationCount})";
+
+        protected override int GetMaxIterationCount() => maxIterationCount;
+
+        protected override IEnumerable<string> GetWarnings()
```
adamsitnik: nit: maybe we should call this method `Validate` instead of `GetWarnings` ?

src/BenchmarkDotNet/Engines/StoppingCriteria/AutoWarmupStoppingCriteria.cs
```diff
@@ -0,0 +1,97 @@
...
+                yield return $"Min Iteration Count ({minIterationCount}) is negative";
+            if (maxIterationCount < 0)
+                yield return $"Max Iteration Count ({maxIterationCount}) is negative";
+            if (minFluctuationCount < 0)
+                yield return $"Min Fluctuation Count ({minFluctuationCount}) is negative";
+            if (minIterationCount > maxIterationCount)
+                yield return $"Min Iteration Count ({minFluctuationCount}) is greater than Max Iteration Count ({maxIterationCount})";
+        }
+
+        public override StoppingResolution Evaluate(IReadOnlyList<Measurement> measurements)
```
adamsitnik: nit: I would put this public method to the top of the file (right after the ctor), this is the most important piece of code here

src/BenchmarkDotNet/Engines/StoppingCriteria/AutoWarmupStoppingCriteria.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            int n = measurements.Count;
+
+            if (n >= maxIterationCount)
+                return StoppingResolution.CreateFinished(maxIterationMessage);
+            if (n < minIterationCount)
+                return StoppingResolution.NotFinished;
+
+            int dir = -1; // The default "pre-state" is "decrease mode"
```
adamsitnik: what is `dir`? to me it's always a "directory" 

This code is very clean and easy to understand, I would just rename this variable and `nextDir` to full name (is it `direction` ?)

src/BenchmarkDotNet/Engines/EngineWarmupStage.cs
```diff
@@ -1,83 +1,24 @@
...
-                int nextDir = Math.Sign(measurements[i].Nanoseconds - measurements[i - 1].Nanoseconds);
-                if (nextDir != dir || nextDir == 0)
-                {
-                    dir = nextDir;
-                    changeCount++;
-                }
-            }
-
-            return changeCount >= 4;
+            var criteria = DefaultStoppingCriteriaFactory.Instance.CreateWarmup(engine.TargetJob, engine.Resolver, iterationMode, runStrategy);
```
adamsitnik: how can be the criteria customized? `DefaultStoppingCriteriaFactory.Instance` always points to a `DefaultStoppingCriteriaFactory` and is a read-only field

src/BenchmarkDotNet/Engines/StoppingCriteria/DefaultStoppingCriteriaFactory.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using BenchmarkDotNet.Characteristics;
+using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Engines
+{
+    public class DefaultStoppingCriteriaFactory
```
adamsitnik: I can see that this class has `virtual` methods but I can't see how the criteria could be customized

src/BenchmarkDotNet/Engines/StoppingCriteria/FixedStoppingCriteria.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        protected override int GetMaxIterationCount() => iterationCount;
+
+        protected override IEnumerable<string> GetWarnings()
+        {
+            if (iterationCount < 0)
+                yield return $"Iteration count ({iterationCount}) is negative";
+        }
+
+        public override StoppingResolution Evaluate(IReadOnlyList<Measurement> measurements)
```
adamsitnik: nit: this is the most important method, I would put it right after the ctor

src/BenchmarkDotNet/Engines/StoppingCriteria/IStoppingCriteria.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// Engine needs this value for setting the maximum capacity of the returned list of measurements.
+        /// The correct capacity helps to avoid infrastructure allocations during benchmarking.
+        /// </summary>
+        int MaxIterationCount { get; }
+        
+        /// <summary>
+        /// An array of user-friendly warnings which notify about incorrect parameters.
+        /// </summary>
+        [NotNull]
+        string[] Warnings { get; }
```
adamsitnik: nit: I would make it an `IReadOnlyList` instead of Array to be more immutable

src/BenchmarkDotNet/Engines/StoppingCriteria/StoppingCriteriaBase.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public abstract class StoppingCriteriaBase : IStoppingCriteria
+    {
+        private readonly Lazy<string> lazyTitle;
+        private readonly Lazy<int> lazyMaxIterationCount;
+        private readonly Lazy<string[]> lazyWarnings;
+
+        public string Title => lazyTitle.Value;
+        public int MaxIterationCount => lazyMaxIterationCount.Value;
+        public string[] Warnings => lazyWarnings.Value;
```
adamsitnik: nit: I think that if we have methods like `GetXYZ` we don't need `XYZ` property.

src/BenchmarkDotNet/Engines/StoppingCriteria/StoppingResolution.cs
```diff
@@ -0,0 +1,21 @@
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Engines
+{
+    public struct StoppingResolution
```
adamsitnik: nit: I think that `StoppingResult` would be a better name, we already use the `Result` convention in few places

src/BenchmarkDotNet/Engines/StoppingCriteria/AutoWarmupStoppingCriteria.cs
AndreyAkinshin: Good idea, will do it.

src/BenchmarkDotNet/Engines/StoppingCriteria/AutoWarmupStoppingCriteria.cs
AndreyAkinshin: Ok.

src/BenchmarkDotNet/Engines/EngineWarmupStage.cs
AndreyAkinshin: I will add API for customization in the future. It will be pretty hard to implement the whole refactoring at once. In this PR, I just introduce `IStoppingCriteria` as a concept.

src/BenchmarkDotNet/Engines/StoppingCriteria/FixedStoppingCriteria.cs
AndreyAkinshin: Ok.

src/BenchmarkDotNet/Engines/StoppingCriteria/StoppingCriteriaBase.cs
AndreyAkinshin: In this case, all of the `GetXYZ` methods are protected while the `XYZ` properties are public API. We need them to fill the private lazy fields which are used in the property getters. This approach helps to control that all of the properties are actually lazy.

adamsitnik:
@AndreyAkinshin thanks for the comments, now everything looks good to me!

Issue #983, opened 2018-12-07
Give a warning when the [Benchmark] method is static


adamsitnik:
BDN supports only instance methods, when our users try to do sth like:

```cs
public class Some
{
   [Benchmark]
   public static void This() { }
}
```

they get a silent error.

@AndreyAkinshin or maybe we should allow for static method calls?

AndreyAkinshin:
@adamsitnik, I'm not sure that static benchmarks is a good idea. It will complicate our generation template and it can introduce some undesirable effects in nanobenchmarks (such a feature should be carefully checked). Let's just warn users about an incorrect signature.

Rizzen:
I can do it.

AndreyAkinshin:
@Rizzen thanks!

PR #985, opened 2018-12-08
Give a warning when the [Benchmark] method is static

Rizzen:
For https://github.com/dotnet/BenchmarkDotNet/issues/983

### DIFF
--- a/tests/BenchmarkDotNet.Tests/Validators/CompilationValidatorTests.cs
+++ b/tests/BenchmarkDotNet.Tests/Validators/CompilationValidatorTests.cs
@@ -70,7 +70,7 @@ namespace BenchmarkDotNet.Tests.Validators
                                                                .ToList();
             
             // Assert
-            Assert.Equal(validationErrors.Any(), hasErrors);
+            Assert.Equal(hasErrors, validationErrors.Any());
         }
       
         private static Delegate BuildDummyMethod<T>(string name)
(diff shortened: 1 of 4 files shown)


Rizzen:
@adamsitnik @AndreyAkinshin We have some static benchmark method cases in InProcessBenchmarkAllCasesReflectionEmitSupported and InProcessBenchmarkAllCasesDelegateCombineSupported. Does it make any sense to have such test cases now?

AndreyAkinshin:
@Rizzen, I think we should fix or remove these tests.

tests/BenchmarkDotNet.Tests/Validators/CompilationValidatorTests.cs
```diff
@@ -28,12 +29,25 @@ public void BenchmarkedMethodNameMustNotContainWhitespaces()
...
+        [InlineData(typeof(BenchmarkClassWithStaticMethod), true)]
+        [InlineData(typeof(BenchmarkClass<PublicClass>), false)]
+        public void Benchmark_Class_Methods_Must_Be_Non_Static(Type type, bool hasErrors)
+        {
+            // Act
+            var validationErrors = CompilationValidator.Default.Validate(BenchmarkConverter.TypeToBenchmarks(type))
+                                                               .ToList();
+            
+            // Assert
+            Assert.Equal(validationErrors.Any(), hasErrors);
```
AndreyAkinshin: Please swap the arguments.
The first argument is the expected value; the second argument is the actual value.

AndreyAkinshin:
@Rizzen thanks!

Issue #986, opened 2018-12-11
NRE in `Summary` ctor


qbit86:
I've got NRE at `BenchmarkDotNet.Columns.MetricColumn.<IsAvailable>b__19_0(BenchmarkReport report)` running this benchmark: [https://goo.gl/4RGThA](https://github.com/qbit86/spontaneous-benchmark/blob/8b780ebbd3398e80edef918d4a09c6f376874a86/benchmarks/Spontaneous.LanguageBenchmarks/ComparingToNullBenchmark.cs)
```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at BenchmarkDotNet.Columns.MetricColumn.<IsAvailable>b__19_0(BenchmarkReport report)
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Lookup`2.Create(IEnumerable`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)
   at System.Linq.GroupedEnumerable`2.GetEnumerator()
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToArray()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at BenchmarkDotNet.Reports.SummaryTable..ctor(Summary summary, ISummaryStyle style)
   at BenchmarkDotNet.Reports.Summary..ctor(String title, IList`1 reports, HostEnvironmentInfo hostEnvironmentInfo, IConfig config, String resultsDirectoryPath, TimeSpan totalTime, ValidationError[] validationErrors)
   at BenchmarkDotNet.Running.BenchmarkRunner.Run(BenchmarkRunInfo benchmarkRunInfo, Dictionary`2 buildResults, IResolver resolver, ILogger logger, List`1 artifactsToCleanup, String rootArtifactsFolderPath, StartedClock& runChro
nometer)
   at BenchmarkDotNet.Running.BenchmarkRunner.Run(BenchmarkRunInfo[] benchmarkRunInfos, IConfig commonSettingsConfig)
   at BenchmarkDotNet.Running.BenchmarkRunner.Run(BenchmarkRunInfo benchmarkRunInfo)
[...]

PR #987, opened 2018-12-12
Fix NRE in MetricColumn (#986)

qbit86:
Fixes NRE after unchecked access to `BenchmarkReport.Metrics` property.

### DIFF
--- a/src/BenchmarkDotNet/Columns/DefaultColumnProvider.cs
+++ b/src/BenchmarkDotNet/Columns/DefaultColumnProvider.cs
@@ -84,7 +84,6 @@ namespace BenchmarkDotNet.Columns
         {
             public IEnumerable<IColumn> GetColumns(Summary summary) => summary
                 .Reports
-                .Where(report => report.Metrics != null)
                 .SelectMany(report => report.Metrics.Values.Select(metric => metric.Descriptor))
                 .Distinct(MetricDescriptorEqualityComparer.Instance)
                 .Select(descriptor => new MetricColumn(descriptor));
--- a/src/BenchmarkDotNet/Columns/MetricColumn.cs
+++ b/src/BenchmarkDotNet/Columns/MetricColumn.cs
@@ -22,13 +22,17 @@ namespace BenchmarkDotNet.Columns
 
         public bool IsDefault(Summary summary, BenchmarkCase benchmarkCase) => false;
 
-        public bool IsAvailable(Summary summary) => summary.Reports.Any(report => report.Metrics.ContainsKey(descriptor.Id));
+        public bool IsAvailable(Summary summary) => summary.Reports.Any(report => report.Metrics?.ContainsKey(descriptor.Id) ?? false);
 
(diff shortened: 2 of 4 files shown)


AndreyAkinshin:
@qbit86, thanks for the PR!

I guess it will be better to make `BenchmarkReport.Metrics` non-nullable instead of catching all the cases where we forgot to add a null check.

P.S. Also it will be nice to add a test for this case.

qbit86:
> I guess it will be better to make `BenchmarkReport.Metrics` non-nullable instead of catching all the cases where we forgot to add a null check.

Ok, then feel free to decline this PR, I'll prepare another.

AndreyAkinshin:
> Ok, then feel free to decline this PR, I'll prepare another.

@qbit86 you can use the same PR. Just add relevant commits yo your master branch.

qbit86:
> Also it will be nice to add a test for this case.

I narrowed down the [repro benchmark](https://github.com/qbit86/spontaneous-benchmark/blob/a2a2ff1689a6d2364f87fbe8a95c00aac0879286/benchmarks/Spontaneous.LanguageBenchmarks/BdnMetricsNreRepro.cs), but I don't know where to place it, in which test to run it, how to specify expected result vs. actual, etc.

qbit86:
...But at least existing checks have passed with no regression on CI.

AndreyAkinshin:
@qbit86 the best option is a unit test in the `BenchmarkDotNet.Tests` project. You can find some good examples in the `Reports` folders.
If it's not possible, we can introduce another integration test in the `BenchmarkDotNet.IntegrationTests` project (check out usages of BenchmarkTestExecutor). However, we are trying to not introduce additional integration tests (if it's possible) because they take too much time.

qbit86:
@AndreyAkinshin
> we are trying to not introduce additional integration tests (if it's possible) because they take too much time.

I guess, indeed, the very essence of this issue can be isolated from actual benchmark running.

> the best option is a unit test in the `BenchmarkDotNet.Tests` project. You can find some good examples in the `Reports` folders.

I believe this belongs to `Summary` tests, since it is `Summary` ctor in the middle of the stack trace who suffers more from passed null, not `BenchmarkReport` itself.

I've added `SummaryTests` class to `Reports/` folder and checked that test fails on old version of code and passes on fixed version.

qbit86:
@AndreyAkinshin 

> I can't reproduce the failure.
> If I revert both changes (in `BenchmarkReport` and `DefaultColumnProviders`), the test is still green.

To which commit did you try to revert changes? It may not reproduce since I've actually fixed the issue twice — once with naive approach of null-checking on call-site, then reverted it and add principal fix making property non-nullable. So it won't reproduce if you only revert last changes.

qbit86:
@AndreyAkinshin 
> I can't reproduce the failure.

Ok, I'll try to make the test red again :)

AndreyAkinshin:
@qbit86 thanks, merged.

PR #992, opened 2018-12-24
Use .NET Standard 2.0 CommandLineParser, Update to net461 for NS2.0 support

glennawatson:
- Added support for CommandLineParser 2.3.1 Net Standard 2.0 library -- this is for future work on support Xamarin releases 
- Now use NET461 for the libraries, this is required for .Net Standard 2.0 support.
- Now use $(TargetFramework.StartsWith('net4')) in our csproj files, this allows upgrades to .NET 4 based libraries without the pain currently required.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -18,7 +18,7 @@
     <UseSharedCompilation>false</UseSharedCompilation>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(IsVisualBasic)' != 'true' And '$(TargetFramework)' == 'net46' ">
+  <PropertyGroup Condition=" '$(IsVisualBasic)' != 'true' And $(TargetFramework.StartsWith('net4')) ">
     <DefineConstants>$(DefineConstants);CLASSIC</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(IsVisualBasic)' != 'true' And ('$(TargetFramework)' == 'netcoreapp2.0' Or '$(TargetFramework)' == 'netcoreapp2.1') ">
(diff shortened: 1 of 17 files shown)


glennawatson:
@adamsitnik you can't compile .net standard 2.0 libraries in your application with net46. 

See https://blogs.msdn.microsoft.com/dotnet/2017/08/14/announcing-net-standard-2-0/
Broad platform support. .NET Standard 2.0 is supported on the following platforms:
.NET Framework 4.6.1
.NET Core 2.0
Mono 5.4
Xamarin.iOS 10.14
Xamarin.Mac 3.8
Xamarin.Android 7.5
Upcoming version of UWP (expected to ship later this year)

So if you want to use either th command line parser ns2 or th command line API your going to hit the same wall you need to upgrade to net461 otherwise you'll get compile errors.

glennawatson:
https://github.com/commandlineparser/commandline/issues/292 is the issue I am hitting. It's solved with https://github.com/commandlineparser/commandline/pull/307 but again you can't get that fix without going to net461 due to it being the first version to support net standard 2.0. Net46 is not supported in that scenario

ViktorHofer:
Just as a note, CommandLineParser.NS20 was published by me some months ago to overcome the time till my PR is accepted. Unfortunately that still hasn't happened.

adamsitnik:
@AndreyAkinshin are you ok with the update from .NET 4.6 to 4.6.1?

I think that we will have to do it anyway since Roslyn's next version is going to support .NET Standard 2.0 only: https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp/2.11.0-beta1-final

(the lowest .NET supporting .NET Standard 2.0 is .NET 4.6.1)

adamsitnik:
@AndreyAkinshin @glennawatson thanks!

PR #996, opened 2019-01-05
Rephrase Notes section and fix markdown

Maximusya:
The sample markdown as it was - was broken and misleading.
I guess it is https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html?tabs=tabid-1%2Ctabid-a#note-warningtipimportant to blame for abstract self-contradictory samples with copy-paste artifacts.

### DIFF
--- a/docs/articles/contributing/documentation.md
+++ b/docs/articles/contributing/documentation.md
@@ -15,37 +15,37 @@ BenchmarkDotNet uses [DocFX](https://dotnet.github.io/docfx/) as a documentation
     
 ## Notes
 
-DocFX uses the following syntax for different types of notes:
+DocFX uses the [following syntax](https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html?tabs=tabid-1%2Ctabid-a#note-warningtipimportant) inside block quote for different types of notes:
 
 ```markdown
 > [!NOTE]
-> <note content>
+> note content
 > [!TIP]
-> <note content>
+> tip content
 > [!WARNING]
-> <warning content>
+> warning content
 > [!IMPORTANT]
-> <important content>
+> important content
 > [!Caution]
-> <caution content>
+> caution content
 ```
 
 It will be transformed to:
 
 > [!NOTE]
-> <note content>
+> note content
 
 > [!TIP]
-> <note content>
+> tip content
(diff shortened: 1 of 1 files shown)


adamsitnik:
Hi @Maximusya 

Why did you remove the `<` `>` signs? It does not look broken on our web page https://benchmarkdotnet.org/articles/contributing/documentation.html

Maximusya:
First the broken part: the html generated contains invalid/unintended and silently ignored by a browser html entities.
> <p>DocFX uses the following syntax for different types of notes:</p>
> <pre><code class="lang-markdown">&gt; [!NOTE]
> &gt; &lt;note content&gt;
> &gt; [!TIP]
> &gt; &lt;note content&gt;
> &gt; [!WARNING]
> &gt; &lt;warning content&gt;
> &gt; [!IMPORTANT]
> &gt; &lt;important content&gt;
> &gt; [!Caution]
> &gt; &lt;caution content&gt;
> </code></pre>

```html
<p>It will be transformed to:</p>
<div class="NOTE">
<h5>Note</h5>
<note content="">
</note></div>
<div class="TIP">
<h5>Tip</h5>
<note content="">
</note></div>
<div class="WARNING">
<h5>Warning</h5>
<warning content="">
</warning></div>
<div class="IMPORTANT">
<h5>Important</h5>
<important content="">
</important></div>
<div class="CAUTION">
<h5>Caution</h5>
<caution content="">
</caution></div>
```

I could have edited only the markdown that generates actual DocFx notes so that they contain strings `note content`, `tip content` etc - by simply removing angle brackets in DocFx markdown only. But then if &lt; &gt; in sample syntax (within block quotes) remain, it would mislead and defeat the purpose of copypaste-ready samples. `It will be transformed to` promise would not hold. The mentioned DocFx link lies too in a similar way.
But if we edit both sample syntax and the DocFx markdown, we will help a newcomer (hell, I was a DocFx newcomer today!) to learn by example and we would keep promises.

Or to put it another way: Markdown block quote angle brackets + usual dummy content template angle brackets + HTML angle brackets put too much cognitive load  both on a programmer and on parsers.

Maximusya:
Thanks for your kind words!

PR #1001, opened 2019-01-06
CoreRT toolchain update

adamsitnik:
1. port #854 PR to CoreRT toolchain as well so the output file path doe not get overwritten by .props files
2. take the current TFM as default, not netcoreapp2.1 for CoreRT
3. if we fail to do the standard publish, let's try with --no-build
4. CoreRT toolchain must respect `--packages`


adamsitnik:
All builds are green, I have done a LOT of manual testing so I am merging it!

PR #1005, opened 2019-01-08
Fix dotnet core version detection

### LINKED ISSUE #995: Running benchmark fails when targeting netcoreapp2.2
Running a benchmark fails when targeting netcoreapp2.2. That is because the *.Autogenerated.csproj. targets netcoreapp2.1 and references the original benchmark project that targets a higher netcoreapp version.

__Repro__:
1. Create a new netcore CLI application that targets netcoreapp2.2
2. Add a reference to BenchmarkDotNet v0.11.3
3. Add any Benchmark w/o explicit configuration (the Md5VsSha256 example from the docs will do)
4. Build the benchmark project: `dotnet build -f netcoreapp2.2 -c Release -r win-x64 --no-incremental --force` 
5. Run the benchmark like so: `.\bin\Release\netcoreapp2.2\win-x64\Target22.exe`

__Expected__:
The benchmark completes without errors.

__Actual__:
Running the benchmark (repro step 5) fails early when building the autogenerated code because the autogenerated csproj targets netcoreapp2.1 and references the original projects which targets 2.2.

__Workaround__:
Use an explicit configuration like so;

        private class DotNetCore22Config : ManualConfig
        {
            public DotNetCore22Config()
            {
                Add(Job.Default.With(CsProjCoreToolchain.NetCoreApp22));
            }
        }

Let me know if you need a complete sample project to reproduce.

jasper-d:
- Fixes #995 by using version from TargetFrameworkAttribute
- Removes obsolete Docker specific code

### DIFF
--- a/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
@@ -160,18 +160,14 @@ namespace BenchmarkDotNet.Portability
             return null;
         }
 
-        internal static string GetNetCoreVersion() => InDocker ? GetDockerNetCoreVersion() : GetBaseNetCoreVersion();
-
-        private static string GetDockerNetCoreVersion() => string.IsNullOrEmpty(DockerSdkVersion) ? DockerAspnetSdkVersion : DockerSdkVersion;
-
-        private static string GetBaseNetCoreVersion()
+        internal static Version GetNetCoreVersion()
         {
             var frameworkName = Assembly.GetEntryAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName;
 
             if (frameworkName == null)
                 return null;
 
-            return new FrameworkName(frameworkName).Version.ToString();
+            return new FrameworkName(frameworkName).Version;
         }
         
         internal static string GetRuntimeVersion()
@@ -206,7 +202,7 @@ namespace BenchmarkDotNet.Portability
             }
             else if (IsNetCore)
             {
-                string runtimeVersion = GetNetCoreVersion() ?? "?";
+                string runtimeVersion = GetNetCoreVersion()?.ToString() ?? "?";
 
                 var coreclrAssemblyInfo = FileVersionInfo.GetVersionInfo(typeof(object).GetTypeInfo().Assembly.Location);
                 var corefxAssemblyInfo = FileVersionInfo.GetVersionInfo(typeof(Regex).GetTypeInfo().Assembly.Location);
--- a/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
@@ -6,8 +6,8 @@ using System.IO;
 using System.Linq;
 using System.Management;
 using System.Reflection;
-using System.Runtime;
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 using System.Text.RegularExpressions;
 using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Extensions;
@@ -166,12 +166,12 @@ namespace BenchmarkDotNet.Portability
 
         private static string GetBaseNetCoreVersion()
         {
-            var assembly = typeof(GCSettings).GetTypeInfo().Assembly;
-            var assemblyPath = assembly.CodeBase.Split(new[] { '/', '\\' }, StringSplitOptions.RemoveEmptyEntries);
-            int netCoreAppIndex = Array.IndexOf(assemblyPath, "Microsoft.NETCore.App");
-            if (netCoreAppIndex > 0 && netCoreAppIndex < assemblyPath.Length - 2)
-                return assemblyPath[netCoreAppIndex + 1];
-            return null;            
+            var frameworkName = Assembly.GetEntryAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName;
+
+            if (frameworkName == null)
+                return null;
+
+            return new FrameworkName(frameworkName).Version.ToString();
         }
         
         internal static string GetRuntimeVersion()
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -160,18 +160,14 @@ internal static CpuInfo GetCpuInfo()
             return null;
         }
 
-        internal static string GetNetCoreVersion() => InDocker ? GetDockerNetCoreVersion() : GetBaseNetCoreVersion();
+        internal static Version GetNetCoreVersion()
+        {
+            var frameworkName = Assembly.GetEntryAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName;
```
adamsitnik: > Please note that in some cases Assembly.GetExecutingAssembly() will yield a different result compared to the previous logic (potentially breaking user benchmarks).
This happens for instance when running unit tests. The target framework attribute of the executing assembly (testhost.dll) has the value .NETCoreApp,Version=v1.0, hence the new logic will detect version 1.0.

@jasper-d I tested that and you are right. However, I can see that `GetExecutingAssembly` returns valid result. I tested it for xunit runner and standalone netcoreapp. So please change `GetEntryAssembly` to `GetExecutingAssembly`

big thanks for testing this scenario and mentioning this!

```suggestion
            var frameworkName = Assembly.GetExecutingAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName;
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -160,18 +160,14 @@ internal static CpuInfo GetCpuInfo()
             return null;
         }
 
-        internal static string GetNetCoreVersion() => InDocker ? GetDockerNetCoreVersion() : GetBaseNetCoreVersion();
+        internal static Version GetNetCoreVersion()
```
adamsitnik: Since this method does not return .NET Core version anymore I think that we should rename it to just `GetTargetFrameworkVersion`

```suggestion
        internal static Version GetTargetFrameworkVersion()
```

src/BenchmarkDotNet/Toolchains/DotNetCli/NetCoreAppSettings.cs
```diff
@@ -99,30 +99,21 @@ internal static NetCoreAppSettings GetCurrentVersion()
             if (RuntimeInformation.IsFullFramework)
                 return Default;
 
-            string netCoreAppVersion = null;
+            var netCoreAppVersion = RuntimeInformation.GetNetCoreVersion();
 
-            try
+            switch (netCoreAppVersion)
```
adamsitnik: looks great! much better that the old code! 👍

adamsitnik:
@jasper-d could you please also add a unit test that verifies the method output? Out `BenchmarkDotNet.Tests` project targets net46, netcoreapp20 and 21 so the test can look like this:

```cs
Version expectedVersion = 
#if net46
   new Version(4, 6);
#elseif netcoreapp2_0
   new Version(2, 0);
#elseif netcoreapp2_1
   new Version(2, 1);
#endif

Assert.Equal(expectedVersion, RuntimeInformation.GetTargetFrameworkVersion());
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
jasper-d: @adamsitnik Sorry, I meant GetEntryAssembly when I wrote GetExecutingAssembly. 😕

Anyway, according to my understanding of GetExecutingAssembly and a quick test, it's even worse. It will return the actual assembly it is called from, so in this case it will always return the BDN assembly.

The following test will fail when running inside a netcoreapp2.2 test project. When using GetEntryAssembly `result` will be 1.0 (testhost.dll), when using GetExecutingAssembly `result` will be 2.1 (BenchmarkDotNet.dll).

```
        [Fact]
        public void GetTargetFrameworkVersionReturns22()
        {
            var currentEnvironment = new FrameworkName(Assembly.GetExecutingAssembly().GetCustomAttribute<TargetFrameworkAttribute>().FrameworkName).Version; //2.2

            var result = BenchmarkDotNet.Portability.RuntimeInformation.GetTargetFrameworkVersion();

            Assert.True(new Version(2, 2) == currentEnvironment, "Assert environment"); //Sanity check, passes always
            Assert.Equal(currentEnvironment, result); //Fails
        }
```

jasper-d: I would propose to fix the original implementation

adamsitnik: @jasper-d you are right. Now we know how both solutions work, but none of them  covers 100% of cases ;(

jasper-d: @adamsitnik So how do we move forward from here? I would propose to close this PR and make a new one that fixes the previous logic, taking into account the varying directory names. Maybe logging a warning to inform users about failed version detection would be helpful as well, just in case.

I still like the GetEntryAssembly approach because it makes the code significantly simpler and removes the need for special handling of containerised applications (would probably worth to test rkt as well). 
I don't know how other people use BDN, but I assume that for the vast majority the entry assembly will be a console application they wrote for their benchmarking and not some other DLL (like testhost). Hence, I think it would be a viable option to update this PR with your change requests (testing will be difficult though). Then come back to it when a new major version of BDN with other breaking changes is going to be released. In cases where the entry assembly is different, users could then use explicit configurations.

Your thoughts?

jasper-d: @adamsitnik Any opinion?

jasper-d:
Closing due to inactivity. I'll create a new PR if there is some information on how it should be done.

adamsitnik:
@jasper-d I am sorry for the inactivity

Today I tried [AppContext.TargetFrameworkName](https://apisof.net/catalog/System.AppContext.TargetFrameworkName), but it does not work in all the scenarios...

PR #1006, opened 2019-01-08
BenchmarkDotNet as global tool

### LINKED ISSUE #213: Add a "benchmark" cmd to dotnet
So it seems like you can [add extra commands](https://github.com/dotnet/cli/blob/rel/1.0.0-preview2/Documentation/developer-guide.md#adding-a-command) to the dotnet tooling, how cool would it be to be able to do `dotnet benchmark`? 

This could then interact with the `BenchmarkSwitcher` so you can do `--class=Benchmarks`, `--diagnosers=Memory`, etc

From my understanding we would have to propose it as a [CLI issue](https://github.com/dotnet/cli/issues), but if we were serious about it, I think someone should do a proof-of-concept first.

Anyone agree? Anyone want to take it on?

CodeTherapist:
This resolves #213.

### DIFF
--- a/BenchmarkDotNet.sln
+++ b/BenchmarkDotNet.sln
@@ -15,11 +15,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BenchmarkDotNet.Diagnostics
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BenchmarkDotNet.IntegrationTests.CustomPaths", "tests\BenchmarkDotNet.IntegrationTests.CustomPaths\BenchmarkDotNet.IntegrationTests.CustomPaths.csproj", "{0031728E-A5D4-47C1-9C1A-6C859A765C9D}"
 EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "BenchmarkDotNet.Samples.FSharp", "samples\BenchmarkDotNet.Samples.FSharp\BenchmarkDotNet.Samples.FSharp.fsproj", "{A329F00E-4B9D-4BC6-B688-92698D773CBF}"
+Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "BenchmarkDotNet.Samples.FSharp", "samples\BenchmarkDotNet.Samples.FSharp\BenchmarkDotNet.Samples.FSharp.fsproj", "{A329F00E-4B9D-4BC6-B688-92698D773CBF}"
 EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "BenchmarkDotNet.IntegrationTests.FSharp", "tests\BenchmarkDotNet.IntegrationTests.FSharp\BenchmarkDotNet.IntegrationTests.FSharp.fsproj", "{367FAFE1-A1C8-4AA1-9334-F4762E128DBB}"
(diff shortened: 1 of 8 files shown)


adamsitnik:
Hi @CodeTherapist 

big thanks for doing this!

How can this be tested and deployed as a dotnet global tool?

@AndreyAkinshin do you like the  `BenchmarkDotNet.Tool` name? maybe ` `BenchmarkDotNet.Runner` would be better?

CodeTherapist:
> How can this be tested and deployed as a dotnet global tool?

**Testing**
I could add a unit test that executes the `.exe` against a sample benchmark dll - that would at least prove that the tool is able to be executed for the most common scenario. On the other hand, it is a really thin layer around the well tested `BenchmarkSwitcher`.

**Deployment**
The distribution of the tools are nuget packages. That would mean the nuget package of the tool must be uploaded to nuget.org to be available for everyone. As far as I understood, the nuget package should be created by the [cake build](https://github.com/dotnet/BenchmarkDotNet/blob/master/build.cake#L158).

**Naming**
Maybe I'm too cautious. I used the name `BenchmarkDotNet.Tool` to avoid any confusion with the existing class named [BenchmarkRunner](https://github.com/dotnet/BenchmarkDotNet/blob/7e7ddebed9acbf258c957c47afcf3332124d62ee/src/BenchmarkDotNet/Running/BenchmarkRunner.cs).

AndreyAkinshin:
I think `BenchmarkDotNet.Tool` is a good name because many other .NET global tools use the same convention (e.g., see [Cake.Tool](https://www.nuget.org/packages/Cake.Tool/), [GitVersion.Tool](https://www.nuget.org/packages/GitVersion.Tool), [NuGetUtils.Tool.Restore](https://www.nuget.org/packages/NuGetUtils.Tool.Restore/)). We can find a big list of different tool names here: https://github.com/natemcmaster/dotnet-tools It seems that there is no the universal convention, but you can find many titles with `Tools`, `GlobalTool`, and so on. I also don't like the `BenchmarkDotNet.Runner` title because of the clash with existed runner infrastructure.

@CodeTherapist could you also add a documentation article about this tool?

CodeTherapist:
@AndreyAkinshin Yes, sure! Since that is my first contribution to this project, I'm unsure where to put that in the current docs structure. Maybe in [docs/articles/guides](https://github.com/dotnet/BenchmarkDotNet/tree/master/docs/articles/guides)? Where do you think is it to be placed right?

AndreyAkinshin:
> Maybe in docs/articles/guides?

Yeah, it's a good place.

adamsitnik:
@AndreyAkinshin I did not know about this convention, it sounds good to me now.

btw @CodeTherapist could you try to import some of the MsBuild properies from https://github.com/dotnet/BenchmarkDotNet/blob/master/build/common.props file?

We import it in following way in all our projects:

https://github.com/dotnet/BenchmarkDotNet/blob/f354659bcfe4b043d69b7812cc69c63c26676da9/src/BenchmarkDotNet/BenchmarkDotNet.csproj#L2

CodeTherapist:
@adamsitnik Yes, I will do. 
What do you think about to use the built-in MSBuild extensibility feature (#1011)?

CodeTherapist:
> could you also add a documentation article about this tool?

@AndreyAkinshin Added documentation! Please let me know if that is enough for a first version.

adamsitnik:
@AndreyAkinshin please take a look and merge if you are ok with the code. I am!

adamsitnik:
@ViktorHofer so now we are going to have a runner tool ;)

AndreyAkinshin:
@CodeTherapist @adamsitnik It would be nice to rename the command from "benchmarkdotnet" to just "benchmark". It will be easier to type `dotnet benchmark` in the command line. What do you think?

P.S. I need some time to test this tool in different environments. But I'm definitely going to merge it in v0.11.4.

CodeTherapist:
@AndreyAkinshin I thought about both variations: `benchmark` and `benchmarkdotnet`. I decided to the last because it matches the product name - but I'm fine also with `benchmark`.

PR #1012, opened 2019-01-10
Changed TargetCount to IterationCount in docs

Sitiritis:
Changed `TargetCount` attribute name to `IterationCount` in Configs.Jobs section of documentation as this attribute was renamed in class `BenchmarkDotNet.Jobs.RunMode` in this commit dotnet@5e6e331


adamsitnik:
@Sitiritis good catch, thanks for doing this!

PR #1015, opened 2019-01-13
[Proof of Concept][WIP] Cross platform disassembler

adamsitnik:
So far both disassemblers lived in the resources of BDN as pre-built **Windows .exes**. I want to make the disassembler cross-platform so this is not acceptable anymore.

In order to do that I created two global tools out of our disassemblers. The goal is to have them as a command "dotnet disassembler_x64" and "dotnet disassembler_x86".

I know that `x86` tests are going to fail:

- x64 cli can't build an x86 .exe 
- if we set RuntimeIdentifier to win7-x86 it can
- but global tools don't support self-contained apps ("Pack as tool does not support self contained")

I need to wrap my head around the x86 tooling issue, but right now I am curious for the x64 results on Unix systems (to verify that the new ClrMD can disassemble methods on Unix)


adamsitnik:
> how you are going to call disassembly diagnosers for .NET Framework benchmarks on machines without installed .NET Core SDK?

@AndreyAkinshin this is a very good question, however, I don't have any good answer for it.

My goal was to validate if ClrMD can really disassemble methods on Unix.

As of today, it can not: https://github.com/Microsoft/clrmd/issues/143

The error from our test runs:

```log
 Failed to disassemble with following exception:
 Cannot marshal 'parameter #2': Invalid managed/unmanaged type combination (Marshaling to and from COM interface pointers isn't supported).
    at Microsoft.Diagnostics.Runtime.DbgEngDataReader.DebugCreate(Guid& InterfaceId, Object& Interface)
    at Microsoft.Diagnostics.Runtime.DbgEngDataReader.CreateIDebugClient() in /_/src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs:line 162
    at Microsoft.Diagnostics.Runtime.DbgEngDataReader..ctor(Int32 pid, AttachFlag flags, UInt32 msecTimeout) in /_/src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs:line 33
    at Microsoft.Diagnostics.Runtime.DataTarget.AttachToProcess(Int32 pid, UInt32 msecTimeout, AttachFlag attachFlag) in /_/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs:line 137
    at BenchmarkDotNet.Disassembler.Program.Handle(Settings settings) in /home/travis/build/dotnet/BenchmarkDotNet/src/BenchmarkDotNet.Tools.Disassembler.x64/Program.cs:line 61
    at BenchmarkDotNet.Disassembler.Program.Main(String[] args) in /home/travis/build/dotnet/BenchmarkDotNet/src/BenchmarkDotNet.Tools.Disassembler.x64/Program.cs:line 40
```

adamsitnik:
@AndreyAkinshin I have closed the PR, please don't remove the branch for a while (I still want to test some scenarios)

AndreyAkinshin:
> this is a very good question, however, I don't have any good answer for it.

@adamsitnik keep in mind that we should support this case. It would be aswesome to have xplat disasm, but we shouldn't drop the .NET Framework support. (I'm OK with ugly hacks if they are needed.)

adamsitnik:
> I'm OK with ugly hacks if they are needed

I am recently less eager to add any hacks. Most of the users expect us to spend our free time to implement some hacks and keep them happy because they are too lazy to install some tools or update .NET framework. Also, things like the disassembler or profilers are optional plugins used mostly by the early adopters who are usually OK with some extra requirements from our side.

Once ClrMD implements Linux support I am going to choose some good solution, but I definitely prefer clean solution over hacks.

0xd4d:
@adamsitnik

> but right now I am curious for the x64 results on Unix systems (to verify that the new ClrMD can disassemble methods on Unix)

CLRMD doesn't have a disassembler, the [method](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Disassembler.x64/Program.cs#L270) BDN is using is part of an interface implemented by native code and this is Windows only.

adamsitnik:
@0xd4d thanks! I did not know that. I guess it's going to make implementing the disassembler for non-Windows even harder.

0xd4d:
@adamsitnik There are open source .NET libraries that can disassemble x86/x64 code, eg. [mine here](https://github.com/0xd4d/iced)

AndreyAkinshin:
@0xd4d your disassembler is pretty cool!
@adamsitnik what do you think about using iced in BenchmarkDotNet? It should allow getting the same disasm style (which can be configurable) and providing additional info about instructions.

adamsitnik:
> what do you think about using iced in BenchmarkDotNet? It should allow getting the same disasm style (which can be configurable) and providing additional info about instructions.

@AndreyAkinshin I totally agree, this is a great idea!

PR #1030, opened 2019-01-29
feature: add msbuild.sdk.extra builds for more platforms

glennawatson:
This is the start on starting support for Xamarin targets.

MSBuild.Sdk.Extras is a library by Oren Novotny https://github.com/onovotny/MSBuildSdkExtras that allows you to use the new csproj format with more TargetFramework's.

This allows me to build executables in those various platforms.

The next step is build something similar to https://github.com/xunit/devices.xunit which I am halfway through which will allow the benchmarks to run on device/emulator.

This also fixes an issue with my project (https://github.com/reactiveui/reactiveui) where was unable to benchmark due to Tizen TargetFramework in my project.

### DIFF
--- a/src/BenchmarkDotNet/Templates/MsBuildSdkExtrasCsProj.txt
+++ b/src/BenchmarkDotNet/Templates/MsBuildSdkExtrasCsProj.txt
@@ -1,0 +1,27 @@
+﻿<Project Sdk="MSBuild.Sdk.Extras/1.6.65">
+  <PropertyGroup>
+    <AssemblyTitle>$PROGRAMNAME$</AssemblyTitle>
+    <TargetFramework>$TFM$</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <PlatformTarget>$PLATFORM$</PlatformTarget>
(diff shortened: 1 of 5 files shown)


adamsitnik:
@glennawatson I would prefer to merge the PR when the entire feature is ready. I assume that for you it would be easier to work with a pre-release package rather than fork of BDN?

Also, since this requires `MSBuild.Sdk.Extras` NuGet package I think that we should put the new Toolchain to new BDN package with explicit dependency to `MSBuild.Sdk.Extras`

glennawatson:
Hi @adamsitnik -- so suggesting I make my own pre-release package or something you can provide?

The rest of your comments make sense though.

adamsitnik:
@glennawatson IMHO you could create a local NuGet package using dotnet publish, add the folder where it's stored to NuGet.config file and work with a local package. As soon as you have everything working, send one big PR. I don't want to merge it now because we are going to release 0.11.4 this week and it would include not-complete Toolchain

glennawatson:
@adamsitnik No worries, I'll re-open this once the other parts are done.

adamsitnik:
@glennawatson big thanks for working on this! it's gonna be a really BIG feature!

PR #1032, opened 2019-01-31
Target .NET Standard 2.0 only

adamsitnik:
I was able to make everything work with single TFM.

I am not going to merge it to master now, I just want to check if all is green.


adamsitnik:
@AndreyAkinshin I tested these crazy things using 2k+ benchmarks from performance repo for .NET 4.6.1, Mono, Core 2.0/2.1/2.2/3.0 and CoreRT. The workaround I added for assembly binding redirects is far from perfect, but it works. 

I am going to merge the PRs now

abelbraaksma:
> I was able to make everything work with single TFM.

@adamsitnik  Apologies for commenting on an old, closed thread, but what does TFM mean in this context? I couldn't find it on several acronym sites, or at least [nothing that made sense](https://acronyms.thefreedictionary.com/TFM).

WojciechNagorski:
I think Target Framework Moniker e.g.: https://github.com/dotnet/BenchmarkDotNet/blob/2ba30330ec5ec8a9abe284c927277a6fe0fc5a0c/src/BenchmarkDotNet/Toolchains/CoreRt/CoreRtToolchain.cs#L13

PR #1036, opened 2019-02-01
Remove InternalsVisibleTo for Samples application

wojtpl2:
Sample application shouldn't use internal class. Instead of `RuntimeInformation.Is64BitPlatform()` I've used `Environment.Is64BitProcess`

https://github.com/dotnet/corefx/blob/94e9d02ad70b2224d012ac4a66eaa1f913ae4f29/src/Common/src/CoreLib/System/Environment.cs#L111

Now global tool works well.


adamsitnik:
good point! @wojtpl2 thanks!

PR #1040, opened 2019-02-02
Making the new InProcessEmitToolchain work after my recent refactor and .NET Standard 2.0 port

adamsitnik:
@ig-sinicyn please take a look, I need to run for a birthday party, will post an explanation of my changes later (but commit messages should be enough more or less)

the plan is that I am going to squash my changes into the newInProcess branch and then merge (no squash) newInProcess  into master if you are OK with my changes and the CI is green


adamsitnik:
@ig-sinicyn some explanation:

1. After my recent refactor (#912) BenchmarkCase has now Config property, so we don't need to pass it separately to the runners
2. After moving to .NET Standard 2.0 only (#1032) I could not easily save the assembly to the disk for the testing purposes. I have used reflection to do that and `#if` to make sure it's not published to NuGet.org package. It's far from perfect, but we need it only for tests and it's just a lesser evil.
3. Static benchmarks are not supported by design (#985). I think that one day we will most probably support them, so I removed the tests only, not the logic that you have implemented.
4. RoslynToolchain works only for Full .NET Framework and Mono, the reason is that we simply don't know which assemblies to reference (see https://github.com/dotnet/roslyn/issues/16211 for more details) so I made the test that compares the produced IL running only for Full .NET Framework

@ig-sinicyn PTAL at the PR, I would like to merge it soo so @AndreyAkinshin can publish 0.11.4 next week

adamsitnik:
@ig-sinicyn thank you! and once again apologies for so delayed review!

Issue #1039, opened 2019-02-02
Some tests are broken on Net 461 (culture-dependent thing)


ig-sinicyn:
master branch, VS 2017 15.9.6, target net461, current UI culture is 'ru-RU'

Some tests are failing under both VS & Resharper test runners, first few of them:

* Tests 
  AllSetupAndCleanupMethodRunsAsyncGenericValueTaskSetupTest, 
  AllSetupAndCleanupMethodRunsAsyncValueTaskSetupTest 
  AnArrayCanBePassedToBenchmarkAsSpan
  ArgumentsArePassedToBenchmarks
  TaskReturningMethodsAreAwaited
  UserCanSpecifyCustomBuildConfiguration

  benchmark build fails with message
  ```
  error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.
  ```

* Tests  `WhenInvalidCommandLineArgumentIsPassedAnErrorMessageIsDisplayedAndNoBenchmarksAreExecuted` 
  `UnknownConfigMeansFailure`

  fail with ArgumentNullException.

*  Test CanDisassembleInlinableBenchmarks, error 'Assert.Contains() Failure
 Not found: (filter expression)'

*  Tests from LocalCoreClrToolchainTests, LocalCoreRtToolchainTests
 error 'System.IO.DirectoryNotFoundException : C:\Projects\corert\bin\Windows_NT.x64.Release provided as newIlcPath does NOT exist'

etc

adamsitnik:
Could you please perform a full clean (`git clean -xfd` for example) and try to re-run?

It passes on AppVeyor and Azure DevOps, so an old bin is my first guess.

ig-sinicyn:
Just a minute, will try.

ig-sinicyn:
OK, I've made PR for culture-related part of the issue.

Some tests still failing when code is built as debug (`CanDisassembleInlinableBenchmarks` as example).

adamsitnik:
Fixed by #1042

ig-sinicyn:
@adamsitnik some tests still fail for debug builds, CanDisassembleInlinableBenchmarks as example. Is it ok?

adamsitnik:
[...]

PR #1042, opened 2019-02-03
Use invariant culture for csc messages

ig-sinicyn:
Fixes culture-related errors mentioned in #1039.
Roslyn uses LocalizableString undercover; all methods that return human-readable messages should be called with specific culture passed as an arg.

I've added CanExecuteWithNonDefaultUiCulture test to prevent possible regressions.

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/RoslynToolchainTest.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/RoslynToolchainTest.cs
@@ -1,0 +1,56 @@
+﻿using System.Globalization;
+using System.Threading;
+
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Tests.Loggers;
+using BenchmarkDotNet.Toolchains.Roslyn;
+
+using Xunit;
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+    public class RoslynToolchainTest : BenchmarkTestExecutor
+    {
+        public RoslynToolchainTest(ITestOutputHelper output) : base(output) { }
+
+        /// <summary>Prooftest for #1039.</summary>
+        [Theory]
+        [InlineData("en-US")]
+        [InlineData("fr-FR")]
+        [InlineData("ru-RU")]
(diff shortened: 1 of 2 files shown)


tests/BenchmarkDotNet.IntegrationTests/RoslynToolchainTest.cs
```diff
@@ -0,0 +1,56 @@
...
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+    public class RoslynToolchainTest : BenchmarkTestExecutor
+    {
+        public RoslynToolchainTest(ITestOutputHelper output) : base(output) { }
+
+        /// <summary>Prooftest for #1039.</summary>
+        [Theory]
```
adamsitnik: Roslyn toolchain does not work on .NET Core, please make this test framework-dependent

```suggestion
        [TheoryFullFrameworkOnly]
```

ig-sinicyn:
@adamsitnik I've made the changes but the status still is "Changes requested". I'm new with github flow, am I doing something wrong?

tests/BenchmarkDotNet.IntegrationTests/RoslynToolchainTest.cs
```diff
@@ -0,0 +1,56 @@
...
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+    public class RoslynToolchainTest : BenchmarkTestExecutor
+    {
+        public RoslynToolchainTest(ITestOutputHelper output) : base(output) { }
+
+        /// <summary>Prooftest for #1039.</summary>
+        [TheoryFullFrameworkOnly]
```
adamsitnik: I am sorry, my previous suggestion was incomplete: we need to specify namespace (or add a using to the file) and the reason why we want to skip

```suggestion
        [Tests.XUnit.TheoryFullFrameworkOnly("Roslyn toolchain does not support .NET Core")]
```

adamsitnik:
@ig-sinicyn I am sorry, my previous suggestion was incomplete: we need to specify namespace (or add a using to the file) and the reason why we want to skip

> I've made the changes but the status still is "Changes requested".

I need to manually click "approve" fo update the status. As soon as you are going to apply my latest suggestion I do that

PR #1043, opened 2019-02-03
minor InProcess fix: diff now checks for implementation flags

ig-sinicyn:
More fixes:) Forget to add impl flags comparison into NaiveEmitDiff. Of course, there was a change that diff did not spotted (SHA-1: 348f87661e25ddbcefa729b6dc724cef8e72347f)

Also, CustomTypeThatAlsoExistsInTheSystemNamespaceAsArgument test now calls a proper benchmark


src/BenchmarkDotNet/Toolchains/InProcess.Emit.Implementation/Runnable/RunnableReuse.cs
```diff
@@ -102,7 +102,8 @@ private static IEngineFactory GetEngineFactory(BenchmarkCase benchmarkCase)
                 TargetJob = benchmarkCase.Job,
                 OperationsPerInvoke = benchmarkCase.Descriptor.OperationsPerInvoke,
                 MeasureGcStats = benchmarkCase.Config.HasMemoryDiagnoser(),
-                Encoding = benchmarkCase.Config.Encoding
+                Encoding = benchmarkCase.Config.Encoding,
+                BenchmarkName = FullNameProvider.GetBenchmarkName(benchmarkCase)
```
ig-sinicyn: Same as in https://github.com/dotnet/BenchmarkDotNet/blob/f10752ace1bf4c5e4316c6f9b7b34ca25ac46c19/src/BenchmarkDotNet/Toolchains/InProcess/InProcessRunner.cs#L151

tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
```diff
@@ -514,7 +514,7 @@ public void Test(DateTime passed)
         }
 
         [Theory, MemberData(nameof(GetToolchains))]
-        public void CustomTypeThatAlsoExistsInTheSystemNamespaceAsArgument(IToolchain toolchain) => CanExecute<WithDateTime>(toolchain);
+        public void CustomTypeThatAlsoExistsInTheSystemNamespaceAsArgument(IToolchain toolchain) => CanExecute<CustomTypeThatAlsoExistsInTheSystemNamespace>(toolchain);
```
adamsitnik: good catch!

PR #1044, opened 2019-02-03
Add "dotnet new" benchmark project template

### LINKED ISSUE #1028: Add new template for "dotnet new benchmark"
I think that we should add a new template for dotnet cli that would create following app when used:

```cs

namespace BenchmarkApp
{
    class Program
    {
    }

    public class TypeWithBenchmarks
    {
        [Benchmark]
        public int YourBenchmark()
        {
            // implement your benchmark here
            return 0;
        }
    }
}
```

CodeTherapist:
The template supports the following parameters:

benchmarktemplate

Usage example:

`dotnet new benchmark -n=MyCompany.Namespace.Benchmarks -b=MyBechmark -f="netstandard2.0;net472"`

The template is a library, not as proposed a `exe` - I suggest that users should take advantage of the dotnet global tool to run the benchmarks, instead of creating a console app.
The `config` parameter will conditionally add a benchmark config class and reference it inside the the benchmark class. 

- [x] Initial release
- [x] Documentation

The documentation will be delivered as soon the template is approved.

Closes #1028


CatGirlsAreLife:
Nice, can I test this template locally on Windows 10 build 18317 x64?

templates/BenchmarkDotNet.BenchmarkPorjectTemplate.CSharp/content/.template.config/template.json
```diff
@@ -0,0 +1,86 @@
...
+        "description": "The name of the benchmark class.",
+        "defaultValue": "MyBenchmark",
+        "FileRename" : "_BenchmarkName_",
+        "replaces": "$(BenchmarkName)"
+      },
+      "framework": {
+        "type": "parameter",
+        "datatype": "string",
+        "description": "The target framework for the project.",
+        "defaultValue": "netstandard2.0",
```
CatGirlsAreLife: What if they want to target multiple frameworks like ``netstandard2.0;netcoreapp2.0;netcoreapp2.1;netcoreapp2.2;netcoreapp3.0;net460;net461;net462;net47;net471;net472``?
I think this template should be aware of said ``;`` separated frameworks and also replace the ``<TargetFramework>$(Framework)</TargetFramework>`` of the xml with ``<TargetFrameworks>$(Framework)</TargetFrameworks>``

CodeTherapist:
@CatGirlsAreLife Yes, you could test this on windows. To do so install the template from the directory location like this:

`dotnet new -i [path-to-the-nuspec-directory-of-the-template]`

After that, you can use it as I described already.

templates/BenchmarkDotNet.BenchmarkPorjectTemplate.CSharp/content/.template.config/template.json
CodeTherapist: Hmm, basically I like the idea. On the other hand, I'm not sure if the user would understand how it behaves when the toolchains are explict set by attributes.

@adamsitnik What do you think is the best way? Has the attribute priority over the `csproj` configuration when BenchmarkDotNet is generating the benchmark project?

templates/BenchmarkDotNet.BenchmarkPorjectTemplate.CSharp/content/BenchmarkConfig.cs
```diff
@@ -0,0 +1,26 @@
...
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.Exporters.Csv;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+
+
+namespace _BenchmarkProjectName_
+{
+
```
YohDeadfall: nit: Empty line.

templates/BenchmarkDotNet.BenchmarkPorjectTemplate.CSharp/content/BenchmarkConfig.cs
```diff
@@ -0,0 +1,26 @@
+using BenchmarkDotNet.Analysers;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.Exporters.Csv;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+
+
```
YohDeadfall: nit: One of these two empty lines can be removed too.

templates/BenchmarkDotNet.BenchmarkPorjectTemplate.CSharp/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp.nuspec
```diff
@@ -0,0 +1,19 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+  <metadata>
+    <id>BenchmarkDotNet.BenchmarkProjectTemplate.CSharp</id>
+    <version>0.11.3</version>
+    <description>A Benchmark project template.</description>
+    <authors>BenchmarkDotNet</authors>
+    <copyright>.NET Foundation and contributors</copyright>
+    <tags>benchmark;benchmarking;performance;template</tags>
+    <iconUrl>https://raw.githubusercontent.com/dotnet/BenchmarkDotNet/master/docs/guide/logo.png</iconUrl>
+    <projectUrl>https://github.com/dotnet/BenchmarkDotNet</projectUrl>
```
Therzok:
```suggestion
    <projectUrl>https://benchmarkdotnet.org</projectUrl>
```

Therzok:
@adamsitnik @AndreyAkinshin is there interest in publishing this?

@CodeTherapist mind if I add some commits on top of yours in another PR?

templates/BenchmarkDotNet.BenchmarkPorjectTemplate.CSharp/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp.nuspec
CodeTherapist: @Therzok Thanks & Done!

CodeTherapist:
@Therzok For contributing and helping me out, very appreciated.

Added a VB.NET and a F# template as well. Even I'm primarily using C# to get cool things done - I decided to provide them with the exact same capability and behavior as the C# template.

After installation:

BDN-Templates

Building them as is:

BDN-Build

**Create the template nuget packages**

```
nuget pack BenchmarkDotNet.BenchmarkProjectTemplate.CSharp.nuspec
nuget pack BenchmarkDotNet.BenchmarkProjectTemplate.VB.nuspec
nuget pack BenchmarkDotNet.BenchmarkProjectTemplate.FSharp.nuspec
```

**Installing the template nuget packages**

```
dotnet new -i BenchmarkDotNet.BenchmarkProjectTemplate.CSharp
dotnet new -i BenchmarkDotNet.BenchmarkProjectTemplate.VB
dotnet new -i BenchmarkDotNet.BenchmarkProjectTemplate.FSharp
```

**Using the template**

```
dotnet new benchmark -n=MyCompany.Namespace.CSharp.Benchmarks -b=MyBechmark -c=true -f="netstandard2.0;net472" -lang C#
```
Accordingly, you can use `-lang F#` or `-lang VB`.

TheAngryByrd:
I had some issues with the F# template.  I sent a PR to your branch to fix them https://github.com/CodeTherapist/BenchmarkDotNet/pull/1

CodeTherapist:
> thanks for the pull request and sorry for such a late reply. I was pretty busy during the last several  months and didn't have enough time for BenchmarkDotNet.

No issue, we are all usually busy 👍 Thanks for your answers and your time, reviewing my PR.

I agree on all your statements except for number six: I think we should encourage the users to use the awesome dotnet tool (#213) to run the benchmarks quickly - I will add it as optional option. 
Beside that I think a first-class integration into the most used IDE's (VS, Rider and VS Code) would make sense and would increase acceptance to use benchmark dotnet.

AndreyAkinshin:
> I agree on all your statements except for number six: I think we should encourage the users to use the awesome dotnet tool

> Beside that I think a first-class integration into the most used IDE's (VS, Rider and VS Code) would make sense and would increase acceptance to use benchmark dotnet.

Yeah, it will be great. I already working on the advanced BenchmarkDotNet support in Rider.

CodeTherapist:
> - [x] 1. Currently, we have three different NuGet packages with templates which should be installed independently. It will be better to have only one NuGet package to them all.
> 
> - [x] 2. It's better to use the current version of BenchmarkDotNet in the csproj file instead of 0.1*. 
> 
> - [x] 3. Other templates (like console, classlib, xunit, mvc) use the --no-restore option for skipping restore. It would be nice to use the same name to be consistent.
> 
> - [x] 4. It will be better to reference BenchmarkDotNet.Diagnostics.Windows only if a benchmark is created on Windows.
> 
> - [x] 5. I prefer to use the current BenchmarkDotNet version as a template version (in the nuspec file). It will simplify syncing between templates and new versions of the library.
>
> - [ ] 6. Since we don't have proper support of BenchmarkDotNet projects in IDEs (VS, Rider) as we have for unit tests, it will be better to create a netcoreapp project by default with the following Main method:

@AndreyAkinshin Thank for your review - Everything done, except number 6.
I'm currently finishing the documentation.

templates/Content/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/.template.config/template.json
```diff
@@ -0,0 +1,86 @@
...
+      "Benchmark",
+      "Library"
+    ],
+    "precedence": "3000",
+    "tags": {
+      "language": "C#",
+      "type": "project"
+    },
+    "description": "A project template for creating benchmarks.",
+    "author": "BenchmarkDotnet",
```
AndreyAkinshin: In all templates, `"author"` should be `".NET Foundation and contributors"`.

templates/Content/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/_BenchmarkProjectName_.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
AndreyAkinshin: Currently, `dotnet build` on the generated template produces output in `bin\Debug` which may be confusing because we have optimized assemblies. It will be better to include `<Configuration>Release</Configuration>` in all csproj to avoid such kind of confusions.

CodeTherapist: Makes sense - the PR update will follow in few seconds...

AndreyAkinshin:
> except number 6

What do you think about "netcoreapp with Main method" as an option? (We can keep the existing behaviour with netstandard/classlib as the default)

CodeTherapist:
>> except number 6

> What do you think about "netcoreapp with Main method" as an option? (We can keep the existing behaviour with netstandard/classlib as the default)

Sure! Also thought about that and said that few comments earlier.
How about `--consoleapp` as options name?

AndreyAkinshin:
> How about --consoleapp as options name?

LGTM.

Could you please also merge master into your branch and resolve the conflicts?

AndreyAkinshin:
@CodeTherapist in general, it works great! I will merge the PR as soon as you introduce the `--consoleapp` option.

templates/Content/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/.template.config/template.json
```diff
@@ -0,0 +1,87 @@
...
+    "language": "C#",
+    "type": "project"
+  },
+  "description": "A project template for creating benchmarks.",
+  "author": ".NET Foundation and contributors",
+  "generatorVersions": "[1.0.0.0-*)",
+
+  "sourceName": "_BenchmarkProjectName_",
+  "preferNameDirectory": true,
+  "defaultName": "MyBenchmark",
```
AndreyAkinshin: One more thing: it confuses me that currently we have "YourBenchmark" method in "MyBenchmark" class. Could we rename the class from "MyBenchmark" to "BenchmarkSuite" and "YourBenchmark" to just "Benchmark"?

CodeTherapist: You are right - the current naming is bad and I didn't thought about it a lot when I started with this work.

I changed that in the latest PR update to this:

* The default project name is `BenchmarkSuite` (directory name is preferred).
* The default class name is `Benchmarks`.
* The default method name is `Scenario1`.

In my humble opinion, the Benchmarks are kind of "scenarios" that we try to measure and maybe compare with another one.

Let me know, if you think, that this is a good naming - as often, naming is hard.

CodeTherapist:
@AndreyAkinshin: The `--console-app` is implemented, docs are updated and conflicts are resolved.

The templates are not tested at the moment during the build. I think we should do this to ensure further changes do not break the templates. What do you think?

AndreyAkinshin:
@CodeTherapist everything works great! Thank you very much for your contribution!

ulex:
@CodeTherapist first of all, thanks for the template, very appreciated!
let me share the experience of using it. I've found this this issue when I was wondering why the default isn't a console application. The first place where I found the llink to the global tool was the current issue. It is nice, that there is a console tool for running benchmark, but there is no way to figure about it from documentation. I checked these places:
1. [Getting Started](https://benchmarkdotnet.org/articles/guides/getting-started.html) shows example with command-line application
2. [How to run your benchmarks](https://benchmarkdotnet.org/articles/guides/how-to-run.html) explictly states, that  
    > There are several ways to run your benchmarks. What is important is that BenchmarkDotNet works **only** with Console Apps.
3. Even the page about templates [didn't mention global tool at all](https://benchmarkdotnet.org/articles/guides/dotnet-new-templates.html)

IMO it seems like either more promotion of global tools is required, or default value should be changed.

GF-Huang:
I try to create project by command line `dotnet new benchmark`, it creates a library project, it's crazy to the docs!!!

image

Issue #715, opened 2018-04-10
"Scenarios" support - run given exe as benchmark, gather multiple results


adamsitnik:
An example: compile hello world, compile Roslyn with Roslyn.

More info soon

Issue #568, opened 2017-10-17
Provide support for running using an existing .NET exe


agocke:
So, here's my scenario:

I would like to use Benchmark.NET to measure the performance of the Roslyn compiler end-to-end, meaning my benchmark is basically `csc -noconfig @repro.rsp`.

I'll grant you, this isn't an obvious fit for BenchmarkDotNet, but technically I'm still benchmarking .NET code and there are a lot of things, like providing easy swapping of runtimes, that BenchmarkDotNet is great for. Perhaps most importantly, I don't really want to redo all the statistical analysis that's already built into BenchmarkDotNet.

So anyway, the problem is that I essentially need to do a process launch. The normal operation of BenchmarkDotNet would do two process launches, but I've managed to kind of get this working using the InProcess running mode (and manually working around the default timeout -- why is there a timeout on the InProcess mode??), but this is really hacked up.

Anyway, here's my proposal:

We should add support for the `BenchmarkAttribute` on properties and fields of type `ProcessStartInfo`. The contract here is that `ProcessStartInfo` would contain an arguments array that includes as the first argument the path to the managed EXE. BenchmarkDotNet would then plug in a process executor that would slot in the appropriate runtime and runtime options and then execute the given StartInfo. Other things would work somewhat analogously.

This would allow me to keep all my code in the external EXE, which is much easier to compile standalone, and also easily allow me to swap out runtimes.

shargon:
Should be fine if we can do all on the same execution (without new process). For visual studio profile performance is very useful

PR #1050, opened 2019-02-05
Scenarios support

adamsitnik:
This PR addresses #568 and #715

Sometimes we want to benchmark the total execution time of some long-running, stand-alone operation. Example: How long does it take to execute Hello World program using .NET Core 2.0? 3.0? CoreRT? Mono?

In BDN we have all we need to make it work, I decided to give it a try and with a few hours of work I was able to implement the `[Scenario]` support.

Idea:

- it's a new Kind of benchmarks
- all the auto-generated code does is simply calling the provided benchmark once
- the time is not reported by the child process but by the host - it's the total time it took to run the process
- no warmup
- no overhead deduction
- no heuristic - we run `launchCount`- many times

For the following code:

```cs
public class MyFirstScenario
{
    [Scenario]
    public void HelloWorld() => Console.WriteLine("Hello World!");
}
```

BDN generates:

```cs
namespace BenchmarkDotNet.Autogenerated
{
    public class UniqueProgramName : global::BenchmarkDotNet.Samples.MyFirstScenario
    {
        public static void Main()
        {
            BenchmarkDotNet.Autogenerated.UniqueProgramName instance = BenchmarkDotNet.Autogenerated.UniqueProgramName();
            
            instance.HelloWorld();
        }
    }
}

```


adamsitnik:
Sample results: How long does it take to execute Hello World application

|     Method | Runtime |             Toolchain |      Mean |     Error |    StdDev | Ratio | RatioSD |
|----------- |-------- |---------------------- |----------:|----------:|----------:|------:|--------:|
| HelloWorld |     Clr |                net461 |  53.92 ms |  3.025 ms |  3.484 ms |  1.00 |    0.00 |
| HelloWorld |    Core |         netcoreapp2.0 |  69.55 ms |  1.528 ms |  1.760 ms |  1.29 |    0.06 |
| HelloWorld |    Core |         netcoreapp2.1 |  83.15 ms | 14.758 ms | 16.995 ms |  1.53 |    0.18 |
| HelloWorld |    Core |         netcoreapp3.0 |  87.95 ms |  2.210 ms |  2.545 ms |  1.63 |    0.06 |
| HelloWorld |  CoreRT | Core RT 1.0.0-alpha-* |  42.13 ms | 17.289 ms | 19.910 ms |  0.77 |    0.26 |
| HelloWorld |    Mono |               Default | 135.67 ms | 20.810 ms | 23.965 ms |  2.50 |    0.24 |

adamsitnik:
@AndreyAkinshin what do you think about this idea? PTAL at the PR

jorive:
@adamsitnik Few suggestions that I think can make the scenario feature more extensible:
1. Allow passing command line arguments to the scenario. ~~`public static void Main()`~~ -> `public static void Main(string[] args)`, then forward it to the scenario.
2. Allow returning the scenario exit code: ~~`instance.HelloWorld();`~~ -> `return instance.HelloWorld();`.
3. Enable running already built processes (do not assume the source code can be modified).

adamsitnik:
> Allow returning the scenario exit code

this is already implemented, but not shown in the example. The Scenario can return `void`, `int`, `Task`, `Task<int>` and `ValueTask<int>`

https://github.com/dotnet/BenchmarkDotNet/blob/2e2b2ca0c304c9e38ceeaab0aad967b19b46049c/src/BenchmarkDotNet/Templates/ScenariosBenchmarkProgram.txt#L26-L31

> Allow passing command line arguments to the scenario. 

I was thinking about it, but I decided that it will be better to allow using the built-in `[Arguments]` and `[ArgumentsSource]` and `[Params]` instead of introducing such option. 

> Enable running already built processes

I agree that it would be great to have it, however I simply don't have enough time to implement is as for now.

adamsitnik:
I am going to close this PR as it looks like there is no need for it as of today.

@AndreyAkinshin please  don't remove the `scenarios` branch, maybe one day we are going to need it ;p

PR #1055, opened 2019-02-07
Allow reflecting on DebuggableAttribute on CoreRT

MichalStrehovsky:
See https://github.com/dotnet/corert/pull/6955#discussion_r254766270.

BenchmakeDotNet reflects on this attribute here:

https://github.com/dotnet/BenchmarkDotNet/blob/a76f438a81e3e0235a2e9b1249a5280038189689/src/BenchmarkDotNet/Extensions/AssemblyExtensions.cs#L28-L38

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/CoreRt/Generator.cs
+++ b/src/BenchmarkDotNet/Toolchains/CoreRt/Generator.cs
@@ -175,7 +175,8 @@ $@"<?xml version=""1.0"" encoding=""utf-8""?>
 </Project>";
 
         /// <summary>
-        /// mandatory to make it possible ot call GC.GetAllocatedBytesForCurrentThread() using reflection (not part of .NET Standard)
+        /// mandatory to make it possible to call GC.GetAllocatedBytesForCurrentThread() using reflection (not part of .NET Standard)
+        /// and 
         /// </summary>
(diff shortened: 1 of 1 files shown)


jkotas:
Benchmark.NET is targeting .NET Standard 2.0:  https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/BenchmarkDotNet.csproj#L5. These methods are available in .NET Standard 2.0.

Would it be better to just avoid the reflection alltogether?

adamsitnik:
@MichalStrehovsky big thanks for the PR! I am glad I am not the only user of our CoreRtToolchain anymore!

@jkotas you are right, the comments in the code are from the dnx or netcoreapp1.0 era ;)

adamsitnik:
I am going to merge this now and fix the way we use this attribute.

I am going to also add one more integration test to make sure we support the latest version of CoreRT.

So far our tests were running against const version of CoreRT:

https://github.com/dotnet/BenchmarkDotNet/blob/6fb83099144462986bf9b9455c16bfd4a128e01e/tests/BenchmarkDotNet.IntegrationTests/CoreRtTests.cs#L27

MichalStrehovsky:
> I am glad I am not the only user of our CoreRtToolchain anymore!

You should add telemetry. I'm sure there's many of us :).

PR #1057, opened 2019-02-08
CoreRT toolchain improvements

adamsitnik:


MichalStrehovsky:
This is failing because ILC didn't get a complete assembly closure (there's a field of a type that lives in `PresentationFramework` assembly, but `PresentationFramework` assembly wasn't specified as input to ILC.

This is bad input, but I submitted dotnet/corert#6970 to make ILC not fail. The compiler can handle bad inputs otherwise (we'll generate throwing methods for things that reference such unresolvable assemblies to simulate what CLR would do).

adamsitnik:
@MichalStrehovsky thanks for providing this fix! when can I expect a new NuGet package with this fix to be published to myget.org?

MichalStrehovsky:
> when can I expect a new NuGet package

We build twice a day, but I triggered a new build manually to get it faster: 1.0.0-alpha-27408-02 should have the fix.

adamsitnik:
@MichalStrehovsky with the new default settings it takes around 5 minutes to build a simple benchmarking program and our CI times out

I read the docs that you mentioned https://github.com/dotnet/corert/blob/master/Documentation/using-corert/optimizing-corert.md

I would like to disable the new default behavior and make it configurable. 

@MichalStrehovsky do you think that using following settings as defaults is a good idea from BDN point of view? We don't need reflection.

```xml
<RootAllApplicationAssemblies>false</RootAllApplicationAssemblies>
<IlcGenerateCompleteTypeMetadata>false</IlcGenerateCompleteTypeMetadata>
<IlcGenerateStackTraceData>false</IlcGenerateStackTraceData>
```

MichalStrehovsky:
This should be enough to prevent the explosion of things we build: `<RootAllApplicationAssemblies>false</RootAllApplicationAssemblies>`

`<IlcGenerateCompleteTypeMetadata>` makes troubleshooting reflection-related errors easier and only adds a couple percent to size. `<IlcGenerateStackTraceData>` makes exception stack traces better at the cost of another couple percent. That said, disabling those makes everything smaller. Smaller is more cache friendly, hence better perf.

adamsitnik:
All CIs except of recently unstable Travis are green so I am merging this fix.

@MichalStrehovsky big thanks for the help!

The default settings are now:

```xml
<RootAllApplicationAssemblies>false</RootAllApplicationAssemblies>
<IlcGenerateCompleteTypeMetadata>true</IlcGenerateCompleteTypeMetadata>
<IlcGenerateStackTraceData>true</IlcGenerateStackTraceData>
```

and are customizable from the `CoreRtToolchain.CreateBuilder()` level

MichalStrehovsky:
🚀 Thank you!

PR #1058, opened 2019-02-08
[Proof of Concept][WIP] Clean CPU cache - Run each iteration of test on different instance of test object.

WojciechNagorski:
This is different solution for #553. I will continue working on this PR If this solution is approved by you. 

**Previous solution**
Last PR #971 doesn't work because:
1. You can't clear cache by allocation or accessing pre-allocated memory because you never know which cache line will be overwritten. 
2. Test is running many times in one iteration of the test. I mean:
https://github.com/dotnet/BenchmarkDotNet/blob/190b9be0f4e1b1919943ae07c91dc2fbfd087fe0/src/BenchmarkDotNet/Engines/Engine.cs#L162
and 
https://github.com/dotnet/BenchmarkDotNet/blob/190b9be0f4e1b1919943ae07c91dc2fbfd087fe0/src/BenchmarkDotNet/Templates/BenchmarkType.txt#L138-L141
Because of that cleaning cache should be run before each run of `workloadDelegate`. Not only first one.

**New solution**
In this PR I launch each run of `workloadDelegate` on different instance of test object. This approach causes CPU cache is invalidated during each test run.

**Examples**
I created `IntroColdCpuCache` In `BenchmarkDotNet.Samples`. You can run this test:
```ini
cd c:\Work\BenchmarkDotNet\samples\BenchmarkDotNet.Samples
λ dotnet run -c Release -f netcoreapp2.1 -- --filter *IntroColdCpuCache*
```

There is the result:
```log
                                                                       Method | CleanCache |        Mean |       Error |      StdDev | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op | CacheMisses/Op |
----------------------------------------------------------------------------- |----------- |------------:|------------:|------------:|------:|--------:|------------:|------------:|------------:|--------------------:|---------------:|
         'Clean Cache affect performance, Benchmark doesn't make allocation.' |      False |    800.1 ns |    616.1 ns |    33.77 ns |  1.00 |    0.00 |           - |           - |           - |                   - |              0 |
         'Clean Cache affect performance, Benchmark doesn't make allocation.' |       True |  3,574.6 ns | 11,392.3 ns |   624.45 ns |  4.45 |    0.59 |           - |           - |           - |                   - |             47 |
                                                                              |            |             |             |             |       |         |             |             |             |                     |                |
         'Clean Cache doesn't affect performance, Benchmark make allocation.' |      False | 11,609.9 ns | 20,018.0 ns | 1,097.25 ns |  1.00 |    0.00 |     18.8679 |           - |           - |             41016 B |              2 |
         'Clean Cache doesn't affect performance, Benchmark make allocation.' |       True | 13,820.3 ns | 41,590.5 ns | 2,279.71 ns |  1.20 |    0.26 |     17.2414 |           - |           - |             41016 B |             29 |
                                                                              |            |             |             |             |       |         |             |             |             |                     |                |
 'Clean Cache doesn't affect performance, Benchmark doesn't make allocation.' |      False |    571.3 ns |    481.4 ns |    26.38 ns |  1.00 |    0.00 |           - |           - |           - |                   - |              0 |
 'Clean Cache doesn't affect performance, Benchmark doesn't make allocation.' |       True |    718.5 ns |    650.2 ns |    35.64 ns |  1.26 |    0.09 |           - |           - |           - |                   - |              4 |
```
**TODO**
There is still a lot to do:
- [ ] Support of InProc
- [ ] Support of test method with parameters
- [ ] Support of `GlobalSetup`, `GlobalCleanup`, `IterationSetup`, `IterationCleanup`
- [ ] Check method returning `Task` or `Task<>`
- [ ] Check method returning `ref` value
- [ ] Check all diagnosers


wojtpl2:
All tests are passed on `Travis CI` build! `AppVeyor` build failed for `InProcessBenchmarkEmitsSameIL` test, because I haven't changed `InProcess` yet. Before I start investing more time in this task, I would like you to confirm my concept.

adamsitnik:
@AndreyAkinshin could you provide some feedback here?

AndreyAkinshin:
@wojtpl2 sorry for the late reply, I was pretty busy lately.

In general, I don't like this approach because it complicates our template and it may also introduce unexpected performance phenomena because of the additional `ChangeObjectUnderTest()` call. Also, a huge number of instances in the `ObjectsUnderTest` array may also affect the GC behavior.

Unfortunately, I don't know a better solution for now. I need some time to play with cache clearing and performing my own experiments.

wojtpl2:
@AndreyAkinshin We can introduce a new template for benchmarks with a clean CPU cache. Thanks to this, we won't change old behavior for normal tests. @adamsitnik suggested a similar solution in #1050. I mean file `ScenariosBenchmarkProgram.txt`.

adamsitnik:
@wojtpl2 I am not sure if we should implement the CPU cache cleanup strategies. It looks like your experiments have shown that it simply does not affect the results? I know that it sounds bad after you have spent so much time working on it. I hope that it's not going to affect your contributing enthusiasm.

@AndreyAkinshin do you agree?

wojtpl2:
@adamsitnik Now it works. There is result:
```
                                                                       Method | CleanCache |        Mean |       Error |      StdDev | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op | CacheMisses/Op |
----------------------------------------------------------------------------- |----------- |------------:|------------:|------------:|------:|--------:|------------:|------------:|------------:|--------------------:|---------------:|
         'Clean Cache affect performance, Benchmark doesn't make allocation.' |      False |    800.1 ns |    616.1 ns |    33.77 ns |  1.00 |    0.00 |           - |           - |           - |                   - |              0 |
         'Clean Cache affect performance, Benchmark doesn't make allocation.' |       True |  3,574.6 ns | 11,392.3 ns |   624.45 ns |  4.45 |    0.59 |           - |           - |           - |                   - |             47 |
 ```
But I agree with @AndreyAkinshin that this solution has some weaknesses. E. g.  How it affects the GC behavior.

I learned a lot in this task. Nothing will happen if you close this task. I know that this happens sometimes.

adamsitnik:
@WojciechNagorski once again big thanks for you work on this.

I am going to close it as it looks like it would not introduce any visible benefits

PR #1061, opened 2019-02-11
Proper cleanup on Ctrl+C/console Window exit

adamsitnik:
1. don't use native windows events to detect shutdown, switch to 100% full managed implementation
2. don't allow MSBuild to create zombie processes by disabling parallel builds


adamsitnik:
Motivation: get rid of:

"A callback was made on a garbage collected delegate of type 'BenchmarkDotNet!BenchmarkDotNet.Helpers.NativeWindowsConsoleHelper+ConsoleCtrlHandler::Invoke'"

on Ctrl+C

adamsitnik:
@AndreyAkinshin I have tested all possible crazy scenarios and it works properly in every single case that I was able to come up with.

I know that disabling parallel builds with MSBuild does not sound great, but every time we build a very small console app and it should not hurt the performance (the time it takes to run the CI tests stayed the same). I did it because it prevents MSBuild from spawning some extra processes that were so far always remaining alive after the Ctrl+C . Now we have 0 zombies no matter what ;)

PR #1080, opened 2019-02-23
Improved Environment Variables API, resolves #1069

### LINKED ISSUE #1069: Improve API for job environment variables
Consider the following job:

```cs
var job = new Job()
    .With(new[] { new EnvironmentVariable("a", "b") })
    .With(new[] { new EnvironmentVariable("c", "d") });
```

The corresponding `With` method is defined as follows:

```cs
public static Job With(this Job job, IReadOnlyList<EnvironmentVariable> environmentVariables) =>
    job.WithCore(j => j.Environment.EnvironmentVariables = environmentVariables);
```

It means that it always override the whole `EnvironmentVariables` list. In the above example, `job` has only one a single environment variable `"c"`. I think that this behavior looks pretty confusing. It will be better to add new environment variables instead of rewriting the whole list.

I want to make a breaking change and replace the "overriding" logic by "adding" logic. However, we also should provide a way to remove an environment variable from a job. I was thinking about the following API:

```cs
// Add the specified environment variables
With(this Job job, IReadOnlyList<EnvironmentVariable> environmentVariable)
// Add the specified environment variable
WithEnvironmentVariable(this Job job, EnvironmentVariable variable)
WithEnvironmentVariable(this Job job, string key, string value)
// Remove the specified environment variable
WithoutEnvironmentVariable(this Job job, string key)
// Remove all environment variables
WithoutEnvironmentVariables(this Job job)
```

Also, if an environment variable is previously defined, it should be overridden by the new value (instead of having two variables with the same key in the `EnvironmentVariables` list).

AndreyAkinshin:


src/BenchmarkDotNet/Jobs/EnvironmentVariable.cs
```diff
@@ -1,17 +1,43 @@
-﻿namespace BenchmarkDotNet.Jobs
+﻿using System;
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Jobs
 {
     public class EnvironmentVariable
```
adamsitnik: missing `IEquatable` interface implementation 

```suggestion
    public class EnvironmentVariable : IEquatable<EnvironmentVariable>
```

src/BenchmarkDotNet/Jobs/JobExtensions.cs
```diff
@@ -180,8 +180,57 @@ public static class JobExtensions
...
+        /// <returns>The new job with additional environment variable</returns>
+        public static Job WithEnvironmentVariable(this Job job, [NotNull] string key, [NotNull] string value)
+            => job.With(new EnvironmentVariable(key, value));
+
+        /// <summary>
+        /// Creates a new job based on the given job without any environment variables.
+        /// </summary>
+        /// <param name="job">The original job</param>
+        /// <returns>The new job which doesn't have any environment variables</returns>
+        public static Job WithoutEnvironmentVariables(this Job job) => job.With(new EnvironmentVariable[0]);
```
adamsitnik: nit: `Array.Empty<T>()` does not allocate, `new EnvironmentVariable[0]` allocates new empty array

```suggestion
        public static Job WithoutEnvironmentVariables(this Job job) => job.With(Array.Empty<EnvironmentVariable>());
```

tests/BenchmarkDotNet.Tests/Configs/EnvironmentVariableTests.cs
```diff
@@ -0,0 +1,81 @@
...
+            Assert.Throws<ArgumentNullException>(() => new EnvironmentVariable(null, null));
+        }
+
+        [Fact]
+        public void ClearVariables()
+        {
+            var job = new Job()
+                .WithEnvironmentVariable("a", "b")
+                .WithoutEnvironmentVariables();
+            Assert.Equal(0, job.Environment.EnvironmentVariables.Count);
```
adamsitnik: nit: `Assert.Empty` is more readable

```suggestion
            Assert.Empty(job.Environment.EnvironmentVariables);
```

AndreyAkinshin:
@adamsitnik thanks for the review.

PR #1090, opened 2019-03-04
XmlExporter.Full fails

### LINKED ISSUE #1086: XmlExporter.Full fails with StackOverflowException
Running a Benchmark with the fullxml exporter ("-e fullxml") causes the benchmark process to terminate with the message:
`Process is terminated due to StackOverflowException.`

Benchmark code (though it doesn't seem to matter, failed with all benchmarks I tried):
```csharp
public class BenchmarkSample
{
    [Benchmark]
    public int Parse() => int.Parse("1");
}
```

Version: 0.11.4

daveMueller:
Fixes #1086


adamsitnik:
@daveMueller thank you!

PR #1093, opened 2019-03-06
make InProcessEmitToolchain the default one

adamsitnik:
Today I saw some people using `[InProcessAttribute]` and I realized that it's still using the old one.

I think that we should switch by default to the new one.

### DIFF
--- a/docs/articles/samples/IntroInProcess.md
+++ b/docs/articles/samples/IntroInProcess.md
@@ -4,10 +4,9 @@ uid: BenchmarkDotNet.Samples.IntroInProcess
 
 ## Sample: IntroInProcess
 
-InProcessToolchain is our toolchain which does not generate any new executable.
+InProcessEmitToolchain is our toolchain which does not generate any new executable.
 It emits IL on the fly and runs it from within the process itself.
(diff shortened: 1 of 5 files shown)


ig-sinicyn:
@adamsitnik What to do with the old InProcessToolchain?

It has two codegen branches.
* BenchmarkActionCodegen.ReflectionEmit that uses dynamic methods and is obsolete now.
* BenchmarkActionCodegen.DelegateCombine that intended to work on platforms that do not support IL emitting.

Should I keep bot branches, remove the first one or remove the InProcessToolchain entirely?

adamsitnik:
@ig-sinicyn that is a very good question.

Maybe we should:

- create new `InProcessNoEmitToolchain` based on `BenchmarkActionCodegen.DelegateCombine`
- add an `[Obsolete]` comment to `InProcessToolchain` and explain that there is `InProcessEmitToolchain` and `InProcessNoEmitToolchain` which should be used instead? Make it a warning for one release and then switch to error after another release.

What do you think?

ig-sinicyn:
Agreed:)
Will do the `InProcessNoEmitToolchain` on weekends.

PR #1097, opened 2019-03-10
Allow for Config per method, introduce OS and OSArchitecture filters

adamsitnik:
in the dotnet/performance repo, I need to enable/disable some benchmarks per OS/architecture/something else like AVX2.IsAvailable. As of today, we allow to enable the configs per entire type, so if I want to disable one benchmark out of N for Linux, I have to move it to a dedicated type and apply some config to it via attributes. This is not very user friendly. 

My proposal: **allow for configs per methods**.

I have changed how `BenchmarkConverter` works and now it looks for the method attributes as well.

What is very important, I did not change `AttributeTargets` flags for any of our Attributes except of `FilterConfigBaseAttribute` and `JobMutatorConfigBaseAttribute` so basically this feature currently works only for Filter and Job attributes.

/cc @ig-sinicyn


src/BenchmarkDotNet/Attributes/Filters/OperatingSystemsFilterAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        // CLS-Compliant Code requires a constructor without an array in the argument list
+        public OperatingSystemsFilterAttribute() { }
+
+        /// <param name="allowed">if set to true, the OSes beloning to platforms are enabled, if set to false, disabled</param>
+        public OperatingSystemsFilterAttribute(bool allowed, params PlatformID[] platforms)
+            : base(new SimpleFilter(_ =>
+            {
+                return allowed
+                    ? platforms.Any(platform => RuntimeInformation.IsOSPlatform(Map(platform)))
```
eerhardt: Maybe you could take the `string` of a OSPlatform instead of `PlatformID`? (basically the string that gets passed into https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.osplatform.create)

I don't think PlatformID is going to work.

From https://docs.microsoft.com/en-us/dotnet/api/system.platformid

> MacOSX | 6 | The operating system is Macintosh. This value was returned by Silverlight. On .NET Core, its replacement is Unix.

From that, it sounds like you won't be able to distinguish between macOS and Linux on .NET Core using PlatformID.

adamsitnik: @eerhardt I am not sure about string. For example: let's say that I dont't want to run the benchmark on Ubuntu. Should I put `Unix`? `Linux`? or `Ubuntu`? or `Ubuntu16.04`?

But I agree that PlatformID is not perfect, I had to implement a mapping on my own.

Maybe I should just introduce a new enum?

eerhardt: > Should I put Unix? Linux? or Ubuntu? or Ubuntu16.04?

Do you want that deep of control? Do you want to be able to say "All Unixes" vs. "just Ubuntu, but not RedHat or OpenSUSE"? If so, it seems like a perfect place to use RIDs...

adamsitnik: > Do you want that deep of control?

No, I just wanted to show that without an enum, the user needs to guess the right string value.

adamsitnik: It's been two years since I created this PR, but now I really need tha for Randomization in the perf repo so I've updated it. I decided to include my own simple enum, just so users don't have to type OS names manually (an avoid mistakes for "OSX" != "macOS" etc)

AndreyAkinshin (approved):
LGTM

PR #1103, opened 2019-03-16
possible fix for MosCpu Info Provider

idormenco:
Hi,
I have run some static code analysis tools other this project and found that `nominalClockSpeed`  was not updated .
I computed  `nominalClockSpeed` as minimum between current and maximum clock speed.

Comments and suggestions are more than welcome.
Hopefully it will help.

Best regards,
Ion Dormenco


adamsitnik:
@Rizzen @AndreyAkinshin could you please take a look?

AndreyAkinshin:
Hey @idormenco. Thanks for the PR and sorry for the late response.

Unfortunately, your fix is not correct. It seems that there is no way to get the values of `minClockSpeed` and `nominalClockSpeed` via `ManagementObjectSearcher`. `nominalClockSpeed` is not CurrentClockSpeed, and `minClockSpeed` is not `Min(nominalClockSpeed, maxClockSpeed)`. These values can be calculated in other CpuInfoProviders, but they are unknown in `MosCpuInfoProvider`.

idormenco:
Hi @AndreyAkinshin thanks for your reply .
I tried to help :) maybe next time .

Sorry for borrowing you .
Best regards,
Ion

Issue #1111, opened 2019-03-22
Change the format of printed Full .NET Framework Version


adamsitnik:
As of today BDN prints something like this:

```log
DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.8.3752.0
```

We have three numbers here: 4.7*, 4.0* and 4.8*.

I took a look at the code and the numbers represent the following:

```
.NET Framework {InstalledDotNetSdKVersion} (CLR {Environment.Version}), 64bit {ProcessModules.JitModuleVersion}
```

The `Environment.Version` does not provide any useful data, I  think that we should simplify the output to the following:

```
.NET Framework SDK 4.7.2 (CLR v4.8.3752.0), 64bit RyuJIT
```

@AndreyAkinshin do you agree?

/cc @billwert

AndreyAkinshin:
@adamsitnik 
1. Do we really need the "SDK" postfix in ".NET Framework SDK"? What about computers without installed SDK (InProcessToolchain should work fine there)
2. Are you sure that it's correct to write "CLR v4.8.3752"? Currently, all .NET Framework 4.x uses that same folder: `C:\Windows\Microsoft.NET\Framework\v4.0.30319\`

billwert:
For #2, The install directory will never change again. Using the version off of clrjit.dll or clr.dll is much more informative.

Thanks,
-Bill
________________________________

adamsitnik:
Fixed by #1114

PR #1114, opened 2019-03-25
simplify the reported Full .NET Framework version, fixes 1111

adamsitnik:
@AndreyAkinshin this is my proposal for #1111 

It looks like `Microsoft.DotNet.PlatformAbstractions.RuntimeEnvironment.FrameworkDescription` returns exactly what we need. 

I decided to remove:

- the SDK version: it was confusing, we don't run the benchmarks against installed SDK but always against latest .NET installed on the machine. Also, as you have noticed the SDK does not need to be installed to run a .NET app
- `System.Environment.Version` - this version number is confusing (even for people who work on .NET Team for 10+ years) because it requires one additional translation to map it to the .NET Framework Version
- `JitModule` version because it returns the same version as `Microsoft.DotNet.PlatformAbstractions.RuntimeEnvironment.FrameworkDescription` (duplication)

```cmd
dotnet run -c Release --info
```

Was:

```log
  [Host] : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3324.0
```

Is:

```log
  [Host] : .NET Framework 4.7.3324.0, 64bit RyuJIT
```


AndreyAkinshin:
In general, I like the idea of simplification in this place. However, I think that ".NET Framework 4.7.3324.0" also looks confusing. People know that there is .NET Framework 4.7, 4.7.1, and 4.7.2. In this context, "4.7.3324.0" looks strange. Can we do something with it?

P.S. Please, don't merge it until I approve this PR. I want to make sure that it looks great before we get these changes in master.

adamsitnik:
> Can we do something with it?

We would have to map the version number reported by .NET Framework to the simplified version. 
I don't think it's worth it. 

Also, I am not sure if my box does not have any internal MS-only preview version of .NET Framework installed. What numbers do you get on your box?

billwert:
".NET Framework 4.7.3324.0" is nice because it represents the servicing release that the user has installed. It has been very useful to me to have it. Keeping the string to this would be great:
```
[Host] : .NET Framework 4.7.2 (4.7.3324.0), 64bit RyuJIT
```

I think keeping the rough branding of 4.7.2 or 4.8 in the string is also useful. We talk about releases in those terms, and our [tools](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/application-compatibility) for determining what quirks are enabled in a given release use those terms. My $0.02.

adamsitnik:
@billwert thank you for your input! Now we got 2:1 for the mapping vs no mapping so I added it ;)

sample results now:

```cs
BenchmarkDotNet=v0.11.4.20190520-develop, OS=Windows 10.0.17763.437 (1809/October2018Update/Redstone5)
  [Host] : .NET Framework 4.7.2 (4.7.3362.0), 64bit RyuJIT
  Dry    : .NET Framework 4.7.2 (4.7.3362.0), 64bit RyuJIT
```

@AndreyAkinshin could you please take a look at this PR again?

btw I also added .NET 4.8 to the list of known .NET Runtimes

PR #1117, opened 2019-03-28
Add support for fully synchronous mode for WebAssembly

jeromelaban:
- `DefaultConfig.SupportsThreading` indicates if the current runtime supports threading. Returns false on current WebAssembly environments.
- `InProcessExecutor` and `InProcessToolchain` have an optional `synchronous` parameter to disable any use of threads.
- Expand the fallback catch in `GetVirtualMachineHypervisor` to ensure that a missing runtime assembly won't fail the report generation

### DIFF
--- a/src/BenchmarkDotNet/Configs/DefaultConfig.cs
+++ b/src/BenchmarkDotNet/Configs/DefaultConfig.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Text;
 using BenchmarkDotNet.Analysers;
 using BenchmarkDotNet.Columns;
@@ -87,5 +88,11 @@ namespace BenchmarkDotNet.Configs
         public IEnumerable<HardwareCounter> GetHardwareCounters() => Array.Empty<HardwareCounter>();
 
         public IEnumerable<IFilter> GetFilters() => Array.Empty<IFilter>();
+
+        /// <summary>
+        /// Determines if the current runtime supports threading.
+        /// </summary>
+        /// <remarks>This property returns <c>false</c> when running under WebAssembly, otherwise <c>true</c>.</remarks>
+        public static bool SupportsThreading { get; } = !RuntimeInformation.IsOSPlatform(OSPlatform.Create("WEBASSEMBLY"));
     }
 }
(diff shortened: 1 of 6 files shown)


jeromelaban:
This is a tentative PR for the support for Synchronous testing, which enables WebAssembly (mono-wasm, specifically) support.

One part I'm not sure is the location of `DefaultConfig.SupportsThreading`. I have not found a good place for this property, though it's only used at one place. Should be located inside the BenchmarkRunnerClean.cs file directly, as a private property ?

Thanks !

src/BenchmarkDotNet/Configs/DefaultConfig.cs
```diff
@@ -87,5 +88,11 @@ public IEnumerable<IValidator> GetValidators()
         public IEnumerable<HardwareCounter> GetHardwareCounters() => Array.Empty<HardwareCounter>();
 
         public IEnumerable<IFilter> GetFilters() => Array.Empty<IFilter>();
+
+        /// <summary>
+        /// Determines if the current runtime supports threading.
+        /// </summary>
+        /// <remarks>This property returns <c>false</c> when running under WebAssembly, otherwise <c>true</c>.</remarks>
+        public static bool SupportsThreading { get; } = !RuntimeInformation.IsOSPlatform(OSPlatform.Create("WEBASSEMBLY"));
```
adamsitnik: Could you please move it to our `RuntimeInformation` class?

https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Portability/RuntimeInformation.cs

src/BenchmarkDotNet/Toolchains/InProcess/InProcessExecutor.cs
```diff
@@ -54,69 +65,97 @@ public InProcessExecutor(TimeSpan timeout, BenchmarkActionCodegen codegenMode, b
...
+                }
+                finally
+                {
+                    process.TrySetPriority(oldPriority, parameters.Logger);
+                    thread.TrySetPriority(oldThreadPriority, parameters.Logger);
+
+                    if (affinity != null && oldAffinity != null)
+                    {
+                        process.TrySetAffinity(oldAffinity.Value, parameters.Logger);
+                    }
```
adamsitnik: I think that it would be better to rename the existing `ExecuteCore` method to `ExecuteCoreAsync` and introduce a new method for synchronous run. This big `if/else` block is hard to read.

adamsitnik:
@jeromelaban big thanks for your PR! How can we test it with web assembly?

jeromelaban:
@adamsitnik Thanks for the comments!

Testing with WebAssembly using mono's tooling is not particularly library friendly, which is why I'm working on https://github.com/nventive/Uno.Wasm.Bootstrap to allow for easy deployment of Wasm apps. There's no automated testing, though puppeteer sound like a good option there.

In [my fork contains a sample app](https://github.com/nventive/BenchmarkDotNet/blob/uno/samples/BenchmarkDotNet.Samples.Wasm/Program.cs), though as I don't know if you're willing to take such a dependency yet I did not include it in my PR, even if it's for a sample app :)

PR #1126, opened 2019-04-02
Styling in docfx fashion

robertmuehsig:
Instead of "@alert..." (which seems not to work as expected) use the DocFx flavor.

(Not sure what the actual change is in the last line - I just used the GitHub UI)


AndreyAkinshin:
@robertmuehsig thanks!

PR #1129, opened 2019-04-07
Consume CoreRT from the new NuGet feed

MichalStrehovsky:
CoreRT no longer publishes to MyGet.

See dotnet/corert#7268.


adamsitnik:
Thank you!

PR #1131, opened 2019-04-12
NativeMemoryDiagnoser

wojtpl2:
Fix for #457. I am open to all suggestions and comments. 

## Information

I added new columns: `Total native memory` and `Native memory leak`. Example of output:
```ini
|                           Method |       Mean |         Error |      StdDev |   Gen 0 |  Gen 1 |  Gen 2 | Allocated | Total native memory | Native memory leak |
|--------------------------------- |-----------:|--------------:|------------:|--------:|-------:|-------:|----------:|--------------------:|-------------------:|
|                 AllocateIntArray | 145.592 us | 1,873.6901 us | 102.7033 us |  0.1526 | 0.1526 | 0.1526 |         - |              40,000 |             40,000 |
|    AllocateAndDeallocateIntArray |   5.772 us |     0.8384 us |   0.0460 us |       - |      - |      - |         - |              40,000 |                  - |
|              AllocateStructArray |  66.127 us |   265.2075 us |  14.5369 us |  0.1831 | 0.1831 | 0.1831 |         - |              80,000 |             80,000 |
| AllocateAndDeallocateStructArray |  13.895 us |     3.7949 us |   0.2080 us |       - |      - |      - |         - |              80,000 |                  - |
|    AllocateAndDeallocateManyType | 140.390 us |    87.7633 us |   4.8106 us |       - |      - |      - |         - |             240,000 |            120,000 |
|                ManagedAllocation |  14.924 us |     6.9775 us |   0.3825 us | 12.6495 |      - |      - |   40024 B |                   - |                  - |
```
`NativeMemoryDiagnoser` can track all native allocations and memory leaks. This diagnoser also prints native type names. This is console output:
image

I compared the result with PerfView and I got the same result. 

## Sources
Thanks to @kayle and his https://github.com/Microsoft/perfview/pull/857 I could get the type of native objects. 

Code from `NativeMemoryDiagnoser` is inspired by https://github.com/Microsoft/perfview/blob/master/src/PerfView/PerfViewData.cs#L5719-L5944 

## How to run
I added a new project `BenchmarkDotNet.IntegrationTests.NativeAllocation` which contains all tests. 

You can run this project in this way:
```ini
cd tests\BenchmarkDotNet.IntegrationTests.NativeAllocation
dotnet run -c Release -- --filter *IntroNativeMemory*
```
I didn't want to put this test into `BenchmarkDotNet.Samples` because this project has to be compiled for the x64 platform.


wojtpl2:
Builds are failed because of the new C++ project.

adamsitnik:
Hi @wojtpl2 ! big thanks for the PR! it looks great!

The only thing that worries me is the new C++ project.

Would it be possible to test the new diagnoser by using `Marshal` type to allocate unmanaged memory?

```cs
IntPtr unmanagedHandle = Marshal.AllocHGlobal(256);
Marshal.FreeHGlobal(unmanagedHandle);
```

wojtpl2:
I'll check. I want to clear and remove some code from perfview. 

I want to also fix built but I need more time with it.

wojtpl2:
@adamsitnik `Marshal.AllocHGlobal` and `Marshal.FreeHGlobal` also work. 

```ini
|                         Method |      Mean |      Error |    StdDev |   Gen 0 | Gen 1 | Gen 2 | Allocated | Total native memory | Native memory leak |
|------------------------------- |----------:|-----------:|----------:|--------:|------:|------:|----------:|--------------------:|-------------------:|
|                  AllocIntArray |  46.65 us | 161.566 us | 8.8560 us |       - |     - |     - |         - |              40,000 |             40,000 |
|           AllocAndFreeIntArray |  32.69 us |  24.865 us | 1.3630 us |       - |     - |     - |         - |              40,000 |                  - |
|               AllocStructArray | 127.63 us | 100.791 us | 5.5247 us |       - |     - |     - |         - |              80,000 |             80,000 |
|        AllocAndFreeStructArray |  88.10 us |  15.757 us | 0.8637 us |       - |     - |     - |         - |              80,000 |                  - |
|           MarshalAllocIntArray |  46.02 us |  31.281 us | 1.7146 us |       - |     - |     - |         - |              40,000 |             40,000 |
|    MarshalAllocAndFreeIntArray |  25.20 us |  22.858 us | 1.2529 us |       - |     - |     - |         - |              40,000 |                  - |
|        MarshalAllocStructArray | 110.16 us |  74.743 us | 4.0969 us |       - |     - |     - |         - |              80,000 |             80,000 |
| MarshalAllocAndFreeStructArray |  80.83 us |   9.791 us | 0.5367 us |       - |     - |     - |         - |              80,000 |                  - |
|           AllocAndFreeManyType | 274.48 us |  45.900 us | 2.5159 us |       - |     - |     - |         - |             240,000 |            120,000 |
|                   ManagedAlloc |  16.76 us |   1.835 us | 0.1006 us | 12.6343 |     - |     - |   40024 B |                   - |                  - |
```

`Marshal.AllocHGlobal` allocate only space without any type. Output from `NativeMemoryDiagnoser` for this memory looks:
image

For a memory that is allocated from C++, `NativeMemoryDiagnoser` prints the type names:
image

wojtpl2:
@adamsitnik https://github.com/Microsoft/perfview also have C++ projects:
image

I'll take care of it after the 24 of April.

adamsitnik:
@wojtpl2 thanks!

kayle:
Thanks for following up on that PerfView commit. I had started thinking about how to track allocations in benchmarks, but haven't had time to submit a PR. Here's a few questions/considerations I had for this area:

1. The docs indicate that heap profiling will likely cause severe slowdowns. Is it helpful/possible to force a single iteration with no warmup? Maybe hide the time metrics entirely or warn that they shouldn't be used?
1. Would it be better to have separate managed/native diagnosers, or a single diagnoser that can optionally collect both types of information. In theory, the managed type allocations are easier to track. I had started a prototype here (https://github.com/kayle/BenchmarkDotNet/commit/7d4a2379089794974e1c6644d92c4bfe14500b50)
1. Regarding adding a c++ project to this repository, I suspect there's a pinvoke available that would allocate on the native heap (or picking a class that uses SafeHandle internally since those need to free native resources).
1. How much disk I/O happens when collecting these ETW events? Ideally, we could process ETW with real-time listeners by default, and optionally save the ETL to disk for manual inspection.
1. I'm guessing a common case is to expect 0 allocations during a benchmark. It'd be nice to offer a simple pass/fail option for this case.
1. Leak tracking seems like more of a correctness issue than performance. I know you already did the hard part of the work, but it might be simpler and less work for this repo to only track raw allocations. To diagnose leaks, users could load the ETL file in perfview.

I'm happy to help out here. Let me know if there's a specific task that I should look closer at.

wojtpl2:
I've finally found some time. I've had a hard time.

I moved `IntroNativeMemory.cs` into `BenchmarkDotNet.Samples` and I used `Bitmap` and `Graphics` from `System.Drawing` to make native allocation instead of using custom `C++` dll. 

I created two tests:
```C#
        [Benchmark]
        public void BitmapWithLeaks()
        {
            var flag = new Bitmap(200, 100);
            var graphics = Graphics.FromImage(flag);
            var blackPen = new Pen(Color.Black, 3);
            graphics.DrawLine(blackPen, 100, 100, 500, 100);
        }

        [Benchmark]
        public void Bitmap()
        {
            using (var flag = new Bitmap(200, 100))
            {
                using (var graphics = Graphics.FromImage(flag))
                {
                    using (var blackPen = new Pen(Color.Black, 3))
                    {
                        graphics.DrawLine(blackPen, 100, 100, 500, 100);
                    }
                }
            }
        }
```
Below is the result of these tests. 
image

image

****TODOs****
1. I have to compare this result with `PerfView`. 

****Additional information****
Thanks @kayle for your suggestions. Bellow are my answers:
>The docs indicate that heap profiling will likely cause severe slowdowns. Is it helpful/possible to force a single iteration with no warmup? Maybe hide the time metrics entirely or warn that they shouldn't be used?
1. NativeMemoryDiagnoser is used only in the extra diagnostic run. I set `performExtraBenchmarksRun=true` in `EtwProfilerConfig`.
>Would it be better to have separate managed/native diagnosers, or a single diagnoser that can optionally collect both types of information. In theory, the managed type allocations are easier to track. I had started a prototype here (kayle@7d4a237)
2. Currently managed memory diagnoser does not use ETW at all. If we create new ETW Managed memory diagnoser then BenchmarkDotNet will have two different diagnosers. I'm not sure if it will be accepted by @adamsitnik or @AndreyAkinshin.
>Regarding adding a c++ project to this repository, I suspect there's a pinvoke available that would allocate on the native heap (or picking a class that uses SafeHandle internally since those need to free native resources).
3. It is a good suggestion. I used the `System.Drawing` instead of custom `C++` project. I'll try to find some example with native type. 
>How much disk I/O happens when collecting these ETW events? Ideally, we could process ETW with real-time listeners by default, and optionally save the ETL to disk for manual inspection.
4. It is a good suggestion. First I was trying to do real-time listeners but I didn't have all the information. Now I think it will be easy to do. I'll try.
>I'm guessing a common case is to expect 0 allocations during a benchmark. It'd be nice to offer a simple pass/fail option for this case.
5. It is a good idea. I can do it but first I would like to finish current PR.
>Leak tracking seems like more of a correctness issue than performance. I know you already did the hard part of the work, but it might be simpler and less work for this repo to only track raw allocations. To diagnose leaks, users could load the ETL file in perfview.
6. I think that @adamsitnik or @AndreyAkinshin should comment on this point. I will not have a problem removing this feature. But for me, the best solution will be tracking memory leaks without information about type. In this case, I could remove a big part of the code.

wojtpl2:
I see that appveyor build on master does not pass. There are some failing tests:
image

src/BenchmarkDotNet/Diagnosers/NativeMemoryDescriptor.cs
```diff
@@ -0,0 +1,33 @@
...
+// BenchmarkDotNetBenchmarkDotNet
+// NativeMemoryDescriptor.cs
+// Copyright  - 2019
+
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Reports;
+
+namespace BenchmarkDotNet.Diagnosers
+{
+    internal class NativeMemoryDescriptor : IMetricDescriptor
```
adamsitnik: nit: it would be nice to rename it to `AllocatedNativeMemoryDescriptor` to be consistent with the existing `AllocatedMemoryMetricDescriptor`

src/BenchmarkDotNet/Diagnosers/NativeMemoryDescriptor.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public string Id => nameof(NativeMemoryDescriptor);
+        public string DisplayName => $"Total native memory";
```
adamsitnik:
```suggestion
        public string DisplayName => $"Allocated native memory";
```

src/BenchmarkDotNet/Diagnosers/NativeMemoryDescriptor.cs
```diff
@@ -0,0 +1,33 @@
...
+    internal class NativeMemoryLeakDescriptor : IMetricDescriptor
+    {
+        public string Id => nameof(NativeMemoryLeakDescriptor);
+        public string DisplayName => $"Native memory leak";
+        public string Legend => $"Native memory leak size in byte.";
+        public string NumberFormat => "N0";
+        public UnitType UnitType => UnitType.Dimensionless;
+        public string Unit => "Count";
+        public bool TheGreaterTheBetter => false;
+    }
```
adamsitnik: same as above: it would be nice to use Byte and Size for this metric as well

tests/BenchmarkDotNet.IntegrationTests/BenchmarkDotNet.IntegrationTests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
adamsitnik: nit: could you please revert this change?

src/BenchmarkDotNet.Diagnostics.Windows/EtwProfiler.cs
```diff
@@ -102,32 +101,44 @@ private void Start(DiagnoserActionParameters parameters)
...
+                if (parameters.Process?.StartInfo.FileName != null)
+                {
+                    foreach (var provider in config.Providers)
+                    {
+                        provider.options.ProcessNameFilter = new List<string>() { parameters.Process?.StartInfo.FileName };
+                    }
+                }
+
                 userSession = new UserSession(parameters, config, CreationTime).EnableProviders();
+                heapSession = new HeapSession(parameters, config, CreationTime).EnableProviders();
```
adamsitnik: important: with this change `HeapSession` is always enabled, even if using `EtwProfiler` as standalone. Could you please extend the `EtwProfilerConfig` with a new flag and make this optional and used by default only be the `NativeMemoryDiagnoser` ?

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,257 @@
...
+            yield return new Metric(new NativeMemoryDescriptor(), result.totalMemory);
+            yield return new Metric(new NativeMemoryLeakDescriptor(), result.memoryLeak);
+        } 
+
+        public IEnumerable<ValidationError> Validate(ValidationParameters validationParameters) => etwProfiler.Validate(validationParameters);
+
+        /// <summary>
+        /// Implements a simple one-element cache for find the heap to look in.  
+        /// </summary>
+        private static Dictionary<Address, long> GetHeap(Address heapHandle, Dictionary<Address, Dictionary<Address, long>> heaps, ref Dictionary<Address, long> lastHeapAllocs, ref Address lastHeapHandle)
```
adamsitnik: nit: the name of the method is misleading, when I saw it I thought that you are using ETW to connect to a live process and get some heap information. Perhaps `CreateHeapCache` would be a better name?

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,257 @@
...
+                    if (!start)
+                    {
+                        return;
+                    }
+
+                    var allocs = lastHeapAllocs;
+                    if (data.HeapHandle != lastHeapHandle)
+                        allocs = GetHeap(data.HeapHandle, heaps, ref lastHeapAllocs, ref lastHeapHandle);
+
+                    allocs[data.AllocAddress] = data.AllocSize;
```
adamsitnik: should it be increasing, not overwriting the value?

```suggestion
                    allocs[data.AllocAddress] += data.AllocSize;
```

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,257 @@
...
+                logger.Write(line.Kind, line.Text);
+        }
+
+        public void Handle(HostSignal signal, DiagnoserActionParameters parameters)
+        {
+            etwProfiler.Handle(signal, parameters);
+        }
+
+        //Code is inspired by https://github.com/Microsoft/perfview/blob/master/src/PerfView/PerfViewData.cs#L5719-L5944
+        private (long totalMemory, long memoryLeak) ParseEtlFile(BenchmarkCase parameters)
```
adamsitnik: this method is very big, you could consider moving it to a separate file, similar to `TraceLogParser`

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,257 @@
...
+                };
+
+                var heapParser = new HeapTraceProviderTraceEventParser(eventSource);
+                // We index by heap address and then within the heap we remember the allocation stack
+                var heaps = new Dictionary<Address, Dictionary<Address, long>>();
+                Dictionary<Address, long> lastHeapAllocs = null;
+
+                Address lastHeapHandle = 0;
+
+                long cumMetric = 0;
```
adamsitnik: can we have a better name for it? Maybe `nativeLeakSize`  ?

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,257 @@
...
+                        return;
+                    }
+
+                    var allocs = lastHeapAllocs;
+                    if (data.HeapHandle != lastHeapHandle)
+                        allocs = GetHeap(data.HeapHandle, heaps, ref lastHeapAllocs, ref lastHeapHandle);
+
+                    allocs[data.AllocAddress] = data.AllocSize;
+
+                    cumMetric += data.AllocSize;
```
adamsitnik: I think that this add operations should be inside `checked` block to make sure we don't show invalid data on long overflow. You could also use `ulong` to represent the values to increase the max handled size

wojtpl2:
I will fix it today.

src/BenchmarkDotNet.Diagnostics.Windows/EtwProfiler.cs
```diff
@@ -102,32 +101,44 @@ private void Start(DiagnoserActionParameters parameters)
...
         {
             WaitForDelayedEvents();
 
             try
             {
                 kernelSession.Stop();
+                heapSession.Stop();
                 userSession.Stop();
-
+
```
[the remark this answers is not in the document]
wojtpl2: I've installed a plugin and this situation should not happen again ;)

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,257 @@
...
+
+        //Code is inspired by https://github.com/Microsoft/perfview/blob/master/src/PerfView/PerfViewData.cs#L5719-L5944
+        private (long totalMemory, long memoryLeak) ParseEtlFile(BenchmarkCase parameters)
+        {
+            (long totalMemory, long memoryLeak) result = (0, 0);
+
+            var traceFilePath = etwProfiler.BenchmarkToEtlFile[parameters];
+            var etlxFile = Path.ChangeExtension(traceFilePath, ".etlx");
+            var options = new TraceLogOptions();
+            TraceLog.CreateFromEventTraceLogFile(traceFilePath, etlxFile, options);
```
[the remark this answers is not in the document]
wojtpl2: `TraceLog.CreateFromEventTraceLogFile` returns `etlxFile`
image
But I aggre that this suggestion is good. I will change it.

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
wojtpl2: I think that in this place we should overwrite the value. 

It is a new allocation on data.AllocAddress. You can not have multiple allocations on the same address. In `HeapTraceFree` event is `allocs.Remove(data.FreeAddress);`

There is original code from prefview:
https://github.com/microsoft/perfview/blob/234398a44fbfee5b828272b8201d19d10c7f3578/src/PerfView/PerfViewData.cs#L5760

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
wojtpl2: TODO.

src/BenchmarkDotNet.Diagnostics.Windows/EtwProfiler.cs
wojtpl2: TODO

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
wojtpl2: I added checked. Can we allocate more than `long` size memory?

wojtpl2:
for `netcoreapp2.1`:
image
image

but I don't know yet why it doesn't work for `net461`:
`dotnet run -c Release -f net461 -- --filter *Bitmap`
image

wojtpl2:
If I comment line:
`TraceEventSession.MergeInPlace(FilePath, TextWriter.Null);`
where e.g.
`FilePath = BenchmarkDotNet.Samples.IntroNativeMemory.Bitmap-20190712-115234.etl`
and then I merge collected files using PerfView, everything works ok. It means that `EtwProfiler` collected data from `heapSession` correctly. I do not know what's happening yet.

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
adamsitnik: > Can we allocate more than long size memory?

Not without freeing some part of it, but in total, yes. Example:

```cs
[Benchmark]
public void AllocateMoreThanLong()
{
    for (long i = 0; i < long.MaxValue; i++)
    {
        IntPtr unmanagedHandle = Marshal.AllocHGlobal(2);
        Marshal.FreeHGlobal(unmanagedHandle);
    }
}
```

This is, of course, an edge case, but we should rather throw than print invalid numbers. If we ever get a bug report about `OverflowException` we can switch from long to `BigInteger`

src/BenchmarkDotNet.Diagnostics.Windows/EtwProfiler.cs
```diff
@@ -138,6 +148,6 @@ private void Stop(DiagnoserActionParameters parameters)
         /// ETW sessions receive events with a slight delay.
         /// This increases the likelihood that all relevant events are processed by the collection thread by the time we are done with the benchmark.
         /// </summary>
-        private static void WaitForDelayedEvents() => Thread.Sleep(TimeSpan.FromMilliseconds(500));
+        private static void WaitForDelayedEvents() => Thread.Sleep(TimeSpan.FromMilliseconds(1000));
```
adamsitnik: @wojtpl2 is this change really needed? do we receive more events if we wait for 1s instead of 0.5s?

src/BenchmarkDotNet.Diagnostics.Windows/EtwProfilerConfig.cs
```diff
@@ -10,33 +12,38 @@ namespace BenchmarkDotNet.Diagnostics.Windows
...
 
+        public bool CreateHeapSession { get; }
+
         /// <param name="performExtraBenchmarksRun">if set to true, benchmarks will be executed one more time with the profiler attached. If set to false, there will be no extra run but the results will contain overhead. True by default.</param>
         /// <param name="bufferSizeInMb">ETW session buffer size, in MB. 256 by default</param>
         /// <param name="cpuSampleIntervalInMiliseconds">The rate at which CPU samples are collected. By default this is 1 (once a millisecond per CPU). There is a lower bound on this (typically 0.125 ms)</param>
         /// <param name="intervalSelectors">interval per hardware counter, if not provided then default values will be used.</param>
         /// <param name="kernelKeywords">kernel session keywords, ImageLoad (for native stack frames) and Profile (for CPU Stacks) are the defaults</param>
         /// <param name="providers">providers that should be enabled, if not provided then default values will be used</param>
+        /// <param name="createHeapSession">value indicating whether to create heap session</param>
```
adamsitnik: nit: I would add few more words to make it more clear to the end users that most probably they don't need to set it to true.

```suggestion
        /// <param name="createHeapSession">value indicating whether to create heap session. False by default, used internally by NativeMemoryDiagnoser.</param>
```

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,63 @@
...
+        private readonly EtwProfiler etwProfiler;
+
+        [PublicAPI] // parameterless ctor required by DiagnosersLoader to support creating this profiler via console line args
+        public NativeMemoryDiagnoser() => etwProfiler = new EtwProfiler(CreateDefaultConfig());
+
+        public IEnumerable<string> Ids => new[] { nameof(NativeMemoryDiagnoser) };
+
+        public IEnumerable<IExporter> Exporters => Array.Empty<IExporter>();
+
+        public IEnumerable<IAnalyser> Analysers => Array.Empty<IAnalyser>();
```
adamsitnik: nit: lack of new line 

```suggestion
        public IEnumerable<IAnalyser> Analysers => Array.Empty<IAnalyser>();

```

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,63 @@
...
+using BenchmarkDotNet.Validators;
+using JetBrains.Annotations;
+using Microsoft.Diagnostics.Tracing.Parsers;
+
+namespace BenchmarkDotNet.Diagnostics.Windows
+{
+    public class NativeMemoryDiagnoser : IDiagnoser
+    {
+        private static readonly string LogSeparator = new string('-', 20);
+        internal readonly LogCapture Logger = new LogCapture();
```
adamsitnik: nit: Logger can be private?

```suggestion
        private readonly LogCapture Logger = new LogCapture();
```

src/BenchmarkDotNet.Diagnostics.Windows/NativeMemoryDiagnoser.cs
```diff
@@ -0,0 +1,63 @@
...
+                return Enumerable.Empty<Metric>();
+
+            return new NativeMemoryLogParser(traceFilePath, results.BenchmarkCase, Logger).Parse();
+        }
+
+        public IEnumerable<ValidationError> Validate(ValidationParameters validationParameters) => etwProfiler.Validate(validationParameters);
+
+        private static EtwProfilerConfig CreateDefaultConfig()
+        {
+            var kernelKeywords = KernelTraceEventParser.Keywords.Default | KernelTraceEventParser.Keywords.VirtualAlloc | KernelTraceEventParser.Keywords.VAMap;
```
adamsitnik: nit: it would be nice to have some comment explaining why we need `VirtualAlloc` and `VAMap`

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
```diff
@@ -15,6 +15,21 @@
...
 namespace BenchmarkDotNet.Diagnostics.Windows
 {
+    internal class HeapSession : Session
+    {
+        public HeapSession(DiagnoserActionParameters details, EtwProfilerConfig config, DateTime creationTime)
+            : base(FullNameProvider.GetBenchmarkName(details.BenchmarkCase) + "Heap", details, config, creationTime)
+        {
+        }
+
+        protected override string FileExtension => ".userheap.etl";
```
adamsitnik: nit: lack of new line

```suggestion
        protected override string FileExtension => ".userheap.etl";

```

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
```diff
@@ -43,21 +58,21 @@ public KernelSession(DiagnoserActionParameters details, EtwProfilerConfig config
...
+                | KernelTraceEventParser.Keywords.ImageLoad // handles stack frames from native modules, SUPER IMPORTANT!
                 | KernelTraceEventParser.Keywords.Profile; // CPU stacks
 
             if (Details.Config.GetHardwareCounters().Any())
                 keywords |= KernelTraceEventParser.Keywords.PMCProfile; // Precise Machine Counters
 
             try
             {
-                TraceEventSession.EnableKernelProvider(keywords, KernelTraceEventParser.Keywords.Profile);
+                TraceEventSession.EnableKernelProvider(keywords, keywords); //TODO ? In this way it is created in PerfView
```
adamsitnik: regarding the "TODO" comment: do we need this change? does it make any difference?

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        private readonly ILogger logger;
+
+        public NativeMemoryLogParser(string etlFilePath, BenchmarkCase benchmarkCase, ILogger logger)
+        {
+            this.etlFilePath = etlFilePath;
+            this.benchmarkCase = benchmarkCase;
+            this.logger = logger;
+        }
+
```
adamsitnik: nit: two empty lines

```suggestion
```

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+using Address = System.UInt64;
+
+namespace BenchmarkDotNet.Diagnostics.Windows.Tracing
+{
+    public class NativeMemoryLogParser
+    {
+        private static readonly string LogSeparator = new string('-', 20);
+        private readonly string etlFilePath;
+        private readonly BenchmarkCase benchmarkCase;
+
```
adamsitnik: nit: empty line between private class members declarations

```suggestion
```

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using Microsoft.Diagnostics.Tracing.Etlx;
+using Microsoft.Diagnostics.Tracing.Parsers.Kernel;
+using Address = System.UInt64;
+
+namespace BenchmarkDotNet.Diagnostics.Windows.Tracing
+{
+    public class NativeMemoryLogParser
```
adamsitnik: thanks for moving it to a separate type! IMHO the separation of responsibilities is much better now 👍

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+                };
+
+                eventSource.Process();
+
+                logger.WriteLine();
+                logger.WriteLineHeader(LogSeparator);
+                logger.WriteLineInfo($"{benchmarkCase.DisplayInfo}");
+                logger.WriteLineHeader(LogSeparator);
+
+                var totalMemory = totalAllocation / totalOperation;
```
adamsitnik: nit: a better name would be: memoryAllocatedPerOperation (the value is scaled, it's not total)

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+                var memoryLeak = nativeLeakSize / totalOperation;
+
+                logger.WriteLine($"Total allocated native memory: {totalMemory.ToSizeStr(SizeUnit.B)}");
```
adamsitnik: similar as above: it's not total, it's scaled per operation
```suggestion
                logger.WriteLine($"Native memory allocated per single operation: {totalMemory.ToSizeStr(SizeUnit.B)}");
```

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+                heapParser.HeapTraceAlloc += delegate(HeapAllocTraceData data)
+                {
+                    if (!start)
+                    {
+                        return;
+                    }
+
+                    var allocs = lastHeapAllocs;
+                    if (data.HeapHandle != lastHeapHandle)
+                        allocs = CreateHeapCache(data.HeapHandle, heaps, ref lastHeapAllocs, ref lastHeapHandle);
```
adamsitnik: nit: all other single line conditional blocks of code are wrapped with `{  }`, it would be nice to keep the consistency

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+                logger.WriteLineInfo($"{benchmarkCase.DisplayInfo}");
+                logger.WriteLineHeader(LogSeparator);
+
+                var totalMemory = totalAllocation / totalOperation;
+                var memoryLeak = nativeLeakSize / totalOperation;
+
+                logger.WriteLine($"Total allocated native memory: {totalMemory.ToSizeStr(SizeUnit.B)}");
+                if (nativeLeakSize != 0)
+                {
+                    logger.WriteLine($"Total memory leak: {memoryLeak.ToSizeStr(SizeUnit.B)}");
```
adamsitnik:
```suggestion
                    logger.WriteLine($"Native memory leak per single operation: {memoryLeak.ToSizeStr(SizeUnit.B)}");
```

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -0,0 +1,206 @@
...
+
+                eventSource.Process();
+
+                logger.WriteLine();
+                logger.WriteLineHeader(LogSeparator);
+                logger.WriteLineInfo($"{benchmarkCase.DisplayInfo}");
+                logger.WriteLineHeader(LogSeparator);
+
+                var totalMemory = totalAllocation / totalOperation;
+                var memoryLeak = nativeLeakSize / totalOperation;
```
adamsitnik: nit: memoryLeakPerOperation would be a better name

src/BenchmarkDotNet.Diagnostics.Windows/EtwProfiler.cs
wojtpl2: No, I'm going to revert this change. It pushed it by my mistake.

wojtpl2:
I was thinking that it was a problem with merging. I checked it again and the problem was in kernel session. 

To EnableWindowsHeapProvider function I send name with the path but I should have sent only name with extension.

```diff
internal override Session EnableProviders()
        {
-            var osHeapExe = Path.ChangeExtension(Details.Process.StartInfo.FileName, ".exe");
+            var osHeapExe = Path.GetFileName(Path.ChangeExtension(Details.Process.StartInfo.FileName, ".exe"));
            TraceEventSession.EnableWindowsHeapProvider(osHeapExe);
            return this;
        }
```

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
wojtpl2: Thanks for this comment!

 I checked it and the second parameter means:
`stackCapture = Specifies which events should have their stack traces captured when an event is logged`
After my change, all kernelKeywords from the configuration would have a captured stack. So I reverted this change. 

Thanks to this comment I checked what more I could cut. In the last commit, I sent some optimization.
Currently, for the user session, BDN collects only information about `EngineEventSource`. 
For kernel session I collect  `VirtualAlloc` and `VAMap`:
```diff
-            var kernelKeywords = KernelTraceEventParser.Keywords.Default | KernelTraceEventParser.Keywords.VirtualAlloc | KernelTraceEventParser.Keywords.VAMap;
+            var kernelKeywords = KernelTraceEventParser.Keywords.VirtualAlloc | KernelTraceEventParser.Keywords.VAMap;
```
`KernelSession` adds some extra kernel keywords 
https://github.com/dotnet/BenchmarkDotNet/blob/ec296dc45de798f7407852d5ab7febe2b457eca4/src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs#L51-L53
and in this case, it could be removed. But I think that it could be improved in the next PR becouse it touch `EtwProfiler` and `ConcurrencyVisualizerProfiler`.

wojtpl2:
@adamsitnik It is ready to review. I have fixed all your suggestion and I have sent some optimization (please check https://github.com/dotnet/BenchmarkDotNet/pull/1131#discussion_r303834412) 

I will send the documentation in the next PR if this one is merged.

adamsitnik:
I've just tested it and it works great! :shipit:

wojtpl2:
Thanks! I'm very happy. I've learned a lot!
Now I'm going to update the documentation in the next PR.

adamsitnik:
> Thanks! I'm very happy. I've learned a lot!

Awesome! Once again big thanks.

> Now I'm going to update the documentation in the next PR.

Could you please mention that if benchmark has a native memory leak it should be using `Job.Short`? (I've run into an issue with that when I was testing it)

Could you also consider implementing the `IProfiler` for it and adding the console argument support? Sth like:

```cmd
--profiler nativememory
```

Issue #723, opened 2018-04-18
MemoryDiagnoser should include memory allocated by all Threads that were live during benchmark execution


adamsitnik:
Suggested by @stephentoub

> Another issue I have with the memory columns is the allocation one.  For async methods, it's usually very misleading, since it's only measuring the calling thread.
So I'm manually removing that from the results where it doesn't make sense.
I'd suggest the memory diagnoser by default remove it as well when the benchmark returns Task
But have some way to turn it back on
Or something like that

Current problem: there is no API available for .NET Core that we could use. As of today BDN is using `System.GC.GetAllocatedBytesForCurrentThread()` for .NET Core. There is no way to iterate through all the threads and get this value. Moreover, some threads could have exited in the meantime.

The problem does not occur for .NET  framework, where we use `AppDomain.CurrentDomain.MonitoringTotalAllocatedMemorySize`

haf:
This is an issue for me when trying to benchmark Logary, because all GC happens on Hopac threads. However, these threads are known at runtime during setup: https://github.com/Hopac/Hopac/blob/master/Libs/Hopac.Platform/Init.fs#L13

NickCraver:
I've fired up the API discussion in https://github.com/dotnet/corefx/issues/30644

hpbieker:
It would be nice to get this fixed so that we could get proper memory numbers for async code where some of the callbacks ends up on some other thread.

adamsitnik:
This issue has been fixed in #1155 but it's going to work only with .NET Core 3.0 preview6+.

The .NET Team does not plan to backport the new API to older versions of .NET Core.

gfoidl:
Should https://github.com/dotnet/performance/blob/master/docs/benchmarkdotnet.md (L233)

> - **For .NET Core the Allocated Memory is only for the current thread**, see [#723](https://github.com/dotnet/BenchmarkDotNet/issues/723) for more.

be updated to reflect this?

adamsitnik:
@gfoidl yes, good catch! Could you please send a PR?

gfoidl:
Yes, will send one.

PR #1155, opened 2019-05-16
Use new .NET Core 3.0 API to get the total number of allocated bytes for all threads

### LINKED ISSUE #1153: Use GC.GetTotalAllocatedBytes when available in MemoryDiagnoser
Currently MemoryDiagnoser via GcStats uses GC.GetAllocatedBytesForCurrentThread on .NET Core to get the number of allocated bytes, but that ends up yielding misleading information for async tests or tests that otherwise involve work being performed on multiple threads.  .NET Core 3.0 adds GC.GetTotalAllocatedBytes:
https://github.com/dotnet/corefx/blob/5310b5646aa1c8b18b6546dc5980c30c35d5f4e4/src/System.Runtime/ref/System.Runtime.cs#L1276
It'd be great to use that instead when it's detected to be available.

adamsitnik:
preview6 of .NET Core 3.0 exposes a new method that allows to get the number of allocated bytes for all threads. By using it we can fix #1153 and #723

Sample:

```cs
public class MultiThreadedAllocation
{
    public const int Size = 1_000_000;
    public const int ThreadsCount = 10;

    private Thread[] threads;

    [IterationSetup]
    public void SetupIteration()
    {
        threads = Enumerable.Range(0, ThreadsCount)
            .Select(_ => new Thread(() => GC.KeepAlive(new byte[Size])))
            .ToArray();
    }

    [Benchmark]
    public void Allocate()
    {
        foreach (var thread in threads)
        {
            thread.Start();
            thread.Join();
        }
    }
}
```

```cmd
dotnet run -c Release -f netcoreapp2.1 --filter *MultiThreadedAllocation* --runtimes netcoreapp3.0 netcoreapp2.1 net461 --memory 
```

|   Method | Runtime |     Toolchain |     Gen 0 |     Gen 1 |     Gen 2 |  Allocated |
|--------- |-------- |-------------- |----------:|----------:|----------:|-----------:|
| Allocate |     Clr |        net461 | 2000.0000 | 2000.0000 | 2000.0000 | 10024816 B |
| Allocate |    Core | netcoreapp2.1 | 2000.0000 | 2000.0000 | 2000.0000 |          - |
| Allocate |    Core | netcoreapp3.0 | 2000.0000 | 2000.0000 | 2000.0000 | 10001360 B |

### DIFF
--- a/src/BenchmarkDotNet/Engines/GcStats.cs
+++ b/src/BenchmarkDotNet/Engines/GcStats.cs
@@ -13,7 +13,7 @@ namespace BenchmarkDotNet.Engines
         public static readonly long AllocationQuantum = CalculateAllocationQuantumSize();
 
         private static readonly Func<long> GetAllocatedBytesForCurrentThreadDelegate = CreateGetAllocatedBytesForCurrentThreadDelegate();
-        private static readonly Func<bool, long> GetGetTotalAllocatedBytesDelegate = CreateGetGetTotalAllocatedBytesDelegate();
+        private static readonly Func<bool, long> GetTotalAllocatedBytesDelegate = CreateGetTotalAllocatedBytesDelegate();
 
         public static readonly GcStats Empty = new GcStats(0, 0, 0, 0, 0);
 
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Engines/GcStats.cs
```diff
@@ -12,7 +12,8 @@ public struct GcStats
 
         public static readonly long AllocationQuantum = CalculateAllocationQuantumSize();
 
-        private static readonly Func<long> GetAllocatedBytesForCurrentThreadDelegate = GetAllocatedBytesForCurrentThread();
+        private static readonly Func<long> GetAllocatedBytesForCurrentThreadDelegate = CreateGetAllocatedBytesForCurrentThreadDelegate();
+        private static readonly Func<bool, long> GetGetTotalAllocatedBytesDelegate = CreateGetGetTotalAllocatedBytesDelegate();
```
stephentoub: GetGet => Get
CreateGetGet => CreateGet

AndreyAkinshin:
@adamsitnik 👍

PR #1157, opened 2019-05-17
Proposal: Invocation setup

adamsitnik:
Many users ask me how to set up every benchmark invocation and still have the BDN do estimate the number of invocations per iteration. What we offer with `[IterationSetup]` is not enough, users typically want to benchmark a short operation and if we run the benchmark only once per iteration, they get a lot of warnings about too short iteration time.

My proposal is to add the possibility to set up and cleanup every benchmark invocation.

The idea is simple: do it for both actual workload and overhead estimation. 

I used delegates to make sure the methods don't get inlined.

@AndreyAkinshin what do you think? if you like this idea I am going to implement the support for InProcessToolchain as well. I know that it's not perfect solution, but I would really like to solve this problem.


adamsitnik:
> I have two major concerns

I agree, they are both real. But I am afraid that the best thing we can do about them is to print warnings about the limitations? 

Or use Roslyn to read the IL and throw whenever they have conditions?

AndreyAkinshin:
It seems that most users just ignore warnings. It's better to design API which doesn't allow shooting in the foot. Currently, I don't know use cases which are impossible to benchmark without `InvocationSetup` (if you know such use cases, I'm ready to discuss them). Since we don't have this feature right now, BenchmarkDotNet forces users to design benchmarks another way. And it's a good thing because the final benchmark will be more reliable with a low risk of unnoticed mistakes.

If we decide to add such a feature, we should have pretty good reasons for that.

PR #1163, opened 2019-05-24
Make netcoreapp2.1 default

### LINKED ISSUE #1162: Incorrect value of BenchmarkDotNet.Toolchains.DotNetCli.NetCoreAppSettings.Default
Let's say that we have a `<TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>` project and we want to execute .NET Core benchmark under the `net461` TFM. In this case, `BenchmarkDotNet.Toolchains.DotNetCli.NetCoreAppSettings.Default` which will be used for TFM in the generated benchmark. Currently, it's `NetCoreApp20` which leads to a compilation error in the generated benchmark (because the base project doesn't support `netcoreapp2.0`).

Because of this problem, some of the Windows tests turned red after removing the `netcoreapp2.0` TFM from `BenchmarkDotNet.IntegrationTests` (see https://github.com/dotnet/BenchmarkDotNet/issues/1141#issuecomment-495259857, 6f91ea59474bf6cce2d7ff8de422854cb4a61b06).

Relevant builds:

* https://dev.azure.com/dotnet/BenchmarkDotNet/_build/results?buildId=17597
* https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/builds/24762511

Red tests:

image

@adamsitnik do you have any ideas what we can do with this problem?

adamsitnik:
Helps with #1162


adamsitnik:
@AndreyAkinshin I am going to merge this change to make our CI green but keep the discussion at https://github.com/dotnet/BenchmarkDotNet/issues/1162 open until we find a good solution

PR #1175, opened 2019-06-03
Added information about what a Gen X column means

jigargandhi:
Added information about what a Gen X means. Right now the definition is ambiguous of whether it is the number of objects in that generation or the number of collections in that generation

### DIFF
--- a/docs/articles/configs/diagnosers.md
+++ b/docs/articles/configs/diagnosers.md
@@ -62,7 +62,7 @@ You can also use one of the following attributes (apply it on a class that conta
 [ConcurrencyVisualizerProfiler]
 ```
 
-In BenchmarkDotNet, 1kB = 1024B, 1MB = 1024kB, and so on. The column Gen X means number of collections done in that generation.
+In BenchmarkDotNet, 1kB = 1024B, 1MB = 1024kB, and so on. The column Gen X means number of collections per 1000 operations for that generation
 
 ## Restrictions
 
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
Hey @jigargandhi! Thanks for the PR.
Indeed, it makes sense to add extra information about "Gen X" columns in the documentation. However, the new sentence is not completely correct. We should remove the "to clean up memory" part because a GC collection doesn't always clean up memory. This column shows the number of GC Generation collects *per 1000 operations*. We print this information below the summary table in the legend section: https://github.com/dotnet/BenchmarkDotNet/blob/0a4550f6ab0b821f31f59806b682dace8296ba54/src/BenchmarkDotNet/Diagnosers/MemoryDiagnoser.cs#L65

AndreyAkinshin:
Hey @jigargandhi! Would you like to update your PR?

jigargandhi:
Hey,
yes I will update it in a day or two

AndreyAkinshin:
@jigargandhi the new version is still not correct. This column displays the number of collections **per 1000 operations**.

AndreyAkinshin:
Also, the current sentence doesn't have a mention of GC (the corresponding line in the legend has this information: "GC Generation {generationId} collects per 1000 operations"); a period at the end of the sentence is missing.

AndreyAkinshin:
@jigargandhi thanks!

PR #1180, opened 2019-06-12
Update docstring for IterationSetup/IterationCleanup

billwert:
Try to make it a bit more explicit that the meaning of the attributes changed. It is often tripped on.


AndreyAkinshin:
@billwert thanks for the PR!

Issue #1182, opened 2019-06-24
Benchingmarking .NET 4.8 Causes Errors


RealDotNetDave:
I am trying to run my benchmark test for .NET Core 3 and Clr 4.8 Preview 6 and am experiencing errors.

In the csproj file, I have it set up like this:
    <TargetFrameworks>netcoreapp3.0;net48</TargetFrameworks>

These same benchmarks work fine for .NET Core 3 and Clr 4.7.2.

Standard error: 
   Restore failed in 113.2 ms for C:\src\Books\RockYourCode\CodePerformance\src\PerfBin\netcoreapp3.0\286d48e8-ea59-4a4d-91cf-74303b8dcad7\BenchmarkDotNet.Autogenerated.csproj.
C:\src\Books\RockYourCode\CodePerformance\src\PerfBin\netcoreapp3.0\286d48e8-ea59-4a4d-91cf-74303b8dcad7\BenchmarkDotNet.Autogenerated.csproj : error NU1201:   - netcoreapp3.0 (.NETCoreApp,Version=v3.0)
C:\src\Books\RockYourCode\CodePerformance\src\PerfBin\netcoreapp3.0\286d48e8-ea59-4a4d-91cf-74303b8dcad7\BenchmarkDotNet.Autogenerated.csproj : error NU1201:   - net48 (.NETFramework,Version=v4.8)
C:\src\Books\RockYourCode\CodePerformance\src\PerfBin\netcoreapp3.0\286d48e8-ea59-4a4d-91cf-74303b8dcad7\BenchmarkDotNet.Autogenerated.csproj : error NU1201: Project dotNetTips.CodePerf.Example.App is not compatible with net472 (.NETFramework,Version=v4.7.2) / win7-x64. Project dotNetTips.CodePerf.Example.App supports:
C:\src\Books\RockYourCode\CodePerformance\src\PerfBin\netcoreapp3.0\286d48e8-ea59-4a4d-91cf-74303b8dcad7\BenchmarkDotNet.Autogenerated.csproj : error NU1201:   - netcoreapp3.0 (.NETCoreApp,Version=v3.0)
C:\src\Books\RockYourCode\CodePerformance\src\PerfBin\netcoreapp3.0\286d48e8-ea59-4a4d-91cf-74303b8dcad7\BenchmarkDotNet.Autogenerated.csproj : error NU1201:   - net48 (.NETFramework,Version=v4.8)
[...]

Issue #1115, opened 2019-03-26
Running using dotnet benchmark uses wrong core runtime 


yahorsi:
I have installed dotnet bemchmark tool, made very simple benchmark and trying to run using:
dotnet benchmark DictionaryBenchmark.dll  --filter *

and getting the error:
 Standard error: 
   Restore failed in 225.53 ms for C:\Dev\DictionaryBenchmark\DictionaryBenchmark\bin\Release\netcoreapp2.2\8b16bbdd-3fbd-4847-84a3-ee53a4ec0f06\BenchmarkDotNet.Autogenerated.csproj.
C:\Dev\DictionaryBenchmark\DictionaryBenchmark\bin\Release\netcoreapp2.2\8b16bbdd-3fbd-4847-84a3-ee53a4ec0f06\BenchmarkDotNet.Autogenerated.csproj : error NU1201: Project DictionaryBenchmark is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1). Project DictionaryBenchmark supports: netcoreapp2.2 (.NETCoreApp,Version=v2.2)
  Restore completed in 25.88 ms for C:\Dev\DictionaryBenchmark\DictionaryBenchmark\DictionaryBenchmark.csproj.

The app fo sure uses core 2.2 in the proj file.
I'm not sure what I'm doing wrong here and did;t found in docs an easy way to specify the runtime.

yahorsi:
Well I figured -runtimes switch but it would be great to mention in in the docs on how you run the benchmarks :)

Issue #1056, opened 2019-02-07
Fails to build when targeting .NET Core 3.0 and .NET Framework


Symbai:
Created a new .NET Core 3.0 console application in VS 2019 and added BenchmarkDotNet via nuget. Changed line in proj.csproj:
```
<TargetFramework>netcoreapp3.0</TargetFramework>
to
<TargetFrameworks>net472;netcoreapp3.0</TargetFrameworks>
```

I receive the following in console output (after CLR job has been successful completed):
```
// Build Exception: Standard output:
completed "proj.csproj".
proj \ bin \ Release \ net472 \ 21078fb3-513f-4d7f-807a-57e8bfabfb62 \ BenchmarkDotNet.Autogenerated.csproj: error NU1201: The proj project is not compatible with **netcoreapp2.0** (.NETCoreApp, version = v2.0). The proj project supports the following:
proj \ bin \ Release \ net472 \ 21078fb3-513f-4d7f-807a-57e8bfabfb62 \ BenchmarkDotNet.Autogenerated.csproj: error NU1201: - net472 (.NET Framework, Version = v4.7.2)
proj \ bin \ Release \ net472 \ 21078fb3-513f-4d7f-807a-57e8bfabfb62 \ BenchmarkDotNet.Autogenerated.csproj: error NU1201: - netcoreapp3.0 (.NETCoreApp, Version = v3.0)
   Error recovering to "106.14 ms" for "proj \ bin \ Release \ net472 \ 21078fb3-513f-4d7f-807a-57e8bfabfb62 \ BenchmarkDotNet.Autogenerated.csproj".
```

For some reason it tries to compile it on .Net Core 2.0 and obviously as the project doesn't support it (target framework is set to .NET Core 3.0), it fails. However when I change the project to .NET Core 2.0 it (not much surprising) "magically" works and complete the core benchmark. Like this:

`<TargetFrameworks>net472;netcoreapp2.0</TargetFrameworks>`

Now I have another project using BenchmarkDotNet which targets netcoreapp3.0 only and I can run benchmarks on it so I have no idea what's different or the problem.

//edit, fyi:
**Works:**
`<TargetFrameworks>net472;netcoreapp2.0</TargetFrameworks>`
`<TargetFrameworks>netcoreapp3.0</TargetFrameworks>`
**Fails:**
`<TargetFrameworks>net472;netcoreapp3.0</TargetFrameworks>`

adamsitnik:
[...]

PR #1188, opened 2019-07-02
Enforce the users to provide full target framework moniker when using attributes to define multiple runtime jobs

adamsitnik:
@AndreyAkinshin this is my proposal for fixing the problem discussed in #1162

And fix/prevent issues similar to:

- https://github.com/dotnet/BenchmarkDotNet/issues/1056
- https://github.com/dotnet/BenchmarkDotNet/issues/993
- https://github.com/dotnet/BenchmarkDotNet/issues/1182
- https://github.com/dotnet/BenchmarkDotNet/issues/1115

When users write following code to define two jobs:

```cs
[ClrJob, CoreJob]
```

Run it as full .NET Framework and BDN fails to determine the correct TFM for .NET Core.

IMHO we should transfer this responsibility to the users. This is going to save us from errors and make it more obvious for the end user what they are testing.

This change affects only the users who want to run benchmarks against multiple runtimes using attributes `[ClrJob, CoreJob, CoreRtJob]`. The default case, when user just wants to test single, current Runtime is not affected.

Sample:

```cs
[TargetFrameworkJob(TargetFrameworkMoniker.Net461, baseline: true)]
[TargetFrameworkJob(TargetFrameworkMoniker.Mono)]
[TargetFrameworkJob(TargetFrameworkMoniker.Netcoreapp21)]
```


adamsitnik:
> Currently, TargetFrameworkMoniker is not actually a target framework moniker because we have Mono on this list. It can be confusing. What do you think about the RuntimeMoniker name? I

I would prefer to keep the `TargetFrameworkMoniker` because this is what we use behind the scenes in auto-generated .csproj (except of Mono). I hope that using this name is going to make it obvious for the end-users about what they get when they use this feature (a project that sets `<TargetFrameworkMoniker>` to provided value)

> Next, I don't really like the TargetFrameworkJobAttribute name

I agree that the name is long, but I want to set the right expectations for this feature.

>  I suggest capitalizing the first letter of each "subword" and write NetCoreApp.

OK 👍 

> If we rename TargetFrameworkMoniker to RuntimeMoniker, we can also include InProcess in this list

But this is not the goal of this feature and PR. If user wants to test different Runtimes using attributes, then this is the attribute to use. If user wants to use `InProcessToolchain`, then (s)he should use the `[InProcess]` attribue

> Also, it will be great to introduce monikers without a specific version.

This is exactly what I want to avoid. There were too many bug reports about existing feature (using `[ClrJob, CoreJob]`). If user wants to compare different runtimes, then the TFMs must be specified in explicit way. No magic, no guessing. What I can implement is better error handling and better explanation message for the end users when they provide invalid values.

>  In most cases, when I want to run a benchmark on the full .NET Framework, I'm expecting the latest available .NET Framework on my machine. Also, if I specify a specific version of the TFM (e.g., Net48), I reduce the portability of this benchmark (it will not work correctly on a machine without .NET Framework 4.8 installed). For .NET Core, it makes sense to introduce two monikers for the latest stable and preview versions (e.g., NetCoreAppLatestStable and NetCoreAppLatestPreview).

> The final thing: I want the Current value that reuses the TFM of the host process. E.g., if I have net48;netcoreapp2.2 application, I want to run benchmarks on .NET Framework 4.8 when I execute the host process with net48 and on .NET Core 2.2 when I execute the host process with netcoreapp2.2

In such case you should not be using any additional config, just run the benchmark and BDN will run it for the same Runtime as the host process. This is the behaviour BDN always had and this PR is not changing it.

> Currently, when some of the requested monikers are unsupported on the current machine, BenchmarkDotNet doesn't handle it nicely. E.g., when we are trying to perform several jobs including CoreRT, BenchmarkDotNet just throws an unhandled exception and doesn't print the summary table with proper explanation. I think it will be better just to skip unsupported monikers, execute the rest of the benchmarks, and print a nice error message under the summary table. In this case, we can safely use the CoreRT job in our examples and don't be afraid that somebody will not be able to execute such a benchmark.

It's rather impossible to detect that somehting is not supported without running the benchmark first. I can definitely implement a better error handling and provide better error messages for the end user.

> It's a pretty huge change in our API, so I suggest to bump the minor version of the library and include it in BenchmarkDotNet v0.12.0. Also, it makes sense to upgrade our obsolete attributes. With the current state of the PR, they just don't affect the config. 

> We should try to provide a behavior that is backward compatible with previous versions.

I agree that this is a breaking change, but it affects only the users who use BDN to compare the performance of different .NET Runtimes. I don't have a strong opinion on the version number, both 0.11.6 and 0.12.0 are fine to me.

> It will be an unpleasant surprise for users who just upgrade the BenchmarkDotNet without reading release notes: it may take a lot of time to figure out why these attributes don't work anymore.

This is why the Obsolete error message provides the name of the attribute that users should be using instead:

"Plase use TargetFrameworkAttribute instead."

I expect that everyone who gets the error message is going to replace the `[ClrJob]`/`[CoreJob]` with suggested `[TargetFramework]` and figure out the only mandatory argument using IDE.

adamsitnik:
Ok, I've added better error handling.

Sample:

```cmd
dotnet run -c Release -f net461 --runtimes netcoreapp2.0 --filter *
```

```log
// Build Error: The project which defines benchmarks targets 'net461', you can not benchmark 'netcoreapp2.0'.
To be able to benchmark 'netcoreapp2.0' you need to use <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks> in your project file ('C:\Projects\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj').
```

adamsitnik:
I am going to add some good docs tomorrow.

adamsitnik:
Ok, I've added the docs, more unit tests and .NET Core 3.1 and 5.0 support.

Sample error for missing SDK:

```log
// Build Error: The current .NET SDK does not support targeting .NET Core 5.0. You need to install it or pass the path to dotnet cli via the `--cli` console line argument.
```

Sample error for project not supporting given TFM:

```log
// Build Error: The project which defines benchmarks does not target 'netcoreapp2.0'.
You need to add 'netcoreapp2.0' to <TargetFrameworks> in your project file ('C:\Projects\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj').
Example: <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
```

docs/articles/configs/toolchains.md
```diff
@@ -5,65 +5,103 @@ name: Toolchains
...
+```
+
+What is going to happen if you provide multiple Full .NET Framework monikers? Let's say:
+
+```cmd
+dotnet run -c Release -f net461 net472 net48
 ```
 
-BenchmarkDotNet allows you to take full advantage of that. With single config, we can execute the benchmarks for all the frameworks that you have listed in your csproj file.
+Full .NET Framework always runs every .NET executable using the latest .NET Framework available on a given machine. If you try to run the benchmarks for a few .NET TFMs, they are all going to be executed using the latest .NET Framework from your machine. The only difference is that they are all going to have different features enabled depending on target version they were compiled for. You can read more about this [here](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/version-compatibility) and [here](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/application-compatibility). This is **Windows Operating System behavior which can not be controlled by BenchmarkDotNet or any other tool**.
```
billwert: This is not technically anything to do with Windows - it has to do with how we choose to ship and package .NET Framework. (Indeed, we had a TFM for .NET Framework 2.0 or 1.1 it would load that runtime if it were installed.)

PR #1206, opened 2019-07-21
add *.fsproj to the list of supported project types

JaggerJo:


docs/articles/overview.md
```diff
@@ -9,7 +9,7 @@ name: Overview
 
 Create new console application and install the [BenchmarkDotNet](https://www.nuget.org/packages/BenchmarkDotNet/) NuGet package. We support:
 
-* *Projects:* `*.csproj` (classic and modern with PackageReferences), `*.xproj`/`project.json`
+* *Projects:* `*.csproj`/`*.fsproj` (classic and modern with PackageReferences), `*.xproj`/`project.json`
```
adamsitnik: If we add `.fsproj` then `.vbproj` is still going to be missing. Perhaps we should simplify it to the following sentence:

```suggestion
* *Projects:* classic and modern with PackageReferences
```

Below we have "Languages: C#, F#, VB"

adamsitnik:
@JaggerJo thanks!

PR #1208, opened 2019-07-23
Support of --profiler NativeMemory command line argument

wojtpl2:
In this PR I support command line argument:
```
--profiler NativeMemory
```

@adamsitnik I wonder if `NativeMemoryDiagnoser` shouldn't be named `NativeMemoryProfiler`?
We have `MemoryDiagnoser` but there is also `EtwProfiler` and `ConcurrencyVisualizerProfiler`.

### DIFF
--- a/docs/articles/configs/diagnosers.md
+++ b/docs/articles/configs/diagnosers.md
@@ -28,7 +28,7 @@ The current Diagnosers are:
   It allows you to not only benchmark, but also profile the code. It's using TraceEvent, which internally uses ETW and exports all the information to a trace file. The trace file contains all of the stack traces captured by the profiler, PDBs to resolve symbols for both native and managed code and captured GC, JIT and CLR events. Please use one of the free tools: PerfView or Windows Performance Analyzer to analyze and visualize the data from trace file. You can find this diagnoser in a separate package with diagnosers for Windows (`BenchmarkDotNet.Diagnostics.Windows`): [![NuGet](https://img.shields.io/nuget/v/BenchmarkDotNet.svg)](https://www.nuget.org/packages/BenchmarkDotNet.Diagnostics.Windows/)
 - Concurrency Visualizer Profiler (`ConcurrencyVisualizerProfiler`)
   It uses `EtwProfiler` to profile the code using ETW and create not only `.etl` file but also a CVTrace file which can be opened by Concurrency Visualizer plugin from Visual Studio.
-- Native Memory Diagnoser (`NativeMemoryDiagnoser`)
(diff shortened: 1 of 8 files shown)


adamsitnik:
> I wonder if NativeMemoryDiagnoser shouldn't be named NativeMemoryProfiler?

I am not sure. On the one hand, it's using EtwProfiler API to get the data. On the other, it reports the metrics and doesn't print a file path to the trace file so it's transparent to the end-user.

Does the trace file contain enough information to track the type names? Or something very useful in general? If so, you can rename it to Profiler but also print the paths to the trace files.

wojtpl2:
I'm going to check it. If it is possible to trace the type names, I will change the name to Profiler.

adamsitnik:
> I'm going to check it. If it is possible to trace the type names, I will change the name to Profiler.

Great, please let me know once you got it so we can merge this PR and release 0.11.6

wojtpl2:
@adamsitnik I know what I have to do to track the type names, using PerfView. It requires small changes in `NativeMemoryDiagnoser`, but after this changes .etl file will be a little bigger because we need to save information about stack traces. Do you think it will be a useful function?

adamsitnik:
> Do you think it will be a useful function?

Yes! I think that it would be nice if users could find out not only the native allocation size but also what was allocated if they open the trace file in a profiler.

adamsitnik:
@wojtpl2 please let me know when this is ready to test ;)

wojtpl2:
Not yet but today morning it will be. I just need test it and describe how it works.

wojtpl2:
It is ready to test.
In this test, I allocate an array of `int` twice and an array of `Point` once. 
image
- An array of `int` with 200 length -> 200 * 4B = 800B
- An array of `int` with 150 length -> 150 * 4B = 600B
- An array of `Point` (value `int` X and `int` Y) with 200 length -> 200 * 2 * 4B = 1600B

SUM 3000B
There are output from BDN:
image
image
This is confirmation of these values:
image
There are print screans from `PerfView` with information about types. 
image
image

I didn't have to make any changes in `MemoryNativeProfiler` because heap session logs stack traces. Then `PerfView` can read type from stack trace:
image

wojtpl2:
@adamsitnik I have a question when you want to release a new version of BDN? I would like to know how much time I have to write my blog about this topic.

adamsitnik:
>  I have a question when you want to release a new version of BDN? I would like to know how much time I have to write my blog about this topic.

My plan is to finish https://github.com/dotnet/BenchmarkDotNet/pull/1188 on Monday and then we should be able to release a new version by the end of next week

Issue #1183, opened 2019-06-24
Plots of benchmarks without params have a double dash (--) in the name


simonech:
For each Job, it adds plots about density, cumulative mean, and so on. But two files are named `BenchmarkDotNet.Artifacts/results/MyBench.Sleeps-Time50--density.png` and `BenchmarkDotNet.Artifacts/results/MyBench.Sleeps-Time50--facetDensity.png`, with the `--` instead of single. Like some iteration variable is empty (since later there are names with `-Default-`
![image](https://user-images.githubusercontent.com/61557/59975824-efbc5c00-95bc-11e9-8576-a09b832c2c64.png)

I've noticed that this last error only happens without parameters. When there are parameters involved, there are no `--`  files.

marcnet80:
Hi, @AndreyAkinshin. Is this issue still actual? I can't reproduce it on Windows 10, R-3.6.1.
`   internal string GetArtifactFullName(Summary summary)
        {
            string fileName = GetFileName(summary);
            return $"{Path.Combine(summary.ResultsDirectoryPath, fileName)}-{FileCaption}{FileNameSuffix}.{FileExtension}";
        }`

AndreyAkinshin:
@marcnet80 yes, the issue is still actual. I guess, the problem is in this line: https://github.com/dotnet/BenchmarkDotNet/blob/v0.11.5/src/BenchmarkDotNet/Templates/BuildPlots.R#L112
When `params` is empty, we get double dash in the file name.

marcnet80:
@AndreyAkinshin, thank you for pointing, values for params are coming from generated  ..-measurements.csv in this line of code: 
https://github.com/dotnet/BenchmarkDotNet/blob/v0.11.5/src/BenchmarkDotNet/Templates/BuildPlots.R#L104
I open ..-measurements.csv and see that values for **Params** columns is empty, that's why @simonech  has `--` 
![image](https://user-images.githubusercontent.com/46241368/61690884-00cec900-ad33-11e9-873d-973276d4fc73.png)
I can fix it in R-script, by checking params and adding additional variable or maybe better add default value for Params column in C#: 
https://github.com/dotnet/BenchmarkDotNet/blob/v0.11.5/src/BenchmarkDotNet/Exporters/Csv/CsvMeasurementsExporter.cs#L98
Something like this (of course wrapp it in Presenter):
[...]

PR #1212, opened 2019-07-24
Plots with a double dash (--) in the names (#1183)

marcnet80:
https://github.com/dotnet/BenchmarkDotNet/issues/1183

### DIFF
--- a/src/BenchmarkDotNet/Templates/BuildPlots.R
+++ b/src/BenchmarkDotNet/Templates/BuildPlots.R
@@ -10,13 +10,13 @@ library(tidyr)
 library(grid)
 library(gridExtra)
 
-isEmpty = function(val){
+isEmpty <- function(val){
    is.null(val) | val == ""
 }
 
-getParamsValues = function(params){ 
-   separator = "-"
-   values = params[!isEmpty(params)]
+getParamsValues <- function(params){ 
+   separator <- "-"
+   values <- params[!isEmpty(params)]
    paste(replace(values, TRUE, paste0(separator, values)), collapse = "")
 }
 
@@ -118,7 +118,7 @@ for (file in files) {
         xlab(paste("Time,", timeUnit)) +
         geom_density(alpha=.5, bw="SJ")
       printNice(paramsDensityPlot)
-      paramsValues = getParamsValues(c(target,params))
+      paramsValues <- getParamsValues(c(target,params))
       ggsaveNice(gsub("-measurements.csv", paste0(paramsValues, "-density.png"), file), paramsDensityPlot)
 
       paramsFacetDensityPlot <- paramsDensityPlot + facet_wrap(~Job_Id)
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
@marcnet80 I think, it will be better to use `-` in this case instead of `-Default-`. When we don't have any parameters, `Default` doesn't make any sense here, but it increases the length of the file name.

src/BenchmarkDotNet/Templates/BuildPlots.R
```diff
@@ -10,6 +10,17 @@ library(tidyr)
 library(grid)
 library(gridExtra)
 
+isEmpty <- function(val){
+   is.null(val) || val == ""
+}
+
+verifyParams <- function(params){ 
+  value = "Default"
```
AndreyAkinshin: Please, use `<-` instead of `=` for assignment operations.

marcnet80: @AndreyAkinshin, please look my last changes. Thank you for reviewing.

AndreyAkinshin: You still use `=` as the assignment operator. Please, use `<-` instead.

marcnet80: it's weird, I replaced assignments operator just for my code: https://github.com/dotnet/BenchmarkDotNet/pull/1212/commits/82c8abb60acc4b711f114e6ee995da28ba41f20f. Should I replaced all others `<-` in R.script?

src/BenchmarkDotNet/Templates/BuildPlots.R
```diff
@@ -103,17 +113,17 @@ for (file in files) {
...
     for (params in unique(df$Params)) {
       paramsDf <- df %>% filter(Params == params)
-
       paramsDensityPlot <- ggplot(paramsDf, aes(x=Measurement_Value, fill=Job_Id)) +
         ggtitle(paste(title, "/", target, "/", params)) +
         xlab(paste("Time,", timeUnit)) +
         geom_density(alpha=.5, bw="SJ")
       printNice(paramsDensityPlot)
-      ggsaveNice(gsub("-measurements.csv", paste0("-", target, "-", params, "-density.png"), file), paramsDensityPlot)
+      paramsValues <- getParamsValues(c(target,params))
```
AndreyAkinshin: My last concern: I don't like `paramsValues` and `getParamsValues` names. They are misleading because here we have a combination of the target and the params. It will be better to rename it.

marcnet80: Hi, @AndreyAkinshin. I thought about naming, **target** and **params**, both are parts of final picture name, maybe `prefix` and `createPrefix`/`combineParts` will be better? Any suggestions?

AndreyAkinshin: `prefix` looks good to me.

marcnet80: Ok, replaced. Thank you.

Issue #1220, opened 2019-08-10
[Docs] RScript / R_HOME setup


tebeco:
Hello,

As part of an experimentation we tried to see how data were exported using `RExporter`.
Was hit by 2 little things :
* the link to `R` installer is a bit small since it'as applied on "one letter only" (barelly visible on High Res).
* Took me time to understand what was intended in `R_HOME` (it was already updated here #339)

Finally managed to work but i had to open the source code to understand that it would concat `/bin` to this value ^^.

#### Side note
Even with `RScript.exe` in the `PATH`, it seems that if `R_HOME` exist and points to a wrong path (i initially tried `C:\Program Files\R\R-3.6.1\bin`) it won't try to check in the `PATH`.
If a run takes few hours, the error appears latter :

* Should the Rscript detection fallback to the `PATH` is `R_HOME` seems wrong ?
* Could the `RExporter` be detected before the `Benchmark` start
  * Should it output an error / warning
  * Should it prompt the User to continue or not ?
  * What to do on the CI ? (would waste ~xxx Min on the CI to fail at the end)

AndreyAkinshin:
> Should the Rscript detection fallback to the PATH is R_HOME seems wrong ?

I'm OK with such a fallback.

> Could the RExporter be detected before the Benchmark start

Yes, it's a good idea. We have validators for such tasks (check out SetupCleanupValidator as an example). PR with such checks is welcome!

adamsitnik:
Fixed by #1222 a while ago, closing

PR #1221, opened 2019-08-10
Update RExporter documentation (link visibility / R_HOME value)

tebeco:
As i tried, and managed to use `RExporter`, i had 2 minor issues listed here : #1220 

Here is a proposal of doc changes that attempt to help others developers trying ot use `RExporter`.
I'm not sure about the `R_HOME` part to be honest.

#### Side note
Even with `RScript.exe` in the `PATH`, it seems that if `R_HOME` exist and points to a wrong path (i initially tried `C:\Program Files\R\R-3.6.1\bin`) it won't try to check in the `PATH`.
If a run takes few hours, the error appears latter :

* Should the Rscript detection fallback to the `PATH` is `R_HOME` seems wrong ?
* Could the `RExporter` be detected before the `Benchmark` start
  * Should it output an error / warning
  * Should it prompt the User to continue or not ?
  * What to do on the CI ? (would waste ~xxx Min on the CI to fail at the end)

I could attempt to add few of these additional changes to this PR

### DIFF
--- a/docs/articles/configs/exporters.md
+++ b/docs/articles/configs/exporters.md
@@ -20,7 +20,7 @@ By default, files with results will be located in
 
 ## Plots
 
-You can install [R](https://www.r-project.org/) to automatically get nice plots of your benchmark results.
+[You can install R](https://www.r-project.org/) to automatically get nice plots of your benchmark results.
 First, make sure `Rscript.exe` or `Rscript` is in your path,
   or define an R_HOME environment variable pointing to the R installation directory (containing the `bin` directory).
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
I close this PR because the corresponding commits were merged into master in https://github.com/dotnet/BenchmarkDotNet/pull/1222

> Could the RExporter be detected before the Benchmark start 

Yes, it's a good idea. We have validators for such tasks (check out `SetupCleanupValidator` as an example). PR with such checks is welcome!

tebeco:
@AndreyAkinshin yep the code was duplicated on purpose as the other PR could have been rejected but was a bit more agressive than this one

Issue #1154, opened 2019-05-16
Add a ConcurrencyDiagnoser?


stephentoub:
.NET Core 3.0 adds several members that would be useful to be exposed for use with certain kinds of performance tests, e.g. Monitor.LockContentionCount and ThreadPool.CompletedWorkItemCount.  Would it make sense to add a ConcurrencyDiagnoser that exposed (end - start) reads of these?

AndreyAkinshin:
@stephentoub it's a pretty interesting idea, I like it!

PR #1227, opened 2019-08-19
Threading Diagnoser

adamsitnik:
This fixes https://github.com/dotnet/BenchmarkDotNet/issues/1154

The `ThreadingDiagnoser` uses new APIs [exposed](https://github.com/dotnet/corefx/issues/35500) in .NET Core 3.0 to report:

* Completed Work Items: The number of work items that have been processed in ThreadPool (per single operation)
* Lock Contentions: The number of times there **was contention** upon trying to take a Monitor's lock (per single operation)

Sample:

```cs
[ThreadingDiagnoser] // ENABLE the diagnoser
public class IntroThreadingDiagnoser
{
    [Benchmark]
    public void CompleteOneWorkItem()
    {
        ManualResetEvent done = new ManualResetEvent(initialState: false);

        ThreadPool.QueueUserWorkItem(m => (m as ManualResetEvent).Set(), done);

        done.WaitOne();
    }

   [Benchmark]
    public void DoNothing() { }
}
```

|              Method |        Median | Completed Work Items | Lock Contentions |
|-------------------- |--------------:|---------------------:|-----------------:|
| CompleteOneWorkItem | 8,111.6074 ns |               1.0000 |                - |
|           DoNothing |     0.0000 ns |               0.0000 |                - |

I've implemented integration tests to make sure it works as expected for .NET Core 3.0, CoreRT and the `InProcessEmitToolchain`.

/cc @stephentoub

### DIFF
--- a/samples/BenchmarkDotNet.Samples/IntroThreadingDiagnoser.cs
+++ b/samples/BenchmarkDotNet.Samples/IntroThreadingDiagnoser.cs
@@ -1,0 +1,19 @@
+﻿using BenchmarkDotNet.Attributes;
+using System.Threading;
+
+namespace BenchmarkDotNet.Samples
+{
+    [ThreadingDiagnoser] // ENABLE the diagnoser
+    public class IntroThreadingDiagnoser
+    {
+        [Benchmark]
+        public void CompleteOneWorkItem()
+        {
+            ManualResetEvent done = new ManualResetEvent(initialState: false);
+
+            ThreadPool.QueueUserWorkItem(m => (m as ManualResetEvent).Set(), done);
+
+            done.WaitOne();
+        }
+    }
+}
--- a/src/BenchmarkDotNet/Configs/ImmutableConfig.cs
+++ b/src/BenchmarkDotNet/Configs/ImmutableConfig.cs
@@ -93,6 +93,9 @@ namespace BenchmarkDotNet.Configs
         public IDiagnoser GetCompositeDiagnoser() => new CompositeDiagnoser(diagnosers);
 
         public bool HasMemoryDiagnoser() => diagnosers.Contains(MemoryDiagnoser.Default);
+        public bool HasThreadingDiagnoser() => diagnosers.Contains(ThreadingDiagnoser.Default);
+
+        public bool HasExtraStatsDiagnoser() => HasMemoryDiagnoser() || HasThreadingDiagnoser();
 
         public IDiagnoser GetCompositeDiagnoser(BenchmarkCase benchmarkCase, RunMode runMode)
         {
--- a/src/BenchmarkDotNet/Configs/ImmutableConfig.cs
+++ b/src/BenchmarkDotNet/Configs/ImmutableConfig.cs
@@ -93,6 +93,7 @@ namespace BenchmarkDotNet.Configs
         public IDiagnoser GetCompositeDiagnoser() => new CompositeDiagnoser(diagnosers);
 
         public bool HasMemoryDiagnoser() => diagnosers.Contains(MemoryDiagnoser.Default);
+
         public bool HasThreadingDiagnoser() => diagnosers.Contains(ThreadingDiagnoser.Default);
 
         public bool HasExtraStatsDiagnoser() => HasMemoryDiagnoser() || HasThreadingDiagnoser();
--- a/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
@@ -29,6 +29,9 @@ namespace BenchmarkDotNet.ConsoleArguments
         [Option('m', "memory", Required = false, Default = false, HelpText = "Prints memory statistics")]
         public bool UseMemoryDiagnoser { get; set; }
 
+        [Option('t', "threading", Required = false, Default = false, HelpText = "Prints threding statistics")]
+        public bool UseThreadingDiagnoser { get; set; }
+
         [Option('d', "disasm", Required = false, Default = false, HelpText = "Gets disassembly of benchmarked code")]
         public bool UseDisassemblyDiagnoser { get; set; }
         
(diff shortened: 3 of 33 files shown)


src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
```diff
@@ -29,6 +29,9 @@ public class CommandLineOptions
         [Option('m', "memory", Required = false, Default = false, HelpText = "Prints memory statistics")]
         public bool UseMemoryDiagnoser { get; set; }
 
+        [Option('t', "threading", Required = false, Default = false, HelpText = "Prints threding statistics")]
```
wojtpl2: Typo thre**a**ding

src/BenchmarkDotNet/Configs/ImmutableConfig.cs
```diff
@@ -93,6 +93,9 @@ public sealed class ImmutableConfig : IConfig
         public IDiagnoser GetCompositeDiagnoser() => new CompositeDiagnoser(diagnosers);
 
         public bool HasMemoryDiagnoser() => diagnosers.Contains(MemoryDiagnoser.Default);
+        public bool HasThreadingDiagnoser() => diagnosers.Contains(ThreadingDiagnoser.Default);
```
wojtpl2: Missing newline, before this line.

src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            yield return new Metric(CompletedWorkItemCountMetricDescriptor.Instance, results.ThreadingStats.CompletedWorkItemCount / (double)results.ThreadingStats.TotalOperations);
+            yield return new Metric(LockContentionCountMetricDescriptor.Instance, results.ThreadingStats.LockContentionCount / (double)results.ThreadingStats.TotalOperations);
+        }
+
+        public IEnumerable<ValidationError> Validate(ValidationParameters validationParameters)
+        {
+            if (!RuntimeInformation.IsNetCore || NetCoreAppSettings.Current.Value.IsOlderThan(TargetFrameworkMoniker.NetCoreApp30))
+            {
+                yield return new ValidationError(false, "ThreadingDiagnoser supports only .NET Core 3.0+");
```
wojtpl2: In this place, you could use `"{nameof(ThreadingDiagnoser)} supports only .NET Core 3.0+"`. But it is only my suggestion, not a mistake.

src/BenchmarkDotNet/Engines/ThreadingStats.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System;
+using System.Threading;
+
+namespace BenchmarkDotNet.Engines
+{
+    public struct ThreadingStats
+    {
+        internal const string ResultsLinePrefix = "Threading: ";
+
+        // BDN targets .NET Standard 2.0, these properties are not part of it
```
wojtpl2: `..these properties are part of .Net Core 3.0+`. I think it will be more understandable for future readers.

adamsitnik:
@wojtpl2 thanks! 

> Tomorrow I'm going to run and test it.

please let me know if it works as expected and then I am going to merge the PR

samples/BenchmarkDotNet.Samples/IntroThreadingDiagnoser.cs
```diff
@@ -0,0 +1,19 @@
+﻿using BenchmarkDotNet.Attributes;
+using System.Threading;
+
+namespace BenchmarkDotNet.Samples
+{
+    [ThreadingDiagnoser] // ENABLE the diagnoser
+    public class IntroThreadingDiagnoser
```
wojtpl2: Currently, this sample test is useless because `TargetFrameworks` doesn't contain `netcoreapp3.0` in `BenchmarkDotNet.Samples.csproj`. If someone would like to run this test, they have to add missing `TargetFrameworks`.

tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
```diff
@@ -0,0 +1,156 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+    public class ThreadingDiagnoserTests
```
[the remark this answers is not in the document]
adamsitnik: I am aware of that, I was running the test locally using .NET Core 3.0.

NET Core 3.0 is still in preview, I think that we should wait with the update until it's released (33 days from now).

src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public void DisplayResults(ILogger logger) { }
+
+        public RunMode GetRunMode(BenchmarkCase benchmarkCase) => RunMode.NoOverhead;
+
+        public void Handle(HostSignal signal, DiagnoserActionParameters parameters) { }
+
+        public IEnumerable<Metric> ProcessResults(DiagnoserResults results)
+        {
+            yield return new Metric(CompletedWorkItemCountMetricDescriptor.Instance, results.ThreadingStats.CompletedWorkItemCount / (double)results.ThreadingStats.TotalOperations);
```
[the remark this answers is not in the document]
adamsitnik: > I think you should add in this place:

the problem is that this check is done in the parent process, not the benchmark. If user run something like:

```cmd
dotnet run -c Release -f netcoreapp2.1 --runtimes netcoreapp3.0 --threading
```

the host process is going to be .NET Core 2.1 while the benchmark process 3.0

this is why I made it only a warning (not an error). I'll try to implement a better error message

samples/BenchmarkDotNet.Samples/IntroThreadingDiagnoser.cs
adamsitnik: again I was aware of that, I should have explain that in the PRs description.

tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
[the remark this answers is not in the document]
wojtpl2: Ok, you have right.

adamsitnik:
I am going to merge it now and send a separate PR with the proper version detection and a bit of refactor ;p

PR #1230, opened 2019-08-22
further runtime detection improvements

adamsitnik:
In #1188 I've changed the multiple runtimes testing config and from then we expect the users to specify full runtime information when testing multiple runtimes. Namely target framework monikers.

The introduced in #1188 `TargetFrameworkMoniker` is an enum, so it's always limited to the list of Runtimes that we know.  In this PR I've extended:
* `TargetFrameworkMoniker` enum with `NotRecognized` value
* the existing `Runtime` class with this enum and ms build moniker name. 

I did that to make the runtime version detection more robust: to versions of .NET Core we don't know yet. 

I've also made the toolchain selection smarter for cases when user defines NuGet packages or custom build configuration name.


adamsitnik:
I've just finished testing, everything works as expected. The PR fixes #995

src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        internal static CoreRuntime GetCurrentVersion()
+        {
+            if (!RuntimeInformation.IsNetCore)
+            {
+                throw new NotSupportedException("It's impossible to reliably detect the version of .NET Core if the process is not a .NET Core process!");
+            }
+
+            string netCoreAppVersion = RuntimeInformation.GetNetCoreVersion();
+            if (!string.IsNullOrEmpty(netCoreAppVersion) && Version.TryParse(netCoreAppVersion, out _))
```
filipnavara: This doesn't work for current prerelease versions (eg. `5.0.0-alpha1.19426.15`). It should have the same `.TakeWhile(c => char.IsDigit(c) || c == '.')` treatment as the version extracted later in `FromProductVersion`.

adamsitnik: @filipnavara great catch! I've send a separate PR with the fix https://github.com/dotnet/BenchmarkDotNet/pull/1233

adamsitnik:
> We could make new strong type API

@wojtpl2 you are right, the `.With(magicTypeName)` methods work only if you know the magic type name. 

> The old API can be marked as [Obsolete] so that the old programs work with new BDN.

I think that we can have both, the existing magic `.With` and the new strong-typed API. We try to minimize the number of breaking changes, here I simply had to do that.

Could you please create an issue with a short proposal?

Issue #1168, opened 2019-05-25
Consider using default value instead of hardcoded '-' in MetricColumn.GetValue()


sleemer:
What we have currently:
```   
 Method |       Mean |    StdDev |  Gen 0 | Allocated |
---------- |----------- |---------- |------- |---------- |
 Iterative | 31.0739 ns | 0.1091 ns |      - |       0 B |
      LINQ | 83.0435 ns | 1.0103 ns | 0.0069 |      32 B | 
```

What we need:
```   
 Method |       Mean |    StdDev |  Gen 0 | Allocated |
---------- |----------- |---------- |------- |---------- |
 Iterative | 31.0739 ns | 0.1091 ns | 0.0000 |       0 B |
      LINQ | 83.0435 ns | 1.0103 ns | 0.0069 |      32 B | 
```

Why we need it:
We run benchmarks as a step of our TeamCity build. And having consistent csv report file will simplify exporting results into influxdb using csv plugin for telegraf.

AndreyAkinshin:
@sleemer thanks for the request!
@adamsitnik what do you think about it?

adamsitnik:
I am ok with changing the CSV file output to be more consistent, but I would prefer not to change the way we display the results in the console/GH/HTML. I like the fact that we display `-` - it's more obvious and easier to read (I don't need to look at the digits)

AndreyAkinshin:
@sleemer is it OK if we change the default value only in the CSV reports?

PR #1240, opened 2019-09-03
Amend SummaryStyle to support visualization of '0' in reports

sleemer:
In order to use generated csv reports for publishing results to TSDB (e.g. Influx) we need to change the way '0' are visualized. By default they are dashed '-'. The goal of this PR is to change it for CsvExporter, but leave it as it is for the other exporters (console/GH/HTML).

Addresses #1168


AndreyAkinshin:
@sleemer thanks!

PR #1248, opened 2019-09-17
Support very long string as benchmark arguments

### LINKED ISSUE #1431: System.IO.FileNotFoundException with EtwProfiler
Hello,

I am receiving the following exception when running the EtwProfiler profiler in Windows 10 as administrator:

```
System.IO.FileNotFoundException
  HResult=0x8007007B
  Message=The filename, directory name, or volume label syntax is incorrect. (0x8007007B)
  Source=System.Private.CoreLib
  StackTrace:
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode)
   at Microsoft.Diagnostics.Tracing.Session.TraceEventSession.EnableKernelProvider(Keywords flags, Keywords stackCapture)
   at BenchmarkDotNet.Diagnostics.Windows.KernelSession.EnableProviders()
   at BenchmarkDotNet.Diagnostics.Windows.EtwProfiler.Start(DiagnoserActionParameters parameters)
   at BenchmarkDotNet.Diagnostics.Windows.EtwProfiler.Handle(HostSignal signal, DiagnoserActionParameters parameters)
   at BenchmarkDotNet.Diagnosers.CompositeDiagnoser.Handle(HostSignal signal, DiagnoserActionParameters parameters)
   at BenchmarkDotNet.Toolchains.DotNetCli.DotNetCliExecutor.Execute(BenchmarkCase benchmarkCase, BenchmarkId benchmarkId, ILogger logger, ArtifactsPaths artifactsPaths, IDiagnoser diagnoser, String executableName, IResolver resolver)
   at BenchmarkDotNet.Toolchains.DotNetCli.DotNetCliExecutor.Execute(ExecuteParameters executeParameters)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.RunExecute(ILogger logger, BenchmarkCase benchmarkCase, BenchmarkId benchmarkId, IToolchain toolchain, BuildResult buildResult, IResolver resolver, IDiagnoser diagnoser, Boolean& success)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Execute(ILogger logger, BenchmarkCase benchmarkCase, BenchmarkId benchmarkId, IToolchain toolchain, BuildResult buildResult, IResolver resolver)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.RunCore(BenchmarkCase benchmarkCase, BenchmarkId benchmarkId, ILogger logger, IResolver resolver, BuildResult buildResult)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Run(BenchmarkRunInfo benchmarkRunInfo, Dictionary`2 buildResults, IResolver resolver, ILogger logger, List`1 artifactsToCleanup, String resultsFolderPath, String logFilePath, StartedClock& runChronometer)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.RunWithDirtyAssemblyResolveHelper(String[] args, IConfig config)
   at com.github.akovac35.AdapterInterceptor.Benchmarks.Program.Main(String[] args) in C:\tmp\AdapterInterceptor\test\com.github.akovac35.AdapterInterceptor.Benchmarks\Program.cs:line 13

  This exception was originally thrown at this call stack:
    System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(int)
    Microsoft.Diagnostics.Tracing.Session.TraceEventSession.EnableKernelProvider(Microsoft.Diagnostics.Tracing.Parsers.KernelTraceEventParser.Keywords, Microsoft.Diagnostics.Tracing.Parsers.KernelTraceEventParser.Keywords)
    BenchmarkDotNet.Diagnostics.Windows.KernelSession.EnableProviders()
    BenchmarkDotNet.Diagnostics.Windows.EtwProfiler.Start(BenchmarkDotNet.Diagnosers.DiagnoserActionParameters)
    BenchmarkDotNet.Diagnostics.Windows.EtwProfiler.Handle(BenchmarkDotNet.Engines.HostSignal, BenchmarkDotNet.Diagnosers.DiagnoserActionParameters)
    BenchmarkDotNet.Diagnosers.CompositeDiagnoser.Handle(BenchmarkDotNet.Engines.HostSignal, BenchmarkDotNet.Diagnosers.DiagnoserActionParameters)
    BenchmarkDotNet.Toolchains.DotNetCli.DotNetCliExecutor.Execute(BenchmarkDotNet.Running.BenchmarkCase, BenchmarkDotNet.Running.BenchmarkId, BenchmarkDotNet.Loggers.ILogger, BenchmarkDotNet.Toolchains.ArtifactsPaths, BenchmarkDotNet.Diagnosers.IDiagnoser, string, BenchmarkDotNet.Characteristics.IResolver)
    BenchmarkDotNet.Toolchains.DotNetCli.DotNetCliExecutor.Execute(BenchmarkDotNet.Toolchains.Parameters.ExecuteParameters)
    BenchmarkDotNet.Running.BenchmarkRunnerClean.RunExecute(BenchmarkDotNet.Loggers.ILogger, BenchmarkDotNet.Running.BenchmarkCase, BenchmarkDotNet.Running.BenchmarkId, BenchmarkDotNet.Toolchains.IToolchain, BenchmarkDotNet.Toolchains.Results.BuildResult, BenchmarkDotNet.Characteristics.IResolver, BenchmarkDotNet.Diagnosers.IDiagnoser, ref bool)
    BenchmarkDotNet.Running.BenchmarkRunnerClean.Execute(BenchmarkDotNet.Loggers.ILogger, BenchmarkDotNet.Running.BenchmarkCase, BenchmarkDotNet.Running.BenchmarkId, BenchmarkDotNet.Toolchains.IToolchain, BenchmarkDotNet.Toolchains.Results.BuildResult, BenchmarkDotNet.Characteristics.IResolver)
    ...
    [Call Stack Truncated]
```

Benchmark starts a job but after a while it fails, probably just before a new job is started. Would you happen to know what the problem could be?

Problem can be reproduced by executing problem/test/com.github.akovac35.AdapterInterceptor.Benchmarks/```dotnet run -c Release --filter AdapterInterceptor_Invocation_Benchmark```

[problem.zip](https://github.com/dotnet/BenchmarkDotNet/files/4534726/problem.zip)

Using "BenchmarkDotNet.Diagnostics.Windows" Version="0.12.1".

Regards,
Aleksander

adamsitnik:
Fixes #1431

/cc @eerhardt


src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
```diff
@@ -147,6 +150,20 @@ private string GetFilePath(DiagnoserActionParameters details, DateTime creationT
...
+        {
+            string methodName = FullNameProvider.GetMethodName(benchmarkCase);
+            if (methodName.Length <= MaxBenchmarkNameLength)
+                return methodName;
+
+            methodName = FullNameProvider.GetMethodName(benchmarkCase, maxArgumentLength: 24); // try to limit argument length
+            if (methodName.Length <= MaxBenchmarkNameLength)
+                return methodName;
+
+            return methodName.Substring(0, MaxBenchmarkNameLength);
```
AndreyAkinshin: It seems that such an approach may lead to the same "BenchmarkName"s for different benchmarks. Can it be a problem?

tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
```diff
@@ -636,10 +636,27 @@ public void Test(SomeEnum defined, SomeEnum undefined, SomeEnum undefinedNegativ
...
+
+        public class WithVeryLongString
         {
-            var config = CreateSimpleConfig(job: Job.Dry.With(toolchain));
-            CanExecute<T>(config);
+            private readonly string LongString = new string('a', 200_000);
+
+            public IEnumerable<object> Arguments()
+            {
+                yield return LongString;
```
AndreyAkinshin: Let's add a second argument here with the following value:
```cs
private readonly string LongString2 = new string('a', 200_000 - 1) + "b";
```

adamsitnik: good idea! #done

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
adamsitnik: Very good catch! In such case, we are going to overwrite the trace file.

This is obviously far from perfect. I could generate some unique file name (using guid for example) but it would make it harder for the end-user to connect the trace file name with the benchmark.

@eerhardt let's say that you want to run following benchmark and use `EtwProfiler` to profile it:

```cs
public IEnumerable<object> Arguments()
{
    yield return new string('a', 200_000);
    yield return new string('a', 200_000 - 1) + "b";
}

[Benchmark]
[ArgumentsSource(nameof(Arguments))]
public void Some(string value)
```

We can't use full benchmark name with the super long string as a trace file name and we can not take just the first part of the string because the file name would be the same for two benchmarks.

Would using a guid in such case as a trace file name be suprising to the end user?

eerhardt: I would suggest keeping the beginning of the name the same, so I could tell that these 2 benchmarks were `Some` benchmarks. But then appending the Guid to the end would be understandable to me.

So either leaving out the args all together:

`MyNamespace.MyClass.MyMethod.NewGuid`

or chopping the arguments `NewGuid` characters before the max, and then appending `NewGuid` to the end.

`MyNamespace.MyClass.MyMethod.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaNewGuid`

I don't think purely using a `Guid` for the whole name would be understandable.

AndreyAkinshin: It will be nice if generate the same name for the same benchmarking. I'm suggesting to use a hashcode of the whole string at the end of the "shortified" name instead of a random GUID.

adamsitnik: > It will be nice if generate the same name for the same benchmarking. I'm suggesting to use a hashcode of the whole string at the end of the "shortified" name instead of a random GUID

This is a good suggestion, however, afaik in .NET Core, the hash codes are the same for given string value only for current program lifetime. If you close the app and run it again, then the value is different (afaik for some security reasons)

eerhardt: One possibility would be to use a different hash algorithm, ex. SHA or [MurmurHash](https://en.wikipedia.org/wiki/MurmurHash).

We use Murmur hashing in ML.NET if you want to copy the code:

https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Core/Utilities/Hashing.cs

adamsitnik:
@AndreyAkinshin @eerhardt it's been a while but the PR is now ready for review.

eerhardt:
> // file copied from https://github.com/dotnet/runtime/blob/master/src/libraries/Common/tests/System/IO/PathFeatures.cs

You probably want to retain the copyright on this and Hashing.cs files.

---
Refers to: src/BenchmarkDotNet/Extensions/PathFeatures.cs:1 in 183927e. [](commit_id = 183927ec13cb958e5db3b50635b6661eeaab07cc, deletion_comment = False)

src/BenchmarkDotNet/Extensions/Hashing.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// Implements the murmur hash 3 algorithm, using a mock UTF-8 encoding.
+        /// The UTF-8 conversion ignores the possibilities of unicode planes other than the 0th.
+        /// That is, it simply converts char values to one, two, or three bytes according to
+        /// the following rules:
+        /// * 0x0000 to 0x007F : 0xxxxxxx
+        /// * 0x0080 to 0x07FF : 110xxxxx 10xxxxxx
+        /// * 0x0800 to 0xFFFF : 1110xxxx 10xxxxxx 10xxxxxx
+        /// NOTE: This MUST match the StringBuilder version below.
+        /// </summary>
+        public static uint MurmurHash(uint hash, ReadOnlySpan<char> span, bool toUpper = false)
```
eerhardt, line 24: (nit) do you want to make this private, so other code can only see the above method?

src/BenchmarkDotNet/Extensions/Hashing.cs
```diff
@@ -0,0 +1,121 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// file copied from https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Core/Utilities/Hashing.cs
```
adamsitnik, line 5:
```suggestion
// file copied from https://github.com/dotnet/machinelearning/blob/b31bdee6671bfe50460f9279609dd948f8ce081c/src/Microsoft.ML.Core/Utilities/Hashing.cs
```

src/BenchmarkDotNet/Extensions/PathFeatures.cs
```diff
@@ -0,0 +1,102 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// file copied from https://github.com/dotnet/runtime/blob/master/src/libraries/Common/tests/System/IO/PathFeatures.cs
```
adamsitnik, line 5:
```suggestion
// file copied from https://github.com/dotnet/runtime/blob/bf2e135c12cbd34aeba2fa4a31d0e84184041a17/src/libraries/Common/tests/System/IO/PathFeatures.cs
```

adamsitnik:
> it's better to reference a specific commit or a specific tag

done

PR #1252, opened 2019-09-26
Add possibility to disable the log file

### LINKED ISSUE #198: [Feature request] No logger for benchmark run?
As a proposal: add an ability to disable [logging for each benchmark run](https://github.com/PerfDotNet/BenchmarkDotNet/blob/571b3dec965d6341ed5a3432c5ac75458f2c7457/BenchmarkDotNet/Running/BenchmarkRunner.cs#L50).

Use case: as usual:)  
Unit test runner, dozens of perftests, tons of garbage in output dir :)
Most of it is removed as we use custom in-process runner, the ones that left are the log files and empty `results` folder (created by [this line](https://github.com/PerfDotNet/BenchmarkDotNet/blob/571b3dec965d6341ed5a3432c5ac75458f2c7457/BenchmarkDotNet/Running/BenchmarkRunner.cs#L98)).

CodeTherapist:
The proposed solution emphasizes:

* Minimal change
* Keeping the current behavior
* No IO when the log file is disabled (perf+)
* Simple to use for the users (relies on well-known API, `IConfig.With(ConfigOptions.DisableLogFile);`)

- [x] Unit Tests
- [x] Xml Doc (Intelisense)
- [x] Documentation (docs)

Documentation will be done, as soon the proposed solution is accepted and final.

Closes #198

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/BenchmarkSwitcherTest.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/BenchmarkSwitcherTest.cs
@@ -13,6 +13,7 @@ using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Tests.Loggers;
 using BenchmarkDotNet.Tests.XUnit;
 using Xunit.Abstractions;
+using System.IO;
 
 namespace BenchmarkDotNet.IntegrationTests
 {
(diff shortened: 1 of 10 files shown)


AndreyAkinshin:
LGTM.

tests/BenchmarkDotNet.Tests/Configs/ConfigOptionsTests.cs
```diff
@@ -17,6 +17,9 @@ public class ConfigOptionsTests
         [Fact]
         public void DefaultConfigDoesNotStopOnFirstError() => Assert.False(DefaultConfig.Instance.Options.HasFlag(ConfigOptions.StopOnFirstError));
 
+        [Fact]
+        public void DefaultConfigDoesNotDisableLogFile() => Assert.False(DefaultConfig.Instance.Options.HasFlag(ConfigOptions.DisableLogFile));
```
adamsitnik: Awesome! thanks for adding this test! 👍

tests/BenchmarkDotNet.IntegrationTests/BenchmarkSwitcherTest.cs
```diff
@@ -124,7 +125,59 @@ public void WhenUserAsksToPrintAListAndProvidesAFilterWePrintFilteredList()
...
+            finally
+            {
+                if (!string.IsNullOrWhiteSpace(logFilePath))
+                {
+                    File.Delete(logFilePath);
+                }
+            }
+        }
+
+
```
adamsitnik: nit: empty extra line

```suggestion
```

tests/BenchmarkDotNet.IntegrationTests/BenchmarkSwitcherTest.cs
```diff
@@ -124,7 +125,59 @@ public void WhenUserAsksToPrintAListAndProvidesAFilterWePrintFilteredList()
...
+                var summary = summaries.Single();
+                logFilePath = summary.LogFilePath;
+                Assert.False(File.Exists(logFilePath), $"Logfile '{logFilePath}' should not exist, but it does.");
+            }
+            finally
+            {
+                if (!string.IsNullOrWhiteSpace(logFilePath))
+                {
+                    File.Delete(logFilePath);
```
adamsitnik: I would move this logic before the try block, to ensure that the file does not exist before calling `RunAll`

CodeTherapist: I was thinking about that issue as well, but unfortunately the file name pattern generation isn't accessible. Thus, generating the file name outside within the test context would break the test, even if the assertion of it would still be valid.

Maybe I missed the piece of how I would do that or my thoughts are fundamentally wrong - could you please elaborate on this?

adamsitnik: Good point, I missed that. I think that we can stay with the current implementation.

PR #1260, opened 2019-10-01
Fix ca1825 rule - Avoid zero-length array allocations

wojtpl2:
In this PR I fix CA1825 rule - Avoid zero-length array allocations
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1825?view=vs-2019

PR for Hacktoberfest 😄


wojtpl2:
I had to revert changes from BenchymarkDotNet.Annotations because .Net Standard doesn't have Array.Empty<T>

PR #1263, opened 2019-10-01
Configuration compatibility validation

### LINKED ISSUE #1238: RunAllJoined Causing Exception
If I run my benchmarks with Run or RunAll, they both work fine. If I run RunAllJoined (below) it causes an exception. I'm using version 0.11.5.1161

```
var config = DefaultConfig.Instance.With(Job.Default.With(CsProjCoreToolchain.NetCoreApp22)).With(Job.Default.With(CsProjClassicNetToolchain.Net48));
BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).RunAllJoined(config);

Unhandled Exception: System.InvalidOperationException: Sequence contains more than one element
   at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source)
   at BenchmarkDotNet.Reports.Summary.GetConfiguredOrdererOrDefaultOne(IEnumerable`1 configs)
   at BenchmarkDotNet.Reports.Summary..ctor(String title, ImmutableArray`1 reports, HostEnvironmentInfo hostEnvironmentInfo, String resultsDirectoryPath, String logFilePath, TimeSpan totalTime, ImmutableArray`1 validationErrors)
   at BenchmarkDotNet.Reports.Summary.Join(List`1 summaries, ClockSpan clockSpan)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.RunWithDirtyAssemblyResolveHelper(String[] args, IConfig config)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.RunAllJoined(IConfig config)
   at dotNetTips.Utility.Benchmarks.Program.Main() in C:\src\git\dotNetTips.Utility\Benchmarks\dotNetTips.Utility.Benchmarks\dotNetTips.Utility.Benchmarks\Program.cs:line 24
```

gsomix:
Closes #1238 

It's not complete solution, just a starting point for discussion. I've reviewed @AndreyAkinshin repro and existing validation system. It seems for now Validators work only per benchmark type and closely coupled with their configs.

I propose the following solution:

1. introduce special Validator for config compatibility checks; we need to call it from `BenchmarkRunnerClean.Validate` method in case of `JoinSummary` option enabled;
1. join all benchmark cases and Validators from all types;
1. merge all configs into one;
1. validate the cases with Validators and the config; we will get additional checks for free, e.g. `BaselineValidator`;

@adamsitnik @AndreyAkinshin What do you think? I'm looking forward to your response.

Thanks and happy #hacktoberfest!

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -256,6 +256,17 @@ namespace BenchmarkDotNet.Running
 
             var validationErrors = new List<ValidationError>();
 
+            if (benchmarks.Any(b => b.Config.Options.IsSet(ConfigOptions.JoinSummary)))
+            {
+                var joinedCases = benchmarks.SelectMany(b => b.BenchmarksCases).ToArray();
+
+                validationErrors.AddRange(
+                    ConfigCompatibilityValidator
+                        .FailOnError
+                        .Validate(new ValidationParameters(joinedCases, null))
+                    );
+            }
+
             foreach (var benchmarkRunInfo in benchmarks)
                 validationErrors.AddRange(benchmarkRunInfo.Config.GetCompositeValidator().Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config)));
 
--- a/src/BenchmarkDotNet/Validators/ConfigCompatibilityValidator.cs
+++ b/src/BenchmarkDotNet/Validators/ConfigCompatibilityValidator.cs
@@ -1,0 +1,26 @@
+﻿using System.Linq;
+using System.Collections.Generic;
+using BenchmarkDotNet.Configs;
+
+namespace BenchmarkDotNet.Validators
+{
+    public class ConfigCompatibilityValidator : IValidator
+    {
+        public static readonly ConfigCompatibilityValidator FailOnError = new ConfigCompatibilityValidator();
+
+        public bool TreatsWarningsAsErrors => true;
+
+        public IEnumerable<ValidationError> Validate(ValidationParameters validationParameters)
+        {
+            var orderers =
+                validationParameters
+                    .Benchmarks
+                    .Where(benchmark => benchmark.Config.Orderer != Order.DefaultOrderer.Instance)
+                    .Select(benchmark => benchmark.Config.Orderer)
+                    .Distinct();
+
+            if (orderers.Count() > 1)
+                yield return new ValidationError(true, "You use JoinSummary options, but provided configurations cannot be joined. Only one Orderer per benchmark cases is allowed.");
+        }
+    }
+}
--- a/tests/BenchmarkDotNet.Tests/Validators/ConfigCompatibilityValidatorTests.cs
+++ b/tests/BenchmarkDotNet.Tests/Validators/ConfigCompatibilityValidatorTests.cs
@@ -1,0 +1,122 @@
+﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Order;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Tests.Loggers;
+using BenchmarkDotNet.Validators;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.Tests.Validators
+{
+    public class ConfigCompatibilityValidatorTests
+    {
+        private ITestOutputHelper Output { get; }
+
+        public ConfigCompatibilityValidatorTests(ITestOutputHelper output) => Output = output;
+
+        [Fact]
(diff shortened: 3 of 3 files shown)


adamsitnik:
Hi @gsomix 

I like the idea of introducing a new Validator that gets all the benchmarks (`BenchmarkRunInfo[] benchmarks` from `BenchmarkRunnerClean.Validate`) and performs this special validation.

> merge all configs into one

I am not sure if this is going to be easy|possible. Merging the configs is non-trivial and you might run into handling some edge cases not related to this problem. Moving the logic that you implemented in this PR to a dedicated type should be enough. It should also make it easier for you to write the tests for it.

Thanks for working on this!

gsomix:
Finished! CI is green (almost).

gsomix:
@adamsitnik @AndreyAkinshin Hey, guys, any chance to get it reviewed? Thank you!

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -248,6 +248,17 @@ private static ImmutableArray<ValidationError> Validate(BenchmarkRunInfo[] bench
 
             var validationErrors = new List<ValidationError>();
 
+            if (benchmarks.Any(b => b.Config.Options.IsSet(ConfigOptions.JoinSummary)))
+            {
+                var joinedCases = benchmarks.SelectMany(b => b.BenchmarksCases).ToList();
```
adamsitnik: nit: if you don't plan to add any items to the collection, an array is a better choice than list (and here `ValidationParameters` accepts an `IReadOnlyList`)

```suggestion
                var joinedCases = benchmarks.SelectMany(b => b.BenchmarksCases).ToArray();
```

src/BenchmarkDotNet/Validators/ConfigCompatibilityValidator.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public class ConfigCompatibilityValidator : IValidator
+    {
+        public static readonly ConfigCompatibilityValidator FailOnError = new ConfigCompatibilityValidator();
+
+        public bool TreatsWarningsAsErrors => true;
+
+        public IEnumerable<ValidationError> Validate(ValidationParameters validationParameters)
+        {
+            var benchmarks = validationParameters.Benchmarks;
```
adamsitnik: nit: this local variable is used only once, it could be inlined to shorten the source code

src/BenchmarkDotNet/Validators/ConfigCompatibilityValidator.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            var orderers =
+                benchmarks
+                    .Select(benchmark => benchmark.Config)
+                    .Where(config => config.Orderer != Order.DefaultOrderer.Instance)
+                    .Select(config => config.Orderer)
+                    .Distinct();
```
adamsitnik: nit: `.Select().Where().Select()` is a performance anti-pattern, it's better to filter first and then map which allows us to avoid an additional map operation

src/BenchmarkDotNet/Validators/ConfigCompatibilityValidator.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            if (orderers.Take(2).Count() > 1)
```
adamsitnik: IMHO the `Take` is not needed and adds an extra complexity

```suggestion
            if (orderers.Count() > 1)
```

tests/BenchmarkDotNet.IntegrationTests/JoinedBenchmarksTests.cs
```diff
@@ -0,0 +1,46 @@
+﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Order;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Tests.Loggers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.IntegrationTests
```
adamsitnik: big thanks for adding the test! It looks like the benchmarks should never be executed in this case so it can be moved to the unit tests project (we try to keep the number of intergration tests low)

tests/BenchmarkDotNet.Tests/Validators/ConfigCompatibilityValidatorTests.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace BenchmarkDotNet.Tests.Validators
+{
+    public class ConfigCompatibilityValidatorTests
+    {
+        [Fact]
+        public void JoinedBenchmarksShouldntHaveDifferentOrderers()
```
adamsitnik: Could you please add one more test for a case where there are no extra orderders defined and the validator does not return any errors?

gsomix:
@adamsitnik Thanks! I'll take a look on weekend.

gsomix:
@adamsitnik Thanks you for the review! It's ready for the second round. :)

PR #1255, opened 2019-09-26
Update Build Tools to use .NET Core SDK to 3.0.100

CodeTherapist:
closes #1254 
required for #1253

PR #1264, opened 2019-10-02
Update Build Tools to use .NET Core SDK to 3.0.100 (#1255)

adamsitnik:
/cc @CodeTherapist @Ky7m

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -32,7 +32,7 @@ Param(
 )
 
 $CakeVersion = "0.30.0"
-$DotNetVersion = "2.1.503";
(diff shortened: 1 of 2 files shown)


adamsitnik:
The AzDO and AppVeyor builds are green. Travis time outs...

We could just disable Travis and switch to AzDO but we would have to configure our CI to run the AzDO jobs for all users, not just me and Andrey.

I need to think about it for a moment

AndreyAkinshin:
@adamsitnik ?

adamsitnik:
Closing, if we decide to switch to 3.* it should be 3.1

PR #1265, opened 2019-10-02
make it possible to configure MaxParamterColumnWidth

adamsitnik:
Recently I've been working on improving the performance of `StartsWith` and `EndsWith` on Linux. I am interested in some test cases, not all possible permutations so I am using a `[ParamsSource]` combined with `ValueTuple`.

Example:

```cs
public static IEnumerable<(CultureInfo CultureInfo, CompareOptions CompareOptions, bool highChars)> GetOptions()
{
    // Ordinal and OrdinalIgnoreCase use single execution path for all cultures, so we test it only for "en-US"
    // without enforcing highChars - the execution path would be the same
    yield return (new CultureInfo("en-US"), CompareOptions.Ordinal, false);
    yield return (new CultureInfo("en-US"), CompareOptions.OrdinalIgnoreCase, false);

    yield return (new CultureInfo("en-US"), CompareOptions.None, false); // no high chars = fast path (managed code on Unix)
    yield return (new CultureInfo("en-US"), CompareOptions.None, true); // high chars = slow path (call to ICU on Unix)
    yield return (new CultureInfo("en-US"), CompareOptions.IgnoreCase, false);
    yield return (new CultureInfo("en-US"), CompareOptions.IgnoreCase, true);

    // IgnoreSymbols always uses the slow path
    // https://github.com/dotnet/coreclr/blob/cd6bc26bdc4ac06fe2165b283eaf9fb5ff5293f4/src/System.Private.CoreLib/shared/System/Globalization/CompareInfo.Unix.cs#L911-L912
    yield return (new CultureInfo("en-US"), CompareOptions.IgnoreSymbols, false);

    // too rarely used to test both slow and fast path
    yield return (new CultureInfo("en-US"), CompareOptions.IgnoreNonSpace, false);

    yield return (CultureInfo.InvariantCulture, CompareOptions.None, false);
    yield return (CultureInfo.InvariantCulture, CompareOptions.None, true);
    yield return (CultureInfo.InvariantCulture, CompareOptions.IgnoreCase, false);
    yield return (CultureInfo.InvariantCulture, CompareOptions.IgnoreCase, true);

    // both en-US and Invariant cultures have an optimized execution path on Unix systems:
    // https://github.com/dotnet/coreclr/blob/cd6bc26bdc4ac06fe2165b283eaf9fb5ff5293f4/src/System.Private.CoreLib/shared/System/Globalization/CompareInfo.Unix.cs#L35
    // so we need one more culture to test the "slow path"
    // Polish language has a lot of special characters, for example 'ch', 'rz', 'sz', 'cz' use two chars to express one ;)
    // it also has a lot of characters with accent so we use it as an example of "complex" language
    yield return (new CultureInfo("pl-PL"), CompareOptions.None, false);
}

[ParamsSource(nameof(GetOptions))]
public (CultureInfo CultureInfo, CompareOptions CompareOptions, bool highChars) Options;
```

The problem is that some important information (`CompareOptions` value in this case) gets trimmed in the summary table:

|              Options |
|--------------------- |
| (, Ig(...)alse) [21] |
| (, IgnoreCase, True) |
|      (, None, False) |
|       (, None, True) |
| (en-U(...)alse) [26] |
| (en-U(...)True) [25] |
| (en-U(...)alse) [30] |
| (en-U(...)alse) [29] |
| (en-US, None, False) |
|  (en-US, None, True) |
| (en-U(...)alse) [23] |
| (en-U(...)alse) [33] |
| (pl-PL, None, False) |

This PR makes it possible to customize the max width.

Example:

|                           Options |
|---------------------------------- |
|             (, IgnoreCase, False) |
|              (, IgnoreCase, True) |
|                   (, None, False) |
|                    (, None, True) |
|        (en-US, IgnoreCase, False) |
|         (en-US, IgnoreCase, True) |
|    (en-US, IgnoreNonSpace, False) |
|     (en-US, IgnoreSymbols, False) |
|              (en-US, None, False) |
|               (en-US, None, True) |
|           (en-US, Ordinal, False) |
| (en-US, OrdinalIgnoreCase, False) |
|              (pl-PL, None, False) |

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
@@ -159,7 +159,7 @@ namespace BenchmarkDotNet.ConsoleArguments
         [Option("disableLogFile", Required = false, HelpText = "Disables the logfile.")]
         public bool DisableLogFile { get; set; }
 
-        [Option("maxWidth", Required = false, HelpText = "Max Paramter Column Width, the default is 20.")]
+        [Option("maxWidth", Required = false, HelpText = "Max paramter column width, the default is 20.")]
         public int? MaxParamterColumnWidth { get; set; }
 
         internal bool UserProvidedFilters => Filters.Any() || AttributeNames.Any() || AllCategories.Any() || AnyCategories.Any();
--- a/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
@@ -7,6 +7,7 @@ using BenchmarkDotNet.Diagnosers;
 using BenchmarkDotNet.Engines;
 using BenchmarkDotNet.Helpers;
 using BenchmarkDotNet.Mathematics;
+using BenchmarkDotNet.Parameters;
 using CommandLine;
 using CommandLine.Text;
 using JetBrains.Annotations;
@@ -157,6 +158,9 @@ namespace BenchmarkDotNet.ConsoleArguments
 
         [Option("disableLogFile", Required = false, HelpText = "Disables the logfile.")]
         public bool DisableLogFile { get; set; }
+
+        [Option("maxWidth", Required = false, HelpText = "Max Paramter Column Width, the default is 20.")]
+        public int? MaxParamterColumnWidth { get; set; }
 
         internal bool UserProvidedFilters => Filters.Any() || AttributeNames.Any() || AllCategories.Any() || AnyCategories.Any();
 
--- a/tests/BenchmarkDotNet.Tests/ParameterInstanceTests.cs
+++ b/tests/BenchmarkDotNet.Tests/ParameterInstanceTests.cs
@@ -1,4 +1,6 @@
-﻿using BenchmarkDotNet.Parameters;
+﻿using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Parameters;
+using BenchmarkDotNet.Reports;
 using System;
 using System.Collections.Generic;
 using Xunit;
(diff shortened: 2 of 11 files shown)


src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
```diff
@@ -158,6 +159,9 @@ public class CommandLineOptions
         [Option("disableLogFile", Required = false, HelpText = "Disables the logfile.")]
         public bool DisableLogFile { get; set; }
 
+        [Option("maxWidth", Required = false, HelpText = "Max Paramter Column Width, the default is 20.")]
```
wojtpl2: "Max parameter column width, the default is 20."?

tests/BenchmarkDotNet.Tests/ParameterInstanceTests.cs
```diff
@@ -15,18 +17,20 @@ public class ParameterInstanceTests
         [InlineData("short")]
         public void ShortParameterValuesDisplayOriginalValue(object value)
         {
-            var parameter = new ParameterInstance(definition, value);
+            var config = DefaultConfig.Instance.CreateImmutableConfig();
+            var parameter = new ParameterInstance(definition, value, config);
```
wojtpl2: It is not an issue. Just my opinion.

I think the `ParameterInstance` class doesn't need the whole configuration. `config.SummaryStyle` would be enough (or maybe even `config.SummaryStyle?.MaxParamterColumnWidth`). I would create an `ISummaryStyle` interface.

Currently, this test is not a unit test because you test two elements. Creating the default configuration and `ParameterInstance.ToDisplayText()`

adamsitnik: Good point, I've made `ParameterInstance` depend on `SummaryStyle`, not entire `Config`

PR #1277, opened 2019-10-12
change the runtimes API

adamsitnik:
@AndreyAkinshin this fixes the things we have discussed yesterday at Dotnetos

* Don't introduce NetCoreLatest and its friends
* Rename TargetFrameworkJob -> SimpleJob (don't introduce JobAttribute in 0.12)
* Rename TargetFrameworkMoniker -> RuntimeMoniker

When it comes to:

> * We make [ClrJob] are friends as obsolete without errors (warning only) and use InvalidConfig with nice message

I made it a warning as we talked but I've not introduced the concept of `InvalidConfig`. 

What I did instead is to call the `GetCurrentVersion` method for every runtime.

```cs
public ClrJobAttribute() : base(Job.Default.With(ClrRuntime.GetCurrentVersion()))
```

If the `host runtime == job runtime`, then it's going to work fine (99% of the cases). If the runtimes are different: for example running as Full .NET Framework and trying to use `[CoreJob]` attribute, then `GetCurrentVersion` is going to throw:

https://github.com/dotnet/BenchmarkDotNet/blob/2ba30330ec5ec8a9abe284c927277a6fe0fc5a0c/src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs#L44


docs/_changelog/header/v0.11.6.md
```diff
@@ -187,18 +187,18 @@ class Program
...
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Jobs;
 
 namespace BenchmarkDotNet.Samples
 {
-    [TargetFrameworkJob(TargetFrameworkMoniker.Net48)]
-    [TargetFrameworkJob(TargetFrameworkMoniker.Mono)]
-    [TargetFrameworkJob(TargetFrameworkMoniker.NetCoreApp21)]
-    [TargetFrameworkJob(TargetFrameworkMoniker.NetCoreApp30)]
+    [SimpleJob(runtimeMoniker: RuntimeMoniker.Net48)]
```
AndreyAkinshin: Don't like the `runtimeMoniker` prefix. Could we introduce a few additional `SimpleJobAttribute` ctor overloads with the `RuntimeMoniker` as the first ctor argument?

src/BenchmarkDotNet/Attributes/Obsolete/ClrJobAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿using BenchmarkDotNet.Environments;
+using BenchmarkDotNet.Jobs;
+using System;
+
+namespace BenchmarkDotNet.Attributes
+{
+    [Obsolete("Please use SimpleJobAttribute instead.", false)]
+    public class ClrJobAttribute : JobConfigBaseAttribute
+    {
+        public ClrJobAttribute() : base(Job.Default.With(ClrRuntime.GetCurrentVersion()))
```
AndreyAkinshin: 👍

src/BenchmarkDotNet/Attributes/Obsolete/DryCoreJobAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿using BenchmarkDotNet.Environments;
+using BenchmarkDotNet.Jobs;
+using System;
+
+namespace BenchmarkDotNet.Attributes
+{
+    [Obsolete("Please use DryJobAttribute instead. Use the ctor that requires RuntimeMoniker argument.", false)]
```
AndreyAkinshin: Instead of "Use the ctor that requires RuntimeMoniker argument", it will be better to write a code example that can be copy-pasted.

AndreyAkinshin:
@adamsitnik in general, it looks very good, thanks! I left just a few minor (but important) comments.

adamsitnik:
@AndreyAkinshin thanks for the review! I've pushed the fixes. Please let me know if there is any blocker for 0.11.6 from my side

Issue #1280, opened 2019-10-14
Improvement in memory statistics


wojtpl2:
I think that empty GC memory statistics can mislead users. In #1250 I create comment:

> @adamsitnik I think that BenchmarkDotNet should not print information about GC if it doesn't run extra iteration. I mean this line:
> 
> ```
> GC:  0 0 0 0 0
> Threading:  0 0 0
> ```
> 
> Extra iteration runs only when user adds `MemoryDiagnoser` or `ThreadingDiagnoser`. In other cases we shouldn't print these lines:
> 
> https://github.com/dotnet/BenchmarkDotNet/blob/2ba30330ec5ec8a9abe284c927277a6fe0fc5a0c/src/BenchmarkDotNet/Engines/RunResults.cs#L71-L72
> 
> Or even more precisely:
> `GC: 0 0 0 0 0` should print only with `MemoryDiagnoser`
> `Threading: 0 0 0` should print only with `ThreadingDiagnoser`

@adamsitnik What do you think?

adamsitnik:
If this does not require a lot of changes in the code, then I am ok with that 👍

PR #1281, opened 2019-10-15
Print GC and threading stats only if needed

wojtpl2:
Fix for #1280 

Without stats:
```
cd samples\BenchmarkDotNet.Samples\bin\Release\net461
.\BenchmarkDotNet.Samples.exe --filter IntroBasic
```
image

With stats (with `-m` option):
```
cd samples\BenchmarkDotNet.Samples\bin\Release\net461
.\BenchmarkDotNet.Samples.exe --filter IntroBasic -m
```

image

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/InProcessNoEmitExecutor.cs
+++ b/src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/InProcessNoEmitExecutor.cs
@@ -127,8 +127,10 @@ namespace BenchmarkDotNet.Toolchains.InProcess.NoEmit
             }
 
             var lines = runResults.GetMeasurements().Select(measurement => measurement.ToOutputLine()).ToList();
-            lines.Add(runResults.GCStats.ToOutputLine());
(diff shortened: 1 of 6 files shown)


src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/InProcessNoEmitExecutor.cs
```diff
@@ -127,8 +127,10 @@ private ExecuteResult GetExecutionResult(RunResults runResults, int exitCode, IL
             }
 
             var lines = runResults.GetMeasurements().Select(measurement => measurement.ToOutputLine()).ToList();
-            lines.Add(runResults.GCStats.ToOutputLine());
-            lines.Add(runResults.GCStats.ToOutputLine());
```
wojtpl2: There were GCStats twice. I assume that it was a mistake?

adamsitnik: it was most probably a bug

PR #1286, opened 2019-10-17
Is this supposed to be void?

kuskmen:


adamsitnik:
Hi @kuskmen 

Thank you for your PR.

This is a `constructor` of a type called `Config`, and it should not be `void`

kuskmen:
:no_mouth: :facepalm:

PR #1292, opened 2019-10-24
passed args to benchmark runner

### LINKED ISSUE #1290: BenchmarkRunner should support parsing console line arguments
As of today, only `BenchmarkSwitcher` supports console line arguments.

`BenchmarkRunner` should support it as well.

It would be also great to detect a situation when a user has passed some BDN arguments to the console app but forgot to pass it to BDN. We could do it by checking the `Environment.GetCommandLineArgs()`. I've seen it happen many times ;)

chan18:
fixes #1290


chan18:
did not wrote test cases yet, need a review from @adamsitnik,

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -16,18 +20,22 @@ namespace BenchmarkDotNet.Running
...
+        public static Summary Run<T>(IConfig config = null, string[] args = null)
         {
             using (DirtyAssemblyResolveHelper.Create())
-                return RunWithExceptionHandling(() => RunWithDirtyAssemblyResolveHelper(typeof(T), config));
+                return RunWithExceptionHandling(() => RunWithDirtyAssemblyResolveHelper(typeof(T), config, args ?? Array.Empty<string>()));
         }
 
         [PublicAPI]
-        public static Summary Run(Type type, IConfig config = null)
+        public static Summary Run(Type type, IConfig config = null, string[] args = null)
```
adamsitnik: could you please extend also an overload that takes an `Assembly` as an argument?

`public static Summary[] Run(Assembly assembly, IConfig config = null)`

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -16,18 +20,22 @@ namespace BenchmarkDotNet.Running
     // so it can ignore the version mismatch ;)
     public static class BenchmarkRunner
     {
+        private static readonly IUserInteraction userInteraction = new UserInteraction();
+        private static readonly List<Type> types = new List<Type>();
+        private static readonly List<Assembly> assemblies = new List<Assembly>();
```
adamsitnik: this field is never used for adding new assemblies, please remove it and just pass an empty list to `TypeFilter.GetTypesWithRunnableBenchmarks`

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -16,18 +20,22 @@ namespace BenchmarkDotNet.Running
     // so it can ignore the version mismatch ;)
     public static class BenchmarkRunner
     {
+        private static readonly IUserInteraction userInteraction = new UserInteraction();
+        private static readonly List<Type> types = new List<Type>();
```
adamsitnik: this should be a local variable of the `RunWithDirtyAssemblyResolveHelper` method

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -16,18 +20,22 @@ namespace BenchmarkDotNet.Running
     // so it can ignore the version mismatch ;)
     public static class BenchmarkRunner
     {
+        private static readonly IUserInteraction userInteraction = new UserInteraction();
```
adamsitnik: please make it a local variable, not a static field

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -73,8 +81,52 @@ public static Summary RunSource(string source, IConfig config = null)
...
+
+            if (options.ListBenchmarkCaseMode != ListBenchmarkCaseMode.Disabled)
+            {
+                PrintList(logger, effectiveConfig, allAvailableTypesWithRunnableBenchmarks, options);
+                return null;
+            }
+
+            var benchmarksToFilter = options.UserProvidedFilters
+                ? allAvailableTypesWithRunnableBenchmarks
+                : userInteraction.AskUser(allAvailableTypesWithRunnableBenchmarks, logger);
```
adamsitnik: if the user did not provide any filters, then you should not ask the user for the choice (to keep the current behavior)

Sth like:

```cs
var benchmarks = options.UserProvidedFilters
   ? TypeFilter.Filter(effectiveConfig, allAvailableTypesWithRunnableBenchmarks)
   : BenchmarkConverter.TypeToBenchmarks(type, config);

BenchmarkRunnerClean.Run(benchmarks).Single();
```

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -125,5 +177,16 @@ private static Summary[] RunWithExceptionHandling(Func<Summary[]> run)
                 return new[] { Summary.NothingToRun(e.Message, string.Empty, string.Empty) };
             }
         }
+        private static void PrintList(ILogger nonNullLogger, IConfig effectiveConfig, IReadOnlyList<Type> allAvailableTypesWithRunnableBenchmarks, CommandLineOptions options)
```
adamsitnik: please reuse the method from `BenchmarkSwitcher` and avoid having a copy:

https://github.com/dotnet/BenchmarkDotNet/blob/916dc355d2cd4845923a29afca089650fc6a7573/src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs#L112

You can make the `BenchmarkSwitcher.PrintList` method internal or move this logic to `BenchmarkCasesPrinter`

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
chan18: you mean we need on more,
```  public static Summary[] Run(Assembly assembly, IConfig config = null,string[] args = null) ```
that takes assembly, config and args.

File: src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs (hunk 3 above)
Comment (chan18) [reply src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs]: yes, makes sense.

File: src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs (hunk 5 above)
Comment (chan18) [reply src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs]: this simplify's code a bit.

File: src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs (hunk 6 above)
Comment (chan18) [reply src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs]: moved ``PrintList() ``` method to ```BenchmarkCasesPrinter```

**Comment by chan18** [comment]
@adamsitnik

File: src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs (hunk 1 above)
Comment (adamsitnik) [reply src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs]: > that takes assembly, config and arg

yes, exactly

File: src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs [hunk 7]
```diff
@@ -125,5 +168,6 @@ private static Summary[] RunWithExceptionHandling(Func<Summary[]> run)
                 return new[] { Summary.NothingToRun(e.Message, string.Empty, string.Empty) };
             }
         }
+
```
adamsitnik: nit: extra new line
```suggestion
```

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
```diff
@@ -13,8 +13,7 @@
 using JetBrains.Annotations;
 
 namespace BenchmarkDotNet.Running
-{
-    public class BenchmarkSwitcher
+{    public class BenchmarkSwitcher
```
adamsitnik: nit: formatting

```suggestion
{
    public class BenchmarkSwitcher
```

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
```diff
@@ -108,17 +107,6 @@ private IEnumerable<Summary> RunWithDirtyAssemblyResolveHelper(string[] args, IC
...
-            var printer = new BenchmarkCasesPrinter(options.ListBenchmarkCaseMode);
-
-            var testNames = TypeFilter.Filter(effectiveConfig, allAvailableTypesWithRunnableBenchmarks)
-                .SelectMany(p => p.BenchmarksCases)
-                .Select(p => p.Descriptor.GetFilterName())
-                .Distinct();
-
-            printer.Print(testNames, nonNullLogger);
-        }
+
```
adamsitnik: nit: extra new line
```suggestion
```

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
```diff
@@ -65,7 +64,7 @@ public IEnumerable<Summary> Run(string[] args = null, IConfig config = null)
         [MethodImpl(MethodImplOptions.NoInlining)]
         private IEnumerable<Summary> RunWithDirtyAssemblyResolveHelper(string[] args, IConfig config)
         {
-            var logger = config.GetNonNullCompositeLogger();
+            var logger = config.GetNonNullCompositeLogger();
```
adamsitnik: nit: formatting: trailing whitespaces

```suggestion
            var logger = config.GetNonNullCompositeLogger();
```

chan18:
https://github.com/dotnet/BenchmarkDotNet/blob/55899866dc0e97fe77870a770075c727b68cef3b/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs#L82-L118

needs another overload. I am thining to create another function to supperate the code that parses args and used it in two different,
``` 
RunWithDirtyAssemblyResolveHelper(Type type, IConfig config, string[] args)
```
``` 
RunWithDirtyAssemblyResolveHelper(Assembly assembly, IConfig config = null,string[] args) 
```

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
adamsitnik: I am sorry for not making myself clear. Could you please extend the existing method with a new optional parameter?

https://github.com/dotnet/BenchmarkDotNet/blob/7c31942f6a78c73e14f9d1b7451d866540657139/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs#L52

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -73,9 +84,86 @@ public static Summary RunSource(string source, IConfig config = null)
...
+            var benchmarks = options.UserProvidedFilters
+                           ? TypeFilter.Filter(effectiveConfig, allAvailableTypesWithRunnableBenchmarks)
+                           : new BenchmarkRunInfo[] { BenchmarkConverter.TypeToBenchmarks(type, config) };
 
+            return BenchmarkRunnerClean.Run(benchmarks).Single();
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static Summary RunWithDirtyAssemblyResolveHelper(Assembly assembly, IConfig config, string[] args)
```
adamsitnik: I think that we should remove the code duplication.

We could have a single method that would handle both `Type` and `Assembly` type arguments:

```cs
private static Summary RunWithDirtyAssemblyResolveHelper(Type type, IConfig config, string[] args)
 => RunWithDirtyAssemblyResolveHelper(new [] { type }, Array.Empty<Assembly>(), config, args);

private static Summary RunWithDirtyAssemblyResolveHelper(Assembly assembly, IConfig config, string[] args)
 => RunWithDirtyAssemblyResolveHelper(Array.Empty<Type>(), new [] { assembly }, config, args);

private static Summary RunWithDirtyAssemblyResolveHelper(IEnumerable<Type> types, IEnumerable<Assembly> assemblies, IConfig config, string[] args)
{
  // skipped for brevity
  var (allTypesValid, allAvailableTypesWithRunnableBenchmarks) = TypeFilter.GetTypesWithRunnableBenchmarks(types, assemblies, logger);
}
```

chan18: why would, IEnumerable<Type> types, IEnumerable<Assembly> assemblies will be IEnumerable? when run method is getting passed only one item of type and assembly! Is any other reason? 

Other way i think it might be better! passing ``` assembly.GetRunnableBenchmarks().ToArray() ``` to ``` RunWithDirtyAssemblyResolveHelper(Type type, IConfig config, string[] args) ``` with IEnumerable<Type> types..

by passing assembly types, we dont need to write ``` RunWithDirtyAssemblyResolveHelper() ``` that takes ```types, config and args[] and assembly``` as parameters and using some if validation to apply type rules to run benchmark.

adamsitnik: > why would, IEnumerable types, IEnumerable assemblies will be IEnumerable? when run method is getting passed only one item of type and assembly! Is any other reason?

You are right, we don't need an `IEnumerable` here.

> Other way i think it might be better!

The assemblies have a "special" treatment because they might also contain a generic benchmark types and unfortunately this would break the generic type benchmarks .

https://github.com/dotnet/BenchmarkDotNet/blob/2415fda6fb8eaf2d0f17b1828a7ee9befbd7baab/samples/BenchmarkDotNet.Samples/IntroGenericTypeArguments.cs#L6-L8

https://github.com/dotnet/BenchmarkDotNet/blob/2ba30330ec5ec8a9abe284c927277a6fe0fc5a0c/src/BenchmarkDotNet/Running/TypeFilter.cs#L32-L35

chan18: also, ``` TypeToBenchmarks(Type type, IConfig config = null) ``` , not ``` IEnumerable<Type> types, IEnumerable<Assembly> assemblies```

chan18:
hi, @adamsitnik  please review it. Let me know any other changes or improvements  you want to make!

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -125,5 +140,59 @@ private static Summary[] RunWithExceptionHandling(Func<Summary[]> run)
...
+                BenchmarkCasesPrinter.PrintList(logger, effectiveConfig, allAvailableTypesWithRunnableBenchmarks, options);
+                return null;
+            }
+
+            var BenchmarkRunInfos = new List<BenchmarkRunInfo[]>();
+
+            foreach (Type type in types)
+            {
+                var BenchmarkRunInfo = options.UserProvidedFilters
+                               ? TypeFilter.Filter(effectiveConfig, allAvailableTypesWithRunnableBenchmarks)
```
adamsitnik: repeating this operation in a loop would end up in duplicates

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -73,13 +84,17 @@ public static Summary RunSource(string source, IConfig config = null)
...
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private static Summary RunWithDirtyAssemblyResolveHelper(Type type, IConfig config)
-            => BenchmarkRunnerClean.Run(new[] { BenchmarkConverter.TypeToBenchmarks(type, config) }).Single();
+        private static Summary RunWithDirtyAssemblyResolveHelper(Type type, IConfig config, string[] args)
+            => RunWithDirtyAssemblyResolveHelper(new[] { type }, Array.Empty<Assembly>(), config, args);
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static Summary RunWithDirtyAssemblyResolveHelper(Assembly assembly, IConfig config, string[] args)
+             => RunWithDirtyAssemblyResolveHelper(assembly.GetRunnableBenchmarks().ToArray(), new[] { assembly }, config, args);
```
adamsitnik: `assembly.GetRunnableBenchmarks().ToArray()` is not needed, `TypeFilter.GetTypesWithRunnableBenchmarks` will take care of it

```suggestion
             => RunWithDirtyAssemblyResolveHelper(Array.Empty<Type>(), new[] { assembly }, config, args);
```

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -125,5 +140,59 @@ private static Summary[] RunWithExceptionHandling(Func<Summary[]> run)
...
+            }
+
+            Summary benchmark = BenchmarkRunnerClean.Run(BenchmarkRunInfos.First()).Single();
+
+            foreach (var BenchmarkRunInfo in BenchmarkRunInfos.Skip(1))
+            {
+                benchmark = BenchmarkRunnerClean.Run(BenchmarkRunInfo).Single();
+            }
+
+            return benchmark;
```
adamsitnik: this could be implemented as following:

```cs
            var filteredBenchmarks = TypeFilter.Filter(effectiveConfig, allAvailableTypesWithRunnableBenchmarks);

            if (filteredBenchmarks.IsEmpty())
            {
                userInteraction.PrintWrongFilterInfo(allAvailableTypesWithRunnableBenchmarks, logger, options.Filters.ToArray());
                return Array.Empty<Summary>();
            }

            return BenchmarkRunnerClean.Run(filteredBenchmarks);
```

this will also most probably require us to return an `IEnumerable<Summary>`, not a single `Summary`. The methods that provide single type and call this method can use `.Single()` on the result of this method

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -125,5 +140,59 @@ private static Summary[] RunWithExceptionHandling(Func<Summary[]> run)
                 return new[] { Summary.NothingToRun(e.Message, string.Empty, string.Empty) };
             }
         }
+
+        private static Summary RunWithDirtyAssemblyResolveHelper(IEnumerable<Type> types, IEnumerable<Assembly> assemblies, IConfig config, string[] args)
+        {
+            var logger = config.GetNonNullCompositeLogger();
+            var userInteraction = new UserInteraction();
```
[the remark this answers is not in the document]
chan18: makes sense, resovled it.

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
chan18: but Types will hold different types. I dont understand how this will create duplicate of benchmarkruninfo.

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
chan18: that is in case of  ```UserProvidedFilters```, what if user don't provide any?

we also have another case where we create benchmarkruninfro from types.
``` 
 foreach (Type type in types)
            {
                var BenchmarkRunInfo = options.UserProvidedFilters
                               ? TypeFilter.Filter(effectiveConfig, allAvailableTypesWithRunnableBenchmarks)
                               : new BenchmarkRunInfo[] { BenchmarkConverter.TypeToBenchmarks(type, config) };

                BenchmarkRunInfos.Add(BenchmarkRunInfo);
            }
```

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
adamsitnik: in case `options.UserProvidedFilters` is true, `TypeFilter.Filter(effectiveConfig, allAvailableTypesWithRunnableBenchmarks)` is executed for every `Type` in `types. Which means that if user has provided filter, we filter `allAvailableTypesWithRunnableBenchmarks` types-many times and add the result to a list than many times

adamsitnik:
The build fails with following compilation error:

```log
Running/BenchmarkRunnerDirty.cs(88,16): error CS0029: Cannot implicitly convert type 'BenchmarkDotNet.Reports.Summary[]' to 'BenchmarkDotNet.Reports.Summary'
```

Could you please fix it?

chan18:
sure.

adamsitnik (approved):

PR #1294, opened 2019-10-25
Event counters

WojciechNagorski:
This PR requires #1264 

In this PR I used event counters that enable users to monitor how many benchmarks were run. 

EventCounters

I wanted to learn how to use it. If you think that this feature is useless, you can just close this PR.


adamsitnik:
Hi @WojciechNagorski 

Thank you for your PR. I think that what we have as of today is enough: we print the number of remaining benchmarks:

```log
// ** Remained 0 benchmark(s) to run **
```

If you really want to know how many were executed I think that it would be better to just extend the message that we print instead of adding a custom counter (it's just simpler solution and works on every .NET while counter requires .NET Core 3.0+)

PR #1298, opened 2019-10-29
Create basic.cs

VirajGalappaththi:


AndreyAkinshin:
@VirajGalappaththi we don't need such a sample in the root folder.

PR #1300, opened 2019-10-29
Update link of " official benchmarking guide" to use the new recommended link

eriawan:
Update link of " official benchmarking guide" to use the new recommended link. The link should now point to doc on dotnet/performance repo.

cc @adamsitnik


adamsitnik:
@eriawan thank you!

Issue #487, opened 2017-06-28
Please consider using 'µs' instead of 'us'


jnm2:
It confused [at least one person](https://gitter.im/dotnet/BenchmarkDotNet?at=595363cb76a757f808b391f3) and has kinda bugged me since I can't think of any reason we couldn't use µ. Everything should be UTF8 these days.

Willing to PR.

adamsitnik:
Hello @jnm2 !

It's an old problem (#459)

[According](https://github.com/dotnet/BenchmarkDotNet/issues/459#issuecomment-306380025) to @AndreyAkinshin :
> us means Microsecond. In fact, it should be μs, but us is a common replacement of μs (some terminals doesn't support the μ symbol).

The [current solution](https://github.com/dotnet/BenchmarkDotNet/issues/459#issuecomment-306416280) is that we put information about it in the legend:
>   1 us  : 1 Microsecond (0.000001 sec)

@AndreyAkinshin could you tell something more about the old terminals that don't support it?

jnm2:
I didn't think of non-Unicode termimaps. Could it be detected? Console.SetEncoding?

adamsitnik:
@jnm2 I don't know anything more about the details. The person who does is @AndreyAkinshin . He  is on holidays now, he should be back within next few days.

AndreyAkinshin:
@jnm2, it's a hard problem, I tried to solve it before, but I got some troubles on Linux.

Basically, we have two challenges here.
1. We have to check if current console encoding (see [Console.OutputEncoding](https://docs.microsoft.com/en-us/dotnet/api/system.console.outputencoding?view=netstandard-1.3#System_Console_OutputEncoding), supported in NetStandard1.3) supports unicode symbols. It's not always true (especially on Unix systems). I had some troubles with Unix+mono here, but I don't remember details.
2. We have to check if current console font supports unicode symbols. On Windows, we can get the font via [GetCurrentConsoleFont](https://msdn.microsoft.com/en-us/library/ms683176(v=vs.85).aspx), an example:
```cs
internal class Program
{
    private const int STD_OUTPUT_HANDLE = -11;
    
    [DllImport("kernel32.dll", SetLastError = true)]
    static extern IntPtr GetStdHandle(int nStdHandle);
    
    [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
[...]

PR #1302, opened 2019-10-30
CultureInfo Refactoring

### LINKED ISSUE #1295: Custom format/culture for report output values (for CSV, and maybe HTML, MD)
At the moment in CSV, HTML, and MD version of reports I get the time in the formated like `6,114.79 ns`.
But each time I have to convert it into `6114,79` manually for my local culture (ru-RU) to be able to use these values for building the diagrams in both Google Docs and Microsoft Excel.
Without the conversion, these programs refuse to understand the values.
Related to #1100.

So is there a way to reformat these values from code? And if none can you add one?

AndreyAkinshin:
The goal of these refactoring is to unify some of our API and respect the given value of CultureInfo.
The implementation is still not perfect, but it's an important first step.
Highlights:

* Unify ToString APIs
  All of the ToStr(), ToTimeStr(), ToSizeStr() methods were removed.
  Initially, they were introduced as hacked shortcuts to formatting purposes, but today they become misleading in some cases.
  Instead of them, now we always should use ToString().
  Where it makes sense, the proper overloads like ToString(CultureInfo, format) are available.
* Introduce CultureInfo property in Config
  Now it's possible to set the CultureInfo, the given value will be used everywhere including exporters (Fix #1295)
* Remove MultiEncodingString and Encodings in Configs
  The original goal of these classes was providing a way to enable Unicode support in exporter (see #487 for details).
  Unfortunately, it made many APIs overcomplicated because we had to pass Encoding everywhere.
  If we think carefully, we will understand that the only problem that we actually have relates to the terminal that doesn't support Unicode
    (in other places, we can use Unicode symbols without any problems).
  I decided to delete the MultiEncodingString class, use Unicode symbols by default, and patch them only in ConsoleLogger
    (see AsciiHelper for details).
  Now, users can turn on Unicode support in console output via [UnicodeConsoleLogger]
    (see IntroUnicode for details).
  All other exporters are always able to use Unicode.
  Some of them may patch Unicode symbols to achieve better portability (e.g., see HtmlExporter.HtmlLoggerWrapper.Escape)
* Introduce ILogger.Id and ILogger.Priority
  These APIs allow overriding existing loggers by custom.
  For each Id, only the logger with the highest priority will be chosen.
  With the help of this feature, we can override ConsoleLogger.Ascii with ConsoleLogger.Unicode
* Introduce SizeValue
  This structure is a wrapper for a long value that helps to operate with size values (bytes, kilobytes, etc.)
* Rename TimeInterval->TimeValue
  The original name was confusing because an interval typically has start and end.
  In our situation, it similar to TimeSpan but has its own features and use cases.
  I decided to rename it to TimeValue (to make consistent with SizeValue).
* Introduce CultureInfo.GetActualListSeparator()
  TextInfo.ListSeparator shouldn't be used anymore because it returns incorrect value on .NET Core+Unix
    (see dotnet/runtime#536 for details)


AndreyAkinshin:
@adamsitnik the PR is ready for review.

AndreyAkinshin:
@adamsitnik ping

src/BenchmarkDotNet/Reports/Measurement.cs
```diff
@@ -61,19 +62,44 @@ public Measurement(int launchIndex, IterationMode iterationMode, IterationStage
             IterationIndex = iterationIndex;
         }
 
-        public string ToOutputLine()
```
adamsitnik: nit: imho the old name was more clear about the intentions

src/BenchmarkDotNet/Exporters/PlainExporter.cs
```diff
@@ -17,17 +17,20 @@ public override void ExportToLog(Summary summary, ILogger logger)
...
-                    logger.WriteLineResult(run.ToStr(report.BenchmarkCase.Config.Encoding));
+                foreach (var measurement in measurements)
+                    logger.WriteLineResult(measurement.ToString());
                 foreach (var (mode, stage) in modeStages)
                 {
                     logger.WriteLine();
                     logger.WriteLineHeader($"* Statistics for {mode}{stage}");
-                    logger.WriteLineStatistic(runs.Where(it => it.Is(mode, stage)).GetStatistics().ToTimeStr(report.BenchmarkCase.Config.Encoding, calcHistogram: true));
+                    var statistics = measurements.Where(it => it.Is(mode, stage)).GetStatistics();
+                    var cultureInfo = summary.GetCultureInfo();
```
adamsitnik: nit: `var cultureInfo = summary.GetCultureInfo();` could be moved out of loop

src/BenchmarkDotNet/Attributes/UnicodeConsoleLoggerAttribute.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Loggers;
+
+namespace BenchmarkDotNet.Attributes
+{
+    /// <summary>
+    /// Enable unicode support in console logger
+    /// </summary>
```
adamsitnik: great to have summary comments for the new attributes! 👍

src/BenchmarkDotNet/Mathematics/StatisticalTesting/TostResult.cs
```diff
@@ -17,10 +17,10 @@ public TostResult(Threshold threshold, EquivalenceTestConclusion conclusion, [Ca
             FasterTestResult = fasterTestResult;
         }
 
-        public string ToStr(bool details) => details
-            ? ConclusionStr() + ": " + (SlowerTestResult?.PValueStr ?? "?") + "|" + (FasterTestResult?.PValueStr ?? "?")
-            : ConclusionStr();
+        public string ToPresentation(bool details) => details
```
adamsitnik: imho it would be better to call it `ToString`.

My main motivation is that so far "ToPresentation" name was always used by the code that generates boilerplate code. 

https://github.com/dotnet/BenchmarkDotNet/blob/2ba30330ec5ec8a9abe284c927277a6fe0fc5a0c/src/BenchmarkDotNet/Characteristics/CharacteristicPresenter.cs#L29

src/BenchmarkDotNet/Helpers/AsciiHelper.cs
```diff
@@ -0,0 +1,12 @@
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Helpers
+{
+    internal static class AsciiHelper
+    {
+        public static string ToAscii([CanBeNull] this string s)
+        {
+            return s?.Replace("\u03BC", "u");
```
adamsitnik: it would be nice to have some comment here explaining what is `\u03BC` and why we need to replace it

are there any other characters that we should take care of?

src/BenchmarkDotNet/Loggers/AccumulationLogger.cs
```diff
@@ -1,11 +1,21 @@
...
 namespace BenchmarkDotNet.Loggers
 {
     public class AccumulationLogger : ILogger
     {
         private readonly StringBuilder builder = new StringBuilder();
 
+        public AccumulationLogger()
+        {
+            // All AccumulationLoggers should have unique Ids
+            Id = nameof(AccumulationLogger) + "." + Guid.NewGuid().ToString("N");
```
adamsitnik: thanks for adding the comment! 👍

src/BenchmarkDotNet/Loggers/ConsoleLogger.cs
```diff
@@ -26,6 +39,9 @@ public sealed class ConsoleLogger : ILogger
 
         private void Write(LogKind logKind, Action<string> write, string text)
         {
+            if (!unicodeSupport)
+                text = text.ToAscii();
```
adamsitnik: I wonder what kind of performance impact this change is going to have. We are by default calling `string.Replace("μ", "u")` for every string that is logged to the console

src/BenchmarkDotNet/Loggers/ILogger.cs
```diff
@@ -2,6 +2,14 @@
...
     public interface ILogger
     {
+        string Id { get; }
+        
+        /// <summary>
+        /// If there are several loggers with the same <see cref="Id"/>,
+        /// only logger with the highest priority will be used.
+        /// </summary>
+        int Priority { get; }
+
```
adamsitnik: imho we should rather give a validation error in such cases rather than extending every logger with two fields

tests/BenchmarkDotNet.IntegrationTests/TestCultureInfo.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace BenchmarkDotNet.IntegrationTests
+{
+    internal static class TestCultureInfo
+    {
+        public static readonly CultureInfo Instance;
+
+        static TestCultureInfo()
+        {
+            Instance = (CultureInfo) DefaultCultureInfo.Instance.Clone();
```
adamsitnik: what is the advantage of introducing this type if it's always using the same value as `DefaultCultureInfo`?

tests/BenchmarkDotNet.Tests/TestCultureInfo.cs
```diff
@@ -0,0 +1,15 @@
+using System.Globalization;
+using BenchmarkDotNet.Helpers;
+
+namespace BenchmarkDotNet.Tests
+{
+    internal static class TestCultureInfo
```
adamsitnik: same as above, do we really need this extra type?

src/BenchmarkDotNet/Attributes/EncodingAttribute.cs
```diff
@@ -1,26 +1,43 @@
...
-        private EncodingAttribute(Encoding encoding) => Config = ManualConfig.CreateEmpty().With(encoding);
+        private EncodingAttribute(Encoding encoding)
+        {
+            if (Equals(encoding, Encoding.Unicode))
+            {
+                Config = ManualConfig.CreateEmpty().With(ConsoleLogger.Unicode);
+                return;
+            }
+
+            if (Equals(encoding, Encoding.Unicode))
```
adamsitnik:
```suggestion
            if (Equals(encoding, Encoding.ASCII))
```

src/BenchmarkDotNet/Horology/TimeValue.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System.Globalization;
+using BenchmarkDotNet.Helpers;
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Horology
+{
+    public struct TimeValue
```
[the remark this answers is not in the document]
AndreyAkinshin: Ok, rollbacked to `TimeInterval`.

src/BenchmarkDotNet/Analysers/OutliersAnalyser.cs
```diff
@@ -77,17 +80,19 @@ string Format(int n, string verb)
         }
 
         [CanBeNull]
-        private static string GetRangeMessage([NotNull] double[] values)
+        private static string GetRangeMessage([NotNull] double[] values, CultureInfo cultureInfo)
         {
+            string Format(double value) => TimeValue.FromNanoseconds(value).ToString(cultureInfo, "N2");
```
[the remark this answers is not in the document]
AndreyAkinshin: @adamsitnik I don't think that we should create an additional method in `TimeValue` (aka `TimeInterval`) because we already have `ToString`. Here I introduce a local function just to remove duplications of `(cultureInfo, "N2")`.

src/BenchmarkDotNet/Reports/Measurement.cs
AndreyAkinshin: Existing of an additional `ToOutputLine` method led to some bugs because we had two different string representations of `Measurement`. Currently, we have a single unified CultureInfo-independent representation in all the places.

src/BenchmarkDotNet/Exporters/PlainExporter.cs
AndreyAkinshin: Thanks, fixed.

src/BenchmarkDotNet/Mathematics/StatisticalTesting/TostResult.cs
AndreyAkinshin: Good catch! Fixed!

src/BenchmarkDotNet/Helpers/AsciiHelper.cs
AndreyAkinshin: > it would be nice to have some comment here explaining what is \u03BC and why we need to replace it

Good point, fixed.

> are there any other characters that we should take care of?

Currently, `\u03BC` is the only symbol that should be patched.

src/BenchmarkDotNet/Loggers/ConsoleLogger.cs
AndreyAkinshin: I don't expect a noticeable performance impact because:
1. Typically, we perform a lot of heavy string concatenation in logging.
2. `Console.Write` itself takes a lot of time (much more than string.Replace of a single char).
3. We don't have too much logs and I don't think that we should care about such things.

src/BenchmarkDotNet/Loggers/ILogger.cs
AndreyAkinshin: I'm not sure that I got the idea. Could you explain it in detail?

tests/BenchmarkDotNet.IntegrationTests/TestCultureInfo.cs
AndreyAkinshin: In the future, we can easily mock it for testing special use cases without changing the original `DefaultCultureInfo`.

tests/BenchmarkDotNet.Tests/TestCultureInfo.cs
AndreyAkinshin: Yes, it will simplify mocking in the future.

src/BenchmarkDotNet/Attributes/EncodingAttribute.cs
AndreyAkinshin: Fixed.

AndreyAkinshin:
@adamsitnik thanks for the review!
I answered/fixed all of your comments except the one with additional fields in `ILogger`. I don't how to properly implement logging overriding without the `Id`/`Priority` properties. The only improvement that I have in mind is to move them to a separate `interface IUniqueLogger : ILogger`: in this case, we can skip these fields in all loggers except the `ConsoleLogger`. What do you think?

src/BenchmarkDotNet/Helpers/DefaultCultureInfo.cs
```diff
@@ -0,0 +1,15 @@
+using System.Globalization;
+
+namespace BenchmarkDotNet.Helpers
+{
+    internal static class DefaultCultureInfo
```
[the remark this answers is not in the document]
adamsitnik: I see, removing the tight coupling should help a lot with moving the helper types to the new standalone package. 👍

src/BenchmarkDotNet/Loggers/ILogger.cs
adamsitnik: > "If there are several loggers with the same only logger with the highest priority will be used."

What I meant is that we should rather give a validation error when user configures two loggers of the same type.

src/BenchmarkDotNet/Helpers/SourceCodeHelper.cs
```diff
@@ -42,7 +42,7 @@ public static string ToSourceCode(object value)
 
             switch (value) {
                 case TimeInterval interval:
-                    return "new BenchmarkDotNet.Horology.TimeInterval(" + ToSourceCode(interval.Nanoseconds) + ")";
+                    return "new BenchmarkDotNet.Horology.TimeValue(" + ToSourceCode(interval.Nanoseconds) + ")";
```
adamsitnik: this should make the CI green

```suggestion
                    return "new BenchmarkDotNet.Horology.TimeInterval(" + ToSourceCode(interval.Nanoseconds) + ")";
```

PR #1307, opened 2019-11-08
Fix project file order

vilinski:
## Reproduce

```
mkdir benchmark
cd benchmark
dotnet new benchmark -lang f# -c true --console-app
dotnet run
```

>...\benchmark\Benchmarks.fs(7,28): error FS0039: The type 'BenchmarkConfig' is not defined in 'BenchmarkDotNet.Configs'. Maybe you want one of the following:↔   BenchmarkLogicalGroupRule [C:\git\local\benches\benches.fsproj]
>...benchmark\Benchmarks.fs(7,10): error FS0267: This is not a valid constant expression or custom attribute value [C:\git\local\benches\benches.fsproj]
>
>The build failed. Fix the build errors and run again.

## Solution

Since `BenckmarConfig` is used within `(BenchmarkName).fs` it should precede it


AndreyAkinshin:
@vilinski thanks!

PR #1313, opened 2019-11-19
add possibility to specify env vars via console line arguments

### LINKED ISSUE #1312: Add an option to pass environment variables to the default job
When testing against the mono/netcore runtime we often want to run the actual benchmark with specific runtime configuration. This is achieved through setting the `MONO_ENV_OPTIONS` environment variable. However, we also want to use different `MONO_ENV_OPTIONS` when running `dotnet restore` or `dotnet build` for performance reasons. It would be nice to allow some way to inject the environment variables but only for the actul benchmark job.

What I currently use is the following hack:
https://github.com/filipnavara/performance/commit/d1768c7fdf52cc5905bdefce25964c08ba90f46d

/cc @EgorBo

adamsitnik:
Fixes #1312

/cc @egorbo @filipnavara

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -33,6 +33,7 @@ namespace BenchmarkDotNet.ConsoleArguments
     public static class ConfigParser
     {
         private const int MinimumDisplayWidth = 80;
+        private const char EnvVarKeyValueSeparator = ':';
 
         private static readonly IReadOnlyDictionary<string, Job> AvailableJobs = new Dictionary<string, Job>(StringComparer.InvariantCultureIgnoreCase)
         {
@@ -165,6 +166,12 @@ namespace BenchmarkDotNet.ConsoleArguments
                 return false;
             }
 
+            if (options.EnvironmentVariables.Any(envVar => envVar.IndexOf(EnvVarKeyValueSeparator) <= 0))
+            {
+                logger.WriteLineError($"Environment variable value must be separated from the key using '{EnvVarKeyValueSeparator}'. Use --help to see examples.");
+                return false;
+            }
+
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -258,6 +265,15 @@ private static Job GetBaseJob(CommandLineOptions options, IConfig globalConfig)
             if (options.RunOncePerIteration)
                 baseJob = baseJob.RunOncePerIteration();
 
+            if (options.EnvironmentVariables.Any())
+            {
+                baseJob = baseJob.With(options.EnvironmentVariables.Select(text =>
+                {
+                    var separated = text.Split(EnvVarKeyValueSeparator);
```
filipnavara: Perhaps `text.Split(EnvVarKeyValueSeparator, 2)` to allow `:` to be used in the value?

filipnavara:
Thank you very much!

adamsitnik:
@filipnavara thanks for the review! The `0.12.0.1234` version contains new feature. You can download it from BDN CI feed https://benchmarkdotnet.org/articles/guides/nuget.html#nightly

filipnavara:
Many thanks, @adamsitnik! Now I just need to convince someone to update the version dependency in dotnet/performance. I'm more than happy to submit a PR but I have been told previously that it may not be accepted until someone reviews all BDN changes picked in the update.

adamsitnik:
We have removed the dependency to BDN CI feed from dotnet/performance repo after BDN 0.12.0 was shipped to nuget.org. Adding it back again would be hard due to security reasons. We also don't plan to ship a new version of BDN to nuget.org in the next few weeks, so you should most probably keep using your fork|branch.

Is there anything else that needs to be changed in BDN to make the new mono from dotnet/runtime work? If supporting new Mono requires some changes we might consider shipping new BDN much sooner that we were planning.

filipnavara:
> Is there anything else that needs to be changed in BDN to make the new mono from dotnet/runtime work?

Short answer: Hopefully not.

From the perspective of BDN Mono looks just like .NET 5 runtime. The exact integration into the build system is not complete yet but hopefully it won't require any major changes.

The scripts I use are public (https://github.com/filipnavara/mono-performance/) and it was running for few months now. I am now trying to reduce any custom steps in the build process to avoid potential breakage of the automated process. There were two custom changes applied on dotnet/performance in the process. One of them was the hack for environment variables that is now resolved with this PR. The other one is a patch for more graceful fallback in case libgdiplus is not installed on the machine (to be removed; loading native libraries was broken for few weeks in Mono due to refactoring to properly support .NET Core 3 APIs).

The only remaining issue is that there are certain benchmarks that depend on HW intrinsics. Availability of these HW intrinsics varies between runtimes and platforms. In case they are not available on given platform the benchmarks bail out early with nearly zero running time. This sticks out when comparing the results because Mono looks infinitely faster than CoreCLR on those benchmarks.

PR #1320, opened 2019-11-25
remove obsolete APIs

adamsitnik:
We have quite many Obsolete APIs that our users should not use anymore, I think that we can remove these methods|types.


adamsitnik:
You are right.

I am going to send a separate PR with removing the very old Obsoletes. Thanks for the review!

Issue #1315, opened 2019-11-20
Implement cross platform EventPipeProfiler diagnoser


adamsitnik:
`EtwProfiler` is very popular, but works only on Windows. .NET Core 3.0 shipped with some new cool diagnostics features and tools. We can use the new diagnostics tooling to allow for profiling benchmarks on Linux and macOS.

We have two options:

* use [Microsoft.Diagnostics.NETCore.Client](https://github.com/dotnet/diagnostics/pull/617) library
* use [dotnet trace](https://github.com/dotnet/diagnostics/tree/master/src/Tools/dotnet-trace) global tool

Microsoft.Diagnostics.NETCore.Client has not been published to nuget.org yet, so the person who is willing to work on this could start with `dotnet trace`.

The new diagnoser should produce a speedscope file by default.

/cc @noahfalk @WojciechNagorski

WojciechNagorski:
I would like to do it. I thought about dotnet trace, as you know 😉

adamsitnik:
> I would like to do it.

Awesome! I've assigned you :)

WojciechNagorski:
I have been doing the `EventPipeProfiler` and I've found a problem with `dotnet trace`. There is an issue https://github.com/dotnet/diagnostics/issues/645

When I run `dotnet trace` in the same console without output redirection I get an output file. 
So currently this issue is blocked.

sywhang:
@WojciechNagorski `Microsoft.Diagnostics.NETCore.Client` is now available as a nuget package here: https://www.nuget.org/packages/Microsoft.Diagnostics.NETCore.Client/0.1.0-preview.19606.2

Thought I'd let you know since I know you were blocked by this :)

AndreyAkinshin:
Unfortunately, we can't reference prerelease NuGet packages in the release versions of BenchmarkDotNet. @sywhang any estimates for the release version of `Microsoft.Diagnostics.NETCore.Client`?

WojciechNagorski:
@sywhang Thanks! I'm going to use it to develop `EventPipeProfiler` but, as @AndreyAkinshin wrote, we should not use prerelease NuGet because of https://github.com/NuGet/Home/issues/2944.

WojciechNagorski:
[...]

PR #1321, opened 2019-11-27
The EventPipeProfiler cross-platform profiler

WojciechNagorski:
Changes for #1315

Currently, I use a local build of `Microsoft.Diagnostics.NETCore.Client` from https://github.com/dotnet/diagnostics/pull/617. This library support only .Net Core 2.1 so I decided that I create a new library `BenchmarkDotNet.Diagnostics.NETCore` that contains EventPipeProfiler. 

(I don't like the name "NETCore" but I wanted to do it like in `Microsoft.Diagnostics.NETCore.Client` )

You can run this profiler via command line:
```
dotnet BenchmarkDotNet.Samples.dll --filter *Algo_Md5VsSha256.Md5* -p EventPipe -j Short
```
image

[Example trace file.zip](https://github.com/dotnet/BenchmarkDotNet/files/3896891/Example.trace.file.zip)

image


adamsitnik:
I've tested the proposed `EventPipeProfiler` and it works fine. Great work @WojciechNagorski !

 I've sent https://github.com/dotnet/diagnostics/pull/700 in order to add .NET Standard  2.0 support to `Microsoft.Diagnostics.NETCore.Client` so we could just add this diagnoser to existing package without adding .NET Core target.

WojciechNagorski:
@adamsitnik Thanks. I'm going to finish this task this week.

adamsitnik:
https://github.com/dotnet/diagnostics/pull/700 got merged and it targets .NET Standard 2.0 now, so IMHO we can add this dependency to main BenchmarkDotNet project and just add the new profiler plugin there

WojciechNagorski:
@adamsitnik Great. Today I will move `EventPipeProfiler` to the BenchmarkDotNet project.

WojciechNagorski:
I've pushed the updated version of `EventPipeProfiler`. It is still `WIP`, I need to test all cases.

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
```diff
@@ -0,0 +1,241 @@
...
+            var etlxFilePath = TraceLog.CreateFromEventPipeDataFile(fileToConvert, null, new TraceLogOptions() { ContinueOnError = continueOnError });
+            using (var symbolReader = new SymbolReader(System.IO.TextWriter.Null) { SymbolPath = SymbolPath.MicrosoftSymbolServerPath })
+            using (var eventLog = new TraceLog(etlxFilePath))
+            {
+                var stackSource = new MutableTraceEventStackSource(eventLog)
+                {
+                    OnlyManagedCodeStacks = true // EventPipe currently only has managed code stacks.
+                };
+
+                var computer = new SampleProfilerThreadTimeComputer(eventLog, symbolReader);
```
adamsitnik: there was a bug in SpeedScopeExporter which got solved recently https://github.com/dotnet/diagnostics/pull/661/

```suggestion
                var computer = new SampleProfilerThreadTimeComputer(eventLog, symbolReader)
                {
                    IncludeEventSourceEvents = false // SpeedScope handles only CPU samples, events are not supported
                };
```

src/BenchmarkDotNet/Diagnosers/IProfiler.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace BenchmarkDotNet.Diagnosers
 {
-    internal interface IProfiler : IDiagnoser
+    public interface IProfiler : IDiagnoser
```
adamsitnik: is this change mandatory after you moved `EventPipeProfiler` to main project?

src/BenchmarkDotNet/Engines/EngineEventSource.cs
```diff
@@ -38,7 +39,9 @@ public class Tasks
 
         internal static readonly EngineEventSource Log = new EngineEventSource();
 
-        private EngineEventSource() { }
+        private EngineEventSource() : base(SourceName, EventSourceSettings.EtwSelfDescribingEventFormat)
```
adamsitnik: why do we need this change?

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
```diff
@@ -12,11 +12,11 @@
 
 namespace BenchmarkDotNet.Toolchains
 {
-    internal static class ToolchainExtensions
+    public static class ToolchainExtensions
```
adamsitnik: is this change mandatory after you moved `EventPipeProfiler` to main project?

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
```diff
@@ -12,11 +12,11 @@
...
     {
-        internal static IToolchain GetToolchain(this BenchmarkCase benchmarkCase) => GetToolchain(benchmarkCase.Job, benchmarkCase.Descriptor);
+        public static IToolchain GetToolchain(this BenchmarkCase benchmarkCase) => GetToolchain(benchmarkCase.Job, benchmarkCase.Descriptor);
```
adamsitnik: same

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
```diff
@@ -12,11 +12,11 @@
...
 
-        internal static IToolchain GetToolchain(this Job job) => GetToolchain(job, null);
+        public static IToolchain GetToolchain(this Job job) => GetToolchain(job, null);
```
adamsitnik: same

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
```diff
@@ -0,0 +1,241 @@
...
+    {
+        // parameterless constructor required by DiagnosersLoader to support creating this profiler via console line args
+        public EventPipeProfiler() { }
+
+        /// <summary>
+        /// Creates new instance of EventPipeProfiler
+        /// </summary>
+        /// <param name="profile">A named pre-defined set of provider configurations that allows common tracing scenarios to be specified succinctly.</param>
+        /// <param name="providers">A list of EventPipe providers to be enabled.</param>
+        public EventPipeProfiler(EventPipeProfile? profile = null, IReadOnlyCollection<EventPipeProvider> providers = null)
```
adamsitnik: this ctor is very complex. It would be better to have two ctors:

```cs
public EventPipeProfiler(IReadOnlyCollection<EventPipeProvider> providers)

public EventPipeProfiler(EventPipeProfile profile) : this(Map(profile))

private static IReadOnlyCollection<EventPipeProvider> Map(EventPipeProfile profile) ...

File: src/BenchmarkDotNet/Diagnosers/ListProfilesCommandHandler.cs [hunk 8]
```diff
@@ -0,0 +1,44 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tracing.Parsers;
+
+namespace BenchmarkDotNet.Diagnosers 
+{
+    internal sealed class ListProfilesCommandHandler
+    {
+        internal static IEnumerable<(EventPipeProfile profile, EventPipeProvider[] Providers)> DotNETRuntimeProfiles { get; } = new[]
```
adamsitnik: should this be a dictionary? and the type name should reflect this (mapper?)?

BenchmarkDotNet.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2027
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29102.190
```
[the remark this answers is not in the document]
WojciechNagorski: I will revert the version of Visual Studio. But I think that guilds should be changed, https://github.com/dotnet/project-system/issues/3079#issuecomment-354180353
and 
https://github.com/dotnet/project-system/issues/2863
What do you think?

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
WojciechNagorski: Thanks!

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
WojciechNagorski: Your suggestion look good but it doesn't have the same function as my constructor.

You can call my constructor with parameters:
```c#
EventPipeProfiler(
    EventPipeProfile.CpuSampling, 
    new[] {
        new EventPipeProvider("System.Buffers.ArrayPoolEventSource", EventLevel.Informational, Int64.MaxValue),
    });
```
It give you all providers from `EventPipeProfile.CpuSampling` and add the `System.Buffers.ArrayPoolEventSource` provider. 
In this way, you can also override the provider from the profile. e.g.:
the `EventPipeProfile.CpuSampling` profile adds:
```c#
new[]
{
    new EventPipeProvider("Microsoft-DotNETCore-SampleProfiler", EventLevel.Informational),
    new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Informational, (long) ClrTraceEventParser.Keywords.Default)
}
```
and you can override for example EventLevel or change keywords, using the following code
```
EventPipeProfiler(
    EventPipeProfile.CpuSampling, 
    new[] {
        new EventPipeProvider("Microsoft-DotNETCore-SampleProfiler", EventLevel.Verbous),
    });
```
Currently it works in the same way like `dotnet trace`. You can fine more information here https://docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-trace#options-1

But if you want I will change it. Please let me know.

WojciechNagorski:
@adamsitnik I made the changes you asked for. I have questions for two topics.

WojciechNagorski:
@adamsitnik ping

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
adamsitnik: thanks for the explanation, now it makes sense to me!

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
```diff
@@ -0,0 +1,235 @@
...
+            catch (DiagnosticsClientException e)
+            {
+                logger.WriteLine(LogKind.Error, $"Unable to start a tracing session: {e}");
+                return;
+            }
+            var fileName = ArtifactFileNameHelper.GetFilePath(parameters, DateTime.Now, "nettrace");
+            Path.GetDirectoryName(fileName).CreateIfNotExists();
+            benchmarkToTraceFile[parameters.BenchmarkCase] = fileName;
+
+            collectingTask = new Task(() =>
```
adamsitnik: nit: we have quite a deep nesting here. in a perfect world, we would move this logic to a separate method and avoid closure capture

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
```diff
@@ -0,0 +1,235 @@
...
+                if (runtime.RuntimeMoniker < RuntimeMoniker.NetCoreApp30)
+                {
+                    yield return new ValidationError(true, $"{nameof(EventPipeProfiler)} supports only .NET Core 3.0+", benchmark);
+                }
+            }
+        }
+
+        public string ShortName => "EventPipe";
+
+        private void ConvertToSpeedscope(BenchmarkCase benchmarkCase, string traceFilePath)
```
adamsitnik: this class is responsible for profiling using event pipe (`EventPipeProfiler`) could you please move everything that is related to exporting to speedscope to a dedicated type?

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
```diff
@@ -0,0 +1,235 @@
...
+                    benchmarkToTraceFile[benchmarkCase] = speedscopeFileName;
+                }
+                else
+                {
+                    logger.WriteLine(LogKind.Error, $"An exception occurred during converting {traceFilePath} file to speedscope format: {ex}");
+                }
+            }
+        }
+
+        // Method copied from https://github.com/dotnet/diagnostics/blob/2c23d3265dd8f642a8d6cf4bb8a135a5ff8b00c2/src/Tools/dotnet-trace/TraceFileFormatConverter.cs#L64
```
adamsitnik: thanks for adding the comment!

src/BenchmarkDotNet/Engines/EngineEventSource.cs
```diff
@@ -1,12 +1,13 @@
...
 using JetBrains.Annotations;
-
+using System.Diagnostics.Tracing;
 namespace BenchmarkDotNet.Engines
 {
-    [EventSource(Name = "BenchmarkDotNet.EngineEventSource")]
+    [EventSource(Name = EngineEventSource.SourceName)]
     public class EngineEventSource : EventSource
     {
+        const string SourceName = "BenchmarkDotNet.EngineEventSource";
```
adamsitnik: it looks like there is no need to extract const if it's used only in once place. Could you please revert this change?

src/BenchmarkDotNet/Helpers/ArtifactFileNameHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains;
+
+namespace BenchmarkDotNet.Helpers 
+{
+    internal static class ArtifactFileNameHelper
+    {
+        public static string GetFilePath(DiagnoserActionParameters details, DateTime creationTime, string fileExtension)
```
adamsitnik: good idea to move it to a separate file!

```suggestion
        internal static string GetFilePath(DiagnoserActionParameters details, DateTime creationTime, string fileExtension)
```

WojciechNagorski:
@adamsitnik I will try to fix it now.

adamsitnik:
>  I will try to fix it now.

@WojciechNagorski don't worry about this, I am going to send a small PR with the improvements

WojciechNagorski:
@adamsitnik Ok, so I won't change anything. There you can find what I've done so far: #1382

PR #1322, opened 2019-12-04
add FirstCallColumn that prints the execution time of the first call

adamsitnik:
Sample:

```cs
class Program
{
    static void Main(string[] args) => BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args,
        DefaultConfig.Instance.With(new FirstCallColumn()));
}

public class RealColdStart
{
    private bool firstCall;

    [Benchmark]
    public void Foo()
    {
        if (firstCall == false)
        {
            firstCall = true;
            Console.WriteLine("// First call");
            Thread.Sleep(1000);
        }
        else
            Thread.Sleep(10);
    }
}
```

BenchmarkDotNet=v0.12.0.20191204-develop, OS=Windows 10.0.18363
  [Host]     : .NET Framework 4.8 (4.8.4018.0), X64 RyuJIT
  DefaultJob : .NET Framework 4.8 (4.8.4018.0), X64 RyuJIT

| Method |     Mean |    Error |   StdDev |   Median | First Call |
|------- |---------:|---------:|---------:|---------:|-----------:|
|    Foo | 14.31 ms | 0.561 ms | 1.498 ms | 14.88 ms |   1.0142 s |

cc @eiriktsarpalis


src/BenchmarkDotNet/Columns/FirstCallColumn.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    public class FirstCallColumn : IColumn
+    {
+        public string Id => nameof(FirstCallColumn);
+
+        public string ColumnName => "First Call";
+
+        public bool AlwaysShow => true;
+
+        public ColumnCategory Category => ColumnCategory.Custom;
```
AndreyAkinshin: I think it will be better to use the `Statistics` category with the "latest" priority. From my point of view, this column may be considered as one of the statistics columns, it shouldn't be placed after baseline columns.

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
```diff
@@ -0,0 +1,48 @@
...
+        public UnitType UnitType => UnitType.Time;
+
+        public string Legend => "Excution time of the first call (Jitting included)";
+
+        public string GetValue(Summary summary, BenchmarkCase benchmarkCase)
+        {
+            if (!summary.HasReport(benchmarkCase) || !summary[benchmarkCase].Success || !summary[benchmarkCase].AllMeasurements.Any(IsFirstCall))
+                return "-";
+
+            var measurement = summary[benchmarkCase].AllMeasurements.Single(IsFirstCall);
```
AndreyAkinshin: Theoretically, the Engine may perform several JittingWorkload iterations that will lead to an exception during summary table building. It will be better to select all `IsFirstCall` measurements, sort them by IterationStage, then by `LaunchIndex`, and take the first measurement. If you think that a situation with several "FirstCall" measurements is an error that should be reported, please use analyzers instead; don't throw exceptions in `IColumn.GetValue`.

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
```
AndreyAkinshin: The `System`, `System.Collections.Generic`, and `System.Text` namespaces are unused, they can be safily removed.

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
```diff
@@ -0,0 +1,48 @@
...
+            return measurement.Nanoseconds.ToTimeStr(summary.Style.TimeUnit, benchmarkCase.Config.Encoding);
+        }
+
+        public string GetValue(Summary summary, BenchmarkCase benchmarkCase, SummaryStyle style) => GetValue(summary, benchmarkCase);
+
+        public bool IsAvailable(Summary summary) => summary.Reports.Any(report => report.AllMeasurements.Any(IsFirstCall));
+
+        public bool IsDefault(Summary summary, BenchmarkCase benchmarkCase) => false;
+
+        private static bool IsFirstCall(Measurement m)
```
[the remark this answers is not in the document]
adamsitnik: Great catch, fixed!

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
adamsitnik: good point, I've added sorting

adamsitnik:
@AndreyAkinshin I've fixed all the issues and added docs. 

I did not add an analyzer because there is only one situation when we don't have the first call result: the benchmark has failed to run (obvious what `-` means).

docs/articles/samples/IntroFirstCallColumn.md
```diff
@@ -0,0 +1,58 @@
...
+## Sample: IntroFirstCallColumn
+
+If you want to have the execution time of the first call to be displayed in a dedicated column, please use `FirstCallColumn`.
+
+### Usage
+
+```cs
+// attribute
+[FirstCallColumn]
+public class MyBenchmarkClass
```
AndreyAkinshin: The benchmark body is missed.

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
```diff
@@ -0,0 +1,74 @@
...
+        public string GetValue(Summary summary, BenchmarkCase benchmarkCase, SummaryStyle style) => GetValue(summary, benchmarkCase);
+
+        public bool IsAvailable(Summary summary) => summary.Reports.Any(HasAnyRealWorkloads);
+
+        public bool IsDefault(Summary summary, BenchmarkCase benchmarkCase) => false;
+
+        private static bool HasAnyRealWorkloads(BenchmarkReport report)
+            => report.AllMeasurements.Any(m => m.IterationMode == IterationMode.Workload);
+
+        private static Measurement GetFirstCall(BenchmarkReport report)
```
AndreyAkinshin: If `report.AllMeasurements` doesn't contain Workload measurements (may be happened because of a mistake in another place), `GetValue` will throw an exception. It would be better to handle such a case, and print a placeholder like `NA` or `?`.

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public ColumnCategory Category => ColumnCategory.Statistics;
+
+        public int PriorityInCategory => int.MinValue;
+
+        public bool IsNumeric => true;
+
+        public UnitType UnitType => UnitType.Time;
+
+        public string Legend => "Execution time of the first call (Jitting included)";
```
AndreyAkinshin: I suggest removing the "(Jitting included)" note. It doesn't provide any value for people who are not aware of jitting and may confuse people who know about it (how is it possible to measure the duration of the first call without jitting?). Also, it will be better to write "first method call" or "first benchmark call" instead of just "first call".

docs/articles/samples/IntroFirstCallColumn.md
adamsitnik: @AndreyAkinshin the body will be included below by `[!code-csharp[IntroColdStart.cs](../../../samples/BenchmarkDotNet.Samples/IntroFirstCallColumn.cs)]`

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
adamsitnik: the method is always used in a context where `HasAnyRealWorkloads` was executed first so it won't happen

src/BenchmarkDotNet/Columns/FirstCallColumn.cs
adamsitnik:
```suggestion
        public string Legend => "Execution time of the first call";
```

adamsitnik: Good point, without profiler, it's impossible to measure the duration of the first call without jitting. I've removed the text.

PR #1327, opened 2019-12-10
Add package-refs to reference assemblies for linux build

### LINKED ISSUE #1326: BDN does not build using dotnet sdk from the command line in Linux
When checking out master copy of BDN in Linux and issuing `dotnet build` the build fails due to missing reference assemblies on Linux.

```bash
...
XUnit/TheoryWindowsOnlyAttribute.cs(9,43): error CS0518: Predefined type 'System.String' is not defined or imported [/home/dmg/projects/public/BenchmarkDotNet/
tests/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj]
XUnit/TheoryWindowsOnlyAttribute.cs(9,9): error CS0518: Predefined type 'System.Void' is not defined or imported [/home/dmg/projects/public/BenchmarkDotNet/tes
ts/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj]
    155 Warning(s)
    11220 Error(s)

Time Elapsed 00:00:13.92
```

Full build log:
https://gist.github.com/damageboy/09c715f9159f36d92813f31323d4b981

damageboy:
closes #1326

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -57,9 +57,10 @@
     <InformationalVersion>$(Major).$(Minor).$(Revision).$(BuildNumber)$(PrereleaseLabel)</InformationalVersion>
     <PackageVersion>$(Major).$(Minor).$(Revision).$(BuildNumber)$(PrereleaseLabel)</PackageVersion>
   </PropertyGroup>
-  <PropertyGroup>
-    <TargetFrameworkRootPath Condition="'$(TargetFrameworkRootPath)' == '' AND Exists('/Library/Frameworks/Mono.framework/Versions/Current')">/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks</TargetFrameworkRootPath>
-    <TargetFrameworkRootPath Condition="'$(TargetFrameworkRootPath)' == '' AND Exists('/usr/local/lib/mono/xbuild-frameworks')">/usr/local/lib/mono/xbuild-frameworks</TargetFrameworkRootPath>
-    <TargetFrameworkRootPath Condition="'$(TargetFrameworkRootPath)' == '' AND Exists('/usr/lib/mono/xbuild-frameworks')">/usr/lib/mono/xbuild-frameworks</TargetFrameworkRootPath>
-  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
 </Project>
--- a/samples/BenchmarkDotNet.Samples.FSharp/BenchmarkDotNet.Samples.FSharp.fsproj
+++ b/samples/BenchmarkDotNet.Samples.FSharp/BenchmarkDotNet.Samples.FSharp.fsproj
@@ -19,6 +19,10 @@
   </ItemGroup>
   
   <ItemGroup>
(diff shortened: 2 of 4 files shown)


adamsitnik:
@AndreyAkinshin  good point

We can also make it conditional to be included only for full framework projects

```xml
  <ItemGroup Condition=" $(TargetFramework.StartsWith('net4')) ">
    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
    </PackageReference>
  </ItemGroup>
```

damageboy:
@AndreyAkinshin Of course!
I was looking for a `Directory.Build.props` somewhere there, never occured to me you have something else going on there...

damageboy:
@adamsitnik  @AndreyAkinshin 
I included the assemblies through common.props.
The two disassembler project were not including common.props so I added an `<Import>` to them.
The F# samples project doesn't like including common.props so I patched those refs manually in that single project file.

build/common.props
```diff
@@ -57,9 +57,10 @@
...
-  </PropertyGroup>
-</Project>
\ No newline at end of file
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('net4')) ">
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+</Project>
```
adamsitnik: @mfilippov could you please take a look? I want to make sure that we don't break your fix from #898

damageboy: Just to be clear. The previous way was the old way I was using in my own projects as well before the reference assemblies nuget became a reality.

In other words, I didn't just remove the old stuff unintentionally.

adamsitnik: > I didn't just remove the old stuff unintentionally.

Sure! I just want to make sure that the "new way" works fine for what @mfilippov needed in #898

build/common.props
```diff
@@ -57,9 +57,10 @@
...
     <PackageVersion>$(Major).$(Minor).$(Revision).$(BuildNumber)$(PrereleaseLabel)</PackageVersion>
   </PropertyGroup>
-  <PropertyGroup>
-    <TargetFrameworkRootPath Condition="'$(TargetFrameworkRootPath)' == '' AND Exists('/Library/Frameworks/Mono.framework/Versions/Current')">/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks</TargetFrameworkRootPath>
-    <TargetFrameworkRootPath Condition="'$(TargetFrameworkRootPath)' == '' AND Exists('/usr/local/lib/mono/xbuild-frameworks')">/usr/local/lib/mono/xbuild-frameworks</TargetFrameworkRootPath>
-    <TargetFrameworkRootPath Condition="'$(TargetFrameworkRootPath)' == '' AND Exists('/usr/lib/mono/xbuild-frameworks')">/usr/lib/mono/xbuild-frameworks</TargetFrameworkRootPath>
-  </PropertyGroup>
-</Project>
\ No newline at end of file
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('net4')) ">
```
mfilippov: You don't need condition here. It handled into ref asm packages.

damageboy: Force pushed and fixed!

build/common.props
damageboy: @mfilippov I think @adamsitnik is waiting for you to confirm that this still works fine w.r.t to 898.

PR #1331, opened 2019-12-16
Use 24-hour time in joined summary file name

jroessel:
This prevents confusing ordering in the file system and should be more in line with the sane date format chosen for the file name as well.


adamsitnik:
The CI failure is not related, I am merging

Issue #1316, opened 2019-11-20
Implement Unix Disassembler for .NET Core


adamsitnik:
As far as I understand, we could use

- EventPipe (shipped with .NET Core 3.0) to emit .NET Runtime events on Unix
- TraceEvent to consume [MethodILToNativeMap](https://github.com/microsoft/perfview/blob/f35dccc88a2e6d467d26411bb09705375b823b9d/src/TraceEvent/Parsers/ClrTraceEventParser.cs#L1138) events
- Get the `IL` and native code map:
    - use Mono.Cecil to turn byte representation of IL to text
    - use [iced](https://github.com/0xd4d/iced) to turn byte representation of native code to text

It might turn out that some data is missing and|or I am simply wrong and it's impossible to implement it as of today. So it's a challange ;)

WojciechNagorski:
@adamsitnik Because #1315 is blocked I would like to take this one too. Is this issue free?

adamsitnik:
It's still up for grabs. However, I am not sure whether `Microsoft.Diagnostics.NETCore.Client` is already available as NuGet package, so you need to confirm it first. Thanks!

WojciechNagorski:
You're right but I'm going to start with my local build of `Microsoft.Diagnostics.NETCore.Client` from https://github.com/dotnet/diagnostics/pull/617
 
![image](https://user-images.githubusercontent.com/17333903/69706863-e2217a00-10f8-11ea-9928-0f6ad2269146.png)

In this comment https://github.com/dotnet/diagnostics/issues/645#issuecomment-558345643 I got information:
>productized version is almost finished

damageboy:
I actually think this isn't blocked at all.

I mean, I got [JitDasm](https://github.com/0xd4d/JitDasm) to work on my linux box (I make no claims about macOs).

I don't mind taking a stab at this either... I would approach it this way:

* Update to latest ClrMD
* Follow the guidelines for flushing the data target as set out in [ClrMD issue 303](https://github.com/microsoft/clrmd/issues/303)
* Remove the requirement for `IDebugControl` for the disassembly part:
  * I have no idea, nor do I want to know if this is supported on Linux right now
  * Given the current quality of Microsoft diagnostic tooling on Linux, I think a safer approach for now would be "the less the better", at least when it comes to Linux
  * If I understand correctly, the entire reason "we" have two projects for disassembly is because of `IDebugControl` so moving to managed disassembler will help remove this
[...]

PR #1332, opened 2019-12-17
Improved and Cross platform disassembler

### LINKED ISSUE #899: Tiered compilation and disassembler
BDN 0.11.1

The disassembler always shows the non-optimized method when tiered compilation is enabled (eg. use .NET Core 3.0 daily build). I assume it's because the method is only executed once and then disassembled.

A workaround is to set `COMPlus_TieredCompilation` to `0` before running the benchmarks.

adamsitnik:
Fixes #899 #1316 #1364
Implemented Linux support by consuming the latest ClrMD goodness and hints from @damageboy .

Switched to `Iced` library for formatting assembly code. It was mandatory because on Linux we can read only the byte representation of given assembly code, there is no method that allows for reading text form (like `DebugControl.Disassemble` from ClrMD on Windows). The disadvantage is that `Iced` does not support ARM (at least as of today). However, I am not really sure if ClrMD itself works on ARM. The advantage is that now every user can specify the formatter in `DisassemblyDiagnoserConfig`. Moreover, now when we search for jumps|calls in the assembly code we don't parse text anymore, but use strong type information (see all the usages of Instruction type).

Removed dependency to `Cecil`, mostly to simplify the code (remove the flaky ClrMd<->Cecil naming translation). It also makes the disassembler more reliable since we don't need to read assemby definition like we used to have.

I have also changed the config type a little bit:

- there is no reason for making printing assembly code for the disassembler optional

- I have also reduced the number of files created by the exporter (it was always 4: pretty MD & HTML, raw HTML and combined HTML). Now user can specify in the config what he|she wants to get.

And improved the disassembler itself. Now we handle runtime helper methods and references to method tables properly. Example:

Before:

```assembly
; MicroBenchmarks.WithCallsAfter.Benchmark(Int32)
       push    rsi
       sub     rsp,20h
       mov     rsi,rcx
       cmp     edx,7FFFFFFFh
       jne     M00_L00
       call    MicroBenchmarks.WithCallsAfter.Static()
       mov     rcx,rsi
       call    MicroBenchmarks.WithCallsAfter.Instance()
       mov     rcx,rsi
       call    MicroBenchmarks.WithCallsAfter.Recursive()
       mov     rcx,rsi
       mov     rax,qword ptr [rsi]
       mov     rax,qword ptr [rax+40h]
       call    qword ptr [rax+20h]
       mov     rcx,rsi
       mov     edx,1
       mov     rax,7FF8F4217050h
       add     rsp,20h
       pop     rsi
       jmp     rax
M00_L00:
       mov     rcx,offset System_Private_CoreLib+0xa31d48
       call    coreclr!MetaDataGetDispenser+0x322a0
       mov     rsi,rax
       mov     ecx,0ACFAh
       mov     rdx,7FF8F42F4680h
       call    coreclr!MetaDataGetDispenser+0x17140
       mov     rdx,rax
       mov     rcx,rsi
       call    System.InvalidOperationException..ctor(System.String)
       mov     rcx,rsi
       call    coreclr!coreclr_shutdown_2+0x39f0
       int     3
       add     byte ptr [rax],al
       sbb     dword ptr [00007ff9`26284e30],eax
       add     dword ptr [rax+40h],esp
       add     byte ptr [rax],al
       add     byte ptr [rax],al
       add     byte ptr [rax],al
       add     byte ptr [rax-70BC4CCh],ah
; Total bytes of code 157
```

After:

```assembly
; BenchmarkDotNet.Samples.WithCallsAfter.Benchmark(Int32)
       push      rsi
       sub       rsp,20
       mov       rsi,rcx
       cmp       edx,7FFFFFFF
       jne       M00_L00 
       call      BenchmarkDotNet.Samples.WithCallsAfter.Static() 
       mov       rcx,rsi
       call      BenchmarkDotNet.Samples.WithCallsAfter.Instance() 
       mov       rcx,rsi
       call      BenchmarkDotNet.Samples.WithCallsAfter.Recursive() 
       mov       rcx,rsi
       mov       rax,[rsi]
       mov       rax,[rax+40]
       call      qword ptr [rax+20]
       mov       rcx,rsi
       mov       edx,1
       mov       rax BenchmarkDotNet.Samples.WithCallsAfter.Benchmark(Boolean) 
       add       rsp,20
       pop       rsi
       jmp       rax
M00_L00:
       mov       rcx MT_System.InvalidOperationException 
       call      CORINFO_HELP_NEWSFAST 
       mov       rsi,rax
       mov       ecx,12D
       mov       rdx,7FF954FF83F0
       call      CORINFO_HELP_STRCNS 
       mov       rdx,rax
       mov       rcx,rsi
       call      System.InvalidOperationException..ctor(System.String) 
       mov       rcx,rsi
       call      CORINFO_HELP_THROW 
       int       3
; Total bytes of code 134
```


src/BenchmarkDotNet.Disassembler.x64/Program.cs
```diff
@@ -274,177 +237,146 @@ private static string GetSmartPrefix(string sourceLine, int length)
...
-            {
-                int hr = state.DebugControl.Disassemble(disasmAddress, 0,
-                    disasmBuffer, disasmBuffer.Capacity, out uint disasmSize,
-                    out ulong endOffset);
-                if (hr != 0)
-                    break;
+            int length = (int)(map.EndAddress - map.StartAddress);
+            byte[] buffer = new byte[length];
+
+            if (!state.Runtime.DataTarget.ReadProcessMemory(map.StartAddress, buffer, buffer.Length, out int bytesRead))
```
damageboy: I think this can/should be protected with some extra checking, on Linux, since many users will probably hit this head-on.

Perhaps the correct place is **not here** but rather when the Diagnoser is condigured at the top level when the run is about to begin...?

I'm specifically referring to `ptrace_scope` on linux. If I'm not mistaken, that kernel flag controls whether `ReadProcessMemory` is going to actually work (Again ,95% sure this is the one, I'm doing this from memory). I personally just set it to `0` in my `sysctl.conf`, which means I can debug my own process (same uid)...

Here is some rough documentation copy-pasted from [here](https://linux-audit.com/protect-ptrace-processes-kernel-yama-ptrace_scope/):

```bash
$ cat /proc/sys/kernel/yama/ptrace_scope
1
```

For this particular key there are four valid options: 0-3
* kernel.yama.ptrace_scope = 0: all processes can be debugged, as long as they have same uid. This is the classical way of how ptracing worked.
* kernel.yama.ptrace_scope = 1: only a parent process can be debugged.
* kernel.yama.ptrace_scope = 2: Only admin can use ptrace, as it required CAP_SYS_PTRACE capability.
* kernel.yama.ptrace_scope = 3: No processes may be traced with ptrace. Once set, a reboot is needed to enable ptracing again.

I *think* it makes sense to check that `uid == 0` or that ptrace_scope is `0`.

Otherwise, you will have to make sure that ptrace_scope is `1` AND that the code you WANT to disassemble lanches the disassembler (I have no idea if that is the current case, just a general suggestion...)...

Of course values: `2`,`3` can also be handled appropriately...

Does this sound good?

adamsitnik: @damageboy thanks for pointing this out! If you have not told me about this, I would most probably learn about this with the first user bug report.

I've added proper handling of `ptrace_scope` value. When it's 3 it's an error, when it's 2 it's a warning. With messages ofc.

BTW the LinuxDisassembler works differently than Windows one. We don't spawn a dedicated disassembler process because the parent and benchmark process have always the same architecture on Linux (on Windows we have x86 and x64, on Linux .NET Core supports only x64)

src/BenchmarkDotNet/Disassemblers/InstructionFormatter.cs
```diff
@@ -0,0 +1,139 @@
...
+            var output = new StringBuilderFormatterOutput(internalBuffer);
+
+            if (config.PrintInstructionAddresses)
+            {
+                FormatInstructionPointer(instruction, config.Formatter, pointerSize, output);
+            }
+
+            if (ClrMdDisassembler.TryGetReferencedAddress(instruction, pointerSize, out ulong referencedAddress) && addressesMappings.ContainsKey(referencedAddress))
+            {
+                // this instruction refers to an address that we know how to transalte
```
0xd4d: Seems like this is code to add symbols to the output? If so you should add your own `ISymbolResolver` and the formatter will call your code to get the symbol. Pass it in to the formatter's constructor. There's rarely a need to call any other method than `Format()`, like the code below in the 'else' block does.

0xd4d:
```
       mov       rax BenchmarkDotNet.Samples.WithCallsAfter.Benchmark(Boolean) 
```

It's missing the comma there. `Format()` will add that automatically.

BTW, since the formatter also includes extra information (eg. Mnemonic, Keyword, etc) you could use that to get some nice colorized output. You just have to replace the `StringBuilderFormatterOutput` with your own class that stores the text and color info.

src/BenchmarkDotNet/Disassemblers/InstructionFormatter.cs
adamsitnik: thanks for a great hint! I've switched to `ISymbolSolver` now https://github.com/dotnet/BenchmarkDotNet/pull/1332/commits/d14e354d477ae6cf00b57db668dbfce88493d545

adamsitnik:
@AndreyAkinshin @damageboy the new disassembler is ready. It would be great if you could give it a try and let me know what you think about it.

I've created a small dedicated repo and configured everything there: https://github.com/adamsitnik/TryNewDisassembler

adamsitnik:
@AndreyAkinshin @damageboy did you have some time to give it a try?

PR #1335, opened 2019-12-19
Attribute improvements

WojciechNagorski:
In #1321 I created example `IntroEventPipeProfiler` which uses the `[EventPipeProfiler(EventPipeProfile.CpuSampling)]` attribute. I wanted to mark this class as `ShortRunJob`. The used profiler requires .Net Core 3.0+ so I wanted to select `RuntimeMoniker` using the `[ShortRunJob(RuntimeMoniker.NetCoreApp30)]` attribute in the same way as `[DryRunJob(RuntimeMoniker.NetCoreApp30)]`. But I noticed that there is not an appropriate constructor. 

In this PR I added the following constructors to attributes: `LongRunJobAttribute`, `MediumRunJobAttribute`, `ShortRunJobAttribute`, and `VeryLongRunJobAttribute`.

```c#
public XXXRunJobAttribute(RuntimeMoniker runtimeMoniker)
public XXXRunJobAttribute(RuntimeMoniker runtimeMoniker, Jit jit, Platform platform)
```

I also added missing the `AttributeUsage` attribute.


WojciechNagorski:
Example log:
```
BenchmarkDotNet=v0.12.0.20191219-develop, OS=Windows 10.0.16299.1387 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7-4770 CPU 3.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
Frequency=3312641 Hz, Resolution=301.8739 ns, Timer=TSC
.NET Core SDK=3.0.100
  [Host]                 : .NET Core 2.1.13 (CoreCLR 4.6.28008.01, CoreFX 4.6.28008.01), X64 RyuJIT
  ShortRun-.NET Core 3.0 : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT

Job=ShortRun-.NET Core 3.0  Runtime=.NET Core 3.0  IterationCount=3
LaunchCount=1  WarmupCount=3
```

WojciechNagorski:
I've just had to migrate project SevenZipExtractor to .Net Standard 
https://github.com/adoconnection/SevenZipExtractor/pull/38 and I used there BenchmarkDotNet https://github.com/adoconnection/SevenZipExtractor/pull/39

If this PR were merged, I would be able to configure my benchmark using the following code:
```
    [MemoryDiagnoser]
    [ShortRunJob(RuntimeMoniker.NetCoreApp31)]
    [ShortRunJob(RuntimeMoniker.Net47)]
    public class Benchmarks
    {
        [...]
    }
```

instead of
```
    [Config(typeof(Config))]
    public class Benchmarks
    {
        {
            public Config()
            {
                Add(MemoryDiagnoser.Default);
                Add(Job.ShortRun.With(ClrRuntime.Net47));
                Add(Job.ShortRun.With(CoreRuntime.Core31));
            }
        }

        [...]
    }

```

src/BenchmarkDotNet/Attributes/Jobs/InProcessAttribute.cs
```diff
@@ -1,7 +1,9 @@
+using System;
 using BenchmarkDotNet.Jobs;
 
 namespace BenchmarkDotNet.Attributes
 {
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = true)]
```
adamsitnik: I think that we don't need `AllowMultiple` for `InProcess` attribute: there are basically two options: log output and dont. IMHO having two of them at the same time makes no sense.
```suggestion
```

src/BenchmarkDotNet/Attributes/Jobs/LegacyJitX86JobAttribute.cs
```diff
@@ -1,7 +1,9 @@
-﻿using BenchmarkDotNet.Jobs;
+﻿using System;
+using BenchmarkDotNet.Jobs;
 
 namespace BenchmarkDotNet.Attributes
 {
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = true)]
```
adamsitnik: `AllowMultiple` won't be useful for this type: it does not allow for any kind of customizations
```suggestion
```

src/BenchmarkDotNet/Attributes/Jobs/LegacyJitX64JobAttribute.cs
```diff
@@ -1,7 +1,9 @@
+using System;
 using BenchmarkDotNet.Jobs;
 
 namespace BenchmarkDotNet.Attributes
 {
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = true)]
```
adamsitnik:
```suggestion
```

src/BenchmarkDotNet/Attributes/Jobs/RyuJitX64JobAttribute.cs
```diff
@@ -1,7 +1,9 @@
-﻿using BenchmarkDotNet.Jobs;
+﻿using System;
+using BenchmarkDotNet.Jobs;
 
 namespace BenchmarkDotNet.Attributes
 {
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = true)]
```
adamsitnik:
```suggestion
```

src/BenchmarkDotNet/Attributes/Jobs/RyuJitX86JobAttribute.cs
```diff
@@ -1,9 +1,11 @@
-﻿using BenchmarkDotNet.Jobs;
+﻿using System;
+using BenchmarkDotNet.Jobs;
 using JetBrains.Annotations;
 
 namespace BenchmarkDotNet.Attributes
 {
     [PublicAPI]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = true)]
```
adamsitnik:
```suggestion
```

adamsitnik:
@WojciechNagorski ping

WojciechNagorski:
I will fix it. Sorry for late, I was busy.

DaveSkender:
Quick question on this topic: I'm producing .NET Standard 2.0 projects.  Which `RuntimeMoniker` enum do I use.  I see `RuntimeMoniker.NetCoreApp20`, but no `RuntimeMoniker.NetStandard20`, or something like that.  I thought these were different frameworks, right?  Confused.

timcassell:
@DaveSkender NetStandard2.0 is not a runtime, but a library target. It is a standard that can be consumed in multiple runtimes, like .Net Framework, .Net (Core), and Unity (not supported in BDN). See the [docs](https://docs.microsoft.com/en-us/dotnet/standard/net-standard) for a full list of runtimes that supports the standard version.

DaveSkender:
I sort get that part, and admittedly this has always confused me, but suppose I'm doing multi-targetting:

```xml
<TargetFrameworks>net6.0;netstandard2.1;</TargetFrameworks>
```

And it produces two DLLs, one for .NET 6.0 and one for .NET Standard 2.1

image

And now I want to know how they comparatively perform with BenchmarkDotNet, with these attributes:

image

I might just be confusing the purpose of `RuntimeMoniker`, but what do I do?

timcassell:
Oh I see what you mean. There is an open issue for [recompiling for target runtime](https://github.com/dotnet/BenchmarkDotNet/issues/1773), which is related, but I'm not sure how you would choose between the net6.0 dll and netstandard2.1 dll in the benchmarks. You could try creating a separate project for each, with each project referencing the different dlls, and the benchmark project referencing both of those projects. I tried that before with my own library, but for different versions (not different frameworks) and got an error which I also mentioned in that issue. So I'm not sure if that will work for you.

In any case, you should add your feedback to that issue.

adamsitnik:
@DaveSkender you are most likely looking for https://github.com/dotnet/BenchmarkDotNet/issues/1706

PR #1349, opened 2020-01-10
Fixed display text for array parameters and arguments

### LINKED ISSUE #1348: Different display text for arrays depending on a value source
There is a difference of array representation in summaries depending on a value source. In case when `ArgumentsAttribute` is used an array is displayed as `Array[{array.Length}]`, but in case of `ArgumentsSourceAttribute` - the type name is used.

The bug can be simply reproduced using the `IntroArrayParam` sample:

https://github.com/dotnet/BenchmarkDotNet/blob/a3f76beebeaab9a7ff32f3af6a16d79094fea3f6/samples/BenchmarkDotNet.Samples/IntroArrayParam.cs#L10-L31

|                 Method |          array | value |       Mean |     Error |    StdDev |
|----------------------- |--------------- |------ |-----------:|----------:|----------:|
|           ArrayIndexOf | System.Int32[] |     4 |  5.7462 ns | 0.1177 ns | 0.1101 ns |
|           ArrayIndexOf | System.Int32[] |     4 |  3.9955 ns | 0.0771 ns | 0.0602 ns |
|          ManualIndexOf | System.Int32[] |     4 |  0.8766 ns | 0.0274 ns | 0.0243 ns |
|          ManualIndexOf | System.Int32[] |     4 |  2.9022 ns | 0.0588 ns | 0.0550 ns |
|           ArrayIndexOf | System.Int32[] |   101 | 25.7760 ns | 0.5521 ns | 1.4350 ns |
|          ManualIndexOf | System.Int32[] |   101 | 41.9487 ns | 0.3473 ns | 0.3079 ns |

The expected behavior is the same as `ArgumentsAttribute` has:

```csharp
[Benchmark]
[Arguments(new int[] { 1, 2, 3 }, 4)]
public int ArrayIndexOf(int[] array, int value)
    => Array.IndexOf(array, value);
```

|        Method |          array | value |      Mean |     Error |    StdDev |
|-------------- |--------------- |------ |----------:|----------:|----------:|
|  ArrayIndexOf |       Array[3] |     4 |  4.794 ns | 0.1270 ns | 0.2225 ns |

The same issue happens with parameters.

YohDeadfall:
Fixes #1348.

It would be better to unify display text generation across all parameters, but cannot find a good place for a common method. Maybe a base class?

### DIFF
--- a/src/BenchmarkDotNet/Code/ArrayParam.cs
+++ b/src/BenchmarkDotNet/Code/ArrayParam.cs
@@ -7,6 +7,12 @@ using JetBrains.Annotations;
 
 namespace BenchmarkDotNet.Code
 {
+    internal static class ArrayParam
+    {
+        public static string GetDisplayString(Array array)
+            => $"{array.GetType().GetElementType().Name}[{array.Length}]";
+    }
+
     public class ArrayParam<T> : IParam
     {
         private readonly T[] array;
@@ -20,7 +26,7 @@ namespace BenchmarkDotNet.Code
 
         public object Value => array;
 
-        public string DisplayText => $"Array[{array.Length}]";
+        public string DisplayText => ArrayParam.GetDisplayString(array);
 
         public string ToSourceCode()
             => $"new {typeof(T).GetCorrectCSharpTypeName()}[] {{ {string.Join(", ", array.Select(item => toSourceCode?.Invoke(item) ?? SourceCodeHelper.ToSourceCode(item)))} }}";
--- a/src/BenchmarkDotNet/Code/ArrayParam.cs
+++ b/src/BenchmarkDotNet/Code/ArrayParam.cs
@@ -10,7 +10,7 @@ namespace BenchmarkDotNet.Code
     internal static class ArrayParam
     {
         public static string GetDisplayString(Array array)
-            => $"{array.GetType().GetElementType().Name}[{array.Length}]";
+            => $"{array.GetType().GetElementType().GetDisplayName()}[{array.Length}]";
     }
 
     public class ArrayParam<T> : IParam
--- a/src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
+++ b/src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
@@ -81,7 +81,7 @@ namespace BenchmarkDotNet.Parameters
 
         public object Value { get; }
 
-        public string DisplayText => Value.ToString();
+        public string DisplayText => Value is Array array ? $"Array[{array.Length}]" : Value.ToString();
 
         public string ToSourceCode()
         {
@@ -115,7 +115,7 @@ namespace BenchmarkDotNet.Parameters
 
         public object Value { get; }
 
-        public string DisplayText => Value.ToString();
+        public string DisplayText => Value is Array array ? $"Array[{array.Length}]" : Value.ToString();
 
         public string ToSourceCode()
         {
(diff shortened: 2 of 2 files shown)


YohDeadfall:
Changed display string to show an element type instead of `Array`, but with no namespace for shortness.

/cc @adamsitnik @WojciechNagorski

src/BenchmarkDotNet/Code/ArrayParam.cs
```diff
@@ -7,6 +7,12 @@
 
 namespace BenchmarkDotNet.Code
 {
+    internal static class ArrayParam
```
AndreyAkinshin: There is no need to introduce a separate class; it will be better to just add a new private static method in `ArrayParam<T>`.

src/BenchmarkDotNet/Code/ArrayParam.cs
```diff
@@ -7,6 +7,12 @@
...
+    {
+        public static string GetDisplayString(Array array)
+            => $"{array.GetType().GetElementType().Name}[{array.Length}]";
```
AndreyAkinshin: It will be better to get the type name with the help of `BenchmarkDotNet.Extensions.ReflectionExtensions.GetDisplayName`: this method provides nice presentation for generic classes.

AndreyAkinshin:
@YohDeadfall thanks for the PR! I left a few minor comments about possible improvements. Once you resolve it, I will merge your PR.

src/BenchmarkDotNet/Code/ArrayParam.cs
YohDeadfall: In that case a generic parameter must be specified when the method is called from `SmartArgument` for example. Of course it can be `object`, but using a shared instance when it isn't needed is a bad idea, you know.

src/BenchmarkDotNet/Code/ArrayParam.cs
YohDeadfall: Okay, will do.

src/BenchmarkDotNet/Code/ArrayParam.cs
AndreyAkinshin: Oh, got it. Ok, let's keep the current implementation. I will refactor it later (will try to find a better util class).

src/BenchmarkDotNet/Code/ArrayParam.cs
YohDeadfall: Looks like the class you mentioned is for reflection extensions only. Don't think that moving the method is a good idea. What am thinking about is refactoring smart and array params into a single class.

AndreyAkinshin: Well, you already call reflection methods (`.GetType().GetElementType().Name`), but your code doesn't work well in the case of generic classes. I'm suggesting to reuse the existing formatting logic for such cases. What's wrong with the usage of `ReflectionExtensions`?

YohDeadfall: `ReflectionExtensions` accept a type and other reflection related classes as `this` parameter, while this one should accept an array. Is it okay to mix such things?

AndreyAkinshin: @YohDeadfall sorry, I guess, the text of my initial suggestion wasn't clear enough, so let me rephrase. In *this* thread, I just suggested to replace

```cs
            => $"{array.GetType().GetElementType().Name}[{array.Length}]";
```
by
```cs
            => $"{array.GetType().GetElementType().GetDisplayName()}[{array.Length}]";
```
In this case, we will get a better presentation for cases when the element type is generic (e.g., `List<int>[]`).

In *another* thread, I suggested moving the whole `GetDisplayString(Array array)` method to another utility class. It's not clear for me right now which class should we choose, I will think about it later. Once I find a good place, I will refactor it myself.

YohDeadfall: Ah, I see.

PR #1381, opened 2020-03-03
Minor event pipe profiler improvements

adamsitnik:


src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
```diff
@@ -17,73 +17,57 @@
...
+        private readonly Dictionary<BenchmarkCase, string> benchmarkToTraceFile = new Dictionary<BenchmarkCase, string>();
+        private readonly ImmutableHashSet<EventPipeProvider> eventPipeProviders;
+        private readonly bool performExtraBenchmarksRun;
+
+        private Task collectingTask;
+
         // parameterless constructor required by DiagnosersLoader to support creating this profiler via console line args
-        public EventPipeProfiler() { }
+        // we use performExtraBenchmarksRun = false for better first user experience
+        public EventPipeProfiler() :this(profile: EventPipeProfile.CpuSampling, performExtraBenchmarksRun: false) { }
```
WojciechNagorski: Why this constructor has the `performExtraBenchmarksRun` parameter with `false` value as a default. In the second constructor, this parameter has 'true', I think It is not consistent.

adamsitnik:
@WojciechNagorski could you please take a look?

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
```diff
@@ -92,144 +76,78 @@ public void Handle(HostSignal signal, DiagnoserActionParameters parameters)
...
-                {
-                    yield return new ValidationError(true, $"{nameof(EventPipeProfiler)} supports only .NET Core 3.0+", benchmark);
-                }
-            }
         }
 
-        public string ShortName => "EventPipe";
-
-        private void ConvertToSpeedscope(BenchmarkCase benchmarkCase, string traceFilePath)
+        private static ImmutableHashSet<EventPipeProvider> MapToProviders(EventPipeProfile profile, IReadOnlyCollection<EventPipeProvider> providers)
```
WojciechNagorski: This function has a little different behavior than my old code and the `dotnet trace` tool. 

There you can find the documentation of `dotnet trace`: https://docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-trace

>* --providers \<list-of-comma-separated-providers>
> A comma-separated list of EventPipe providers to be enabled. These providers supplement any providers implied by --profile <profile-name>. If there's any inconsistency for a particular provider, this configuration takes precedence over the implicit configuration from the profile.

So for example a user could send:
1. profile = `EventPipeProfile.CpuSampling`
2. providers = `new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Verbose, (long) ClrTraceEventParser.Keywords.Default)`

```c#
 new EventPipeProfiler(EventPipeProfile.CpuSampling, new [] { new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Verbose, (long) ClrTraceEventParser.Keywords.Default | ClrTraceEventParser.Keywords.JitTracing)});
```

This should give all the providers sent by the user and all providers resulting from the profile except those that were sent in the providers.

In the above example it should use:

1. sent by user - `new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Verbose, (long) ClrTraceEventParser.Keywords.Default | ClrTraceEventParser.Keywords.JitTracing)`
2. from the profile - `new EventPipeProvider("Microsoft-DotNETCore-SampleProfiler", EventLevel.Informational)`

This from the profile won't be used:
`new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Informational, (long) ClrTraceEventParser.Keywords.Default)`

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
adamsitnik: because this ctor is used when users specify console line arguments and they simply cant control it from command line level.

We do the same thing for ETWProfiler: https://github.com/dotnet/BenchmarkDotNet/blob/e92abf845ac916e449bf9d8ccf1d872204951bec/src/BenchmarkDotNet.Diagnostics.Windows/EtwProfiler.cs#L32

So, in general, we do it to improve first user experience (shorter time to get the results)

src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
adamsitnik: @WojciechNagorski great catch! I've added tests for that and fixed it. PTAL again

https://github.com/dotnet/BenchmarkDotNet/pull/1381/commits/68a207d6c3627be3372de6370d17461f6d2a0c6b

PR #1387, opened 2020-03-09
Make BaselineCustomColumn expose "GetValue" as a public API

### LINKED ISSUE #1385: Make BaselineCustomColumn.GetValue public
Hi,
Referring to this:

https://github.com/dotnet/BenchmarkDotNet/blob/1d63d666da70530b03e392af23453bf5cd7c1784/src/BenchmarkDotNet/Columns/BaselineCustomColumn.cs#L27

Is there any particular reason the whole class is public while this specific method is kept internal?

Wouldn't it make sense to make the method public so it could be overriden outside of BDN, or is there some sort of an issue with that?

damageboy:
- Add ability to generate baseline columns based on custom metrics
- Expose the existing GetValue method as public
- Closes #1385


src/BenchmarkDotNet/Columns/BaselineCustomColumn.cs
```diff
@@ -19,12 +21,16 @@ public string GetValue(Summary summary, BenchmarkCase benchmarkCase)
...
+            var currentMetrics = summary[benchmarkCase].Metrics;
 
-            return GetValue(summary, benchmarkCase, baselineStat, currentStat, isBaseline);
+            return GetValue(summary, benchmarkCase, baselineStat, baselineMetrics, currentStat, currentMetrics, isBaseline);
         }
 
-        internal abstract string GetValue(Summary summary, BenchmarkCase benchmarkCase, Statistics baseline, Statistics current, bool isBaseline);
+        [PublicAPI]
+        public abstract string GetValue(Summary summary, BenchmarkCase benchmarkCase, Statistics baseline, IReadOnlyDictionary<string, Metric> baselineMetrics,
+            Statistics current, IReadOnlyDictionary<string, Metric> currentMetrics, bool isBaseline);
```
adamsitnik: good idea to use `IReadOnlyDictionary` 👍

Issue #1336, opened 2019-12-19
Improve contrast in console (dark text on black console)


bender2k14:
When I run `BenchmarkDotNet.Samples.FSharp`, here is what the console looks like.
![2019-12-19_10-59-11_10_BenchmarkDotNet Samples FSharp](https://user-images.githubusercontent.com/34664007/71192956-e00d8f80-224e-11ea-9246-11c975842730.png)

The text is a dark color on a black background.  I am having difficulty seeing the text.

Can this contrast be improved?

adamsitnik:
Hello @bender2k14 

> Can this contrast be improved?

Yes, but it would require your contribution.

The colors are defined [here](https://github.com/dotnet/BenchmarkDotNet/blob/3223c94a92050147a02482a3810860703f0c5171/src/BenchmarkDotNet/Loggers/ConsoleLogger.cs#L64-L75) Please feel free to experiment with them and send a PR.

Thanks,
Adam

bender2k14:
Hello @adamsitnik, thanks for your response.

If you would prefer to reinvent logging (instead taking on dependency for logging), then I recommend copying how [Serilog.Sinks.Console](https://github.com/serilog/serilog-sinks-console#themes) does it.  Specifically, the coloring that you are currently using is fine but only apply it to the log type (which implies that it is also displayed).

What do you think?

adamsitnik:
Do you mean using the same colors as Serilog?

![obraz](https://user-images.githubusercontent.com/6011991/77436183-779cf400-6de3-11ea-99c3-f44df402e247.png)

I am fine with that. @AndreyAkinshin what do you think?

AndreyAkinshin:
@adamsitnik so, do I understand correctly that we are going to replace
```cs
{ LogKind.Help, ConsoleColor.DarkGreen },
```
with
```cs
{ LogKind.Help, ConsoleColor.White },
```
?

adamsitnik:
@AndreyAkinshin yes, but we might also want to set bg color to red for error and yellow text color for warning

bender2k14:
> Do you mean using the same colors as Serilog?

I don't care about the exact colors used.  I am suggesting that the text to be log should be white so that the contrast with the black background is high.

> do I understand correctly that we are going to replace...

No.  Sorry that I have been unclear.  I created PR #1404 that makes the kind of change that I am taking about.  It also includes a screenshot of the new behavior.

adamsitnik:
[...]

PR #1404, opened 2020-03-24
adds log kind abbreviation to output of console logger and then only colors that

bender2k14:
Intended to address #1336

Here is a screenshot when running the project `BenchmarkDotNet.Samples.FSharp`

2020-03-24_10-10-44_86_BenchmarkDotNet Samples FSharp

### DIFF
--- a/src/BenchmarkDotNet/Loggers/ConsoleLogger.cs
+++ b/src/BenchmarkDotNet/Loggers/ConsoleLogger.cs
@@ -16,50 +16,77 @@ namespace BenchmarkDotNet.Loggers
 
         private readonly bool unicodeSupport;
         private readonly Dictionary<LogKind, ConsoleColor> colorScheme;
+        private readonly Dictionary<LogKind, string> logKindAbbreviations;
+
+        private bool newLine = true;
 
         [PublicAPI]
         public ConsoleLogger(bool unicodeSupport = false, Dictionary<LogKind, ConsoleColor> colorScheme = null)
         {
             this.unicodeSupport = unicodeSupport;
             this.colorScheme = colorScheme ?? CreateColorfulScheme();
+            this.logKindAbbreviations = CreateLogKindAbbreviations();
         }
 
         public string Id => nameof(ConsoleLogger);
 
         public int Priority => unicodeSupport ? 1 : 0;
 
-        public void Write(LogKind logKind, string text) => Write(logKind, Console.Write, text);
+        public void Write(LogKind logKind, string text) {
+            if (newLine)
+                WriteAbbreviation(logKind);
+            newLine = false;
+            WriteText(Console.Write, text);
+        }
 
-        public void WriteLine() => Console.WriteLine();
+        public void WriteLine() {
+            Console.WriteLine();
+            newLine = true;
+        }
 
-        public void WriteLine(LogKind logKind, string text) => Write(logKind, Console.WriteLine, text);
+        public void WriteLine(LogKind logKind, string text) {
+            if (newLine)
+                WriteAbbreviation(logKind);
+            WriteText(Console.WriteLine, text);
+            newLine = true;
+        }
 
         public void Flush() { }
 
-        private void Write(LogKind logKind, Action<string> write, string text)
+        private void WriteAbbreviation(LogKind logKind)
         {
-            if (!unicodeSupport)
-                text = text.ToAscii();
-
             var colorBefore = Console.ForegroundColor;
 
+            Console.Write("[");
             try
             {
                 var color = GetColor(logKind);
                 if (color != Console.ForegroundColor && color != Console.BackgroundColor)
                     Console.ForegroundColor = color;
-
-                write(text);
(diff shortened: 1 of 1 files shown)


adamsitnik:
@bender2k14 may I ask you for a screen shot of powershell command prompt? I wonder how this looks with the default PWS blue colors

AndreyAkinshin:
@bender2k14 To be honest, I don't like the idea of `[XXX]` prefixes because of the following reasons:
1. These prefixes look confusing, nobody will understand that `[STA]` means "statistics" and `[HDR]` means "header"
2. They don't provide any useful information for user.

The whole idea of our color schemes is color differentiation between different kinds of information (e.g., we use it to distinguish adjacent lines in the summary table; the approach with prefixes will not help to distinguish lines)

bender2k14:
Sure @adamsitnik.

Can you tell me how to do that?

To get the existing screenshot, I opened the solution in Visual Studio, set `BenchmarkDotNet.Samples.FSharp` as the start up project, and then started a new instance (by right clicking the project and selecting "Debug -> Start new instance").

adamsitnik:
You need to open the folder with fsharp samples project, press shift and right click with the mouse:

obraz

Choose "Open PowerShell windows here" and type `dotnet run -c Release -f net461` and hit enter

bender2k14:
Thanks for the feedback @AndreyAkinshin.

> 1. These prefixes look confusing, nobody will understand that [STA] means "statistics" and [HDR] means "header".

Sure, that is easily avoided by using "statistics" and "header" instead the the abbreviations.  I was just trying to provide a proof-of-concept that copies how [Serilog.Sinks.Console](https://github.com/serilog/serilog-sinks-console) does it.

> 2. They don't provide any useful information for user.

I don't think I understand what you mean here.  You are already using color to convey the log kind.  I am suggesting that your log kinds should also be communicated to the user in some way in addition to the user of color.  I _don't_ think that you are saying that the log kind are _not_ "useful information to the user" and therefore should _not_ be conveyed to the user in any way including via color.

Relying solely on color to differentiate the log kind of log events is problematic.  The color support of programs (e.g. terminals and PowerShell), the color support on monitors, and the color vision of people varies widely.  You are currently using `ConsoleColor.DarkGreen` for log events of kind `Help`.  I could not find an authoritative source for the corresponding hex value, but I think it is `006400`.  According to the [Web Content Accessibility Guidelines (WCAG)](https://en.wikipedia.org/wiki/Web_Content_Accessibility_Guidelines), your current use of [`DarkGreen` on `Black` fails all their accessibility tests](https://webaim.org/resources/contrastchecker/?fcolor=006400&bcolor=000000).

> The whole idea of our color schemes is color differentiation between different kinds of information (e.g., we use it to distinguish adjacent lines in the summary table; the approach with prefixes will not help to distinguish lines)

It sounds to me like your problems have better and more common solutions.  If your problem is differentiating general sections of text, then I think the correct solution is to put them in separate paragraphs for separate them with a section heading.  If your problem is to differentiating adjacent lines in a table, then you should use a [stripped table](https://www.w3schools.com/css/tryit.asp?filename=trycss_table_striped).

bender2k14:
Ah, yes.  I remember now that I have done that before.  Thanks for the reminder @adamsitnik.  

Here you go.

2020-03-24_11-53-15_88_powershell

AndreyAkinshin:
@bender2k14,

1. Speaking of the default help text that you show on the screenshots, it doesn't make sense to print `[HLP]` on the EACH line because all of the lines are help lines. It's just visual noise. However, I agree that `ConsoleColor.DarkGreen` is not the best choice here. We can change it.
2. Our logging levels are not actually LOGGING levels in the classic sense. It's nice when users can easily distinguish Statistics/Header/Info using colors, but it's not critical information. If we remove all the colors from all the lines, the output will be still readable (it's pretty easy to identify headers, statistics, the summary table, etc.). The color scheme is just an *additional* way to improve readability.
3. I agree that the current scheme is not optimal because of the limitations of `ConsoleColor`. We can try to improve it for terminals that support true colors, but we still should have fallback console colors for 16 color terminal.
4. I don't think that the stripped table (when we modify the text background) is a good choice for the terminal. We already use empty lines to separate the most important groups of lines; the colors are used as an *additional* way to distinguish subgroups inside groups (it's not a way to highlight adjacent lines). In the HtmlExporter, we use regular and bold fonts for this purpose.

bender2k14:
Thanks for the lengthy reply @AndreyAkinshin.  I agree with all points, especially point 2.  I am not suggesting the removal of all colors.  Color can indeed be a useful addition to makes things more readable, but color can also makes things less readable (like the screenshot I provided in https://github.com/dotnet/BenchmarkDotNet/issues/1336#issue-540439794).

AndreyAkinshin:
>  I agree with all points

@bender2k14 great. So, I close this pull request.

> but color can also makes things less readable

Agree. In case if you want to create another pull request with updated color scheme, I will be happy to discuss it!

PR #1407, opened 2020-03-26
Improve warnings for small operations number

### LINKED ISSUE #1003: improve MinIterationTime and MultimodeDistribution warnings
So, I feel like I'm getting some invalid benchmark results, and if so, I'm out of ideas on how to address it.

What I'm seeing is about a dozen warnings of the following form...

```
Add.Dictionary: InvocationCount=1, UnrollFactor=1 -> MinIterationTime = 768.2200 ns which is very small. It's recommended to increase it.
...
Add.Dictionary: InvocationCount=1, UnrollFactor=1 -> It seems that the distribution is bimodal (mValue = 3.49)
```
([FULL BENCHMARK LOG](https://hastebin.com/kemetevuda.makefile))

I had a look at #767 where the author was having what appears to be the exact same issue, on account of a manual configuration, and the fact that the benchmark in question produced side affects.

I have no customized configuration, so I figure my issue has GOT to be the fact that my benchmark also produces side-effects. I attempted to counter the side-effects by use of an `[IterationSetup]` method, but I can only assume I don't know what I'm doing there, since I still get warnings, AND since another benchmark in this same project, which doesn't produce side-effects, doesn't generate any warnings, and takes MUCH longer to run (presumably because it's doing more than 1 iteration).

[Here's the benchmark in question](https://github.com/JakenVeina/ImmutableHashDictionary/blob/master/ImmutableHashDictionary.Benchmark/Add.cs)

Alternatively, the entire [codebase](https://github.com/JakenVeina/ImmutableHashDictionary) should be cloneable and runnable, as-is, to reproduce the issue.

Am I just completely misunderstanding how to counteract side-effects?

CodeFuller:
Fixes #1003

### DIFF
--- a/src/BenchmarkDotNet/Analysers/MinIterationTimeAnalyser.cs
+++ b/src/BenchmarkDotNet/Analysers/MinIterationTimeAnalyser.cs
@@ -25,7 +25,7 @@ namespace BenchmarkDotNet.Analysers
                 yield break;
             var minActualIterationTime = TimeInterval.FromNanoseconds(target.Min(m => m.Nanoseconds));
             if (minActualIterationTime < MinSufficientIterationTime)
-                yield return CreateWarning($"The minimum observed iteration time is {minActualIterationTime} which is very small. It's recommended to increase it.", report);
+                yield return CreateWarning($"The minimum observed iteration time is {minActualIterationTime} which is very small. It's recommended to increase it to at least {MinSufficientIterationTime} using more operations.", report);
         }
     }
 }
--- a/src/BenchmarkDotNet/Analysers/MultimodalDistributionAnalyzer.cs
+++ b/src/BenchmarkDotNet/Analysers/MultimodalDistributionAnalyzer.cs
@@ -32,6 +32,6 @@ namespace BenchmarkDotNet.Analysers
 
         [NotNull]
         private Conclusion Create([NotNull] string kind, double mValue, [CanBeNull] BenchmarkReport report, CultureInfo cultureInfo)
-            => CreateWarning($"It seems that the distribution {kind} (mValue = {mValue.ToString("0.##", cultureInfo)}). The issue could be fixed by using more operations.", report);
+            => CreateWarning($"It seems that the distribution {kind} (mValue = {mValue.ToString("0.##", cultureInfo)})", report);
     }
 }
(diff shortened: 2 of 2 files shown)


src/BenchmarkDotNet/Analysers/MinIterationTimeAnalyser.cs
```diff
@@ -25,7 +25,7 @@ protected override IEnumerable<Conclusion> AnalyseReport(BenchmarkReport report,
                 yield break;
             var minActualIterationTime = TimeInterval.FromNanoseconds(target.Min(m => m.Nanoseconds));
             if (minActualIterationTime < MinSufficientIterationTime)
-                yield return CreateWarning($"The minimum observed iteration time is {minActualIterationTime} which is very small. It's recommended to increase it.", report);
+                yield return CreateWarning($"The minimum observed iteration time is {minActualIterationTime} which is very small. It's recommended to increase it to at least {MinSufficientIterationTime} using more operations.", report);
```
AndreyAkinshin: Good catch.

src/BenchmarkDotNet/Analysers/MultimodalDistributionAnalyzer.cs
```diff
@@ -32,6 +32,6 @@ protected override IEnumerable<Conclusion> AnalyseReport(BenchmarkReport report,
 
         [NotNull]
         private Conclusion Create([NotNull] string kind, double mValue, [CanBeNull] BenchmarkReport report, CultureInfo cultureInfo)
-            => CreateWarning($"It seems that the distribution {kind} (mValue = {mValue.ToString("0.##", cultureInfo)})", report);
+            => CreateWarning($"It seems that the distribution {kind} (mValue = {mValue.ToString("0.##", cultureInfo)}). The issue could be fixed by using more operations.", report);
```
AndreyAkinshin: The additional sentence can be confused in some cases. Indeed, it's possible to get a false-positive warning about multimodality when the number of operations/invocations/iterations is not big enough. However, in many cases, the performance distribution is multimodal by nature, we can't do anything with it by adjusting benchmark parameters.

CodeFuller: Good point. I have reverted the change for this warning.

AndreyAkinshin:
@CodeFuller thanks!

PR #1426, opened 2020-04-16
Fix typo in log message

martincostello:
Fix typo in log message when command times out.


AndreyAkinshin:
@martincostello thanks!

PR #1443, opened 2020-04-29
Handle assemblies loaded via a stream

jeremyosterhoudt:
Hello,

We are using a test framework called Gauge (https://gauge.org/) to do regression testing and wanted to roll performance tests in with those regression tests.  In our case we want to use Gauge steps to control and kick off different BenchmarkDotNet benchmarks (so we have a common top level framework for all of our regression tests).  

One issue we have run into is Gauge loads assemblies via streams, this causes the assembly location to be an empty string when running against net core/net standard.  More information about the assembly location being empty in this case can be found in Microsoft's documentation (https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location).

This PR provides a way to handle that scenario by providing a default location for the project build artifacts.  Other non-CsProjGenerators seem to do this by default so we have only updated this generator.

It's possible this could be handled a different way, but we've found this solution is working well for us.

Any feedback is welcomed.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
@@ -35,13 +35,12 @@ namespace BenchmarkDotNet.Toolchains.CsProj
 
         protected override string GetBuildArtifactsDirectoryPath(BuildPartition buildPartition, string programName)
         {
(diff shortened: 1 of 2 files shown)


jeremyosterhoudt:
@adamsitnik 

Thanks for the feedback.

I followed the instructions to generate a nuget spec using `dotnet pack -c Release` and used that spec in my project.  Everything is running as expected.

Screen Shot 2020-04-30 at 11 21 57 AM


Thanks,
Jeremy

PR #1451, opened 2020-05-07
Fix native memory profiler

### LINKED ISSUE #1427: NativeMemoryProfiler reports false positive leak
Hello all,
Thank you for the work on the project, it is much appreciated!

I am reporting an issue we are experiencing with the NativeMemoryProfiler of BenchmarkDotNet. It detects small but constant native memory leaks that seem to be false positives.
The attached reproducer showcases the issue. The benchmark only instantiates a mock object and computes a SHA256 (to avoid spending too little time in the benchmark). The resulting detected native memory leak is 0.06KB on my machine.

I believe the mocking library we use does not do weird things with native memory, so this gives me a strong feeling that this could be a false positive. Please note that this also happens with an industrial library we use in our software and which is heavily tested for memory leaks. While the library is using native memory, we are confident that the provider extensively tested it for any leak, which is another evidence that this could be a false positive. Unfortunately I cannot provide this library for inspection.

I am available for clarifications, should the need arise.

Version of BenchmarkDotNet: 0.12.1
Version of the Windows Diagnostics nuget: 0.12.1
Version of .net: core 3.1
Minimal reproducer: [MemoryProfilerReproducer.zip](https://github.com/dotnet/BenchmarkDotNet/files/4493306/MemoryProfilerReproducer.zip)

WojciechNagorski:
This PR fixes #1427 for .Net Core and .Net Framework.

The following benchmark is problematic:
```c#
        [Benchmark]
        public string StackTrace()
        {
            var stackTrace = new StackTrace();

            return stackTrace.GetType().Name;
        }
```

I don't know why but `new StackTrace()` causes different workings of the benchmark. 

#### .Net Core
image
There are allocation (propably from JIT) between ActulaWorkStart and Stop :
image
It is a event of JIT of WorkloadActualStop method.
```
ThreadID="13,340" ProcessorNumber="0" MethodID="140,712,104,538,952" ModuleID="140,712,102,993,560" MethodToken="100,665,663" MethodILSize="10" MethodNamespace="BenchmarkDotNet.Engines.EngineEventSource" MethodName="WorkloadActualStop" MethodSignature="instance void  (int64)" ClrInstanceID="8"
```
The following screen shows that that allocations do not come from our benchmark:
image

So I solved it by filtering events only from benchmark stacks. I wanted to fix it by running a benchmark twice but I got a different JIT event.

#### .NET Framework without fix
In this case, there are allocations between ActualWork Start and Stop. 
image
But I don't know anything about the origin of these allocations.
image

#### .Net Framework with fix
I found a fix during analyzing the PerfView code. I don't know why but these allocations disappear after adding `ClrTraceEventParser.Keywords.JittedMethodILToNativeMap`

image

#### Summary
I tested if using my repository https://github.com/WojciechNagorski/NativeMemorySample. If you want to check it you need to change references to a custom build of BenchmarkDotNet.


WojciechNagorski:
@jjolidon, @jmmorato, @Meetsch, or @pleonex Can you check if my fix works in your environment?

jmmorato:
Hi @WojciechNagorski,

Quick test after clone your branch and replace NuGet references:

BenchmarkDotNet=v0.12.0.20200508-develop, OS=Windows 10.0.18362.418 (1903/May2019Update/19H1)
Intel Core i7-7820HQ CPU 2.90GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
  [Host]   : .NET Framework 4.8 (4.8.3752.0), X64 RyuJIT
  ShortRun : .NET Framework 4.8 (4.8.3752.0), X64 RyuJIT

Job=ShortRun  InvocationCount=1  IterationCount=3  
LaunchCount=1  UnrollFactor=1  WarmupCount=3  

Method | Payload |     Mean |    Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated | Allocated native memory | Native memory leak |
-------------------- |-------- |---------:|---------:|----------:|------:|------:|------:|----------:|------------------------:|-------------------:|
 FileTransferPayload |   10000 | 1.453 ms | 7.577 ms | 0.4153 ms |     - |     - |     - |         - |                691132 B |           376790 B |


I'm using the followin configuration:

```csharp
        /// <summary>
        /// Initializes a new instance of the <see cref="BenchmarkConfig"/> class.
        /// </summary>
        public BenchmarkConfig()
        {
#if DEBUG
            AddValidator(JitOptimizationsValidator.DontFailOnError);
#endif
            AddJob(Job.ShortRun);
            AddColumnProvider(DefaultConfig.Instance.GetColumnProviders().ToArray());
            AddLogger(DefaultConfig.Instance.GetLoggers().ToArray());
            AddExporter(PlainExporter.Default);
            AddExporter(CsvExporter.Default);
            AddExporter(MarkdownExporter.Default);
            AddDiagnoser(MemoryDiagnoser.Default);
            AddDiagnoser(new NativeMemoryProfiler());
        }
```

WojciechNagorski:
@jmmorato Is the result ok or not? 
Can you share a source of the benchmark and csproj?

jjolidon:
Hey,
Thanks a lot for the work on this!
I have tested with my simple reproducer and your fixes, it seems the leak is smaller but still present:
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.14393.3595 (1607/AnniversaryUpdate/Redstone1)
Intel Core i7-7820HQ CPU 2.90GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
Frequency=2835938 Hz, Resolution=352.6170 ns, Timer=TSC
.NET Core SDK=3.1.201
  [Host]   : .NET Core 3.1.3 (CoreCLR 4.700.20.11803, CoreFX 4.700.20.12001), X64 RyuJIT
  ShortRun : .NET Core 3.1.3 (CoreCLR 4.700.20.11803, CoreFX 4.700.20.12001), X64 RyuJIT
|      Method |   Gen 0 |  Gen 1 | Gen 2 | Allocated | Allocated native memory | Native memory leak |
|------------|-------:|-------:|------:|----------:|------------------------:|-------------------:|
| MyBenchmark | 2.8687 | 0.6714 |     - |   11.8 KB |                 2.23 KB |            0.05 KB |

(I have removed unrelated columns).
Sorry to bring in the bad news, but the bug doesn't seem to be fixed. Still, it doesn't mean your changes are necessarily useless, but there seems to be more to this issue.

WojciechNagorski:
@jjolidon I will check it one more time ;)

jjolidon:
OH WAIT, I had the wrong version restored, it would seem. Clearing the nuget cache solved that. Sorry for my mistake... Your branch fixes the leak shown by my reproducer, thanks again for the work!

BenchmarkDotNet=v0.12.0.20200508-develop, OS=Windows 10.0.14393.3595 (1607/AnniversaryUpdate/Redstone1)
Intel Core i7-7820HQ CPU 2.90GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
Frequency=2835938 Hz, Resolution=352.6170 ns, Timer=TSC
.NET Core SDK=3.1.201
  [Host]   : .NET Core 3.1.3 (CoreCLR 4.700.20.11803, CoreFX 4.700.20.12001), X64 RyuJIT
  ShortRun : .NET Core 3.1.3 (CoreCLR 4.700.20.11803, CoreFX 4.700.20.12001), X64 RyuJIT

Job=ShortRun  Force=True  IterationCount=3  
LaunchCount=1  WarmupCount=3  

|      Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated | Allocated native memory | Native memory leak |
|------------ |---------:|---------:|---------:|-------:|-------:|------:|----------:|------------------------:|-------------------:|
| MyBenchmark | 33.81 μs | 25.66 μs | 1.406 μs | 2.8687 | 0.6714 |     - |   11.8 KB |                 2.18 KB |                  - |

WojciechNagorski:
@jjolidon I have the same result. Thanks!
@adamsitnik It is ready to review.

jmmorato:
> Is the result ok or not?

Well... the result is not ok for me in terms of memory leak reported :P but after your fix I cannot ensure that the problem is not in my side without more investigation. 

> Can you share a source of the benchmark and csproj?

I'm creating a wrapper library ([OpenDDSharp](https://github.com/jmmorato/openddsharp)) over a C++ library ([OpenDDS](https://github.com/objectcomputing/OpenDDS)). Right now, the development environment is not easy to setup (I need to work on that), so I prefer to share with you a minimal reproducer using the already compiled NuGet package [PerformanceTest.zip](https://github.com/dotnet/BenchmarkDotNet/files/4607801/PerformanceTest.zip) (You'll need to change the Benchmark.DotNet project references).

I just discovered your great diagnoser few weeks ago and start to play around with it. I'm really interested in your fix and/or improvements, so anything I could do to help you just let me know. 

Thanks for your job.

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
```diff
@@ -71,6 +71,8 @@ internal override Session EnableProviders()
             if (Details.Config.GetHardwareCounters().Any())
                 keywords |= KernelTraceEventParser.Keywords.PMCProfile; // Precise Machine Counters
 
+            TraceEventSession.StackCompression = true;
```
adamsitnik, line 74: it would be great if you could add some comment on why we need that

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -23,18 +27,32 @@ public class NativeMemoryLogParser
...
             this.benchmarkCase = benchmarkCase;
             this.logger = logger;
+
+            var module = Path.GetFileNameWithoutExtension(benchmarkCase.Descriptor.WorkloadMethod.Module.Name);
+            var function = new Regex($@"{benchmarkCase.Descriptor.WorkloadMethod.DeclaringType?.FullName}\.{benchmarkCase.Descriptor.WorkloadMethod.Name}\(");
+
+            benchmarkedFunctionStacks = new (string module, Regex function)[]
+            {
+                (module, function),
+                (programName, new Regex(@"BenchmarkDotNet\.Autogenerated\.Runnable_[0-9]+.WorkloadActionNoUnroll\(")),
```
adamsitnik, line 45: nit: use `RegexOptions.Compiled` to get the best perf

```suggestion
                (programName, new Regex(@"BenchmarkDotNet\.Autogenerated\.Runnable_[0-9]+.WorkloadActionNoUnroll\(", RegexOptions.Compiled)),
```

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -23,18 +27,32 @@ public class NativeMemoryLogParser
...
+                (programName, new Regex(@"BenchmarkDotNet\.Autogenerated\.Runnable_[0-9]+.WorkloadActionUnroll\("))
```
adamsitnik, line 46:
```suggestion
                (programName, new Regex(@"BenchmarkDotNet\.Autogenerated\.Runnable_[0-9]+.WorkloadActionUnroll\(", RegexOptions.Compiled))
```

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
```diff
@@ -21,7 +21,7 @@ public HeapSession(DiagnoserActionParameters details, EtwProfilerConfig config,
         {
         }
 
-        protected override string FileExtension => ".userheap.etl";
+        protected override string FileExtension => "userheap.etl";
```
adamsitnik, line 24: nice catch! I wonder why we have not found the double dots `..` in the file names before ;p

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -23,18 +27,32 @@ public class NativeMemoryLogParser
...
+
+        public NativeMemoryLogParser(string etlFilePath, BenchmarkCase benchmarkCase, ILogger logger,
+            string programName)
         {
             this.etlFilePath = etlFilePath;
             this.benchmarkCase = benchmarkCase;
             this.logger = logger;
+
+            var module = Path.GetFileNameWithoutExtension(benchmarkCase.Descriptor.WorkloadMethod.Module.Name);
+            var function = new Regex($@"{benchmarkCase.Descriptor.WorkloadMethod.DeclaringType?.FullName}\.{benchmarkCase.Descriptor.WorkloadMethod.Name}\(");
```
adamsitnik, line 40: `type.FullName` is most probably going to return some unusable name for generic types ;/

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
```diff
@@ -91,7 +117,27 @@ public IEnumerable<Metric> Parse()
...
+                            s.function.IsMatch(stack[1])))
+                        {
+                            return true;
+                        }
+
+                        var callerIndex = stackSource.GetCallerIndex(index);
+
+                        return callerIndex != StackSourceCallStackIndex.Invalid
+                               && IsCallStackIn(callerIndex);
+                    }
```
adamsitnik, line 138: this method is quite expensive. For every stack in the stack trace we are:

* calling the method in a recursive way (we should be able to avoid `StackOverflowException` because some process was able to get such a deep call stack)
* allocating an array of strings (`string.Split`)
* comparing the name using 3 regular expressions

> I wanted to fix it by running a benchmark twice but I got a different JIT event.

This was my first idea, it would be great if we could just ignore the results from the first iteration ;)

But if this is impossible, perhaps we could just search for `.WorkloadAction` in the call stacks? Sth like this:

```cs
while (index != StackSourceCallStackIndex.Invalid)
{
    var frame = stackSource.GetFrameIndex(index);
    var name = stackSource.GetFrameName(frame, false);
    
    if (name.IndexOf(".WorkloadAction", StringComparison.Ordinal) > 0)
    {
        return true;
    }
    
    index = stackSource.GetCallerIndex(index);
}

return false;
```

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
WojciechNagorski: To be honest, we do not need it, but PerfView uses it and as you can read here: https://gist.github.com/Zhentar/7f3f60fb51e89fe95abca7b9724964ba:

>"Stack Caching" (or Stack Compression as PerfView calls it) is a feature of ETW designed to reduce trace buffer & etl file sizes by de-duplicating stack traces. 

I can revert it if you want.

src/BenchmarkDotNet.Diagnostics.Windows/Sessions.cs
WojciechNagorski: ;) I've broken it, so I'm fixing it.

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
WojciechNagorski: I need more time to check it.

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
WojciechNagorski: I need more time to investigate it. Each idea takes a long time to test it. I have 12 benchmark in my test repo https://github.com/WojciechNagorski/NativeMemorySample and I have to run it for .Net Framework and .Net Core.

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
adamsitnik: @WojciechNagorski ping

WojciechNagorski: @adamsitnik I'm going to finish this task but currently I have a lot of work. Unfortunately, BDN is just my hobby.

WojciechNagorski: @adamsitnik I've removed this Regex at all.

src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
WojciechNagorski: @adamsitnik I've used your iterative algorithm with minor modifications.

WojciechNagorski:
@adamsitnik I think it is ready for review.

WojciechNagorski:
@adamsitnik Build passed

WojciechNagorski:
I created the repo https://github.com/WojciechNagorski/NativeMemorySample that contains benchmarks that use native memory allocation.

adamsitnik (approved):
@WojciechNagorski LGTM, big thanks for doing this!

WojciechNagorski:
@adamsitnik Thanks for review!

PR #1483, opened 2020-06-19
Add Wasm Tool Chain 

naricc:
This PR adds a tool chain to support Mono/Wasm runs, as well as other changes to Benchmark Dotnet to support WASM.


src/BenchmarkDotNet/Templates/WasmBenchmarkProgram.txt
```diff
@@ -0,0 +1,121 @@
...
+            // before .NET Framework tries to load it for this method
+#if NETFRAMEWORK
+            using(new DirtyAssemblyResolveHelper())
+#endif
+                return AfterAssemblyLoadingAttached(args);
+        }
+
+        private static System.Int32 AfterAssemblyLoadingAttached(System.String[] args)
+        {
+            BenchmarkDotNet.Engines.ConsoleHost host = new BenchmarkDotNet.Engines.ConsoleHost(System.Console.Out, System.IO.TextReader.Null); // this variable name is used by CodeGenerator.GetCoreRtSwitch, do NOT change it
```
naricc, line 27: Input is used for acknowledgements.

naricc: We will need to use a "FakeHost" instead of console host, use it with an ifdef to crate a new one of those insteaad of ConsoleHost

naricc: Introduce a WasmRuntime type to propagate ifdefs; CodeGenerator.cs will have to check this type.

src/BenchmarkDotNet/Environments/BenchmarkEnvironmentInfo.cs
```diff
@@ -30,15 +30,15 @@ public class BenchmarkEnvironmentInfo
         protected BenchmarkEnvironmentInfo()
         {
             Architecture = RuntimeInformation.GetArchitecture();
-            RuntimeVersion = RuntimeInformation.GetRuntimeVersion();
+            RuntimeVersion = "wasm"; // RuntimeInformation.GetRuntimeVersion();
```
naricc, line 33: Need to extend RUntimeInformation.GetRuntimeVersion so that it can detect if we are running wasm.

src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs
```diff
@@ -39,6 +39,9 @@ public static CoreRuntime CreateForNewVersion(string msBuildMoniker, string disp
 
         internal static CoreRuntime GetCurrentVersion()
         {
+            return FromVersion(Version.Parse("1.0.0"));
```
naricc, line 42: RuntimeInformation change should take care of this.

samples/BenchmarkDotNet.Samples/DoNothing.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Threading;
```
naricc, line 1: I created this sample project for debugging and was going to remove it, but it occured to me that it is actually a pretty useful sample to have around. I can delete it if that is preferred though.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public override bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver)
+        {
+            return !System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        }
+
+        [PublicAPI]
+        public static IToolchain From(NetCoreAppSettings settings)
+        {
+            WasmAppSettings wasmAppSettings = (WasmAppSettings)settings;
```
adamsitnik, line 58: the `WasmAppSettings` does not derive from `NetCoreAppSettings` so this cast would always throw. We can just change the parameter type and avoid cast
```suggestion
        public static IToolchain From(WasmAppSettings wasmAppSettings)
        {
```

src/BenchmarkDotNet/Environments/Runtimes/WasmRuntime.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Environments
+{
+    public class WasmRuntime : Runtime, IEquatable<WasmRuntime>
+    {
+        public static readonly WasmRuntime Default = new WasmRuntime("Wasm");
+
+        public WasmRuntime(string name) : base(RuntimeMoniker.Wasm, "wasm", name)
```
adamsitnik, line 10: the moniker should be `net5.0` instead of `wasm` ?

```suggestion
        public WasmRuntime(string name) : base(RuntimeMoniker.Wasm, "net5.0", name)
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -42,6 +42,13 @@ public static bool IsCoreRT
             => ((Environment.Version.Major >= 5) || FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase))
                && string.IsNullOrEmpty(typeof(object).Assembly.Location); // but it's merged to a single .exe and .Location returns null here ;)
 
+        /// <summary>
+        /// "Is the target where we will run the benchmarks WASM?"
+        /// </summary>
+        public static bool IsBenchmarkTargetWasm { get; set; }
```
adamsitnik, line 48: benchmark host can run benchmarks against multiple runtimes by using things like:

```cmd
--runtimes net48 netcoreapp3.1 wasm
```

so this static field would introduce a bug. Please remove it
```suggestion
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -42,6 +42,13 @@ public static bool IsCoreRT
...
+
+        public static bool IsWasm() => IsOSPlatform(OSPlatform.Create("BROWSER")); // { return BenchmarkEnvironmentInfo.RunningWasm; }
```
adamsitnik, line 50: nit: outdated comment

```suggestion
        public static bool IsWasm() => IsOSPlatform(OSPlatform.Create("BROWSER"));
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -144,6 +151,10 @@ internal static string GetRuntimeVersion()
             {
                 return FrameworkVersionHelper.GetFrameworkDescription();
             }
+            else if (IsBenchmarkTargetWasm || IsWasm())
```
adamsitnik, line 154: `IsBenchmarkTargetWasm` should be removed for the reasons explained above

```suggestion
            else if (IsWasm())
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -168,11 +179,15 @@ internal static Runtime GetCurrentRuntime()
                 return MonoRuntime.Default;
             if (IsFullFramework)
                 return ClrRuntime.GetCurrentVersion();
+            if (IsBenchmarkTargetWasm || IsWasm())
```
adamsitnik, line 182: `IsBenchmarkTargetWasm` should be removed for the reasons explained above

```suggestion
            if (IsWasm())
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -168,11 +179,15 @@ internal static Runtime GetCurrentRuntime()
...
+                return WasmRuntime.Default;
             if (IsNetCore)
                 return CoreRuntime.GetCurrentVersion();
             if (IsCoreRT)
                 return CoreRtRuntime.GetCurrentVersion();
 
+
+
             throw new NotSupportedException("Unknown .NET Runtime");
```
adamsitnik, line 191: empty lines...

```suggestion

            throw new NotSupportedException("Unknown .NET Runtime");
```

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -604,5 +606,17 @@ private static void Cleanup(HashSet<string> artifactsToCleanup)
                 }
             }
         }
+
+        private static string removeWasmPrefix(string line)
```
adamsitnik, line 610: Is it possible to disable the printing of "WASM: " in every output line?

If not, please move this logic to [Measurement.Parse](https://github.com/dotnet/BenchmarkDotNet/blob/0445917bf93059f17cb09e7d48cdb5e27a096c37/src/BenchmarkDotNet/Reports/Measurement.cs#L119)

src/BenchmarkDotNet/Running/BuildPartition.cs
```diff
@@ -46,6 +46,8 @@ public BuildPartition(BenchmarkBuildInfo[] benchmarks, IResolver resolver)
             // given job can have CoreRT toolchain set, but Runtime == default ;)
             || (RepresentativeBenchmarkCase.Job.Infrastructure.TryGetToolchain(out var toolchain) && toolchain is CoreRtToolchain);
 
+        public bool IsWasm => Runtime is WasmRuntime;
```
adamsitnik, line 49: we should handle the edge case where some of our users did set Toolchain to Wasm, but forgot to set Runtime to Wasm as well

```suggestion
        public bool IsWasm => Runtime is WasmRuntime;
            // given job can have Wasm toolchain set, but Runtime == default ;)
            || (RepresentativeBenchmarkCase.Job.Infrastructure.TryGetToolchain(out var toolchain) && toolchain is WasmToolchain);
```

src/BenchmarkDotNet/Toolchains/GeneratorBase.cs
```diff
@@ -119,7 +119,16 @@ private ArtifactsPaths GetArtifactsPaths(BuildPartition buildPartition, string r
             string programName = buildPartition.ProgramName;
             string buildArtifactsDirectoryPath = GetBuildArtifactsDirectoryPath(buildPartition, programName);
             string binariesDirectoryPath = GetBinariesDirectoryPath(buildArtifactsDirectoryPath, buildPartition.BuildConfiguration);
-            string executablePath = Path.Combine(binariesDirectoryPath, $"{programName}{GetExecutableExtension()}");
+
+            string executablePath;
```
adamsitnik, line 123: please follow the existing pattern: introduce a new virtual method `GetExecutablePath` with `Path.Combine(binariesDirectoryPath, $"{programName}{GetExecutableExtension()}")` as default implementation and override it in `WasmGenerator` and return `Path.Combine(binariesDirectoryPath, "run-v8.sh")` there

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
```diff
@@ -0,0 +1,146 @@
...
+
+{
+    [PublicAPI]
+    public class WasmExecutor : IExecutor
+    {
+        public WasmExecutor(string customDotNetCliPath) => CustomDotNetCliPath = customDotNetCliPath;
+
+        private string CustomDotNetCliPath { get; }
+
+        private string JavaScriptEngine = "v8";
```
adamsitnik, line 28: this field is not used anywhere. Should it be removed or should it be populated and used to allow for customization?

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
```diff
@@ -0,0 +1,146 @@
...
+                                                        bool redirectStandardInput = false)
+        {
+            const string dotnetMultiLevelLookupEnvVarName = "DOTNET_MULTILEVEL_LOOKUP";
+
+            ProcessStartInfo startInfo = new ProcessStartInfo
+            {
+                FileName = "v8",
+                WorkingDirectory = workingDirectory,
+                Arguments = $"--expose_wasm runtime.js  -- --enable-gc --run {programName}.dll {benchmarkId} ",
+                UseShellExecute = false,
```
adamsitnik, line 129: it looks that we could just extend the `Executor.CreateStartInfo` with handling of the `Wasm` runtime and avoid having a new dedicated `WasmExecutor` type

https://github.com/dotnet/BenchmarkDotNet/blob/0b0ede9565758d16f774c68bb02b8fbd7feea576/src/BenchmarkDotNet/Toolchains/Executor.cs#L104-L115

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -356,6 +358,19 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
...
+                                                                 "net5.0",
+                                                                 options.CliPath.FullName,
+                                                                 null,
+                                                                 null,
+                                                                 options.WasmRuntimePackPath,
+                                                                 options.WasmAppBuilderPath,
+                                                                 options.WasmMainJS,
+                                                                 timeOut ?? NetCoreAppSettings.DefaultBuildTimeout);
+
+                        RuntimeInformation.IsBenchmarkTargetWasm = true;
```
adamsitnik, line 372: `IsBenchmarkTargetWasm` should be removed
```suggestion
```

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -0,0 +1,43 @@
+﻿<Project Sdk="$SDKNAME$" DefaultTargets="BuildApp">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <OutputPath>bin</OutputPath>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+    <TargetFramework>$TFM$</TargetFramework>
+    <TargetArchitecture>wasm</TargetArchitecture>
+    <TargetOS>Browser</TargetOS>
+    <RuntimePackDir>$RUNTIMEPACKDIR$</RuntimePackDir>
+    <AppDir>$(MSBuildThisFileDirectory)\bin\net5.0\browser-wasm\publish</AppDir>
```
adamsitnik, line 10: if `$TFM$` is `net5.0`, we should use it everywhere (to allow for using `net6.0` in the future)

```suggestion
    <AppDir>$(MSBuildThisFileDirectory)\bin\$TFM$\browser-wasm\publish</AppDir>
```

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -0,0 +1,43 @@
...
+
+
+  <UsingTask TaskName="WasmAppBuilder" 
+      AssemblyFile="$WASMAPPBUILDERASSEMBLY$"/>
+
+  <Target Name="BuildApp" DependsOnTargets="Publish">
+    <ItemGroup>
+      <AssemblySearchPaths Include="$(AppDir)"/>
+      <AssemblySearchPaths Include="$(RuntimePackDir)\native"/>
+      <AssemblySearchPaths Include="$(RuntimePackDir)\lib\net5.0"/>
```
adamsitnik, line 24:
```suggestion
      <AssemblySearchPaths Include="$(RuntimePackDir)\lib\$TFM$"/>
```

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -0,0 +1,43 @@
...
+    </ItemGroup>    
+    <WasmAppBuilder
+      AppDir="$(AppDir)/output"
+      MicrosoftNetCoreAppRuntimePackDir="$(RuntimePackDir)"
+      MainAssembly="bin/net5.0/browser-wasm/$(AssemblyName).dll"
```
adamsitnik, line 29:
```suggestion
      MainAssembly="bin/$TFM$/browser-wasm/$(AssemblyName).dll"
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
```diff
@@ -0,0 +1,146 @@
...
+                                                        string programName,
+                                                        string benchmarkId,
+                                                        IReadOnlyList<EnvironmentVariable> environmentVariables = null,
+                                                        bool redirectStandardInput = false)
+        {
+            const string dotnetMultiLevelLookupEnvVarName = "DOTNET_MULTILEVEL_LOOKUP";
+
+            ProcessStartInfo startInfo = new ProcessStartInfo
+            {
+                FileName = "v8",
```
adamsitnik, line 126: `CommandLineOptions.WasmJavascriptEnginePath` allows for specifing a custom value, but anyway "v8" is always used here. We should either not allow for customization or use the provided value here

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+                   .Replace("$MAINJS$", MainJS)
+                   .ToString();
+
+                File.WriteAllText(artifactsPaths.ProjectFilePath, content);
+            }
+
+        }
+
+        protected override string GetBinariesDirectoryPath(string buildArtifactsDirectoryPath, string configuration)
+    => Path.Combine(buildArtifactsDirectoryPath, "bin/net5.0/browser-wasm/publish/output");
```
adamsitnik, line 74: it should not be hardcoded to net5.0 and we should use proper path combining (even if currently we plan to use it only on Linux)

```suggestion
           => Path.Combine(buildArtifactsDirectoryPath, "bin", base.TargetFrameworkMoniker, "browser-wasm", "publish", "output");
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+                             )
+            : base(targetFrameworkMoniker,
+                   cliPath,
+                   packagesPath,
+                   runtimeFrameworkVersion)
+        {
+            RuntimePackPath = runtimePackPath;
+            WasmAppBuilderAssembly = wasmAppBuilderAssembly;
+            MainJS = mainJS;
+
```
adamsitnik, line 42: empty line

```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+using BenchmarkDotNet.Toolchains.CsProj;
+using BenchmarkDotNet.Toolchains.DotNetCli;
+
+
+
+namespace BenchmarkDotNet.Toolchains.MonoWasm
+{
+    public class WasmGenerator : CsProjGenerator
+
+    {
```
adamsitnik, line 21: empty lines...

```suggestion
using BenchmarkDotNet.Toolchains.DotNetCli;

namespace BenchmarkDotNet.Toolchains.MonoWasm
{
    public class WasmGenerator : CsProjGenerator
    {
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+                   .Replace("$CONFIGURATIONNAME$", buildPartition.BuildConfiguration)
+                   .Replace("$SDKNAME$", sdkName)
+                   .Replace("$RUNTIMEPACKDIR$", RuntimePackPath)
+                   .Replace("$WASMAPPBUILDERASSEMBLY$", WasmAppBuilderAssembly)
+                   .Replace("$MAINJS$", MainJS)
+                   .ToString();
+
+                File.WriteAllText(artifactsPaths.ProjectFilePath, content);
+            }
+
```
adamsitnik, line 70:
```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                var (customProperties, sdkName) = GetSettingsThatNeedsToBeCopied(file, projectFile);
+
+                string content = new StringBuilder(ResourceHelper.LoadTemplate("WasmCsProj.txt"))
+                   .Replace("$PLATFORM$", buildPartition.Platform.ToConfig())
+                   .Replace("$CODEFILENAME$", Path.GetFileName(artifactsPaths.ProgramCodePath))
+                   .Replace("$CSPROJPATH$", projectFile.FullName)
+                   .Replace("$TFM$", TargetFrameworkMoniker)
+                   .Replace("$PROGRAMNAME$", artifactsPaths.ProgramName)
+                   .Replace("$RUNTIMESETTINGS$", GetRuntimeSettings(benchmark.Job.Environment.Gc, buildPartition.Resolver))
```
adamsitnik, line 59: can we customize GC settings for Wasm? If not, the `$RUNTIMESETTINGS$` logic should be removed

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+                File.WriteAllText(artifactsPaths.ProjectFilePath, content);
+            }
+
+        }
+
+        protected override string GetBinariesDirectoryPath(string buildArtifactsDirectoryPath, string configuration)
+    => Path.Combine(buildArtifactsDirectoryPath, "bin/net5.0/browser-wasm/publish/output");
+
+        private static string GetPackagesDirectoryPath(bool useTempFolderForRestore, string packagesRestorePath)
+    => packagesRestorePath ?? (useTempFolderForRestore ? Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString()) : null);
```
adamsitnik, line 77: it looks like copy-pasted from the `CoreRtGenerator`:

https://github.com/dotnet/BenchmarkDotNet/blob/3223c94a92050147a02482a3810860703f0c5171/src/BenchmarkDotNet/Toolchains/CoreRt/Generator.cs#L80-L85

is this method needed? I can't see any usages of it

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -0,0 +1,76 @@
...
+        private WasmToolChain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string runtimeJavaScriptPath)
+            : base(name, generator, builder, executor)
+        {
+        }
+
+        public WasmToolChain(string name,
+                             string targetFrameworkMoniker,
+                             string cliPath,
+                             string packagesPath,
+                             string runtimeFrameworkVersion,
```
adamsitnik, line 28: if `runtimeFrameworkVersion` makes no sense for `Wasm`, we can just remove it and make `WasmGenerator` pass `null` to base ctor (the same applies to other unused settings)

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -356,6 +358,19 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
                     builder.TargetFrameworkMoniker(runtime.MsBuildMoniker);
 
                     return baseJob.WithRuntime(runtime).WithToolchain(builder.ToToolchain());
+                case RuntimeMoniker.Wasm:
+                        IToolchain toolChain = new WasmToolChain("Wasm",
+                                                                 "net5.0",
+                                                                 options.CliPath.FullName,
+                                                                 null,
```
adamsitnik, line 365: using the provided packages restore path is important from CI perspective
```suggestion
                                                                 options.RestorePath?.FullName,
```

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -356,6 +358,19 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
...
+                                                                 null,
+                                                                 options.WasmRuntimePackPath,
+                                                                 options.WasmAppBuilderPath,
+                                                                 options.WasmMainJS,
+                                                                 timeOut ?? NetCoreAppSettings.DefaultBuildTimeout);
```
adamsitnik, line 370: it would be great if you could use named arguments here as there are many of them and it would help us avoid bugs

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppSettings.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using BenchmarkDotNet.Toolchains.DotNetCli;
+using JetBrains.Annotations;
+
+
+namespace BenchmarkDotNet.Toolchains.MonoWasm
+{
+    [PublicAPI]
+    public class WasmAppSettings : NetCoreAppSettings
```
adamsitnik, line 9: I think that instead of introducing this type and `WasmSettings`, you can revert all the changes done to `NetCoreAppSettings`, remove `WasmAppSettings`, and expect all the necessary methods to just accept an instance of `NetCoreAppSettings` and an instance of `WasmSettings`. It should simplify the code

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmSettings.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class WasmSettings
+    {
+        public string WasmMainJS { get; }
+
+        public string WasmRuntimePack { get; }
+
+        public string WasmAppBuilder { get; }
+
+        public WasmSettings(string wasmMainJS, string wasmRuntimePack, string wasmAppBuilder)
```
adamsitnik, line 12: what about "v8"?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
naricc: Its not possible to remove with out changes to WASM. Those changes might happen, but I am not sure when. So I will move the logic to Measurement.Parse.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -356,6 +358,19 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
                     builder.TargetFrameworkMoniker(runtime.MsBuildMoniker);
 
                     return baseJob.WithRuntime(runtime).WithToolchain(builder.ToToolchain());
+                case RuntimeMoniker.Wasm:
```
akoeplinger, line 361: ideally this wouldn't be needed and we could just use `RuntimeMoniker.NetCoreApp50` with a `runtimeId=browser-wasm`

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
akoeplinger: @naricc actually this happened a few hours ago: https://github.com/dotnet/runtime/pull/38330 😄 . It should be available shortly in the SDK build from https://github.com/dotnet/installer#installers-and-binaries

I agree with Adam that we should remove this here.

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -0,0 +1,43 @@
...
+    <TargetOS>Browser</TargetOS>
+    <RuntimePackDir>$RUNTIMEPACKDIR$</RuntimePackDir>
+    <AppDir>$(MSBuildThisFileDirectory)\bin\net5.0\browser-wasm\publish</AppDir>
+    <AssemblyName>$PROGRAMNAME$</AssemblyName>
+    <RuntimeIdentifier>browser-wasm</RuntimeIdentifier>
+    $COPIEDSETTINGS$
+  </PropertyGroup>
+
+
+  <UsingTask TaskName="WasmAppBuilder"
```
akoeplinger, line 17: we discussed this in our wasm sync with @lewing today and think we should copy the WasmAppBuilder logic into BDN for now until we can push a proper version up into the .NET SDK. I'll look at that.

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -0,0 +1,43 @@
+﻿<Project Sdk="$SDKNAME$" DefaultTargets="BuildApp">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <OutputPath>bin</OutputPath>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+    <TargetFramework>$TFM$</TargetFramework>
+    <TargetArchitecture>wasm</TargetArchitecture>
+    <TargetOS>Browser</TargetOS>
+    <RuntimePackDir>$RUNTIMEPACKDIR$</RuntimePackDir>
```
akoeplinger, line 9: this doesn't need to be set since doing a `dotnet publish` on the csproj will automatically pull down the runtime pack and put all assets into the publish directory and we can take everything from there.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
naricc: I don't think we can do this now that I added the javaScriptEngineArguments option, since some arguments have to go to the engine, and others to the runtime. Or, at least, it doesn't look like the existing pattern?

naricc:
@adamsitnik Ok, I think I've resolved all the comments. That took a little while :) Please let me know if you have any further comments.

src/BenchmarkDotNet/Toolchains/MonoWasm/PInvokeTableGenerator.cs
```diff
@@ -0,0 +1,156 @@
+// -*- indent-tabs-mode: nil -*-
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
SamMonoRT: Not sure if BDN project is following the pattern the dotnet\runtime repo is, but they removed the third line from this versioning/header comments.

naricc: I've already deleted this file; it was part of the WasmAppBuilder but turned out not to be needed.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -21,6 +21,7 @@
 using BenchmarkDotNet.Toolchains.Results;
 using BenchmarkDotNet.Validators;
 using Perfolizer.Horology;
+using BenchmarkDotNet.Toolchains.MonoWasm;
```
akoeplinger, line 24: nit: looks like this isn't needed in this file.

src/BenchmarkDotNet/Engines/NoAcknowledgementConsoleHost.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.IO;
+using BenchmarkDotNet.Validators;
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Engines
+{   /* This class is used by wasm because wasm cannot read from the console.
```
akoeplinger, line 7: nit: add newline

```suggestion
{
    /* This class is used by wasm because wasm cannot read from the console.
```

src/BenchmarkDotNet/Templates/WasmCsProj.txt
akoeplinger: @naricc it doesn't look like Adam's comment was resolved here.

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="$SDKNAME$" DefaultTargets="Publish">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <OutputPath>bin</OutputPath>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+    <TargetFramework>$TFM$</TargetFramework>
+    <TargetArchitecture>wasm</TargetArchitecture>
+    <TargetOS>Browser</TargetOS>
```
akoeplinger, line 8: you can remove these two, they are only set in the dotnet/runtime build system.

```suggestion
```

src/BenchmarkDotNet/Toolchains/DotNetCli/NetCoreAppSettings.cs
```diff
@@ -67,6 +69,7 @@ public class NetCoreAppSettings
 
         public string CustomDotNetCliPath { get; }
 
+
```
akoeplinger, line 72: nit: unnecessary whitespace

src/BenchmarkDotNet/Toolchains/GeneratorBase.cs
```diff
@@ -119,7 +121,18 @@ private ArtifactsPaths GetArtifactsPaths(BuildPartition buildPartition, string r
...
+            if (buildPartition.IsWasm)
+            {
+                executablePath = Path.Combine(binariesDirectoryPath, "run-v8.sh");
+            }
+            else
+            {
+                executablePath = Path.Combine(binariesDirectoryPath, $"{programName}{GetExecutableExtension()}");
+            }
+
+            executablePath = GetExecutablePath(binariesDirectoryPath, programName);
```
akoeplinger, line 135: the whole block above isn't necessary now that you have `GetExecutablePath`:

```suggestion
            string executablePath = GetExecutablePath(binariesDirectoryPath, programName);
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+    private readonly WasmSettings WasmSettings;
+
+    private readonly string TargetFrameworkMoniker;
+
+  
+
+    private Dictionary<string, Assembly> _assemblies;
+    private AssemblyResolver _resolver;
+
+
```
akoeplinger, line 29: remove the unnecessary newlines:

```suggestion

    private Dictionary<string, Assembly> _assemblies;
    private AssemblyResolver _resolver;

```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+        if (!File.Exists(mainAssemblyPath))
+            throw new ArgumentException($"File MainAssembly='{mainAssemblyPath}' doesn't exist.");
+        if (!File.Exists(WasmSettings.WasmMainJS))
+            throw new ArgumentException($"File MainJS='{WasmSettings.WasmMainJS}' doesn't exist.");
+
+        var paths = new List<string>();
+        _assemblies = new Dictionary<string, Assembly>();
+
+        // Collect and load assemblies used by the app
+        foreach (string dir in assemblySearchPaths)
```
akoeplinger, line 59: we only have one entry in assemblySearchPaths so we don't need this

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            if (!Directory.Exists(dir))
+                throw new ArgumentException($"Directory '{dir}' doesn't exist or not a directory.");
+            paths.Add(dir);
+        }
+        _resolver = new AssemblyResolver(paths);
+        var mlc = new MetadataLoadContext(_resolver, "System.Private.CoreLib");
+
+        Assembly mainAssembly = mlc.LoadFromAssemblyPath (mainAssemblyPath);
+        Add(mlc, mainAssembly);
```
akoeplinger, line 69: you can remove all of the MetadataLoadContext and AssemblyResolver related logic, we don't need it for BDN.

it's enough to simply copy all *.dll files to `output/managed`, e.g. like this:

```suggestion
        _resolver = new AssemblyResolver(paths);
        var mlc = new MetadataLoadContext(_resolver, "System.Private.CoreLib");

        Assembly mainAssembly = mlc.LoadFromAssemblyPath (mainAssemblyPath);
        Add(mlc, mainAssembly);
        string[] _assemblies = Directory.GetFiles(appDir, "*.dll");
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        using (var sw = File.CreateText(Path.Combine(outputDir, "run-v8.sh")))
+        {
+            sw.WriteLine("v8 --expose_wasm runtime.js -- --run " + Path.GetFileName(mainAssemblyPath) + " $*");
+        }
+
+        return true;
+    }
+
+    private void Add(MetadataLoadContext mlc, Assembly assembly)
```
akoeplinger, line 124: you can remove this

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+                Add(mlc, refAssembly);
+            }
+            catch (System.IO.FileNotFoundException)
+            {
+                Console.Error.WriteLine($"WARNING: Could not load {aname.Name}");
+            }
+        }
+    }
+
+    private Dictionary<string, List<string>> copySystemPrivateCoreLib(string appDir, string supportFilesDir)
```
akoeplinger, line 145: I don't think this is needed or does BDN try to open System.Private.CoreLib.dll from the filesystem?

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+            sw.WriteLine("\tenable_debugging: 0,");
+            sw.WriteLine("\tassembly_list: [");
+            foreach (var assembly in _assemblies.Values)
+            {
+                sw.Write("\t\t\"" + Path.GetFileName(assembly.Location) + "\"");
+                sw.WriteLine(",");
+            }
+            sw.WriteLine ("\t],");
+            sw.WriteLine("\tfiles_to_map: [");
+            foreach (KeyValuePair<string, List<string>> keyValuePair in filesToMap)
```
akoeplinger, line 99: I'm pretty sure we don't need to have any entries in `files_to_map` so this can be removed.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+        files.Add(Path.GetFileName(targetPath));
+        fileMap[directory] = files;
+
+        return fileMap;
+    }
+
+}
+
+
+internal class AssemblyResolver : MetadataAssemblyResolver
```
akoeplinger, line 179: can be removed

src/BenchmarkDotNet/BenchmarkDotNet.csproj
```diff
@@ -28,6 +28,7 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.10.0" />
     <PackageReference Include="Microsoft.Diagnostics.NETCore.Client" Version="0.2.61701" />
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.49" PrivateAssets="contentfiles;analyzers" />
+    <PackageReference Include="System.Reflection.MetadataLoadContext" Version="4.7.1" />
```
akoeplinger, line 31: once the MetadataLoadContext logic is removed from WasmAppBuilder you can get rid of this one too.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+        // Create app
+        Directory.CreateDirectory(outputDir);
+        Directory.CreateDirectory(Path.Combine(outputDir, "managed"));
+        foreach (var assembly in _assemblies.Values)
+            File.Copy(assembly.Location, Path.Combine(outputDir, "managed", Path.GetFileName(assembly.Location)), true);
+        foreach (var f in new string[] { "dotnet.wasm", "dotnet.js" })
+            File.Copy(Path.Combine(appDir, f), Path.Combine(outputDir, f), true);
+        File.Copy(WasmSettings.WasmMainJS, Path.Combine(outputDir, "runtime.js"),  true);
+
+        string supportFilesDir = Path.Combine(outputDir, "supportFiles");
```
akoeplinger, line 80: can almost certainly be removed as well

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmBuilder.cs
```diff
@@ -0,0 +1,35 @@
...
+        public override BuildResult Build(GenerateResult generateResult, BuildPartition buildPartition, ILogger logger)
+        {
+            BuildResult buildResult = base.Build(generateResult, buildPartition, logger);
+            WasmAppBuilder.BuildApp(buildPartition.ProgramName, generateResult.ArtifactsPaths.BuildArtifactsDirectoryPath);
+
+            return buildResult;
+        }
+
+
+    }
```
akoeplinger, line 34: remove unecessary newlines:

```suggestion

        public override BuildResult Build(GenerateResult generateResult, BuildPartition buildPartition, ILogger logger)
        {
            BuildResult buildResult = base.Build(generateResult, buildPartition, logger);
            WasmAppBuilder.BuildApp(buildPartition.ProgramName, generateResult.ArtifactsPaths.BuildArtifactsDirectoryPath);

            return buildResult;
        }
    }
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
```diff
@@ -0,0 +1,167 @@
...
+using BenchmarkDotNet.Helpers;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.Parameters;
+using BenchmarkDotNet.Toolchains.Results;
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Toolchains.MonoWasm
+
```
akoeplinger, line 19:
```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,65 @@
...
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Portability;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.CsProj;
+using BenchmarkDotNet.Toolchains.DotNetCli;
+
+namespace BenchmarkDotNet.Toolchains.MonoWasm
+{
+    public class WasmGenerator : CsProjGenerator
+
```
akoeplinger, line 18:
```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
akoeplinger: @naricc this also wasn't addressed.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -0,0 +1,63 @@
...
+                                   cliPath,
+                                   packagesPath,
+                                   wasmSettings.WasmMainJS),
+                 new WasmBuilder(targetFrameworkMoniker, wasmSettings, cliPath, timeout),
+                 new WasmExecutor(wasmSettings.WasmMainJS, wasmSettings.WasmJavaScriptEngine,  wasmSettings.WasmJavaScriptEngineArguments)
+                 )
+        {
+        }
+
+
```
akoeplinger, line 38: one newline is enough:

```suggestion

```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -0,0 +1,76 @@
...
+                                   wasmAppBuilderAssembly,
+                                   mainJS),
+                 new DotNetCliBuilder(targetFrameworkMoniker, cliPath, timeout),
+                 new WasmExecutor(mainJS)
+                 )
+        {
+        }
+
+
+        internal string RuntimeJavaScripthPath { get; }
```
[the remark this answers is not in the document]
akoeplinger: @naricc another unaddressed comment.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
naricc: It does, which is why we added it to the file system in the first place. I never did go and understand why it tries to do that though; maybe fixable on the BDN end?

naricc: It turns out not to be needed any more due to other changes, so I got rid of it.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+                sw.Write("\t\t\"" + Path.GetFileName(assembly) + "\"");
+                sw.WriteLine(",");
+            }
+            sw.WriteLine ("\t],");
+            sw.WriteLine("\tfiles_to_map: [],");
+
+            sw.WriteLine ("}");
+        }
+
+        using (var sw = File.CreateText(Path.Combine(outputDir, "run-v8.sh")))
```
akoeplinger, line 71: if I'm reading the PR correctly you don't need this file since you call v8 explicitly with the `WasmJavaScriptEngine` etc logic in the WasmExecutor?

Alternatively if you keep this file you might be able to go with Adam's suggestion in https://github.com/dotnet/BenchmarkDotNet/pull/1483/files#r444980216 and remove the need for a special WasmExecutor as you could instead just start the `run-v8.sh` script as the "executable".

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
naricc: @adamsitnik Is that actually a reasonable thing to do? There doesn't seem to be any where else where this is conditions on runtimeID, and I am not sure how easy it would be to get the right information in the right place?

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmAppBuilder.cs
naricc: I think leaving it in the C# code is cleaner; I don't want to start hiding stuff in a shell script. So I did it that way and eliminated run-v8.sh

SamMonoRT:
@AndreyAkinshin fyi - we are just waiting on @adamsitnik to sign off and merge this. Not sure if there was any other validation required here.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class WasmBuilder : DotNetCliBuilder
+    {
+        private WasmAppBuilder WasmAppBuilder;
+
+        public WasmBuilder(string targetFrameworkMoniker, WasmSettings wasmSettings, string customDotNetCliPath = null, TimeSpan? timeout = null)
+            :base (targetFrameworkMoniker, customDotNetCliPath, timeout)
+        {
+            List<string> extraAssemblies = new List<string> { };
+
```
adamsitnik, line 20: this variable is not used anywhere

should it be consumed or removed?

```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+            WasmAppBuilder = new WasmAppBuilder(wasmSettings, targetFrameworkMoniker);
+        }
+
```
adamsitnik, line 23: nit: extra line

```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
adamsitnik: to achieve that you could move all the fields from `WasmSettings` type to `WasmRuntime` type and create such runtime when parsing the args in ConfigParser:

https://github.com/dotnet/BenchmarkDotNet/blob/50e5157d0fc5c67e8982c9ff4817c511684dae49/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs#L362-L368

we can leave it as it is and refactor in a separate PR

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -356,6 +358,22 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
...
+                                                                 );
+
+                    IToolchain toolChain = new WasmToolChain(name: "Wasm",
+                                                             targetFrameworkMoniker: wasmRuntime.MsBuildMoniker,
+                                                             cliPath: options.CliPath.FullName,
+                                                             packagesPath: options.RestorePath?.FullName,
+                                                             wasmSettings: wasmSettings,
+                                                             timeout: timeOut ?? NetCoreAppSettings.DefaultBuildTimeout);
+
+                        return baseJob.WithRuntime(runtimeMoniker.GetRuntime()).WithToolchain(toolChain);
```
adamsitnik, line 376: we already have the runtime in `wasmRuntime` local variable:

```suggestion
                        return baseJob.WithRuntime(wasmRuntime).WithToolchain(toolChain);
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,64 @@
...
+                   packagesPath,
+                   null)
+        {
+            MainJS = mainJS;
+        }
+
+        protected override void GenerateProject(BuildPartition buildPartition, ArtifactsPaths artifactsPaths, ILogger logger)
+        {
+                BenchmarkCase benchmark = buildPartition.RepresentativeBenchmarkCase;
+            var projectFile = GetProjectFilePath(benchmark.Descriptor.Type, logger);
```
adamsitnik, line 37:
```suggestion
            BenchmarkCase benchmark = buildPartition.RepresentativeBenchmarkCase;
            var projectFile = GetProjectFilePath(benchmark.Descriptor.Type, logger);
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                File.WriteAllText(artifactsPaths.ProjectFilePath, content);
+            }
+        }
+
+        protected override string  GetExecutablePath(string binariesDirectoryPath, string programName) => Path.Combine(binariesDirectoryPath, "runtime.js");
+
+
+        protected override string GetBinariesDirectoryPath(string buildArtifactsDirectoryPath, string configuration)
+    => Path.Combine(buildArtifactsDirectoryPath, $"bin/{TargetFrameworkMoniker}/browser-wasm/publish/output");
```
adamsitnik, line 61: formatting and OS specific path combining

```suggestion

        protected override string GetBinariesDirectoryPath(string buildArtifactsDirectoryPath, string configuration)
            => Path.Combine(buildArtifactsDirectoryPath, "bin", TargetFrameworkMoniker, "browser-wasm", "publish", "output");
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -0,0 +1,64 @@
...
+
```
adamsitnik, line 62: extra line

```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmSettings.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace BenchmarkDotNet.Toolchains.MonoWasm
+{
+    public class WasmSettings
+    {
+        public string WasmMainJS { get; }
+
+        public string WasmJavaScriptEngine { get; }
+
+        public string WasmJavaScriptEngineArguments { get;  }
+
```
adamsitnik, line 11:
```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -0,0 +1,76 @@
...
+
+
+namespace BenchmarkDotNet.Toolchains.MonoWasm
+{
+    [PublicAPI]
+    public class WasmToolChain : Toolchain
+    {
+        [PublicAPI] public static readonly IToolchain NetCoreApp50Wasm = From(WasmAppSettings.NetCoreApp50Wasm);
+
+        private WasmToolChain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string runtimeJavaScriptPath)
```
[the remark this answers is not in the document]
adamsitnik: the `runtimeJavaScriptPath ` argument is still not used

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -0,0 +1,59 @@
...
+                                                        netCoreAppSettings.CustomDotNetCliPath,
+                                                        netCoreAppSettings.PackagesPath,
+                                                        wasmSettings.WasmMainJS),
+                                      new DotNetCliBuilder(netCoreAppSettings.TargetFrameworkMoniker,
+                                                           netCoreAppSettings.CustomDotNetCliPath,
+                                                           netCoreAppSettings.Timeout),
+                                      new WasmExecutor(wasmSettings.WasmMainJS, wasmSettings.WasmJavaScriptEngine, wasmSettings.WasmJavaScriptEngineArguments),
+                netCoreAppSettings.CustomDotNetCliPath);
+        }
+
```
adamsitnik, line 57:
```suggestion
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -0,0 +1,59 @@
...
+using BenchmarkDotNet.Characteristics;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Loggers;
+
+namespace BenchmarkDotNet.Toolchains.MonoWasm
+{
+    [PublicAPI]
+    public class WasmToolChain : Toolchain
+    {
+        [PublicAPI] public static readonly IToolchain NetCoreApp50Wasm = From(NetCoreAppSettings.NetCoreApp50, new WasmSettings(null, null, null));
```
adamsitnik, line 14: will it work if we don't specify all the paths? If not, this property should be removed

```suggestion
        [PublicAPI] public static readonly IToolchain NetCoreApp50Wasm = From(NetCoreAppSettings.NetCoreApp50, new WasmSettings(null, null, null));
```

adamsitnik:
@naricc as soon as [this](https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/builds/34109217) build goes green (should be 30-40 minutes from now) BDN CI is going to publish `0.12.1.1387` package to our CI feed. After that you should be able to consume it.

### nuget.config

```xml
<packageSources>
</packageSources>
```

### .csproj

```xml
<ItemGroup>
    <PackageReference Include="BenchmarkDotNet" Version="0.12.1.1387" />
</ItemGroup>
```

@naricc could you please provide a short description of how to use this toolchain? (where do I take all the mandatory .js files from? ) thanks!

adamsitnik:
the follow-up PR: https://github.com/dotnet/BenchmarkDotNet/pull/1498

naricc:
@adamsitnik 
So, first you need get the prereqs for mono-wasm, and then build it.  This document is about how to build/run the libraries tests for wasm, but the prereqs and build steps are the same: https://github.com/dotnet/runtime/blob/master/docs/workflow/testing/libraries/testing-wasm.md

The wasmMainJS is in `<dotnet/runtimeroot>/src/mono/wasm/runtime-test.js`; it is not generated or altered as part of the build, it is just a static file. Someday it will probably be packaged up, but the plan for how to do that is not clear yet.

If v8 is on your path, it will be used by default and there is no need to specify a Java script engnine.

For reference, hhere is the default command line I h ave been using for basic "does it run" tests:

```
--filter "IntroBasic" --cli /Users/naricc/workspace/runtime-pristine-2/dotnet.sh --runtimes wasm   --wasmMainJS /Users/naricc/workspace/runtime-pristine-2/src/mono/wasm/runtime-test.js --keepFiles
```

Issue #1448, opened 2020-05-02
NRE in MultimodalDistributionAnalyzer


MaceWindu:
version: 0.12.1

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at BenchmarkDotNet.Analysers.MultimodalDistributionAnalyzer.<AnalyseReport>d__4.MoveNext()
   at BenchmarkDotNet.Analysers.AnalyserBase.<Analyse>d__2.MoveNext()
   at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
   at System.Linq.Enumerable.<DistinctIterator>d__64`1.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.PrintSummary(ILogger logger, ImmutableConfig config, Summary summary)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Run(BenchmarkRunInfo[] benchmarkRunInfos)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.RunWithDirtyAssemblyResolveHelper(String[] args, IConfig config)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.Run(String[] args, IConfig config)
   at LinqToDB.Benchmarks.Program.Main(String[] args)
```

Unfortunatelly, I cannot provide more details right now, as restarting benchmarks resulted in successful execution.

MaceWindu:
Tried to debug it - it fails due to `summary.Style` being `null`

AndreyAkinshin:
Fixed by @suslovk in #1488

PR #1488, opened 2020-06-27
Safe access to CultureInfo

suslovk:
Resolver #1448


AndreyAkinshin:
@suslovk thanks for the fix!

PR #1489, opened 2020-06-29
enforce Deterministic builds for auto-generated .NET Core projects

adamsitnik:
We are already doing this for .NET Framework|Mono:

https://github.com/dotnet/BenchmarkDotNet/blob/9ac7770682a45afb6cf4ec353f9fa3c69ece67ce/src/BenchmarkDotNet/Toolchains/Roslyn/Builder.cs#L42

I don't know why we were not doing this for .NET Core so far, but I think that we should enforce that.


adamsitnik:
as @eerhardt has pointed in an offline conversation, it's true by default for .NET Core projects:

https://github.com/dotnet/sdk/blob/b910664dfd9f4c723606e8caf18b4a69d16d2c71/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L34

but with this PR we don't need to rely on defaults and we simply enforce it

AndreyAkinshin:
@adamsitnik could you please resolve the conflicts?

adamsitnik:
> could you please resolve the conflicts?

@AndreyAkinshin done

PR #1490, opened 2020-06-29
be less strict about verifying the Acknowledgment

### LINKED ISSUE #1136: benchmark / beforeEverything fails when running in docker-win
I am trying to run a benchmark inside a docker-win container using hyper-v. When I run the benchmark with a default config I see an error like:

```
BeforeEverythingElse
exitcode != 0

no results were gathered
```

when I run using the debugInProcess config - the benchmark runs.

I tried the advice here:
https://benchmarkdotnet.org/articles/guides/troubleshooting.html#in-the-same-process

but the execution never seems to make it to my benchmark to even enter the while loop.

debugging the resulting genereted.exe using MDbg.exe just seem to hang forever at the `BeforeAnythingElse` step.

Is there anything that would prevent the benchmarks from running in docker-win?

### LINKED ISSUE #1487: Benchmarks do not execute (as service process / VSTS agent) on non US systems
# What went wrong

When .NET Framework benchmarks are executed from a VSTS agent, running on a _**non US English**_ Windows installation, the benchmark fails with 

```
// **************************
// Benchmark: BlaBla
// *** Execute ***
// Launch: 1 / 1
// Execute: D:\agent\perf-win01\_work\3\s\src\blabla\bin\Release\net48\DefaultJob.exe --benchmarkName "BlaBla" --job "Default" --benchmarkId 0 in 
// BeforeAnythingElse
ExitCode != 0
// Benchmark Process 4344 has exited with code 255
No more Benchmark runs will be launched as NO measurements were obtained from the previous run!
```

```
System.NotSupportedException: Unknown Acknowledgment: ´╗┐Acknowledgment
   at BenchmarkDotNet.Engines.ConsoleHost.SendSignal(HostSignal hostSignal)
   at BenchmarkDotNet.Autogenerated.UniqueProgramName.AfterAssemblyLoadingAttached(String[] args)
   at BenchmarkDotNet.Autogenerated.UniqueProgramName.Main(String[] args)
```

_This problem does not occur for .NET Core, only for .NET Framework!_

Looking at a crashdump, it fails in [`public void SendSignal(HostSignal hostSignal)`](https://github.com/dotnet/BenchmarkDotNet/blob/0445917bf93059f17cb09e7d48cdb5e27a096c37/src/BenchmarkDotNet/Engines/ConsoleHost.cs#L25) at https://github.com/dotnet/BenchmarkDotNet/blob/0445917bf93059f17cb09e7d48cdb5e27a096c37/src/BenchmarkDotNet/Engines/ConsoleHost.cs#L33-L34

The `NotSupportedException` is thrown because it receives an `acknowledgment` string containing garbage at the start: **"´╗┐Acknowledgment" (bytes `B4 00 57 25 10 25 ...`)**

# Possible reason

I expect the reason to be a missmatch in the encoding / codepages used in the outer process _UTF-8 / 1200_ and in the inner job process _Western European (DOS) / 850_.

# Possible fix

In both [`SynchronousProcessOutputLoggerWithDiagnoser`](https://github.com/dotnet/BenchmarkDotNet/blob/2ba30330ec5ec8a9abe284c927277a6fe0fc5a0c/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs#L37-L62) and [`ConsoleHost`](https://github.com/dotnet/BenchmarkDotNet/blob/3223c94a92050147a02482a3810860703f0c5171/src/BenchmarkDotNet/Engines/ConsoleHost.cs#L13) wrap the StandartInput / StandardOutput with a TextReader/Writer that uses the _same_ encoding (UTF-8).

# Possibly related issues

- #1136 
- #1480 
- https://github.com/gfoidl/Base64/issues/139

adamsitnik:
The encoding mismatch between host and benchmark process can cause issues like #1487

I was not sure whether we should enforce some particular encoding for every OS and Runtime (it could have some side effects, be not always supported, require more testing etc) and I decided to relax the check from full equality to basically contains ignore case.

It should fix #1487 and it may fix #1136 as well.

### DIFF
--- a/src/BenchmarkDotNet/Engines/ConsoleHost.cs
+++ b/src/BenchmarkDotNet/Engines/ConsoleHost.cs
@@ -30,8 +30,11 @@ namespace BenchmarkDotNet.Engines
             // I did not use Mutexes because they are not supported for Linux/MacOs for .NET Core
             // this solution is stupid simple and it works
             string acknowledgment = inReader.ReadLine();
-            if (acknowledgment != Engine.Signals.Acknowledgment)
-                throw new NotSupportedException($"Unknown Acknowledgment: {acknowledgment}");
+            if (acknowledgment.IndexOf(Engine.Signals.Acknowledgment, StringComparison.OrdinalIgnoreCase) < 0)
+            {
+                throw new NotSupportedException($"Unknown Acknowledgment: '{acknowledgment}'." + Environment.NewLine +
+                    $"If for some reason you are running the benchmark process manually you just need to type '{Engine.Signals.Acknowledgment}' and hit enter.");
+            }
         }
 
         public void SendError(string message) => outWriter.WriteLine($"{ValidationErrorReporter.ConsoleErrorPrefix} {message}");
--- a/tests/BenchmarkDotNet.IntegrationTests/SignalsTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/SignalsTests.cs
@@ -1,0 +1,101 @@
+﻿using BenchmarkDotNet.Analysers;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Diagnosers;
+using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Environments;
+using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.IntegrationTests.Xunit;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Portability;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains;
+using BenchmarkDotNet.Toolchains.InProcess.Emit;
+using BenchmarkDotNet.Validators;
+using System;
+using System.Collections.Generic;
+using Xunit;
+using Xunit.Abstractions;
+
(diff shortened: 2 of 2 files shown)


ChristophLindemann:
Please see my comments in #1487.

I think you should enforce a fixed encoding. Using UTF8 without BOM should be safe, as it is supported on all platforms.
This is also the default selected by .Net Core and I think PowerShell Core.

You might still want to keep the ordinal comparison.

ChristophLindemann:
I have tried the 0.12.1.1382 package, and it indeed fixes the problem.

I still think, setting the encoding would be a more correct way to solve this, but I accept the current solution as it solves my problem #1487 .

When do you think this can be released?

adamsitnik:
> I think you should enforce a fixed encoding. Using UTF8 without BOM should be safe

You are right, I've merged #1491 which does that. I am going to close this PR.

> I have tried the 0.12.1.1382 package, and it indeed fixes the problem.

Thanks for letting me know!

> When do you think this can be released?

as soon as [this](https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/builds/34111341) build goes green (should be 30-40 minutes from now) BDN CI is going to publish `0.12.1.1389` package to our CI feed. After that you should be able to consume it.

### nuget.config

```xml
<packageSources>
</packageSources>
```

### .csproj

```xml
<ItemGroup>
    <PackageReference Include="BenchmarkDotNet" Version="0.12.1.1389" />
</ItemGroup>
```

PR #1494, opened 2020-06-30
enforce Optimizations when using Custom Build Configurations

### LINKED ISSUE #1493: Using WithCustomBuildConfiguration leads to always running with RyuJIT Debug
Recently I got some help from you guys to set up a way to run benchmarks to use different versions of the same DLL using `WithCustomBuildConfiguration`. This works pretty well. However, there's one gotcha that I don't understand: it appears to always use `RyuJIT DEBUG`. 

It appears to have the effect that certain optimizations do not take place (which is not surprising if it is indeed the RyuJIT DEBUG running). However, I'm not 100% certain if this is just a flaw in the reporting, because when running each job separately (and then it shows just `RyuJIT`), the numbers are within error margin.

Anyway, this is the output I have:

```text
// * Summary *

BenchmarkDotNet=v0.12.1, OS=Windows 7 SP1 (6.1.7601.0)
Intel Xeon CPU X5680 3.33GHz, 2 CPU, 12 logical and 12 physical cores
Frequency=3247138 Hz, Resolution=307.9635 ns, Timer=TSC
.NET Core SDK=3.1.400-preview-015151
  [Host]    : .NET Core 3.1.3 (CoreCLR 4.700.20.11803, CoreFX 4.700.20.12001), X64 RyuJIT DEBUG
  Before    : .NET Core 3.1.3 (CoreCLR 4.700.20.11803, CoreFX 4.700.20.12001), X64 RyuJIT
  Optimized : .NET Core 3.1.3 (CoreCLR 4.700.20.11803, CoreFX 4.700.20.12001), X64 RyuJIT DEBUG

MaxRelativeError=0.01  IterationTime=150.0000 ms  WarmupCount=1
```

Note that for my "Optimized" job I would expect it to use the same 64 bit `RyuJIT`. This is the config that I use (with help from @adamsitnik, from here: https://github.com/dotnet/BenchmarkDotNet/issues/1486). It's F#, but you'll get the idea:

```f#
let baseJob = Job.Default
                .WithWarmupCount(1) // 1 warmup is enough for our purpose
                .WithIterationTime(TimeInterval.FromMilliseconds(250.0)) // the default is 0.5s per iteration
                .WithMaxRelativeError(0.01)

let jobBefore = baseJob.WithId("Before").WithBaseline(true)

let jobAfter = baseJob.WithCustomBuildConfiguration("LocalBuild").WithId("Optimized")

let config = DefaultConfig.Instance.AddJob(jobBefore).AddJob(jobAfter).KeepBenchmarkFiles()

BenchmarkRunner.Run<BenchToString>(config)
|> ignore
```

adamsitnik:
fixes #1493 

cc @abelbraaksma

### DIFF
--- a/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
@@ -20,7 +20,7 @@ namespace BenchmarkDotNet.Portability
 {
     internal static class RuntimeInformation
     {
-        private const string DebugConfigurationName = "DEBUG";
+        internal const string DebugConfigurationName = "DEBUG";
         internal const string ReleaseConfigurationName = "RELEASE";
         internal const string Unknown = "?";
 
--- a/src/BenchmarkDotNet/Templates/CsProj.txt
+++ b/src/BenchmarkDotNet/Templates/CsProj.txt
@@ -15,6 +15,8 @@
     <DebugSymbols>true</DebugSymbols>
     <UseSharedCompilation>false</UseSharedCompilation>
     <CodeAnalysisRuleSet></CodeAnalysisRuleSet>
+    <!-- needed for custom build configurations (only "Release" builds are optimized by default) -->
+    <Optimize Condition=" '$(Configuration)' != 'Debug' ">true</Optimize>
     $COPIEDSETTINGS$
     <!-- we set LangVersion after $COPIEDSETTINGS$ which might contain LangVersion copied from the benchmarks project -->
     <LangVersion Condition="'$(LangVersion)' == '' Or ($([System.Char]::IsDigit('$(LangVersion)', 0)) And '$(LangVersion)' &lt; '7.3')">latest</LangVersion>
--- a/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
+++ b/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
@@ -5,6 +5,7 @@ using System.Text;
 using BenchmarkDotNet.Characteristics;
 using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Loggers;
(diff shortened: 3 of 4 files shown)


abelbraaksma:
Wow, that was fast! I wonder how that will influence the timings (things were already faster with Debug JIT, I wonder what the real improvements are).

Thanks for this!

abelbraaksma:
If I understand this change correctly, it changes the parent scaffolding. The build settings for my own project already had this setting.

But since the scaffolded project is the one that runs, it still matters what jit is chosen, I guess.

adamsitnik:
> Wow, that was fast!

correctness bugs are always getting top priority! I was also attending a few meetings in a row and wanted to be productive ;)

@abelbraaksma in your particular case we had two-projects:

* the project that defined the benchmarks (Optimize set to true in `.csproj`)
* the auto-generated project (Optimize not set to true for custom build configurations)

So the boilerplate code generated by BDN was not optimized, but your code has been. It's not BAD, but definitely far from perfect.

The problem would be much bigger for scenario where the project with benchmarks would reference more projects. They would all have been build for given build configuration with optimizations not enabled. 

```log
Auto-generated (Optimizations disabled) -> Benchmarks (Optimizations enabled) -> Libraries (Optimizations disabled)
```

The fix that I've provided passes `/p:Optimize=true` to MsBuild anytime we call dotnet `build|publish|restore` and once we have it merged, setting `<Optimize>true</Optimize>` won't be needed anymore in the project with benchmarks.

abelbraaksma:
Thanks for the explanation. So seeing "RyuJIT DEBUG" does not mean a checked build of RyuJIT is used, but that the boilerplate project was built using debug settings (or at least without optimizations).

This may or may not explain that I occasionally saw very weird timings (up to 4x slower than the reference project). I also occasionally had unexpected errors when halfway the benchmark. Though often a manual clean of the bin directories solved those issues.

PR #1501, opened 2020-07-16
Add Custom Runtime Pack option

naricc:
This adds an option to specify a custom runtime pack (currently only used by wasm, but in principle useful for other net5.0 appolications). This is used to run against a custom built net5.0 runtime.


naricc:
@adamsitnik (I can't actually add reviewers)

SamMonoRT:
@adamsitnik - is it possible to kick off a re-run of the BenchmarkDotNet - macOS lane to confirm it is not related to changes here. Also, if you have a moment, please review and merge if things look good. Thank you

adamsitnik:
> macOS lane to confirm it is not related to changes here

done

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -30,6 +35,8 @@ protected override void GenerateProject(BuildPartition buildPartition, Artifacts
                     .Replace("$COPIEDSETTINGS$", customProperties)
                     .Replace("$CONFIGURATIONNAME$", buildPartition.BuildConfiguration)
                     .Replace("$SDKNAME$", sdkName)
+                    .Replace("$RUNTIMEPACK$", CustomRuntimePack != null ? CustomRuntimePack : "")
```
adamsitnik, line 38: nit: could be simplified

```suggestion
                    .Replace("$RUNTIMEPACK$", CustomRuntimePack ?? "")
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs
```diff
@@ -12,6 +12,8 @@ public class WasmToolChain : Toolchain
     {
         private string CustomDotNetCliPath { get; }
 
+        private string CustomRuntimePack { get;  }
+
```
adamsitnik, line 16: it's never used
```suggestion
```

naricc:
@adamsitnik So, the benefit is primarily if you want to make a change to the runtime (whether that is mono, or clr, or whatever) and measure some benchmark results against that change. It's very easy to make the changes to the runtime and produce a new runtime pack.

naricc:
Or, to put it another way, runtime packs are how we can point a project at a specific runtime with out going through the laborious process of packaging that runtime into a dotnet host.

adamsitnik:
> runtime packs are how we can point a project at a specific runtime with out going through the laborious process of packaging that runtime into a dotnet host.

this is what I needed, thanks!

adamsitnik:
@naricc  [this build](https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/builds/34162500) should produce `0.12.1.1402` package within 30-40 mintues

SamMonoRT:
Thank you @naricc and @adamsitnik for getting this in. Getting regular perf numbers will go a long way in improving what we ship to customers.

Issue #1504, opened 2020-07-20
Add Support For Normal Globalization to Wasm


naricc:
Mono-Wasm recently added support for full globalization, but taking advantage of it requires a change in the .json manifest for the wasm app. A change is required to the tool chain to generate it.

Not having this doesn't break anything, but its absence means any benchmark comparison using WASM must use invariant globalization to be a fair comparison.

naricc:
I intend to take this issue myself.

PR #1507, opened 2020-07-23
Change mono-config.js format in the Wasm Tool Chain for ICU support

naricc:
This change uses an updated format for the mono-config.js generated in the wasm tool chain. This allows wasm to be used with ICU support for globalization (it otherwise would run in invariant globalization mode).

Fixes: https://github.com/dotnet/BenchmarkDotNet/issues/1504


naricc:
@adamsitnik

akoeplinger:
@AndreyAkinshin do you think we could merge this PR?

AndreyAkinshin:
@akoeplinger looks good to me.
@adamsitnik could you please also take a look?

DrewScoggins:
@adamsitnik is on vacation until August 3rd.

PR #1523, opened 2020-08-24
add .NET 6.0 support, rename .NET Core 5.0 to .NET 5.0

adamsitnik:
This PR:

- adds `net6.0` support
- renames `netcoreapp5.0` to `net5.0` (I personally don't like it but well we have to live with that)
- simplifies displayed version numbers for .NET 5.0+

Before:

```ini
BenchmarkDotNet=v0.12.1.20200824-develop, OS=Windows 10.0.18363.959 (1909/November2019Update/19H2)
.NET Core SDK=5.0.100-preview.8.20362.3
  [Host]     : .NET Core 2.1.21 (CoreCLR 4.6.29130.01, CoreFX 4.6.29130.02), X64 RyuJIT
  Job-ILZMMP : .NET Core 5.0.0 (CoreCLR 5.0.20.36102, CoreFX 5.0.20.36102), X64 RyuJIT
```

After:

```ini
BenchmarkDotNet=v0.12.1.20200824-develop, OS=Windows 10.0.18363.959 (1909/November2019Update/19H2)
.NET Core SDK=5.0.100-preview.8.20362.3
  [Host]     : .NET Core 2.1.21 (CoreCLR 4.6.29130.01, CoreFX 4.6.29130.02), X64 RyuJIT
  Job-PBAJYZ : .NET 5.0.0 (5.0.20.36102), X64 RyuJIT
```

@AndreyAkinshin PTAL

cc @DrewScoggins @billwert @danmosemsft

### DIFF
--- a/src/BenchmarkDotNet/Environments/HostEnvironmentInfo.cs
+++ b/src/BenchmarkDotNet/Environments/HostEnvironmentInfo.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using BenchmarkDotNet.Helpers;
 using BenchmarkDotNet.Loggers;
@@ -98,7 +99,7 @@ namespace BenchmarkDotNet.Environments
             if (RuntimeInformation.IsNetCore && IsDotNetCliInstalled())
             {
                 // this wonderfull version number contains words like "preview" and ... 5 segments so it can not be parsed by Version.Parse. Example: "5.0.100-preview.8.20362.3"
-                if (DotNetSdkVersion.Value[0] >= '5')
+                if (int.TryParse(new string(DotNetSdkVersion.Value.TrimStart().TakeWhile(char.IsDigit).ToArray()), out int major) && major >= 5)
                     yield return $".NET SDK={DotNetSdkVersion.Value}";
                 else
                     yield return $".NET Core SDK={DotNetSdkVersion.Value}";
--- a/src/BenchmarkDotNet/Environments/HostEnvironmentInfo.cs
+++ b/src/BenchmarkDotNet/Environments/HostEnvironmentInfo.cs
@@ -96,7 +96,13 @@ namespace BenchmarkDotNet.Environments
                 yield return $"Frequency={ChronometerFrequency}, Resolution={ChronometerResolution.ToString(cultureInfo)}, Timer={HardwareTimerKind.ToString().ToUpper()}";
 
             if (RuntimeInformation.IsNetCore && IsDotNetCliInstalled())
-                yield return $".NET SDK={DotNetSdkVersion.Value}";
+            {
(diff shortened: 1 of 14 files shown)


AndreyAkinshin:
If understand correctly, .NET Core SDK was renamed to .NET SDK
image
Speaking of this line
```
.NET Core SDK=5.0.100-preview.8.20362.3
```
we should also rename it to `.NET SDK`.

adamsitnik:
> we should also rename it to .NET SDK.

Good point, done

src/BenchmarkDotNet/Environments/HostEnvironmentInfo.cs
```diff
@@ -96,7 +96,7 @@ public override IEnumerable<string> ToFormattedString()
                 yield return $"Frequency={ChronometerFrequency}, Resolution={ChronometerResolution.ToString(cultureInfo)}, Timer={HardwareTimerKind.ToString().ToUpper()}";
 
             if (RuntimeInformation.IsNetCore && IsDotNetCliInstalled())
-                yield return $".NET Core SDK={DotNetSdkVersion.Value}";
+                yield return $".NET SDK={DotNetSdkVersion.Value}";
```
AndreyAkinshin, line 99: We should display `.NET SDK` only for 5.0+; otherwise we should display `.NET Core SDK`.

adamsitnik: We can't just use `if (Version.TryParse(DotNetSdkVersion.Value, out var sdkVersion) && sdkVersion >= new Version(5,0))` because the SDK version contains words like "preview" in the name and 5 segements (`A.B.C.D.E`). Example: `5.0.100-preview.8.20362.3`. Anyway I have added a check for that

src/BenchmarkDotNet/Environments/HostEnvironmentInfo.cs
```diff
@@ -96,7 +96,13 @@ public override IEnumerable<string> ToFormattedString()
                 yield return $"Frequency={ChronometerFrequency}, Resolution={ChronometerResolution.ToString(cultureInfo)}, Timer={HardwareTimerKind.ToString().ToUpper()}";
 
             if (RuntimeInformation.IsNetCore && IsDotNetCliInstalled())
-                yield return $".NET SDK={DotNetSdkVersion.Value}";
+            {
+                // this wonderfull version number contains words like "preview" and ... 5 segments so it can not be parsed by Version.Parse. Example: "5.0.100-preview.8.20362.3"
+                if (DotNetSdkVersion.Value[0] >= '5')
```
AndreyAkinshin, line 101: In November 2025, .NET 10 should be released and this heuristic will be broken. =)
I think it will be better to split the version string by `'.'` and parse the first component. This logic can be encapsulated in a method `int GetMajorVersionComponent(string version)`.

AndreyAkinshin (approved):

PR #1554, opened 2020-10-10
Documentation: add --maxWidth description

joostas:
`--maxWidth` console argument is described under [Command-line tool](https://benchmarkdotnet.org/articles/guides/tool.html) page, but is missing under page [How to use console arguments](https://benchmarkdotnet.org/articles/guides/console-args.html).


adamsitnik (approved):
@joostas thanks!

Issue #1539, opened 2020-09-24
Inconsistency between benchmark filter and hint


kevingosse:
I tried running a benchmark with the command:
```
dotnet run -c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f SpanBenchmark.StartFinishScope
```
But received a puzzling error message:

```
You must have made a typo in 'SpanBenchmark.StartFinishScope'. Suggestions:
        SpanBenchmark.StartFinishScope
```

It turns out that the filter operates on the full type name (in my case, Benchmarks.Trace.SpanBenchmark.StartFinishScope), but the CorrectionSuggester splits the type name on `.` and computes all the possible permutations (Benchmarks, Trace, Benchmarks.Trace, Trace.SpanBenchmarks, ...) , without checking if they're valid.

```
> dotnet run -c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f Trace

You must have made a typo in 'Trace'. Suggestions:
        Trace
```

kevingosse:
The weird part is that it seems to be by design: https://github.com/dotnet/BenchmarkDotNet/blob/master/tests/BenchmarkDotNet.Tests/CorrectionsSuggesterTests.cs#L68

```csharp
        [Fact]
        public void FilterByNamespace_LevenshteinOrdering()
        {
            var suggestedNames = new CorrectionsSuggester(new[]
            {
                typeof(NamespaceA.MyClassA),
                typeof(NamespaceB.MyClassB),
                typeof(AnotherNamespace.MyClassZ),
                typeof(NamespaceB.NamespaceC.MyClassC)
            }).SuggestFor("Nmespace");
            Assert.Equal(new[] { "NamespaceA", "NamespaceB", "NamespaceC" }, suggestedNames);
        }
```
Shouldn't this return "NamespaceA.MyClassA", "NamespaceB.MyClassB", "NamespaceB.NamespaceC.MyClassC" ?

adamsitnik:
> Shouldn't this return

You are right, it should.

I am going to mark it as up-for-grabs and #hacktoberfest as it should be quite simple to fix.

stanciuadrian:
[...]

PR #1555, opened 2020-10-11
Apply name (glob) filters to substring of fully qualified [Benchmark] method name

stanciuadrian:
Closes: #1539.

TODO: document intended behavior with additional unit tests

### DIFF
--- a/src/BenchmarkDotNet/Filters/GlobFilter.cs
+++ b/src/BenchmarkDotNet/Filters/GlobFilter.cs
@@ -26,6 +26,6 @@ namespace BenchmarkDotNet.Filters
         }
 
         // https://stackoverflow.com/a/6907849/5852046 not perfect but should work for all we need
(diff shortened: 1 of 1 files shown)


adamsitnik:
I've responded in the issue thread: https://github.com/dotnet/BenchmarkDotNet/issues/1539#issuecomment-709395535

adamsitnik (changes requested):
> suggests wildcards are mandatory

yes, this is by design and I intend to keep it that way so removing the `^$` anchors is rather not an option (it would be a breaking change and the glob pattern matching with `*` was introduced on purpose).

If we have the following type:

```cs
namespace A
{
   public class B
   {
        [Benchmark]
        public void C() { }
   }
}
```

And the user passes filter `B.C`:

```cmd
--filter B.C
```

The hint should mention that either `A.B.C` or `*B.C`.

PR #1556, opened 2020-10-11
Simplify code: remove sort before adding to HashSet

stanciuadrian:
`allBenchmarkNames` is a `HashSet` so the order of the added elements should not matter.


adamsitnik (approved):
nice catch!

I can see that it gets sorted later: https://github.com/dotnet/BenchmarkDotNet/blob/65e15f30aa0b198bb54d1497940761f5a215b35c/src/BenchmarkDotNet/ConsoleArguments/CorrectionsSuggester.cs#L46

so LGTM

stanciuadrian:
Would you mind adding a `hacktoberfest-accepted` label to my commit? I'm trying to follow it through this year. Thanks!

adamsitnik:
@stanciuadrian done

PR #1568, opened 2020-10-22
Update getting-started.md

### LINKED ISSUE #1474: docs: Getting started guide
Do you think it makes sense to swap the current getting started guide 1 and 2. It seems that the package should be installed into the console application.

michalgalecki:
Fixes #1474


adamsitnik (approved):
LGTM, thank you @michalgalecki !

PR #1569, opened 2020-10-22
Respect size number format in MetricColumn

njy:
The number format is not used when formatting a size value in a MetricColumn


adamsitnik (approved):
LGTM, thank you @njy !

PR #1581, opened 2020-11-02
Json exporter fix

marcnet80:


src/BenchmarkDotNet/Exporters/Json/SimpleJson.cs
```diff
@@ -1554,11 +1554,29 @@ protected virtual bool TrySerializeUnknownTypes(object input, out object output)
...
+            {
+                return value;
+            }
+
+            switch (value)
+            {
+                case double.NaN:
+                case double.NegativeInfinity:
+                case double.PositiveInfinity:
+                    return 0.0;
```
adamsitnik, line 1575: I wonder why we do get `NaN`s here. Most probably by doing division by `0` using doubles?

@marcnet80  In such a case, would it be better (and possible) to return an empty string instead of `0.0`? Just to say "there is no value" instead "the value is 0.0" (it would be hard to tell if there was actual 0 or NaN serialized as 0).

```suggestion
                    return string.Empty;
```

tests/BenchmarkDotNet.Tests/Exporters/JsonExporterTests.cs
```diff
@@ -0,0 +1,26 @@
...
+            //Arrange
+            var logger = new AccumulationLogger();
+            var summary = MockFactory.CreateSummary(DefaultConfig.Instance);
+
+            //Act
+            JsonExporter.Default.ExportToLog(summary, logger);
+
+            //Assert
+            Assert.True(JsonSerializer.TryDeserializeObject(logger.GetLog(), out object obj));
+        }
```
adamsitnik, line 24: we already have tests that check the Json serializer output: https://github.com/dotnet/BenchmarkDotNet/blob/9744089383f881506c8069f9802260e63c448ea1/tests/BenchmarkDotNet.Tests/Exporters/CommonExporterApprovalTests.cs#L85-L88

Could you please remove this test, make `ReplaceUnsupportedNumericValues` internal and just test what it returns for valid and invalid inputs?

adamsitnik (commented):
Hi @marcnet80 

Big thanks for your PR!

Overall it looks good, but I have left some comments. Could you please address them before I merge the PR?

Thanks,
Adam

src/BenchmarkDotNet/Exporters/Json/SimpleJson.cs
marcnet80: Hello, @adamsitnik. Yes,`NaN` occurs in result of doubles zero division here: https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Mathematics/Statistics.cs#L61, it calls Perfolizer method: https://github.com/AndreyAkinshin/perfolizer/blob/master/src/Perfolizer/Perfolizer/Mathematics/Common/Moments.cs#L31

Propose to return escaped empty string like `"\"\""`, because we have `StringBuilder`, in resulted Json will be: `""`.

tests/BenchmarkDotNet.Tests/Exporters/JsonExporterTests.cs
marcnet80: Some tests will be failed due hardcoded `NaN` values in text files from ApprovedFiles folder.
Please, review new changes.

adamsitnik: @marcnet80 the new changes look good to me! but we need to get the CI green.

The easiest way of updating the approved files after on purpose exporter changes is:

* go to `tests\BenchmarkDotNet.Tests`
* run the tests `dotnet test -c Release -f netcoreapp2.1`
* go to `Exporters\ApprovedFiles` and check which `.approved` file is different than `.received`. Ensure that the changes are what we want and remove the old `.approved` file and rename the `.received` file `.approved` 
* re-run the tests
* if everything is green, push the changes

obraz

marcnet80: fixed.

adamsitnik (approved):
LGTM, thank you @marcnet80 !

PR #1582, opened 2020-11-02
introduce ManualConfig.CreateMinimumViable() method

adamsitnik:
when users use `ManualConfig.CreateEmpty` they often don't know that they should add column provider and logger.

Introducing this new method should make it easier for the users to define custom exporters etc by using `ManualConfig`

```cs
var config = ManualConfig.CreateEmpty()
    .AddColumnProvider(DefaultColumnProviders.Instance)
    .AddLogger(ConsoleLogger.Default);
```


AndreyAkinshin (approved):
LGTM

PR #1584, opened 2020-11-03
Prevent dotnet pack from packaging benchmark projects

kendaleiv:
Prevent the dotnet pack command from unexpectedly packaging benchmarking projects built via the dotnet new templates.


adamsitnik (approved):
LGTM, thank you @kendaleiv !

Issue #1513, opened 2020-08-11
Make memory alignment more random


adamsitnik:
While working on a new bot for auto-filing performance regressions in dotnet/runtime repository ([sample issue](https://github.com/dotnet/runtime/issues/37806)) we have found out that quite a few microbenchmarks from dotnet/performance repository are bimodal and the modality tends to be stable for a few days before it switches to the other mode

Example:

![obraz](https://user-images.githubusercontent.com/6011991/89928990-e64ee280-dc08-11ea-85e7-4beb9004f846.png)

So it's very often something like:

```log
_____~~~~~~_____~~~~~~______~~~~~~
```

A while ago @andyayersms has [mentioned](https://github.com/dotnet/BenchmarkDotNet/issues/756) [stabilizer](https://github.com/ccurtsinger/stabilizer/blob/master/README.md) which performs full randomization.

.NET does not allow for full control of memory alignment, but we could at least try to make it more random.

In https://github.com/dotnet/runtime/issues/37814 @jkotas has [provided](https://github.com/dotnet/runtime/issues/37814#issuecomment-667804880) a small repro that shows "the many modal nature of memory copying":

```cs
using System;
using System.Diagnostics;

class Program
{
    static void Work(ReadOnlySpan<string> from, Span<string> to)
    {
        for (int i = 0; i < 1000000; i++) from.CopyTo(to);
    }

    static void Main(string[] args)
    {
        Random r = new Random();
        for (;;)
        {
            var sw = new Stopwatch();
            GC.KeepAlive(new byte[r.Next(32)]); // the trick
            var from = new string[2048];
            GC.KeepAlive(new byte[r.Next(32)]);
            var to = new string[2048];
            if (r.Next(10) == 0) GC.Collect();
            sw.Start();
            Work(from, to);       
            sw.Stop();
            Console.WriteLine(sw.ElapsedMilliseconds);
        }
    }
}
```

So the first step could be to allocate a variable-size byte array between iterations to have more randomized memory alignment of the objects allocated during benchmarking.

The problem is that very often the input is allocated in `[GlobalSetup]` method (to exclude the cost of allocation from the benchmark which is good) which we promise to call only once during benchmarking ;) Perhaps we could add an optional config flag to invoke it once per every iteration? (but somehow avoid the `[IterationSetup]` hell)
Some benchmarks are initialized in constructors, so we might also consider allocating a new instance of benchmarked type for every iteration.

@AndreyAkinshin what do you think? 

[...]

PR #1587, opened 2020-11-06
Memory Randomization

adamsitnik:
A very simple implementation of https://github.com/dotnet/BenchmarkDotNet/issues/1513 that I hope is going to help us answer the question of whether we should invest more in this direction.

I am creating the PR just to have a NuGet package published by our CI so others can easily give it a try.

Sample benchmark:

```cs
public class IntroMemoryRandomization
{
    [Params(512 * 4)]
    public int Size;

    private int[] _array;
    private int[] _destination;

    [GlobalSetup]
    public void Setup()
    {
        _array = new int[Size];
        _destination = new int[Size];
    }

    [Benchmark]
    public void Array() => System.Array.Copy(_array, _destination, Size);
}
```

### Default settings

```cs
dotnet run -c Release -f netcoreapp2.1 --filter IntroMemoryRandomization
```

```ini
[502.859 ns ; 508.045 ns) | @@@@@@@@@@@@@@@
```

### MemoryRandomization set to true and Default Outliers setting (remove upper)

```cmd
dotnet run -c Release -f netcoreapp2.1 --filter IntroMemoryRandomization --memoryRandomization true --maxIterationCount 50
```

```ini
[117.514 ns ; 203.847 ns) | @@@@@@@@@@@@
[203.847 ns ; 287.079 ns) |
[287.079 ns ; 362.172 ns) |
[362.172 ns ; 445.404 ns) | @@@@@@@@@@@@@@@@@@@@@@@@@@@@@
```

### MemoryRandomization set to true and custom Outliers setting: don't remove any

```cmd
dotnet run -c Release -f netcoreapp2.1 --filter IntroMemoryRandomization --memoryRandomization true --outliers DontRemove --maxIterationCount 50
```

```ini
[108.803 ns ; 213.537 ns) | @@@@@@@@@@@@@@@
[213.537 ns ; 315.458 ns) |
[315.458 ns ; 446.853 ns) | @@@@@@@@@@@@@@@@@@@@
[446.853 ns ; 559.259 ns) | @@@@@@@@@@@@@@@
```

@kunalspathak

### DIFF
--- a/samples/BenchmarkDotNet.Samples/IntroMemoryRandomization.cs
+++ b/samples/BenchmarkDotNet.Samples/IntroMemoryRandomization.cs
@@ -1,0 +1,28 @@
+﻿using BenchmarkDotNet.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BenchmarkDotNet.Samples
+{
+    public class IntroMemoryRandomization
+    {
+        [Params(512 * 4)]
+        public int Size;
+
+        private int[] _array;
+        private int[] _destination;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            _array = new int[Size];
+            _destination = new int[Size];
+        }
+
+        [Benchmark]
+        public void Array() => System.Array.Copy(_array, _destination, Size);
+    }
+}
--- a/src/BenchmarkDotNet/Attributes/Mutators/MemoryRandomizationAttribute.cs
+++ b/src/BenchmarkDotNet/Attributes/Mutators/MemoryRandomizationAttribute.cs
@@ -1,0 +1,17 @@
+﻿using BenchmarkDotNet.Jobs;
+using Perfolizer.Mathematics.OutlierDetection;
+
+namespace BenchmarkDotNet.Attributes
+{
+    /// <summary>
+    /// specifies whether Engine should allocate some random-sized memory between iterations
+    /// <remarks>it makes [GlobalCleanup] and [GlobalSetup] methods to be executed after every iteration</remarks>
+    /// </summary>
+    public class MemoryRandomizationAttribute : JobMutatorConfigBaseAttribute
+    {
+        public MemoryRandomizationAttribute(bool enable = true, OutlierMode outlierMode = OutlierMode.DontRemove)
+            : base(Job.Default.WithMemoryRandomization(enable).WithOutlierMode(outlierMode))
+        {
+        }
+    }
+}
--- a/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
@@ -168,6 +168,9 @@ namespace BenchmarkDotNet.ConsoleArguments
         [Option("envVars", Required = false, HelpText = "Colon separated environment variables (key:value)")]
         public IEnumerable<string> EnvironmentVariables { get; set; }
 
+        [Option("memoryRandomization", Required = false, HelpText = "Enables Memory Randomization that causes Engine to allocate random-size memory between iterations and invoke GlobalSetup afterwards.")]
+        public bool MemoryRandomization { get; set; }
+
         [Option("wasmEngine", Required = false, HelpText = "Full path to a java script engine used to run the benchmarks, used by Wasm toolchain.")]
         public FileInfo WasmJavascriptEngine { get; set; }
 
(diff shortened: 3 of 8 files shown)


adamsitnik:
With MemoryRandomization enabled and no outliers removal we get distribution (3 buckets|modes) similar to what we have gathered from multiple runs in the past: https://pvscmdupload.blob.core.windows.net/reports/allTestHistory%2frefs%2fheads%2fmaster_x64_Windows%2010.0.18362%2fSystem.Collections.CopyTo(Int32).Array(Size%3a%202048).html

obraz

adamsitnik:
@kunalspathak if [this](https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/builds/36176081) build gets green (a matter of 30 minutes from now) then a new package `0.12.1.1459` should become available at our CI feed:

```xml
<packageSources>
</packageSources>
```

adamsitnik:
cc @AndyAyersMS

AndyAyersMS:
This looks really promising.

I wonder if we might need something more sophisticated eventually. We don't know is which GC heap(s) the benchmark is accessing. We can impact Gen0/LOH alignments but it's trickier to impact Gen1/Gen2. Stack alignment might also come into play -- perhaps the benchmark runner can do random-sized stackallocs too?

I suspect for Gen0 we only need very small alignments changes, perhaps just fractions of cache line sizes, though we might need to go all the way up to fractions of page sizes (though it is hard to imagine us doing enough iterations to really cover the space of possibilities here). For LOH likewise, some 85K+ base plus a cache-line-sized random amount on top.

This also tells us we might need to pay more attention to controlling alignment of some key data in the runtime (eg frequently accessed static arrays?) Worth thinking about, anyways.

adamsitnik:
> We don't know is which GC heap(s) the benchmark is accessing.

By default every benchmark is single-threaded so it should not be a problem in most cases. When it comes to multithreaded benchmarks we could achieve that by for example having an affinitzed thread per core, but this would require much more work... I'll keep this in the back of my head and when it becomes a problem try to improve the solution.

> We can impact Gen0/LOH alignments but it's trickier to impact Gen1/Gen2.

Very good point! I've modified the code and made sure that the object gets promoted to Gen 1 and Gen 2 by keeping it alive for two GC collections:

```cs
Debug.Assert(GC.GetGeneration(gen0object) == 0);
GC.Collect(0); // get it promoted to Gen 1
GC.Collect(1); // get it promoted to Gen 2

GC.KeepAlive(gen0object);
```

> Stack alignment might also come into play -- perhaps the benchmark runner can do random-sized stackallocs too?

And another great point! I've added a call to the following method just a moment before start the watch:

```cs
private unsafe void RandomizeStackMemory()
{
    Span<byte> stackMemory = stackalloc byte[random.Next(32)];
    Consume(stackMemory);
}

private void Consume(in Span<byte> _) { }
```

(BDN repo uses C# 7.3 so the code is not simple as it could be with C#8 or 9 ;))

> For LOH likewise, some 85K+ base plus a cache-line-sized random amount on top.

I have added that as well:

```cs
var lohObject = new byte[85 * 1024 + random.Next(32)];
Debug.Assert(GC.GetGeneration(lohObject) == 2);

GC.KeepAlive(lohObject);
```

> This also tells us we might need to pay more attention to controlling alignment of some key data in the runtime

Maybe we should make the `ArrayPool<byte>.Shared` arrays being aligned by default as they are used everywhere in the runtime, BCL, and ASP.NET? Assuming that we already don't do that (cc @stephentoub @vsadov)

adamsitnik:
If anyone wants to give it a try then you need to use the `0.12.1.1462` version from 

```xml
<packageSources>
</packageSources>
```

AndyAyersMS:
> made sure that the object gets promoted to Gen 1 and Gen 2

I'm not sure if promoting will help or not... I guess my point was that the code being benchmarked may read from all sorts of objects and as far as I can tell we can't reliably randomize all their addresses.

What you had initially may end up working better, if benchmarks tend to read from objects allocated after the random object. And perhaps the random LOH allocation will help too. But influencing Gen1 / Gen2 addresses seems harder.

VSadov:
I am not sure if targeting specific GC behaviors (i.e. promotions) is necessary. Besides some of those behaviors could change with different tunings, state of the machine, server vs. workstation GC, etc..

It could be fairly certain though that objects allocated together in a sequence will typically be allocated together and likely stay together even when relocated.

I think just allocating batches of objects of varying size should be sufficient, if I get the idea here. 
Also, since LOH objects are separated, it would need to be handled separately.

Varying the size differences within cache line could be enough. 

64bit is the largest granularity that GC would align by itself. 
(SOH objects are aligned to 32bit on 32bit, with few exceptions where 64bit alignment used even on 32bit)

AndreyAkinshin (approved):

PR #1590, opened 2020-11-09
Update message to not suggest an obsolete API

martincostello:
Update the hint for non-optimised references to use the `WithOptions(ConfigOptions)` method, rather than the obsolete `With(ConfigOptions)` overload.


adamsitnik (approved):
LGTM, thank you @martincostello !

Issue #1561, opened 2020-10-18
NativeMemoryProfiler doesn't report allocations in v0.12.1.1432 Nightly


JimBobSquarePants:
I'm running the following benchmark to compare graphics libraries. 

https://github.com/JimBobSquarePants/core-imaging-playground/tree/ac506f591f7448cb4b2250c7fb9717a976faf198

With v0.12.1 I get the following result:

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.572 (2004/?/20H1)
Intel Core i7-8650U CPU 1.90GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.403
  [Host]            : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  .Net Core 3.1 CLI : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT

Job=.Net Core 3.1 CLI  Toolchain=.NET Core 3.1  IterationCount=5
LaunchCount=1  WarmupCount=5
```

|                                Method |      Mean |     Error |   StdDev | Ratio |     Gen 0 | Gen 1 | Gen 2 |  Allocated | Allocated native memory | Native memory leak |
|-------------------------------------- |----------:|----------:|---------:|------:|----------:|------:|------:|-----------:|------------------------:|-------------------:|
|   'System.Drawing Load, Resize, Save' | 396.97 ms | 14.314 ms | 2.215 ms |  1.00 |         - |     - |     - |    10.2 KB |             11375.79 KB |          738.27 KB |
|       'ImageSharp Load, Resize, Save' | 221.30 ms |  6.784 ms | 1.762 ms |  0.56 |         - |     - |     - | 1985.19 KB |                 24.9 KB |            9.14 KB |
|      'ImageMagick Load, Resize, Save' | 437.39 ms | 17.609 ms | 4.573 ms |  1.10 |         - |     - |     - |   53.89 KB |              60956.9 KB |            0.66 KB |
|        'ImageFree Load, Resize, Save' | 235.38 ms |  3.423 ms | 0.889 ms |  0.59 | 6000.0000 |     - |     - |   69.75 KB |             49140.58 KB |           58.91 KB |
[...]

Issue #794, opened 2018-06-13
Don't exclude allocation quantum side effects for .NET Core 2.0+


adamsitnik:
The GC API problem has been fixed for .NET Core 2.0, we don't need it anymore https://github.com/dotnet/coreclr/issues/10207

```cs
[MemoryDiagnoser]
public class TimeConsuming
{
    [Benchmark] 
    public byte[] SixtyFourBytesArray()
    {
        // this benchmark should hit allocation quantum problem
        // it allocates a little of memory, but it takes a lot of time to execute so we can't run in thousands of times!
        
        Thread.Sleep(TimeSpan.FromSeconds(0.5)); 
        
        return new byte[64];
    }
}
```

Was reporting 0 for .NET Core, should show the actual value.

PR #1600, opened 2020-11-19
fix issue #1561

WojciechNagorski:
This PR fixes issue #1561. I've improved NativeMemoryProfiler and it always should show results, but I found a few cases that I need to check deeply. 

I've run benchmarks twice with :

```c#
                .WithWarmupCount(1)
                .WithLaunchCount(1)
                .WithIterationCount(1)
                .WithInvocationCount(1)
                .WithUnrollFactor(1)
```

And for some benchmarks I got different results for `Allocated native memory` `Native memory leaks` for each run (number 1 on the following picture)
image

I will check it, but it requires more time, so I think we should merge this fix now.

@adamsitnik On the picture above you can see that the `Allocated` column is different for each run. Did you know about it?

### DIFF
--- a/src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
+++ b/src/BenchmarkDotNet.Diagnostics.Windows/Tracing/NativeMemoryLogParser.cs
@@ -29,7 +29,7 @@ namespace BenchmarkDotNet.Diagnostics.Windows.Tracing
 
         private readonly string moduleName;
 
-        private readonly string functionName;
+        private readonly string[] functionNames;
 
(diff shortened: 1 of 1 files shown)


adamsitnik (approved):
LGTM, thank you @WojciechNagorski !

adamsitnik:
> On the picture above you can see that the Allocated column is different for each run

It's most probably caused by Allocation Quantum (GC allocates at least 8 KB of memory, #794) or Tiered JIT (https://github.com/dotnet/BenchmarkDotNet/issues/1542#issuecomment-702169827). Running the benchmark more than once typically solves the problem

WojciechNagorski:
@adamsitnik thanks for the information!

PR #1606, opened 2020-12-03
CoreRT feed and version update

### LINKED ISSUE #1605: CoreRT / NativeAOT version
As per [this issue](https://github.com/dotnet/runtimelab/issues/415#issuecomment-737468088) BDN is using an outdated version by default, and I've not been able to find a way to set the version manually.

adamsitnik:
@AndreyAkinshin CoreRT has moved from https://github.com/dotnet/corert/ to https://github.com/dotnet/runtimelab/tree/feature/NativeAOT and we need to update the default compiler version and nuget feed address to avoid issue like #1605

Fixes #1605

/cc @MichalStrehovsky @jkotas

### DIFF
--- a/src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
+++ b/src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
@@ -61,7 +61,7 @@ namespace BenchmarkDotNet.Environments
                 case Version v when v.Major == 3 && v.Minor == 0: return CoreRt30;
                 case Version v when v.Major == 3 && v.Minor == 1: return CoreRt31;
                 case Version v when v.Major == 5 && v.Minor == 0: return CoreRt50;
-                case Version v when v.Major == 6 && v.Minor == 0: return CoreRt50;
+                case Version v when v.Major == 6 && v.Minor == 0: return CoreRt60;
                 default:
                     return new CoreRtRuntime(RuntimeMoniker.NotRecognized, $"net{version.Major}.{version.Minor}", $"CoreRT {version.Major}.{version.Minor}");
             }
--- a/src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
+++ b/src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
@@ -10,62 +10,60 @@ namespace BenchmarkDotNet.Environments
         /// <summary>
         /// CoreRT compiled as netcoreapp2.0
         /// </summary>
-        public static readonly CoreRtRuntime CoreRt20 = new CoreRtRuntime(RuntimeMoniker.CoreRt20, "netcoreapp2.0", "CoreRt 2.0");
+        public static readonly CoreRtRuntime CoreRt20 = new CoreRtRuntime(RuntimeMoniker.CoreRt20, "netcoreapp2.0", "CoreRT 2.0");
         /// <summary>
         /// CoreRT compiled as netcoreapp2.1
         /// </summary>
-        public static readonly CoreRtRuntime CoreRt21 = new CoreRtRuntime(RuntimeMoniker.CoreRt21, "netcoreapp2.1", "CoreRt 2.1");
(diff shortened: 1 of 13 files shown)


MichalStrehovsky:
Thank you!

Maybe the old compiler reference could just be removed in favor of latest? The old compiler and runtime is a snapshot from somewhere halfway between .NET Core 3.0 and 5.0 anyway. The NuGet feed is going to be retired at some point.

jkotas:
I agree with Michal. The experimental project does not have packages synchronized with .NET release. We always point everybody to latest.

adamsitnik:
I wanted to keep the old TFMs (`< net5.0`) using the old compiler because I was getting an error when building a sample app that was targeting old monikers and new compiler:

```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>net5.0;netcoreapp2.1</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="6.0.0-*" />
  </ItemGroup>
</Project>
```

```cmd
dotnet publish -r win-x64 -c Release -f netcoreapp2.1
```

```log
PS C:\Projects\repro\HelloWorld> dotnet publish -r win-x64 -c Release -f netcoreapp2.1
Microsoft (R) Build Engine version 16.9.0-preview-20561-01+56ab63409 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored C:\Projects\repro\HelloWorld\HelloWorld.csproj (in 993 ms).
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  HelloWorld -> C:\Projects\repro\HelloWorld\bin\Release\netcoreapp2.1\win-x64\HelloWorld.dll
C:\Users\adsitnik\.nuget\packages\microsoft.netcore.app\2.1.23\build\netcoreapp2.1\Microsoft.NETCore.App.targets(19,5): error : Project is targeting runtime 'win-x64' but did not resolve any runtime-specific packages for the 'Microsoft.NETCore.App' package.  This runtime may not be supported by .NET Core. [C:\Projects\repro\HelloWorld\HelloWorld.csproj]
```

But it turns out that it's just a warning and the .exe is produced anyway. So I am going to ignore the warning and use new compiler everywhere.

adamsitnik:
Update: the .exe is produced, but not in `/native` subfolder and it's a managed .NET Core .exe, not CoreRT. I'll dig into that later

src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
```diff
@@ -10,62 +10,60 @@ public class CoreRtRuntime : Runtime
...
-                case "net6.0": return CoreRt60;
-                default: // support future version of CoreRT
-                    return new CoreRtRuntime(RuntimeMoniker.NotRecognized, msBuildMoniker, displayName);
+                case Version v when v.Major == 2 && v.Minor == 0: return CoreRt20;
+                case Version v when v.Major == 2 && v.Minor == 1: return CoreRt21;
+                case Version v when v.Major == 2 && v.Minor == 2: return CoreRt22;
+                case Version v when v.Major == 3 && v.Minor == 0: return CoreRt30;
+                case Version v when v.Major == 3 && v.Minor == 1: return CoreRt31;
+                case Version v when v.Major == 5 && v.Minor == 0: return CoreRt50;
+                case Version v when v.Major == 6 && v.Minor == 0: return CoreRt50;
```
AndreyAkinshin, line 64: CoreRt50 should be replaced with CoreRt60

AndreyAkinshin (dismissed):
LGTM

MichalStrehovsky:
Sigh. Not known. Apparently https://github.com/dotnet/runtime/pull/34633 introduced new dependencies that we need to compensate for.

Could you please try if dropping

```xml
    <ItemGroup>
      <LinkerArg Include="-lanl" />
    </ItemGroup>
```

into the project file that is invoking the compiler fixes it? We'll need to put that into the targets that the compiler carries with it.

Apparently we don't test anything that uses networking in the CI.

adamsitnik:
@MichalStrehovsky thanks a lot!

AndreyAkinshin (approved):

AndreyAkinshin (approved):

PR #1610, opened 2020-12-08
Fix add missing culuture info to ToString in RatioStyle

JohannesDeml:
Fixes the following test on machines with different culture info: RatioPrecisionTestWithBaseline(testDataKey: "Trend")


adamsitnik (approved):
LGTM, thank you @JohannesDeml !

PR #1612, opened 2020-12-09
Sorting parameter columns with parameter priority

### LINKED ISSUE #1150: Order of parameter columns changes with different Params* attribute usage
In trying to force an `enum` parameter to be output in an order different than that of the `enum` values (which I think might be #660) I found that the order of the parameter columns in the summary table is not consistent between different usages of the `Params`, `ParamsAllValues`, and `ParamsSource` attributes.  For example, with this code...

    public enum SomeEnum
    {
        First,
        Second,
        Third
    }

    [DryJob()]
    public class Benchmarks
    {
        [Params(1, 10, 100)]
        public int X
        {
            get; set;
        }

        [ParamsAllValues()]
        public SomeEnum Y
        {
            get; set;
        }

        [Benchmark()]
        public void Method1()
        {
        }

        [Benchmark()]
        public void Method2()
        {
        }
    }

    static class Program
    {
        static void Main()
        {
            BenchmarkDotNet.Running.BenchmarkRunner.Run<Benchmarks>();
        }
    }

...I get this output (`Mean` and `Error` columns omitted)...

    |  Method |   X |      Y |
    |-------- |---- |------- |
    | Method1 |   1 |  First |
    | Method2 |   1 |  First |
    | Method1 |   1 | Second |
    | Method2 |   1 | Second |
    | Method1 |   1 |  Third |
    | Method2 |   1 |  Third |
    | Method1 |  10 |  First |
    | Method2 |  10 |  First |
    | Method1 |  10 | Second |
    | Method2 |  10 | Second |
    | Method1 |  10 |  Third |
    | Method2 |  10 |  Third |
    | Method1 | 100 |  First |
    | Method2 | 100 |  First |
    | Method1 | 100 | Second |
    | Method2 | 100 | Second |
    | Method1 | 100 |  Third |
    | Method2 | 100 |  Third |

In most cases that's the output I get, however if I change `Y` from a property to a field and apply the `[Params(...)]` attribute...

    [Params(SomeEnum.First, SomeEnum.Second, SomeEnum.Third)]
    public SomeEnum Y;

...then the two parameter columns get swapped...

    |  Method |      Y |   X |
    |-------- |------- |---- |
    | Method1 |  First |   1 |
    | Method2 |  First |   1 |
    | Method1 |  First |  10 |
    | Method2 |  First |  10 |
    | Method1 |  First | 100 |
    | Method2 |  First | 100 |
    | Method1 | Second |   1 |
    | Method2 | Second |   1 |
    | Method1 | Second |  10 |
    | Method2 | Second |  10 |
    | Method1 | Second | 100 |
    | Method2 | Second | 100 |
    | Method1 |  Third |   1 |
    | Method2 |  Third |   1 |
    | Method1 |  Third |  10 |
    | Method2 |  Third |  10 |
    | Method1 |  Third | 100 |
    | Method2 |  Third | 100 |

I get the same result when `[ParamsSource(...)]` is applied to `X` and `Y` is a property or field with the `[Params(...)]` attribute...

    public static int[] XValues { get; } = new[] { 1, 10, 100 };

    [ParamsSource(nameof(XValues))]
    public int X
    {
        get; set;
    }

    [Params(SomeEnum.First, SomeEnum.Second, SomeEnum.Third)]
    public SomeEnum Y
    {
        get; set;
    }

...or when `[ParamsSource(...)]` is applied to `X` and `Y` is a field with the `[ParamsSource(...)]` attribute...

    public static int[] XValues { get; } = new[] { 1, 10, 100 };

    [ParamsSource(nameof(XValues))]
    public int X
    {
        get; set;
    }

    public static SomeEnum[] YValues { get; } = new[] { SomeEnum.First, SomeEnum.Second, SomeEnum.Third };

    [ParamsSource(nameof(YValues))]
    public SomeEnum Y;

Here's a summary of the different permutations and the resulting column order...

| X member type |    X attribute   | Y member type |     Y attribute     | Parameter columns |
|:-------------:|:----------------:|:-------------:|:-------------------:|:-----------------:|
|    Property   |    `Params()`    |    Property   | `ParamsAllValues()` |        X, Y       |
|    Property   |    `Params()`    |     Field     | `ParamsAllValues()` |        X, Y       |
|    Property   |    `Params()`    |    Property   |      `Params()`     |        X, Y       |
|    Property   |    `Params()`    |     Field     |      `Params()`     |      **Y, X**     |
|    Property   |    `Params()`    |    Property   |   `ParamsSource()`  |        X, Y       |
|    Property   |    `Params()`    |     Field     |   `ParamsSource()`  |        X, Y       |
|    Property   | `ParamsSource()` |    Property   | `ParamsAllValues()` |        X, Y       |
|    Property   | `ParamsSource()` |     Field     | `ParamsAllValues()` |        X, Y       |
|    Property   | `ParamsSource()` |    Property   |      `Params()`     |      **Y, X**     |
|    Property   | `ParamsSource()` |     Field     |      `Params()`     |      **Y, X**     |
|    Property   | `ParamsSource()` |    Property   |   `ParamsSource()`  |        X, Y       |
|    Property   | `ParamsSource()` |     Field     |   `ParamsSource()`  |      **Y, X**     |

JohannesDeml:
This is a draft for adding the possiblity to sort the parameter columns throuah an attribute property. This fixes #1150 

You can see the functionality in `IntroParamsPriority.cs`.

- [x] `ArgumentsAttribute`
- [x] `ArgumentsSourceAttribute`
- [x] `ParamsAllValuesAttribute`
- [x] `ParamsAttribute`
- [x] `ParamsSourceAttribute`

Currently params are sorted in `BenchmarkConverter.GetParameterDefinitions()`, which has the limitation, that they can't be intertwined with argument sorting. I think the best way for sorting would be to somehow pass the priority to the column which would then be already sorted in the end through `SummaryExtensions.GetColumns()`, but I didn't find the code, where I would be able to control the creation of columns for the params / attributes.

## TODO
- [x] Support Params
- [x] Params example
- [x] Params documentation
- [x] Support Attributes
- [x] Attributes example
- [x] Attributes documentation
- [x] Sorting test


JohannesDeml:
@adamsitnik I guess it's best to continue the conversation here. I went ahead and named the Property `Priority` for consistency. However I do think Oder might be a more fitting name in general. With Priority I would expect a higher priority to come first, but the columns are sorted from lowest to highest priority value. For order on the other hand I would expect a ascending sorting.

As for setting the priority value for IColumn: Can you point me to the code entry point, where columns for params are defined?
And do you think it is necessary to add tests for sorting?

adamsitnik:
> However I do think Oder might be a more fitting name in general

To be honest with you I also prefer the `Order` name as `Priority` can be a little confusing. Perhaps we should go with `DisplayOrder` name? @AndreyAkinshin do you have any thoughts on that?

> As for setting the priority value for IColumn: Can you point me to the code entry point, where columns for params are defined?

https://github.com/dotnet/BenchmarkDotNet/blob/10abc4f4daa5142f0257c9d512a38afc2ab28be1/src/BenchmarkDotNet/Columns/DefaultColumnProvider.cs#L75-L82

> And do you think it is necessary to add tests for sorting?

JohannesDeml:
> https://github.com/dotnet/BenchmarkDotNet/blob/10abc4f4daa5142f0257c9d512a38afc2ab28be1/src/BenchmarkDotNet/Columns/DefaultColumnProvider.cs#L75-L82

Thanks, that's exactly the method I was searching for! The priority is now passed to `IColum.PriorityInCategory` where it belonged in the first place. Now it is also able to sort arguments and the priority does not care if it is a param or an argument as one would expect. I also added documentation for both arguments and params.
 
> It's not mandatory, more like a nice to have. You could write a test that mocks a summary with 3 parameters and uses
> 
> https://github.com/dotnet/BenchmarkDotNet/blob/10abc4f4daa5142f0257c9d512a38afc2ab28be1/src/BenchmarkDotNet/Columns/DefaultColumnProvider.cs#L17
> and verifies that the output is sorted

Sounds good!

adamsitnik (approved):
LGTM, thank you very much @JohannesDeml !

I was wondering if `DisplayOrder` would be a better name, but since `IColumn` already defines [PriorityInCategory](
https://github.com/dotnet/BenchmarkDotNet/blob/10abc4f4daa5142f0257c9d512a38afc2ab28be1/src/BenchmarkDotNet/Columns/IColumn.cs#L37-L40) property, it's better to keep the naming consistent and use `Priority` as you did.

adamsitnik (approved):

JohannesDeml:
Thanks a lot for the review and the additional fixes. And thanks for guiding me, I'm really happy we now have a way to sort parameter colums! 🎉

Issue #1512, opened 2020-08-08
Template has hardcoded netcoreapp3.0 TFM


am11:
Using dotnet 5.0 preview 8, the template produces netcoreapp3.0 hardcoded csproj: 

```sh
$ dotnet new --version
5.0.100-preview.8.20362.3

$ dotnet new --install BenchmarkDotNet.Templates::0.12.1
$ dotnet new benchmark --console-app -n MathBenchmarks
$ cd MathBenchmarks
$ head -3 MathBenchmarks.csproj
```

outputs:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netcoreapp3.0</TargetFramework>
```

running it with `dotnet run` gives:

```sh
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '3.0.0' was not found.
  - The following frameworks were found:
      5.0.0-preview.8.20361.2 at [/Users/am11/.dotnet/shared/Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=3.0.0&arch=x64&rid=osx.10.14-x64
```

running after fixing csproj by hand, by replacing netcoreapp3.0 with net5.0, results in the following exception:

```sh
Unhandled exception. System.NotSupportedException: Unknown .NET Runtime
   at BenchmarkDotNet.Portability.RuntimeInformation.GetCurrentRuntime()
   at BenchmarkDotNet.Characteristics.Resolver.<>c__DisplayClass1_0`1.<Register>b__0(CharacteristicObject obj)
   at BenchmarkDotNet.Characteristics.Resolver.Resolve[T](CharacteristicObject obj, Characteristic`1 characteristic)
   at BenchmarkDotNet.Characteristics.CompositeResolver.Resolve[T](CharacteristicObject obj, Characteristic`1 characteristic)
   at BenchmarkDotNet.Characteristics.CharacteristicObject.ResolveValue[T](Characteristic`1 characteristic, IResolver resolver)
   at BenchmarkDotNet.Toolchains.ToolchainExtensions.GetToolchain(Job job, Descriptor descriptor)
   at BenchmarkDotNet.Toolchains.ToolchainExtensions.GetToolchain(BenchmarkCase benchmarkCase)
[...]

PR #1614, opened 2020-12-10
Update templates to default console app to net5.0

kendaleiv:


ExceptionCaught:
hi @adamsitnik  and @kendaleiv , this PR  also relate to the issue #1512. 
However, I think a more approachable way is to take on the target framework based on the `-f` option with a default value of `.net5.0`. 
If we agree so, I would start implementing on the PR for that issue.

adamsitnik:
> I think a more approachable way is to take on the target framework based on the -f option with a default value of .net5.0

I like this idea!

adamsitnik:
Closing in favour of #1630 which added the possibility to read it from the provided arguments

PR #1631, opened 2021-01-07
don't redirect Standard Error, as we don't read it and writing to it by benchmark can cause deadlocks

### LINKED ISSUE #1629: Writing to Console.Error in benchmarked code causes a deadlock
When writing to `Console.Error` within code being benchmarked, the benchmark runner deadlocks. 

Just a hunch but this could be caused by a common gotcha reading redirected stdin and stdout streams: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=net-5.0#remarks

Here's a trivial repro:

```csharp
using System;
using BenchmarkDotNet.Jobs;

namespace benchmarks
{
    class Program
    {
        {
            BenchmarkRunner.Run<Benchmarks>();
        }
    }

    [Config(typeof(Config))]
    public class Benchmarks
    {
        {
            public Config()
            {
                AddJob(Job.Default);
            }
        }

        [Benchmark]
        public void DEADLOCK()
        {
            Console.Error.Write("hi!");
        }
    }
}

```

adamsitnik:
fixes #1629

@AndreyAkinshin PTAL

cc @jonsequitur

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommandExecutor.cs
+++ b/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommandExecutor.cs
@@ -98,7 +98,6 @@ namespace BenchmarkDotNet.Toolchains.DotNetCli
                 startInfo.StandardErrorEncoding = Encoding.UTF8;
             }
 
-
             if (environmentVariables != null)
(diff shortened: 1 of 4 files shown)


AndreyAkinshin (approved):
LGTM

PR #1630, opened 2021-01-07
Update framework symbol on dotnet new template #1512

ExceptionCaught:
This is to address the issue #1512.
.template.json file as follow.
* Additional [generated switch](https://github.com/dotnet/templating/wiki/Available-Symbols-Generators#switch)  symbol `frameworksDefaults` to calculate the default values when `--console-app` is `true` or `false`
* [Generated join](https://github.com/dotnet/templating/wiki/Available-Symbols-Generators#join) with the original `frameworks` to take its value if its not empty.

PR #1632, opened 2021-01-07
update console app templating

ExceptionCaught:
This is to address the missing output type as mentioned [here](https://github.com/dotnet/BenchmarkDotNet/pull/1630/files#r553350990).
Based on the test on my local. the original issue #1512 is actually fixed by bumping up the referenced version to `0.12.1`.

### DIFF
--- a/templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/_BenchmarkProjectName_.csproj
+++ b/templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/_BenchmarkProjectName_.csproj
@@ -1,10 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup Condition="'$(consoleApp)' == 'true'">
-    <TargetFramework>$(Frameworks)</TargetFramework>
+    <TargetFramework>$(Framework)</TargetFramework>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
   <PropertyGroup Condition="'$(consoleApp)' == 'false'">
-    <TargetFrameworks>$(Frameworks)</TargetFrameworks>
+    <TargetFramework>$(Framework)</TargetFramework>
  </PropertyGroup>
   <PropertyGroup>
     <PlatformTarget>AnyCPU</PlatformTarget>
--- a/templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/_BenchmarkProjectName_.csproj
+++ b/templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/_BenchmarkProjectName_.csproj
@@ -1,7 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
-    <TargetFrameworks>$(Frameworks)</TargetFrameworks>
+  <PropertyGroup Condition="'$(consoleApp)' == 'true'">
+    <TargetFramework>$(Frameworks)</TargetFramework>
+    <OutputType>Exe</OutputType>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(consoleApp)' == 'false'">
+    <TargetFrameworks>$(Frameworks)</TargetFrameworks>
+ </PropertyGroup>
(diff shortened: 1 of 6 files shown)


ExceptionCaught:
@adamsitnik and @am11, can you also verify the finding regard to upgrading the referenced version to 0.12.1?

am11:
Thanks. Yes it needs 0.12.1 for net5.0 support, as 0.12.0 lacks it.

ExceptionCaught:
I had a look a the error of Ubuntu build. Not sure the exact root cause, but this is the log. It has been failing consistently for the last four builds.
image

templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/_BenchmarkProjectName_.csproj
```diff
@@ -1,7 +1,11 @@
...
-    <TargetFrameworks>$(Frameworks)</TargetFrameworks>
+  <PropertyGroup Condition="'$(consoleApp)' == 'true'">
+    <TargetFramework>$(Frameworks)</TargetFramework>
+    <OutputType>Exe</OutputType>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(consoleApp)' == 'false'">
+    <TargetFrameworks>$(Frameworks)</TargetFrameworks>
+ </PropertyGroup>
   <PropertyGroup>
     <PlatformTarget>AnyCPU</PlatformTarget>
```
adamsitnik, line 10: Should we always have a single target framework?

```suggestion
  <PropertyGroup Condition="'$(consoleApp)' == 'true'">
    <OutputType>Exe</OutputType>
  </PropertyGroup>
  <PropertyGroup>
    <TargetFramework>$(Frameworks)</TargetFramework>
    <PlatformTarget>AnyCPU</PlatformTarget>
```

ExceptionCaught: hi @adamsitnik , it could be multiple target frameworks.
However, `dotnet run` will throw error without the framework option if there were multiple target frameworks, which may not be ideal for user experience. 
Plus, I guess it will be rare that developers want to have multiple target frameworks when building a console app to run benchmark. unless I was wrong?

adamsitnik: Can user specify multiple target frameworks when creating a new project from template? If not, we should simplify it and users can always modify it later.

am11: I agree with @adamsitnik. Templates like `dotnet new console` or `dotnet new classlib` (and the projects created by VS templates) just offer `<TargetFramework>` (singular). Users can choose to modify project files based on requirements.

ps: on the topic of keeping things simple, it would be very intuitive for `dotnet` cli users  if `dotnet new benchmark.console` and `dotnet new benchmark.classlib` are separated, rather than expecting `--console-app true` with classlib variant as default. At the very least, we should prefer `--console` flag (presence means yes, absence means no) instead of two-worded option with mandatory boolean value. (we are still in version 0.x, so breaking changes will not be frowned upon by end users)

ExceptionCaught: Hi @adamsitnik, I think for the classlib, it supports multiple templates, and some user would choose to do so. 
but as @am11 mentioned, the `console` and `classlib` ones from Microsoft generates `TargetFramework`. 
I will made the change, all with single TargetFramework only. 
And I think it would be better to address the address the suggestion from @am11 on separate templates in different pull request.

ExceptionCaught: hi @adamsitnik  and @am11 , template are updated with single `TargetFramework` for both console and classlib per discussion.

adamsitnik (approved):
LGTM, thank you @ExceptionCaught !

PR #1633, opened 2021-01-08
Update etwprofiler.md

MendelMonteiro:
`With` is obsolete and `AddDiagnoser` should be used instead.


adamsitnik (approved):
@MendelMonteiro thank you!

PR #1635, opened 2021-01-09
Install local SDK without sudo on Unix

am11:
When using `build.sh` on Unix, it prompts for user password upon installation of the .NET SDK. That even contradicts the .NET SDK installer instruction:

```sh
$ ./build.sh
Installing .NET CLI...
Password:
dotnet-install: Note that the intended use of this script is for Continuous Integration (CI) scenarios, where:
dotnet-install: - The SDK needs to be installed without user interaction and without admin rights.
```
(line 2 and 4 indicate the said contradiction)

We don't need administrator's privilege to install the SDK in `{repo root direcotry}/.dotnet`.


adamsitnik (approved):
thank you @am11 !

PR #1651, opened 2021-02-19
Ability to configure benchmark class name in outputs

### LINKED ISSUE #1447: Ability to configure benchmark class name in outputs
An individual benchmark method's name can be overridden for output purposes via the Description property on BenchmarkAttribute (see #1243 for a related issue on this). However, there doesn't seem to be a way to do the same for the entire class.

My use case: we have a superclass of benchmarks being extended for different databases, so an abstract OpenConnectionBenchmarks class gets extended by SqliteOpenConnectionBenchmarks. However, in the outputs (CSV going into a database) it's desirable to have OpenConnectionBenchmarks, as the database is tracked elsewhere.

cdonke:
Implementation of attribute to rename class.

image
image
image
image

Closes #1447

### DIFF
--- a/src/BenchmarkDotNet.Annotations/Attributes/BenchmarkNameAttribute.cs
+++ b/src/BenchmarkDotNet.Annotations/Attributes/BenchmarkNameAttribute.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BenchmarkDotNet.Attributes
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class BenchmarkNameAttribute : Attribute
+    {
+        public BenchmarkNameAttribute(){        }
+        public BenchmarkNameAttribute(string name)
+            => Name = name;
+
+        public string Name { get; set; }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/BenchmarkDotNet.Annotations/Attributes/BenchmarkNameAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BenchmarkDotNet.Attributes
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class BenchmarkNameAttribute : Attribute
```
timcassell, line 8: I think `DescriptionAttribute` would be a better name. It matches the `Description` property of `BenchmarkAttribute`.

AndreyAkinshin:
Here are some ideas from my side:
1. I agree with @timcassell, `BenchmarkName` can be a confusing name in such a situation. I would suggest `BenchmarkDescription`.
2. If we introduce such an attribute, it makes sense to "extract" the `Description` property from the `[Benchmark]` attribute to be consistent and annotate both methods and classes with a new attribute.
3. Currently, a benchmark description overrides the method name. I would suggest splitting the original class/method names and their descriptions in the exporters so that we always have access to the original names during csv/json/xml/etc parsing.
4. In the Summary table I would suggest using "adaptive" default column that shows a description if it's available and the original name otherwise.

PR #1667, opened 2021-03-12
Changed default debug type to pdbonly

YohDeadfall:
Since `DisassemblyDiagnoser` requires `pdbonly` it makes sense to use it by default.


adamsitnik (approved):
LGTM, thank you @YohDeadfall !

PR #1679, opened 2021-03-28
Fix location for NativeAOT publish files

kant2002:
Even if native folder contains generated output, that location contains only ompiled C# code, without additional native dependencies.
I project has Microsoft.Data.SqlClient dependency, then benchmark would not run, since it cannot find Microsoft.Data.SqlClient.SNI.dll
Proper location is to look for publish folder, where all nescessary files to run application copied.


adamsitnik:
> Proper location is to look for publish folder, where all nescessary files to run application copied.

@MichalStrehovsky can you confirm this?

MichalStrehovsky:
> @MichalStrehovsky can you confirm this?

👍

adamsitnik (approved):
LGTM, thank you @kant2002 !

PR #1689, opened 2021-04-16
Dont redirect standard input for WASM

naricc:
This is needed to address this issue in dotnet/performance: https://github.com/dotnet/performance/issues/1720

To reiterate what is said there, since wasm processes cannot and is not reading from standard in, they do not wait for acknowledgements. This means the process can race ahead, and finish before the acknowledgement is even sent, at which point it will fail due to a closed pipe.

This change simply ignores the exception in that case, and prints an error.

### DIFF
--- a/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
+++ b/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
@@ -3,7 +3,9 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using BenchmarkDotNet.Diagnosers;
 using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Running;
+
 
 namespace BenchmarkDotNet.Loggers
 {
@@ -18,7 +20,7 @@ namespace BenchmarkDotNet.Loggers
         {
             if (!process.StartInfo.RedirectStandardOutput)
                 throw new NotSupportedException("set RedirectStandardOutput to true first");
-            if (!process.StartInfo.RedirectStandardInput)
+            if (!(process.StartInfo.RedirectStandardInput || benchmarkCase.GetRuntime() is WasmRuntime))
                 throw new NotSupportedException("set RedirectStandardInput to true first");
 
             this.logger = logger;
@@ -54,7 +56,11 @@ namespace BenchmarkDotNet.Loggers
                 else if (Engine.Signals.TryGetSignal(line, out var signal))
                 {
                     diagnoser?.Handle(signal, diagnoserActionParameters);
-                    process.StandardInput.WriteLine(Engine.Signals.Acknowledgment);
+
+                    if (process.StartInfo.RedirectStandardInput)
+                    {
+                        process.StandardInput.WriteLine(Engine.Signals.Acknowledgment);
+                    }
 
                     if (signal == HostSignal.AfterAll)
                     {
--- a/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
+++ b/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
@@ -6,7 +6,6 @@ using BenchmarkDotNet.Engines;
 using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Running;
 
-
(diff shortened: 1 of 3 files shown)


naricc:
@adamsitnik

src/BenchmarkDotNet/Toolchains/Executor.cs
```diff
@@ -66,8 +66,17 @@ private ExecuteResult Execute(Process process, BenchmarkCase benchmarkCase, Sync
...
                 process.TrySetAffinity(benchmarkCase.Job.Environment.Affinity, logger);
             }
+            Runtime runtime = GetRuntime(benchmarkCase);
 
-            loggerWithDiagnoser.ProcessInput();
+            try
+            {
+                loggerWithDiagnoser.ProcessInput();
+            }
+            catch (System.IO.IOException) when (runtime is WasmRuntime)
```
adamsitnik, line 75: This `catch` block is catching every exception, which means that we could lose some important information (imagine that something else in this method fails)

adamsitnik (changes requested):
It would be better to just don't redirect the standard input for WASM:

https://github.com/dotnet/BenchmarkDotNet/blob/b0ad0ecdb326bef5290872c4d81f2999ec97d238/src/BenchmarkDotNet/Toolchains/Executor.cs#L107

https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs#L21-L22

and just don't use it:

https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs#L57

src/BenchmarkDotNet/Toolchains/Executor.cs
naricc: Removed.

src/BenchmarkDotNet/Loggers/SynchronousProcessOutputLoggerWithDiagnoser.cs
```diff
@@ -3,8 +3,10 @@
 using System.Diagnostics;
 using BenchmarkDotNet.Diagnosers;
 using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Running;
 
+
```
adamsitnik, line 9:
```suggestion
```

src/BenchmarkDotNet/Toolchains/Executor.cs
```diff
@@ -98,9 +98,8 @@ private ProcessStartInfo CreateStartInfo(BenchmarkCase benchmarkCase, ArtifactsP
 
             string exePath = artifactsPaths.ExecutablePath;
 
-            var runtime = benchmarkCase.Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic)
-                ? benchmarkCase.Job.Environment.Runtime
-                : RuntimeInformation.GetCurrentRuntime();
+            var runtime = benchmarkCase.GetRuntime();
+
```
adamsitnik, line 102:
```suggestion
```

src/BenchmarkDotNet/Running/BenchmarkCase.cs
```diff
@@ -26,6 +28,13 @@ private BenchmarkCase(Descriptor descriptor, Job job, ParameterInstances paramet
             Config = config;
         }
 
+        public Runtime GetRuntime()
+        {
+            return Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic)
+                              ? Job.Environment.Runtime
+                              : RuntimeInformation.GetCurrentRuntime();
+        }
+
```
adamsitnik, line 37: nit: since other methods in this type use expression bodies, this one could be as well (for consistency)
```suggestion
        public Runtime GetRuntime() => Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic)
                ? Job.Environment.Runtime
                : RuntimeInformation.GetCurrentRuntime();

```

adamsitnik (dismissed):
LGTM, thank you @naricc !

adamsitnik (approved):

adamsitnik:
[0.12.1.1528](https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/builds/38784063)

PR #1698, opened 2021-05-03
Update "View results" from "Getting started"

rstm-sf:
Update "View results" from "Getting started"

The current output is misleading. Copy/Past from https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/docs/articles/overview.md


adamsitnik:
> The current output is misleading

Why is that? The method names are correct.

rstm-sf:
This output shows the Allocated. If we follow step by step, then there will be no such it and we need to add the attribute

adamsitnik:
> This output shows the Allocated. If we follow step by step, then there will be no such it and we need to add the attribute

Now I get it. Thank you!

PR #1702, opened 2021-05-07
fixes(configuration): Not unique exporter for exporter type

### LINKED ISSUE #1700: [Bug] Not unique exporter for exporter type
## Description

If the same exporter is defined in several levels (assembly and class) with different parameters, they are not merged.

## Steps to Reproduce

1. Download attachment
2. Decompress
3. Run
4. At the end of benchmark you can see in Console output like this:
   ```shell
   // * Export *
   BenchmarkDotNet.Artifacts\results\TestMultipleExportIssue.SimpleBenchmark-report.csv
   BenchmarkDotNet.Artifacts\results\TestMultipleExportIssue.SimpleBenchmark-report-github.md
   BenchmarkDotNet.Artifacts\results\TestMultipleExportIssue.SimpleBenchmark-report.html
   CustomExporter - Benchmark
   CustomExporter - Assembly
   ```

## Expected Behavior
Console output like this:
```shell
// * Export *
BenchmarkDotNet.Artifacts\results\TestMultipleExportIssue.SimpleBenchmark-report.csv
BenchmarkDotNet.Artifacts\results\TestMultipleExportIssue.SimpleBenchmark-report-github.md
BenchmarkDotNet.Artifacts\results\TestMultipleExportIssue.SimpleBenchmark-report.html
CustomExporter - Benchmark
```

Version: [0.12.1](https://github.com/dotnet/BenchmarkDotNet/releases/tag/v0.12.1)

### Attachment
[TestMultipleExportIssue.zip](https://github.com/dotnet/BenchmarkDotNet/files/6440779/TestMultipleExportIssue.zip)

workgroupengineering:
fixes #1700 

## This PR 
- [x] Don't contain the break change
- [x] It has an integration test

### DIFF
--- a/src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
+++ b/src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
@@ -38,7 +38,7 @@ namespace BenchmarkDotNet.Configs
 
             var uniqueHardwareCounters = source.GetHardwareCounters().ToImmutableHashSet();
             var uniqueDiagnosers = GetDiagnosers(source.GetDiagnosers(), uniqueHardwareCounters);
-            var uniqueExporters = GetExporters(source.GetExporters(), uniqueDiagnosers);
+            var uniqueExporters = GetExporters(source.GetExporters(), uniqueDiagnosers, uniqueLoggers);
             var uniqueAnalyzers = GetAnalysers(source.GetAnalysers(), uniqueDiagnosers);
 
             var uniqueValidators = GetValidators(source.GetValidators(), MandatoryValidators, source.Options);
@@ -88,7 +88,9 @@ namespace BenchmarkDotNet.Configs
             return builder.ToImmutable();
         }
 
-        private static ImmutableArray<IExporter> GetExporters(IEnumerable<IExporter> exporters, ImmutableHashSet<IDiagnoser> uniqueDiagnosers)
+        private static ImmutableArray<IExporter> GetExporters(IEnumerable<IExporter> exporters,
+            ImmutableHashSet<IDiagnoser> uniqueDiagnosers,
+            IEnumerable<Loggers.ILogger> loggers = null)
         {
 
             var mergeDictionary = new Dictionary<System.Type, IExporter>();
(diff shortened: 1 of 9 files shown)


src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
```diff
@@ -111,6 +113,27 @@ private static ImmutableArray<IExporter> GetExporters(IEnumerable<IExporter> exp
             for (int i = result.Count - 1; i >=0; i--)
                 if (result[i] is IExporterDependencies exporterDependencies)
                     foreach (var dependency in exporterDependencies.Dependencies)
+                        /*
+                         * I don't know how to correct
```
workgroupengineering, line 117: @adamsitnik Can you give me a suggestion how to solve?

src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
```diff
@@ -111,8 +115,43 @@ private static ImmutableArray<IExporter> GetExporters(IEnumerable<IExporter> exp
             for (int i = result.Count - 1; i >=0; i--)
                 if (result[i] is IExporterDependencies exporterDependencies)
                     foreach (var dependency in exporterDependencies.Dependencies)
-                        if (!result.Contains(dependency))
+                        /*
+                         *  When exporter that depends on an other already present in the configuration print warning.
+                         *
+                         *  Example:
+                         *
+                         *  // Global Current Culture separetor is Semicolon;
```
AndreyAkinshin, line 123: Typo: `separetor` -> `separator`

src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
```diff
@@ -111,8 +115,43 @@ private static ImmutableArray<IExporter> GetExporters(IEnumerable<IExporter> exp
...
+                         *  [CsvMeasurementsExporter(CsvSeparator.Comma)] // force use Comma
+                         *  [RPlotExporter]
+                         *  public class MYBanch
```
AndreyAkinshin, line 126: Typo: `MYBanch` -> `MyBench`

tests/BenchmarkDotNet.Tests/Loggers/DelegateLogger.cs
```diff
@@ -0,0 +1,28 @@
+﻿using BenchmarkDotNet.Loggers;
+using System;
+
+namespace BenchmarkDotNet.Tests.Loggers
+{
+    internal class DelegateLogger : ILogger
+    {
+        private readonly Action<LogKind, string> _writeAction;
```
AndreyAkinshin, line 8: In BenchmarkDotNet, we don't use the underscore prefix for private fields. Please, rename it to `writeAction`.

src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
```diff
@@ -88,18 +88,22 @@ private static ImmutableHashSet<IDiagnoser> GetDiagnosers(IEnumerable<IDiagnoser
...
+            IEnumerable<Loggers.ILogger> loggers = null)
         {
-            var result = new List<IExporter>();
+
+            var mergeDictionary = new Dictionary<System.Type, IExporter>();
 
             foreach (var exporter in exporters)
-                if (!result.Contains(exporter))
-                    result.Add(exporter);
+                mergeDictionary[exporter.GetType()] = exporter;
```
AndreyAkinshin, line 99: If `exporters` contains two exporters of the same type, the first one will be overwritten with the second one without a warning.

AndreyAkinshin (changes requested):
I don't think that the logging approach is the best way to report a warning in this case. In most cases, BenchmarkDotNet users read only the bottom part of the log (with the summary table) and don't scroll up. Thus, this warning will be most likely lost.
On the other hand, we could define `IAnalyser` which reports such a warning. The `IAnalyser` approach has two advantages:
1. It prints all the warning below the summary table (so the warnings are more visible for the users)
2. The list of the warnings could be fetched programmatically (so it's possible to introduce an assertion that the benchmarks don't produce any warnings).

However, this particular case is tricky because we have to report warning on the stage of config building (not at the final stage of the `BenchmarkReport` analysis). I guess, we could introduce an `IReadOnlyList<Conclusion>` property in `IConfig` and use it to report issues related to malformed config in a structured way.

src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
workgroupengineering: Before correcting, I await your decision your decision as to whether it is better to use the Analyzer.

workgroupengineering: I have a idea. We can create a LoggerAnalyser thath implement ILogger and IAnalyser. LoggerAnalyser collect all message have logKind equal Error or Warning. This way we don't need to add any new property to IConfig.

AndreyAkinshin: I don't think it's a good idea to aggregate ALL the errors/warnings under the summary table. Let's say we can't build a specific benchmark. In this case, we dump the full error message into the log. Howerver, at the end of the log, we want to just print a short message "We can't build a benchmark"  without the full build error message (which would be too overwhelmed).

workgroupengineering: So instead of collecting warnings and errors we could count them. If they are greater than zero, we show warning.

workgroupengineering: I think it is better to show a warning the overwrite. When an unexpected result occurs, there are more indications to find the source of the problem.

workgroupengineering:
Please @AndreyAkinshin can you review? If the proposed changes are not to your liking. Please close this PR. :smile:

AndreyAkinshin:
Closed since #1796 looks better.

PR #1719, opened 2021-06-03
Added UsingBrowserRuntimeWorkload

naricc:
A check was added to the SDK to check for the wasm "workloads" to be installed when building wasm apps, but BDN doesn't actually need this. So this PR adds a msbuild property to disable  the check.

Fixes: https://github.com/dotnet/performance/issues/1823


naricc:
@adamsitnik

adamsitnik (approved):
LGTM, thank you for fixing the issue @naricc

PR #1707, opened 2021-05-14
Set ValidateExecutableReferencesMatchSelfContained to false in aotllvm template

naricc:
Set ValidateExecutableReferencesMatchSelfContained to false in aotllvm template. 

Fixes: https://github.com/dotnet/performance/issues/1804

PR #1724, opened 2021-06-13
Add ValidateExecutableReferencesMatchSelfContained

kant2002:
This is required to make CoreRT work on 5.0.300 SDK


adamsitnik:
@MichalStrehovsky could you please confirm that?

MichalStrehovsky:
If we can't compile with the .NET 5 SDK without this, the fix should probably go to the NativeAOT branch instead.

kant2002:
I create this PR because I find #1707. Isn't this means that same logic applied to Mono AOT?

adamsitnik:
#1707 was specific to Mono AOT. 

We have a [test](https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/tests/BenchmarkDotNet.IntegrationTests/CoreRtTests.cs#L17) that ensures that we support latest CoreRT version and our CI is green, which means that this change is not needed.

PR #1734, opened 2021-06-29
Added <WasmNativeWorkload>false</WasmNativeWorkload> to Wasm cs proj

naricc:
This is needed to prevent another check in the SDK that is not needed for BDN, and unblock an issue we are seeing in wasm-aot.


naricc:
@adamsitnik

adamsitnik (approved):
LGTM, but we should most probably either limit the amount of breaking changes introduced to WASM or make the toolchain more resilient or extensible (so we don't need a new BDN version to fix it)

adamsitnik:
@naricc thanks for the fix!

naricc:
@adamsitnik 

> LGTM, but we should most probably either limit the amount of breaking changes introduced to WASM or make the toolchain more resilient or extensible (so we don't need a new BDN version to fix it)

I agree. I am not too happy about all these breaking changes, but I haven't decided the best way to fix it yet.

PR #1739, opened 2021-07-09
Add support for returning unmanaged types from benchmarks

### LINKED ISSUE #1720: Error with benchmark returning a pointer
Having a method that returns a pointer leads to errors CS0208, CS0214 and CS0673 in generated code when running the benchmark.

Minimum reproduction:
```cs
using BenchmarkDotNet.Jobs;

namespace ConsoleApp4
{
	internal static class Program
	{
		private static void Main()
		{
			BenchmarkRunner.Run<Test>();
		}
	}

	[SimpleJob(RuntimeMoniker.Net50)]
	public unsafe class Test
	{
		[Benchmark]
		public void* Method()
		{
			return null;
		}
	}
}
```

The generated code (GitHub doesn't accept .notcs so I had to zip it): 
[b13b0d0e-b4af-451d-b284-f346ee112559.zip](https://github.com/dotnet/BenchmarkDotNet/files/6591537/b13b0d0e-b4af-451d-b284-f346ee112559.zip)

Using BenchmarkDotNet 0.13.0 from the official nuget.

kant2002:
- `System.Void*` is not supported as type in C#, so I map to `void*` in GetCorrectCSharpTypeName similarly to regular `void`
- `System.Void*` and other pointer types cannot be consumed, so I have to add two Consume functions. I do not mark them as `[PublicAPI]` since this is too would be much liberty from me.
- I mark whole autogenerated class as `unsafe` to not introduce special checks for unsafe types when generate code. Hopefully this is accepteable.

Closes #1720

### DIFF
--- a/src/BenchmarkDotNet/Engines/Consumer.cs
+++ b/src/BenchmarkDotNet/Engines/Consumer.cs
@@ -32,6 +32,7 @@ namespace BenchmarkDotNet.Engines
         private ulong ulongHolder;
         private string stringHolder;
         private object objectHolder;
+        private IntPtr ptrHolder;
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         [PublicAPI]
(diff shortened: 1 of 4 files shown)


adamsitnik (approved):
LGTM, big thanks for the fix @kant2002 !

adamsitnik:
the CI failures are not related, I am merging

PR #1744, opened 2021-07-21
Make mono/wasm run on Windows

radekdoulik:
Just log a warning about it. It eases development on Windows.

I tried to run a microbenchmark from the dotnet/performance repo and it
run the build phase just fine. I am not sure yet, whether it will run
fine, so I kept the log message there.


adamsitnik (approved):
LGTM, thank you @radekdoulik

PR #1746, opened 2021-07-22
Make PrepareForWasmBuild work with wasm workload

radekdoulik:
The `WasmBuildAppDependsOn` property gets overwritten by the wasm-tools
workload's https://github.com/dotnet/runtime/blob/main/src/mono/wasm/build/WasmApp.props#L9-L18

So instead, set the `WasmBuildAppAfterThisTarget` property to our
`PrepareForWasmBuild` target and make it run after `Publish` target.

The workload's `WasmApp` has condition on `WasmBuildAppAfterThisTarget`
and should not overwrite it.

### DIFF
--- a/src/BenchmarkDotNet/Templates/WasmAotCsProj.txt
+++ b/src/BenchmarkDotNet/Templates/WasmAotCsProj.txt
@@ -33,9 +33,9 @@
   </ItemGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


radekdoulik:
@adamsitnik could you please review this PR? (looks like I don't have permissions to set reviewers or merge approved PRs)

SamMonoRT (approved):

adamsitnik (approved):

Issue #1727, opened 2021-06-16
Unhelpful rounding in MemoryDiagnoser


stephentoub:
I have a benchmark that allocates and I'm using MemoryDiagnoser to see how much.  When the benchmark allocates 1520 bytes, MemoryDiagnoser by default shows it as "1K".  When the benchmark allocates 1624 bytes, MemoryDiagnoser shows that as "2K" bytes.  Those rounded values are huge divergences (25-35%) from the actual values, making it not particularly helpful, especially when using the benchmark to drive changes that individually end up changing allocation by tens of bytes at a time.  I'd prefer to see the allocation column always default to bytes, but short of that, rounding to whole numbers of kilobytes in a case like this is unfortunate.

PR #1749, opened 2021-07-25
Unhelpful rounding in MemoryDiagnoser #1727

TenCoKaciStromy:
Pull request for [Unhelpful rounding in MemoryDiagnoser](https://github.com/dotnet/BenchmarkDotNet/issues/1727).

There is arbitrary value 9KB which is "small enough" for printing bytes instead of kilobytes.
Maybe it is too much. I have no strong opinion about that.

### DIFF
--- a/src/BenchmarkDotNet/Columns/SizeUnit.cs
+++ b/src/BenchmarkDotNet/Columns/SizeUnit.cs
@@ -36,6 +36,14 @@ namespace BenchmarkDotNet.Columns
                 return B;
             // Use the largest unit to display the smallest recorded measurement without loss of precision.
             long minValue = values.Min();
+            long maxValue = values.Max();
+
+            const long maxSmallBytes = 9 * 1024;
+
+            if (minValue <= maxSmallBytes && maxValue <= maxSmallBytes) {
+              return B; // its useful to show small amount of bytes like bytes and not rounded to kilobytes if all bytes are small
+            }
+
             foreach (var sizeUnit in All)
             {
                 if (minValue < sizeUnit.ByteAmount * BytesInKiloByte)
--- a/src/BenchmarkDotNet/Columns/SizeUnit.cs
+++ b/src/BenchmarkDotNet/Columns/SizeUnit.cs
@@ -40,8 +40,9 @@ namespace BenchmarkDotNet.Columns
 
             const long maxSmallBytes = 9 * 1024;
 
-            if (minValue <= maxSmallBytes && maxValue <= maxSmallBytes) {
-              return B; // its useful to show small amount of bytes like bytes and not rounded to kilobytes if all bytes are small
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1749) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1749) before we can accept your contribution.<br/><br/>:x: TenCoKaciStromy [sign now](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1749)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/BenchmarkDotNet?pullRequest=1749) it.</sub>

src/BenchmarkDotNet/Columns/SizeUnit.cs
```diff
@@ -35,7 +35,16 @@ public static SizeUnit GetBestSizeUnit(params long[] values)
             if (!values.Any())
                 return B;
             // Use the largest unit to display the smallest recorded measurement without loss of precision.
-            long minValue = values.Min();
+            long minValue = values.Min();
+            long maxValue = values.Max();
+
+            const long maxSmallBytes = 9 * 1024;
```
timcassell, line 41: Why not `(10 * 1024) - 1`?

TenCoKaciStromy: There is no special reason. Even 9 KB seemed too much to me. But I have no strong opinion about that.

timcassell: Maybe a better solution would be to print fractional values if it's not measures in bytes.

TenCoKaciStromy: I'm not sure, if I understand right.
I'm new in this codebase, I thought the result is produced by https://github.com/dotnet/BenchmarkDotNet/blob/602d3da11524f842a581183f3fe62c92910c5519/tests/BenchmarkDotNet.Tests/SizeUnitTests.cs#L47 and this method already produce fractional numbers https://github.com/dotnet/BenchmarkDotNet/blob/602d3da11524f842a581183f3fe62c92910c5519/tests/BenchmarkDotNet.Tests/SizeUnitTests.cs#L39

Did I miss something or do you mean something else?

timcassell: I haven't looked at this particularly deeply myself, but the issue that this is trying to fix mentions that for 1KB to 2KB, it does not use fractional. So maybe there is a bug in there that should be reporting fractional that needs to be fixed.

PR #1747, opened 2021-07-22
Trigger build

radekdoulik:
Last 2 builds of main failed. Lets try empty commit to trigger another build.

PR #1750, opened 2021-07-26
Fix pointer-returning benchmarks support for InProcessToolchain

adamsitnik:
After I've merged #1739 one `MemoryDiagnoser` test started failing for `InProcessToolchain` I initially thought it's something related to `TieredJit` and ignored it (my bad).

It turned out that `InProcessToolchain` required a fix to make it work with #1739 

I've added tests for both `void*` and `T*` and ensured they now work for every toolchain (.NET, Core and InProcess).

This should make the CI green again cc @radekdoulik @kant2002

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/ValuesReturnedByBenchmarkTest.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ValuesReturnedByBenchmarkTest.cs
@@ -93,6 +93,9 @@ namespace BenchmarkDotNet.IntegrationTests
 
             [Benchmark]
             public NoNamespace TypeWithoutNamespace() => new NoNamespace();
+
+            [Benchmark]
+            public unsafe void* PointerToAnything() => IntPtr.Zero.ToPointer();
         }
     }
 }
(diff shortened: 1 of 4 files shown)


tests/BenchmarkDotNet.IntegrationTests/ValuesReturnedByBenchmarkTest.cs
```diff
@@ -93,6 +104,24 @@ public class Job { }
...
+            public System.IntPtr IntPtr() => System.IntPtr.Zero;
+
+            [Benchmark]
+            public System.UIntPtr UIntPtr() => System.UIntPtr.Zero;
+
+            [Benchmark]
+            public nint NativeSizeInteger() => 0;
+
+            [Benchmark]
+            public uint UnsignedNativeSizeInteger() => 0;
```
timcassell, line 124: Should be `nuint`?

adamsitnik: great catch, fixed!

adamsitnik:
Closing in favour of #1747

PR #1757, opened 2021-07-29
[wasm] Add WasmMainJSPath in interpreter projects

radekdoulik:
This should fix errors like:

    performance/tools/dotnet/x64/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/6.0.0-rc.1.21378.3/Sdk/WasmApp.targets(139,5): error : $(WasmMainJSPath) property needs to be set


adamsitnik (approved):
LGTM, thank you for the fix @radekdoulik

PR #1765, opened 2021-08-04
Copy GC settings from host process

adamsitnik:
We have been always recognizing the settings of the Host process:

https://github.com/dotnet/BenchmarkDotNet/blob/4bd433d85fff4fb6ba8c4f8df3e685ad669e2519/src/BenchmarkDotNet/Environments/GcResolver.cs#L12-L13

But for some reason we have never re-applied them to benchmark process. Which could cause a confusion in cases where users apply things like:

```xml
<ServerGarbageCollection>true</ServerGarbageCollection>
```

in their project file.

Reported offline by @stephentoub


AndreyAkinshin (approved):

stephentoub:
Thanks :)

PR #1769, opened 2021-08-10
Make wasm-interpreter work like wasm-aot

naricc:
This mostly unifies wasm-aot and wasm-interpreter, so they work largely the same way. It resolves some on-going issues with the existing wasm-interpreter toolchain.


naricc:
@adamsitnik

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -1,37 +1,33 @@
...
     <TargetFramework>$TFM$</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <AppDir>$(MSBuildThisFileDirectory)\bin\$TFM$\browser-wasm\publish</AppDir>
     <AssemblyName>$PROGRAMNAME$</AssemblyName>
     <RuntimeIdentifier>browser-wasm</RuntimeIdentifier>
-    <WasmMainJSPath>$MAINJS$</WasmMainJSPath>
-    <MicrosoftNetCoreAppRuntimePackDir>$RUNTIMEPACK$</MicrosoftNetCoreAppRuntimePackDir>
-    <UsingBrowserRuntimeWorkload>false</UsingBrowserRuntimeWorkload>
+    <PublishTrimmed>false</PublishTrimmed>
+    <SuppressTrimAnalysisWarnings>true</SuppressTrimAnalysisWarnings>
```
radekdoulik, line 14: Not sure whether we want to disable trimming above, is there an issue with linked assemblies?
 If so then `SuppressTrimAnalysisWarnings` property should not be needed as it is only used when trimming and running `illink`.

naricc: I disabled trimming because it can potentially cause problems, and we only *really* need it on wasm-aot to keep aot compile times down. 

But I can get rid of the SupressTrimAnalysisWarnings.

Lxiamail:
@naricc are we done with this PR?

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -425,7 +417,7 @@ private static Job MakeWasmJob(Job baseJob, CommandLineOptions options, TimeSpan
             bool wasmAot = options.AOTCompilerMode == MonoAotCompilerMode.wasm;
 
             var wasmRuntime = new WasmRuntime(
-                mainJs: options.WasmMainJs ?? new FileInfo(Path.Combine(options.RuntimeSrcDir.FullName, "src", "mono", "wasm", "runtime-test.js")),
+                mainJs: new FileInfo(Path.Combine(options.RuntimeSrcDir.FullName, "src", "mono", "wasm", "runtime-test.js")),
```
adamsitnik, line 420: is it being used anywhere now? It seems to be hardcoded in the project file:

```xml
<WasmMainJSPath>$(RuntimeSrcDir)\src\mono\wasm\runtime-test.js</WasmMainJSPath>
```

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -96,14 +96,7 @@ protected void GenerateProjectInterpreter(BuildPartition buildPartition, Artifac
 
         protected override string GetBinariesDirectoryPath(string buildArtifactsDirectoryPath, string configuration)
         {
-            if (Aot)
```
adamsitnik, line 99: I don't understand this change and I would like to be sure that we are not breaking anything. Have you tested both AOT and non-AOT toolchains?

naricc: Basically, before wasm-intepreter mode was using a  simplified custom app builder I built into BDN, because that was the best thing that could be done at the time. 

Wasm-aot used all the app bundling stuff from the the runtime source directory, which put things in a different place than the one I coded into BDN.

With this change, both wasm-interpreter and wasm-aot will use the app builder from the runtime source dir, so the files will end up in the same place.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
naricc: I removed it.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
naricc: I've tested both AOT and interpreter configurations.

src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
```diff
@@ -175,9 +175,6 @@ public class CommandLineOptions
         [Option("wasmEngine", Required = false, HelpText = "Full path to a java script engine used to run the benchmarks, used by Wasm toolchain.")]
         public FileInfo WasmJavascriptEngine { get; set; }
 
-        [Option("wasmMainJS", Required = false, HelpText = "Path to the main.js file used by Wasm toolchain. Mandatory when using \"--runtimes wasm\"")]
-        public FileInfo WasmMainJs { get; set; }
```
adamsitnik, line 179: Please keep in mind that once we delete this parameter and you update BDN in dotnet/performance everything that is still using this argument is going to get an error from the BDN config parser

adamsitnik (approved):
LGTM, thank you @naricc !

src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
naricc: Yes, I am aware. We have made changes in dotnet/performance so nothing is using it.

naricc:
@adamsitnik If you would be so kind as to merge it when the builds are green :)

Lxiamail:
create PR https://github.com/dotnet/performance/pull/1918 to have perf repo depends on this new version of bdn.

Issue #730, opened 2018-04-25
IterationSetup is not running before each benchmark invocation


stephentoub:
Repro using 0.10.14:
```C#
public class Test
{
    public static void Main() => BenchmarkRunner.Run<Test>();

    [IterationSetup]
    public void MySetup() => Console.WriteLine("MySetup");

    [Benchmark]
    public void MyBenchmark()
    {
        Console.WriteLine("MyBenchmark");
        Thread.Sleep(100);
    }
}
```

Example output from the main run:
```
MainTarget  1: 16 op, 1608514852.97 ns, 100.5322 ms/op
MySetup
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MainTarget  2: 16 op, 1605497716.92 ns, 100.3436 ms/op
MySetup
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MainTarget  3: 16 op, 1610182705.89 ns, 100.6364 ms/op
MySetup
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
MyBenchmark
```

This leads to very incorrect benchmark results when MyBenchmark is doing something that ends up caching data that can affect subsequent invocations of MyBenchmark.  The whole purpose of IterationSetup in my case is to ensure that cache is cleared before the benchmark is executed.

adamsitnik:
I have an idea for solving this problem:

Add new attribute: `[BenchmarkSetup]` which would be internally translated to `[IterationSetup]` with the right settings (Monitoring strategy etc)

@stephentoub would `[BenchmarkSetup]` name be less confusing?

stephentoub:
In what scenario is the current behavior useful? It just seems buggy to me.

AndreyAkinshin:
[...]

PR #1776, opened 2021-08-16
Clarify effects of IterationSetup on UnrollFactor and InvocationCount

SnakyBeaky:
As mentioned in https://github.com/dotnet/BenchmarkDotNet/issues/730#issuecomment-389971683

I think it would be useful to include that information in the documentation as well (effects on `UnrollFactor` and `InvocationCount`).


adamsitnik (approved):
LGTM, thank you @SnakyBeaky !

PR #1801, opened 2021-09-09
Include RatioStyle in SummaryStyle.Equals()/GetHashCode(). Fixes #1800.

### LINKED ISSUE #1800: Small bug in SummaryStyle
Commit 349e90084a666e39199ed4eff43f2bb5920a14fd introduced `RatioStyle` and adjusted constructor and fluent APIs accordingly, but not the `Equals()` and `GetHashCode()` implementations.
https://github.com/dotnet/BenchmarkDotNet/blob/156fd6928c166d2d67fe2262c7304da45503c9db/src/BenchmarkDotNet/Reports/SummaryStyle.cs#L62-L91

As far as I can tell, this is only used for the `Distinct().SingleOrDefault()` here:
https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Reports/Summary.cs#L154-L155

I can send a PR, if you like.

1. Side question
Do you guys prefer an issue to be opened for these kind of trivial bugs, or should one just send a PR?
2. Side note
The biggest hurdle in becoming acquainted with *BenchmarkDotNet* was trying to remember which of those pesky fluent APIs modifies the calling instance and which one returns a new instance leaving the caller unchanged. 😉

mawosoft:


adamsitnik (approved):
LGTM, thank you @mawosoft !

PR #1805, opened 2021-09-12
Update PackageReference for System.Management to latest (5.0.0).

mawosoft:
- Also updated Microsoft.Win32.Registry to latest (5.0.0) since it is a direct dependency of the former.
- Fixes bugs in .Net Core >=3.1 builds in RuntimeInformation.GetAntivirusProducts() and GetVirtualMachineHypervisor() where the use of ManagementObjectSearcher throws a silently eaten exception.

Theoretically, this would also have affected MosCpuInfoProvider.MosCpuInfo, but its use is limited to .Net Framework by the caller.

You can easily see this in previous CI logs where the net461 build shows VM=Hyper-V in the system info while the net5.0 build doesn't.


adamsitnik (approved):
LGTM, thank you @mawosoft !

PR #1807, opened 2021-09-12
Updated (almost) all package references to latest.

mawosoft:
- except: ApprovalTests and FxCopAnalyzers (deprecated).

This is an alternate to PR #1805 (bugfix, commit also included here) with some additional, manual housekeeping - trying to do it automatically via Nuget Manager went terribly wrong.

Not sure how comfortable you are with updating these wholesale, but I want to at least see if it passes the CI checks this time. If they pass, I guess it's a coin flip between having outdated refs causing subtle problems that don't immediately catch the eye like the ManagementObjectSearcher thing or implicitly fixing those, but introducing new ones.


mawosoft:
Well, it would have been nice. I'm closing this for now and might take a look at a later date.

PR #1815, opened 2021-09-27
fixes: detection of IEnumerable<object[]> in SmartArgument.ToSourceCode();

### LINKED ISSUE #1813: Cannot run Benchmark with ArgumentsSource and MemoryDiagnoser
## Step to reproduce
1. Edit IntroNativeMemory like this:
   ```csharp
    public class ImageSize
    {
        public int Width { get; set; }
        public int Height { get; set; }
    }


    [ShortRunJob]
    [NativeMemoryProfiler]
    [MemoryDiagnoser]
    public class IntroNativeMemory
    {

        public IEnumerable<object[]> SizeSource()
        {
            yield return new object[] { new ImageSize() { Width = 200, Height = 100 } };
            yield return new object[] { new ImageSize() { Width = 400, Height = 200 } };
        }

        [Benchmark]
        [ArgumentsSource(nameof(SizeSource))]
        public void BitmapWithLeaks(ImageSize size)
        {
            var flag = new Bitmap(size.Width, size.Height);
            var graphics = Graphics.FromImage(flag);
            var blackPen = new Pen(Color.Black, 3);
            graphics.DrawLine(blackPen, 100, 100, 500, 100);
        }
       // other code
     }
   ```
2. Run Benck
3. Will be diplay buil error like this:
    `1aa2555a-d975-4aba-ba82-9d94e026af0a.notcs(3136,27): error CS0030: Cannot convert type 'object[]' to 'BenchmarkDotNet.Samples.ImageSize'`

P.S. : this is a stupid example just to show the error. The real one is more complex.

workgroupengineering:
Fixes #1813

### DIFF
--- a/src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
+++ b/src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
@@ -93,11 +93,24 @@ namespace BenchmarkDotNet.Parameters
         {
             string cast = $"({parameterDefinitions[argumentIndex].ParameterType.GetCorrectCSharpTypeName()})"; // it's an object so we need to cast it to the right type
 
-            string callPostfix = source is PropertyInfo ? string.Empty : "()";
+            string callPostfix = string.Empty;
+            string indexPostfix = string.Empty;
+            Type returnType = default;
 
-            string indexPostfix = parameterDefinitions.Length > 1
-                ? $"[{argumentIndex}]" // IEnumerable<object[]>
-                : string.Empty; // IEnumerable<object>
+            if (source is PropertyInfo propertyInfo)
+            {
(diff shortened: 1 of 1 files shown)


AndreyAkinshin (changes requested):
The suggested fix breaks the following benchmark:
```cs
[DryJob]
public class Foo
{
    public IEnumerable<object[]> Provider()
    {
        yield return new object[] { 1, 2 };
        yield return new object[] { 1, 2, 3 };
    }

    [Benchmark]
    [ArgumentsSource(nameof(Provider))]
    public void Benchmark(object[] array)
    {
        Thread.Sleep(array.Length * 100);
    }
}
```
See https://github.com/dotnet/BenchmarkDotNet/issues/1813#issuecomment-929196645 for details.

PR #1816, opened 2021-09-27
Add net70 runtime support

am11:
Runtime has switched to net7.0. This PR adds support for 7.0 in BDN.

IMO, we should look into removing hard coded mechanism (enums, --tfm / --cli options etc.). There are just too many redundant moving parts and it is not scaling well.

### DIFF
--- a/src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
+++ b/src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
@@ -34,7 +34,11 @@ namespace BenchmarkDotNet.Environments
         /// <summary>
         /// CoreRT compiled as net6.0
         /// </summary>
-        public static readonly CoreRtRuntime CoreRt60 = new CoreRtRuntime(RuntimeMoniker.CoreRt50, "net6.0", "CoreRT 6.0");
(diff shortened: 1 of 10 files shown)


src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs
```diff
@@ -34,7 +34,11 @@ public class CoreRtRuntime : Runtime
         /// <summary>
         /// CoreRT compiled as net6.0
         /// </summary>
-        public static readonly CoreRtRuntime CoreRt60 = new CoreRtRuntime(RuntimeMoniker.CoreRt50, "net6.0", "CoreRT 6.0");
+        public static readonly CoreRtRuntime CoreRt60 = new CoreRtRuntime(RuntimeMoniker.CoreRt60, "net6.0", "CoreRT 6.0");
```
am11, line 37: This seemed like a bug so I fixed it.

AndreyAkinshin (approved):

AndreyAkinshin:
@am11 thanks for the PR!

> IMO, we should look into removing hard coded mechanism (enums, --tfm / --cli options etc.)

Well, it's not an easy problem. I agree that we have a decent number of different TFMs. However, this set is still maintainable. Meanwhile, hard coded values in API help to easily discover available options. Also they provide a strictly typed way to work with TFMs.

@adamsitnik what do you think on this?

PR #1817, opened 2021-10-07
Pass args to runner in console template

am11:
Upgraded console template to pass args tot he runner, as it gets a bit confusing when the passed args have no effect on the generated project.


AndreyAkinshin:
I think we should wait with such a change until .NET 6 is officially released.

AndreyAkinshin:
> Regarding version, I think ideally the template should pick up the latest SDK

Do you have any idea how to do this using the existing templating engine?

> I can repurpose the PR and drop version revving.

Sounds good.

AndreyAkinshin:
Closed in favor of #2338

PR #1824, opened 2021-10-12
Explicitly stop the current EventPipeSession after all tasks

ronbrogan:
Adding the EventPipeProfiler diagnoser to a benchmark and executing benchmarks in-process causes the diagnoser to never finish, appearing as the entire benchmark session has hung. 

This can be reproduced with the following setup:

```csharp
using BenchmarkDotNet.Diagnosers;

BenchmarkRunner.Run<Benchmarks>(new DebugInProcessConfig());

[EventPipeProfiler(EventPipeProfile.CpuSampling)]
public class Benchmarks
{
    [Benchmark]
    public void Work()
    {
        Thread.Sleep(1);
    }
}

```

This PR introduces handling of `HostSignal.AfterAll` to explicitly stop the EventPipeSession.

I don't see any tests exercising this area, but this fix works for my scenarios.

### DIFF
--- a/src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
+++ b/src/BenchmarkDotNet/Diagnosers/EventPipeProfiler.cs
@@ -29,6 +29,7 @@ namespace BenchmarkDotNet.Diagnosers
         private readonly bool performExtraBenchmarksRun;
 
         private Task collectingTask;
+        private EventPipeSession currentSession = null;
 
         // parameterless constructor required by DiagnosersLoader to support creating this profiler via console line args
         // we use performExtraBenchmarksRun = false for better first user experience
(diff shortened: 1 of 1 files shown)


adamsitnik:
@WojciechNagorski could you PTAL?

WojciechNagorski:
@adamsitnik I've never tried `EventPipeProfiler` with `InProcess`. 
In normal mode (not inProcess) it works without closing the session explicitly, I mean without:
```c# 
currentSession?.Stop();
```
because the `exe` process is ended and it causes the end of the session.
I am afraid that after the change, not all data will be saved before closing the session. But I'm not sure, and I don't have enough time to check it. 
@ronbrogan Can you compare the file that was generated without and with your change but without using `InProcess`. Especially the end of the file.

WojciechNagorski:
`EtwProfiler` doesn't support `InProcessToolchain` and I don't think `EventPipeProfiler` should. @adamsitnik what do you think?

PR #1835, opened 2021-11-03
Enable mono llvmaot tool chain to work with net7

naricc:
This change lets the llvmaot tool chain work with net7 (and, in principle, netX for any X), and also allows selecting net6. 

Also add a .net7 nuget feed.

Fixes: https://github.com/dotnet/performance/issues/2100


naricc:
@adamsitnik

adamsitnik (approved):
LGTM, thank you @naricc !

PR #1852, opened 2021-11-28
Fix size rounding

### LINKED ISSUE #1727: Unhelpful rounding in MemoryDiagnoser

YegorStepanov:
Fixes #1727.

For the benchmark:
```C#
[MemoryDiagnoser, DisassemblyDiagnoser]
public class Benchmarks
{
    [Benchmark] public int[] B1() => Enumerable.Range(0, 400).Where(i => i % 2 == 0).Select(i => i).ToArray();
    [Benchmark] public byte[] B2() => new byte[1024 * (1024 + 512)]; // 1.5 MB + 24b overhead
}
``` 
The current export result is:
| Method | Code Size | Allocated |
|--------- |------------:|-----------:|
|         B1 |          2 KB |         2 KB |
|         B2 |          0 KB |  1,536 **KB** |

The new result is:
| Method | Code Size |    Allocated |
|--------- |------------:|-------------:|
|         B1 |      1,772 B |       2,376 B |
|         B2 |          41 B | 1,572,888 B |

The rounding to specific `SizeUnit` can be returned by:
```C#
BenchmarkRunner.Run<Benchmarks>(
    ManualConfig.CreateMinimumViable().WithSummaryStyle(
            SummaryStyle.Default.WithSizeUnit(BenchmarkDotNet.Columns.SizeUnit.KB)));
```

If the rounding is pointless, should be opened another PR for removing/deprecated public `SummaryStyle.WithSizeUnit()` and public `SizeUnit`, `SizeValue` classes that are responsible for converting bytes?

https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Columns/SizeUnit.cs#L9
https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Columns/SizeValue.cs#L8

### DIFF
--- a/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
+++ b/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
@@ -165,7 +165,7 @@ namespace BenchmarkDotNet.Diagnosers
             public string Id => "Native Code Size";
             public string DisplayName => "Code Size";
             public string Legend => "Native code size of the disassembled method(s)";
-            public string NumberFormat => "N0";
+            public string NumberFormat => "0.##";
             public UnitType UnitType => UnitType.Size;
             public string Unit => SizeUnit.B.Name;
             public bool TheGreaterTheBetter => false;
--- a/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
+++ b/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
@@ -165,8 +165,8 @@ namespace BenchmarkDotNet.Diagnosers
             public string Id => "Native Code Size";
             public string DisplayName => "Code Size";
             public string Legend => "Native code size of the disassembled method(s)";
-            public string NumberFormat => "0.##";
-            public UnitType UnitType => UnitType.Size;
+            public string NumberFormat => "N0";
+            public UnitType UnitType => UnitType.CodeSize;
             public string Unit => SizeUnit.B.Name;
             public bool TheGreaterTheBetter => false;
             public int PriorityInCategory => 0;
--- a/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
+++ b/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
@@ -22,6 +22,7 @@ namespace BenchmarkDotNet.Diagnosers
     public class DisassemblyDiagnoser : IDiagnoser
     {
         private static readonly Lazy<string> ptrace_scope = new Lazy<string>(() => ProcessHelper.RunAndReadOutput("cat", "/proc/sys/kernel/yama/ptrace_scope").Trim());
+        private static readonly NativeCodeSizeMetricDescriptor Descriptor = new NativeCodeSizeMetricDescriptor();
 
         private readonly WindowsDisassembler windowsDisassembler;
         private readonly LinuxDisassembler linuxDisassembler;
@@ -52,7 +53,7 @@ namespace BenchmarkDotNet.Diagnosers
         public IEnumerable<Metric> ProcessResults(DiagnoserResults diagnoserResults)
         {
             if (results.TryGetValue(diagnoserResults.BenchmarkCase, out var disassemblyResult))
-                yield return new Metric(NativeCodeSizeMetricDescriptor.Instance, SumNativeCodeSize(disassemblyResult));
+                yield return new Metric(Descriptor, SumNativeCodeSize(disassemblyResult));
         }
 
         public RunMode GetRunMode(BenchmarkCase benchmarkCase)
@@ -157,19 +158,5 @@ namespace BenchmarkDotNet.Diagnosers
 
         private static long SumNativeCodeSize(DisassemblyResult disassembly)
             => disassembly.Methods.Sum(method => method.Maps.Sum(map => map.SourceCodes.OfType<Asm>().Sum(asm => asm.Instruction.Length)));
-
-        private class NativeCodeSizeMetricDescriptor : IMetricDescriptor
-        {
-            internal static readonly IMetricDescriptor Instance = new NativeCodeSizeMetricDescriptor();
-
-            public string Id => "Native Code Size";
-            public string DisplayName => "Code Size";
-            public string Legend => "Native code size of the disassembled method(s)";
-            public string NumberFormat => "0.##";
-            public UnitType UnitType => UnitType.Size;
-            public string Unit => SizeUnit.B.Name;
-            public bool TheGreaterTheBetter => false;
-            public int PriorityInCategory => 0;
-        }
     }
 }
--- a/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
+++ b/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
@@ -21,8 +21,6 @@ namespace BenchmarkDotNet.Diagnosers
 {
     public class DisassemblyDiagnoser : IDiagnoser
     {
-        internal static string DescriptorId { get; } = NativeCodeSizeMetricDescriptor.Instance.Id;
-        
         private static readonly Lazy<string> ptrace_scope = new Lazy<string>(() => ProcessHelper.RunAndReadOutput("cat", "/proc/sys/kernel/yama/ptrace_scope").Trim());
 
         private readonly WindowsDisassembler windowsDisassembler;
--- a/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
+++ b/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
@@ -21,6 +21,8 @@ namespace BenchmarkDotNet.Diagnosers
 {
     public class DisassemblyDiagnoser : IDiagnoser
     {
+        internal static string DescriptorId { get; } = NativeCodeSizeMetricDescriptor.Instance.Id;
+        
         private static readonly Lazy<string> ptrace_scope = new Lazy<string>(() => ProcessHelper.RunAndReadOutput("cat", "/proc/sys/kernel/yama/ptrace_scope").Trim());
 
         private readonly WindowsDisassembler windowsDisassembler;
@@ -165,8 +167,8 @@ namespace BenchmarkDotNet.Diagnosers
             public string Id => "Native Code Size";
             public string DisplayName => "Code Size";
             public string Legend => "Native code size of the disassembled method(s)";
-            public string NumberFormat => "N0";
-            public UnitType UnitType => UnitType.CodeSize;
+            public string NumberFormat => "0.##";
+            public UnitType UnitType => UnitType.Size;
             public string Unit => SizeUnit.B.Name;
             public bool TheGreaterTheBetter => false;
             public int PriorityInCategory => 0;
--- a/src/BenchmarkDotNet/Columns/MetricColumn.cs
+++ b/src/BenchmarkDotNet/Columns/MetricColumn.cs
@@ -1,4 +1,5 @@
 ﻿using System.Linq;
+using BenchmarkDotNet.Diagnosers;
 using BenchmarkDotNet.Reports;
 using BenchmarkDotNet.Running;
 using Perfolizer.Horology;
@@ -32,7 +33,7 @@ namespace BenchmarkDotNet.Columns
                 return "-";
 
             var cultureInfo = summary.GetCultureInfo();
-            if (style.PrintUnitsInContent && descriptor.UnitType == UnitType.CodeSize)
+            if (style.PrintUnitsInContent && descriptor.UnitType == UnitType.Size && descriptor.Id == DisassemblyDiagnoser.DescriptorId)
                 return SizeValue.FromBytes((long)metric.Value).ToString(style.CodeSizeUnit, cultureInfo, descriptor.NumberFormat);
             if (style.PrintUnitsInContent && descriptor.UnitType == UnitType.Size)
                 return SizeValue.FromBytes((long)metric.Value).ToString(style.SizeUnit, cultureInfo, descriptor.NumberFormat);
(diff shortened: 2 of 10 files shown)


mawosoft:
I'm not convinced this issue (#1727) really is one.

Currently, the size-rounding behaves like the time-rounding, determining the "best" unit by default, which can be overwritten by picking a preference via SummaryStyle.WithSizeUnit/TimeUnit().

Changing that behavior seems counterintuitive to me. It also requires more work for a user to restore the current default than picking a fixed unit as preference.

>If the rounding is pointless,

I don't think so, but that's indeed the question that should be answered first.

timcassell:
Like I mentioned in https://github.com/dotnet/BenchmarkDotNet/pull/1749, I believe the issue is that fractional values are not displayed between 1 and 2. Though it seems they do at higher values like 9.5kb. I agree with @mawosoft that the default behavior for unit selection shouldn't change.

YegorStepanov:
@mawosoft, The idea was:
1. change to byte always
2. there are a way to get "wrong" rounding behavior, there are no choice, deprecated it
3. there are public classes, that do nothing, refactor and make it internal
4. breaking change

@timcassell, It's integer always.

I thought byte-only is correct and easy to read:
1. It's hard to mix up `1,000 B` with `1,000,000 B`, but much easier `1,000 B` with `1,000 KB` after a lot of benchmarks.
2. byte-perfect precision. 
3. it's not clear the fraction part is divined by `1000` or `1024`. 
4. rounding are not necessary for benchmarking at all 

Thank you, I'll fix it.

If `Allocated` is `KB`, it forces `Code Size` to `KB`. but not vice versa.
Should `Code Size` column always be byte?

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1852) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1852) before we can accept your contribution.<br/><br/>:x: YegorStepanov [sign now](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1852)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/BenchmarkDotNet?pullRequest=1852) it.</sub>

YegorStepanov:
For benchmarks:
```C#
    BenchmarkRunner.Run<B>();
    BenchmarkRunner.Run<KB>();
    BenchmarkRunner.Run<MB>();
    BenchmarkRunner.Run<KBKB>();

    [DryJob, MemoryDiagnoser, DisassemblyDiagnoser]
    public class B
    {
        [Benchmark] public byte[] A1() => new byte[200];
    }

    [DryJob, MemoryDiagnoser, DisassemblyDiagnoser]
    public class KB
    {
        [Benchmark] public byte[] A2() => new byte[2_000];
    }

    [DryJob, MemoryDiagnoser, DisassemblyDiagnoser]
    public class MB
    {
        [Benchmark] public byte[] A3() => new byte[2_000_000];
    }

    [DryJob, MemoryDiagnoser, DisassemblyDiagnoser]
    public class KBKB
    {
        [Benchmark] public byte[] A7() => new byte[2_000].Select(x => (byte)x).Take(10_000).Skip(1).ToArray().ToList().ToArray();
    }
```

master:
| Method | Code Size | Allocated |
|------- |----------:|----------:|
|      K |      41 B |     224 B |
|     KB |      0 KB |      2 KB |
|     MB |      0 MB |      2 MB |
|   KBKB |      2 KB |      8 KB |

"Change the number specifier to digit placeholder" commit:
| Method | Code Size | Allocated |
|------- |----------:|----------:|
|      B |      41 B |     224 B |
|     KB |   0.04 KB |   1.98 KB |
|     MB |      0 MB |   1.91 MB |
|   KBKB |   1.53 KB |   8.16 KB |

"Decouple Code Size unit from Allocated Memory" commit:
| Method | Code Size | Allocated |
|------- |----------:|----------:|
|      K |      41 B |     224 B |
|     KB |      41 B |   1.98 KB |
|     MB |      41 B |   1.91 MB |
|   KBKB |   1,566 B |   8.16 KB |

Changing specifiers only don't solve the problem completely.

`Allocated` sets [`SummaryStyle.SizeUnit`,](https://github.com/dotnet/BenchmarkDotNet/blob/d312edcbf96ad37586119520ad6b10cbab9a95b2/src/BenchmarkDotNet/Reports/SummaryStyle.cs#L21) but both columns use it. I've decoupled `CodeSize` by duplication `SizeUnit` without [`SummaryStyle.WithSizeUnit`](https://github.com/dotnet/BenchmarkDotNet/blob/d312edcbf96ad37586119520ad6b10cbab9a95b2/src/BenchmarkDotNet/Reports/SummaryStyle.cs#L47) setter and added `UnitType.CodeSize` to distinguish SizeUnits.

timcassell:
That looks good to me, but why doesn't the CodeSize display as `1.53 KB` instead of `1,566 B` for consistency?

src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
```diff
@@ -166,7 +166,7 @@ private class NativeCodeSizeMetricDescriptor : IMetricDescriptor
             public string DisplayName => "Code Size";
             public string Legend => "Native code size of the disassembled method(s)";
             public string NumberFormat => "N0";
```
timcassell, line 168: `"0.##"` for consistency with memory?

YegorStepanov: In the second today commit I changed it to correct double(second table), but in the third commit I've back it back(third table), because `CodeSize` is always integer.

timcassell: Allocations are always integer also, they are both measured in bytes. I think they should be consistent with their output. I like what you did with the fractional values in KB/MB/etc and I think the CodeSize should do the same.

YegorStepanov:
If the `CodeSize` should be rounding, then I should make CodeSize public and add setter-method for `SummaryStyle`?

timcassell:
I'm actually curious why the CodeSize and Allocations currently have their units coupled. I think they should be decoupled without adding a new enum value. Each column should be independent. @adamsitnik any input?

YegorStepanov:
I'm worried about that: (`A` without 24b)
Screenshot 2021-11-30 021325
It's easy to read. But if we add 8 byte, we move to the KB rounding, and it's a magic now:
Screenshot 2021-11-30 021439

YegorStepanov:
What about unimportant digits? 
Currently, `0.####`:
Screenshot 2021-11-30 022028

vs `N4`:
image

vs 'N2':
image

I like `N2`, it's readable. If the minimal alignment is 8b. We get the following precision for `1024 + 8*i`. 
1.00 1.01 1.02 1.02 1.03 1.04 1.05 1.05 1.06 1.07 1.08 1.09 1.09 1.10...
 I don't see 3 equal numbers in the whole row.

timcassell:
I agree, I like `N2`.

YegorStepanov:
Both columns are `N2`:
image

byte-only and `N2`:
image

If both are `N2`:
1. After 1024, The accuracy for `CodeSize` is `~10b` (it's 10 times more then minimal, `1b`). `CodeSize` can be used to measure the performance of code size: There is some algorithm. What if all properties replace with fields?  `if` with branchless `if`? It's impossible for microbenchmark, but if `CodeSize` is decreased, something happened!
2. The rounded value is written to the logs, so if you want to see byte size when CodeSize > 1024,  the setter should be added to `SummaryStyle` to force byte-only.
3. Need write code 😞
4. I think the size of code with `DisassemblyDiagnoser` is rarely `> 1024b`. But if some methods exceed this value, all others become KB. 

I have no data about typical usage of `DisassemblyDiagnoser`, it seems, the extra precision will not harm!

timcassell:
> 1. After 1024, The accuracy for `CodeSize` is `~10b` (it's 10 times more then minimal, `1b`). `CodeSize` can be used to measure the performance of code size: There is some algorithm. What if all properties replace with fields?  `if` with branchless `if`? It's impossible for microbenchmark, but if `CodeSize` is decreased, something happened!

I think it's fine for precision to diminish with larger values.

> 2. The rounded value is written to the logs, so if you want to see byte size when CodeSize > 1024,  the setter should be added to `SummaryStyle` to force byte-only.

I also think it's fine for users to force byte-only if they want, and auto-choose units by default.

> 3. Need write code disappointed

Maybe add a field to the attribute to specify unit? I don't think it's such a big deal, though.

> 4. I think the size of code with `DisassemblyDiagnoser` is rarely `> 1024b`. But if some methods exceed this value, all others become KB.

Code Size can easily exceed 1KB with recursionDepth set to a value larger than 1, depending on what you're benchmarking. Also, the default unit selection goes by the least common denominator, so if there is a single value < 1KB, it will use `B`, not `KB`.

> I have no data about typical usage of `DisassemblyDiagnoser`, it seems, the extra precision will not harm!

I also do not have data for that, but my own benchmarks are in the 10-25 KB Code Size range (with recursionDepth set high), and I wouldn't mind seeing the table display in `KB`.

YegorStepanov:
Decoupling sizes without `UnitType.CodeSize` are very hard:

There are two places where we should distinguish sizes:
https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Extensions/CommonExtensions.cs#L15-L31

https://github.com/dotnet/BenchmarkDotNet/blob/f372668e028161d0d1bf675811a6168967175034/src/BenchmarkDotNet/Columns/MetricColumn.cs#L29-L41

Here, we should find all `CodeSize` and round them in the best way.
The other columns take the data from [`summary`](https://github.com/dotnet/BenchmarkDotNet/blob/7d5a8ff0308201cff3e48b1173585984095e44cf/src/BenchmarkDotNet/Reports/Summary.cs#L17), but it doesn't know anything about `DisAsm`, except [`Summary.Reports.Metric`](https://github.com/dotnet/BenchmarkDotNet/blob/f372668e028161d0d1bf675811a6168967175034/src/BenchmarkDotNet/Reports/Metric.cs#L7), the `Metric.Key` is `IMetricDescriptor.Key`, the `Metric.Value` is `IMetricDescriptor`, so we should someway distinguish [NativeCodeSizeMetricDescriptor](https://github.com/dotnet/BenchmarkDotNet/blob/f372668e028161d0d1bf675811a6168967175034/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs#L161) from [Memory](https://github.com/dotnet/BenchmarkDotNet/blob/f372668e028161d0d1bf675811a6168967175034/src/BenchmarkDotNet/Diagnosers/MemoryDiagnoser.cs#L48), but it's a private class (besides, except for `ID`, they are equals)
https://github.com/dotnet/BenchmarkDotNet/blob/4bd433d85fff4fb6ba8c4f8df3e685ad669e2519/src/BenchmarkDotNet/Reports/SummaryTable.cs#L45-L54

I found the only way to do this: by static `DisassemblyDiagnoser.DescriptorId`, it's ugly but solves all the problems. I do not know how to do it in another way.

So there are a few options:
1. fix only `N0` to `0.##` (not 'N2', because it gives `0.00 B`). The `CodeSize` is dependent from `Memory`, for `KB` it's ok, for `>MB` the user should force `B`/`KB` rounding.
2. Make beautiful code less beautiful by a few lines, but the rounding works correctly.

timcassell:
> 1. fix only `N0` to `0.##` (not 'N2', because it gives `0.00 B`). The `CodeSize` is dependent from `Memory`, for `KB` it's ok, for `>MB` the user should force `B`/`KB` rounding.

0 B should display `-` by default, so I think we should still use `N2` (I also don't think `0.00 B` is bad):
```cs
if (!summary.HasReport(benchmarkCase) || !summary[benchmarkCase].Metrics.TryGetValue(descriptor.Id, out Metric metric) || (metric.Value == 0.0 && !style.PrintZeroValuesInContent))
    return "-";
```

> Decoupling sizes without `UnitType.CodeSize` are very hard:

I'm actually not entirely convinced that it's in the scope of the issue, and should maybe be brought up as a separate issue. This PR could just change `N0` to `N2` for memory and code size, and do the decoupling separately.

timcassell:
> 1. fix only `N0` to `0.##` (not 'N2', because it gives `0.00 B`). The `CodeSize` is dependent from `Memory`, for `KB` it's ok, for `>MB` the user should force `B`/`KB` rounding.

Could do this:

```cs
public string NumberFormat => Unit == "B" ? "N0" : "N2";
```

YegorStepanov:
Okey, at the end of the discussion, I will create 2 pure PR and close this.

Nice idea about enum, but it generates a warning :)

Is `Allocation` naming equal here to 'AllocatedSize'? `SummaryTable.CodeSize` should be renamed too, but AllocatedSizeUnit is long.

I found the one internal implementation of `IMetricDescriptor`, it seems I can do it too. I hope it's acceptable to extract sizes from `Metrics`.

timcassell:
> Nice idea about enum, but it generates a warning :)

Oh yeah, I wrote that in browser. I guess way without warning would look like this:
```cs
public enum UnitType
{
    Dimensionless,
    Time,
    AllocatedSize,
    [Obsolete("Use AllocatedSize")]
    Size = AllocatedSize,
    CodeSize
}
```

> Is `Allocation` naming equal here to 'AllocatedSize'? `SummaryTable.CodeSize` should be renamed too, but AllocatedSizeUnit is long.

Yeah, I'm not picky about `Allocated` vs `Allocation`. Or even just `Alloc`. ;)

YegorStepanov:
I should have guessed myself to it. Yes, you are right.
Thank you for the great review!
The time of this PR is over.

PR #1854, opened 2021-12-02
Fix size specifier

YegorStepanov:
The split pull request of #1852. The Part 1.

master:
| Method | Code Size | Allocated |
|------- |----------:|----------:|
|      A |      0 KB |      1 KB |

| Method | Code Size | Allocated |
|------- |----------:|----------:|
|      C |   1,120 B |     432 B |

PR:
| Method | Code Size | Allocated |
|------- |----------:|----------:|
|      A |   0.04 KB |   1.49 KB |

| Method | Code Size | Allocated |
|------- |----------:|----------:|
|      C |    1120 B |     432 B |

Benchmarks:
```C#
    [MemoryDiagnoser, DisassemblyDiagnoser]
    public class Benchmark1
    {
        [Benchmark] public byte[] A() => new byte[1500];
    }

    [MemoryDiagnoser, DisassemblyDiagnoser]
    public class Benchmark2
    {
        [Benchmark] public byte[] C() => new byte[100].Select(x => x).Reverse().ToArray();
    }
```

### DIFF
--- a/src/BenchmarkDotNet/Diagnosers/MemoryDiagnoser.cs
+++ b/src/BenchmarkDotNet/Diagnosers/MemoryDiagnoser.cs
@@ -52,7 +52,7 @@ namespace BenchmarkDotNet.Diagnosers
             public string Id => "Allocated Memory";
             public string DisplayName => "Allocated";
             public string Legend => "Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)";
-            public string NumberFormat => "N0";
+            public string NumberFormat => "0.##";
             public UnitType UnitType => UnitType.Size;
             public string Unit => SizeUnit.B.Name;
             public bool TheGreaterTheBetter => false;
--- a/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
+++ b/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs
@@ -165,7 +165,7 @@ namespace BenchmarkDotNet.Diagnosers
             public string Id => "Native Code Size";
             public string DisplayName => "Code Size";
             public string Legend => "Native code size of the disassembled method(s)";
-            public string NumberFormat => "N0";
+            public string NumberFormat => "0.##";
             public UnitType UnitType => UnitType.Size;
             public string Unit => SizeUnit.B.Name;
             public bool TheGreaterTheBetter => false;


adamsitnik (approved):
It's definitely an improvement. In a long term we should consider displaying the disassembly size always in bytes.

thank you for your contribution @YegorStepanov !

YegorStepanov:
Thank you!

It's super easy: We should revert a half of this commit :) and distinguish `UnitType.Size(for Disassembly)` from `UnitType.Size(for Memory)` in these 2 places:

https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Extensions/CommonExtensions.cs#L15-L31

https://github.com/dotnet/BenchmarkDotNet/blob/f372668e028161d0d1bf675811a6168967175034/src/BenchmarkDotNet/Columns/MetricColumn.cs#L29-L41

There are 2 ways only: 
1) Add `UnitType.CodeSize`, but it will be used only 2 times in solution and it looks alien for `UnitType`.

2) A bit hacky: Add to [`DisassemblyDiagnoser`](https://github.com/dotnet/BenchmarkDotNet/blob/7b02117781f63b8be9dcf292ee21341dc5af5c07/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs)
```C#
    internal static string DescriptorId { get; } = NativeCodeSizeMetricDescriptor.Instance.Id;
```

We can now add 2 special cases:

```C#
case UnitType.Size when column.Id == DisassemblyDiagnoser.DescriptorId:
             return $"{column.ColumnName} [ B ]"; 
```
```C#
if (style.PrintUnitsInContent && descriptor.UnitType == UnitType.Size && descriptor.Id == DisassemblyDiagnoser.DescriptorId)
         return SizeValue.FromBytes((long)metric.Value).ToString(SizeUnit.B, cultureInfo, descriptor.NumberFormat); 
```

It easy to understand what is going on. And it will be easy to change in the future.

adamsitnik:
@YegorStepanov I would prefer to introduce a new enum.

YegorStepanov:
@adamsitnik, to clarify: you mean the new value in `UnitType` or about different enum structure?

adamsitnik:
@YegorStepanov a new value in UnitType

PR #1859, opened 2021-12-04
Add Allocation Ratio column

### LINKED ISSUE #722: Add scaled column for Allocated Memory

YegorStepanov:
fixes #722

image
image
image
It's an analog of [`BaselineRatioColumn`.](https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Columns/BaselineRatioColumn.cs)

Notes:

If it's a breaking change
```C#
    public virtual ColumnCategory Category => ColumnCategory.Baseline;
    //for 
    public override ColumnCategory Category => ColumnCategory.Metric; //it should be displayed after Allocated column

```

Then it can be done thought EIMI, but I think it's helpful extension for the future. 
```C#
    IColumn.ColumnCategory Category => ColumnCategory.Metric;
```

P.S. `ColumnCategory` is used for the outer sorting:
https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Reports/SummaryExtensions.cs#L13-L24

.

There are 2 ways how to find correct Metric: 
1) Move `AllocatedMemoryMetricDescriptor` from private to internal. There are exist few internal Descriptors (e.g [`AllocatedNativeMemoryDescriptor`](https://github.com/dotnet/BenchmarkDotNet/blob/f372668e028161d0d1bf675811a6168967175034/src/BenchmarkDotNet/Diagnosers/AllocatedNativeMemoryDescriptor.cs)). It creates one time per `Summary` creation now. `Equals` is called by `Id` in [`MetricDescriptorEqualityComparer`](https://github.com/dotnet/BenchmarkDotNet/blob/f372668e028161d0d1bf675811a6168967175034/src/BenchmarkDotNet/Reports/Metric.cs#L39-L47), so it seems to be safe.

Adam in #878 has changed `AllocatedMemoryMetricDescriptor.Id` from `typeof(AllocatedMemoryMetricDescriptor)` to `"Allocated Memory"`. So if class will be moved to new file, the history will be lost. Need add a comment explaining why only this Descriptor uses literal.

2) Incapsulate filtering by adding to `MemoryDiagnoser` (or to the new inner class in it) filter method):
```C#
public bool ICantThinkTheName(IMetricDescriptor d) => d is AllocatedMemoryMetricDescriptor;
```

....

The base class writes `?` to the cell when `Statistics` data are invalid (often for `Dry`, `Short`). I shouldn't inherit from this class?

https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Columns/BaselineCustomColumn.cs#L14-L29

### DIFF
--- a/src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
+++ b/src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
@@ -54,7 +54,7 @@ namespace BenchmarkDotNet.Columns
         {
             string logicalGroupKey = summary.GetLogicalGroupKey(benchmarkCase);
             var nonBaselines = summary.GetNonBaselines(logicalGroupKey);
-            return nonBaselines.Any(c => GetAllocationRatio(summary[c].Metrics, baselineMetric) < 0.01);
+            return nonBaselines.Any(c => GetAllocationRatio(summary[c].Metrics, baselineMetric) is > 0 and < 0.01);
         }
 
         private static double? GetAllocationRatio(
--- a/src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
+++ b/src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using BenchmarkDotNet.Diagnosers;
+using BenchmarkDotNet.Mathematics;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using JetBrains.Annotations;
+
+namespace BenchmarkDotNet.Columns
+{
+    public class BaselineAllocationRatioColumn : BaselineCustomColumn
+    {
+        public override string Id => nameof(BaselineAllocationRatioColumn);
+        public override string ColumnName => "Alloc Ratio";
+
+        public override string GetValue(Summary summary, BenchmarkCase benchmarkCase, Statistics baseline, IReadOnlyDictionary<string, Metric> baselineMetrics,
+            Statistics current, IReadOnlyDictionary<string, Metric> currentMetrics, bool isBaseline)
+        {
+            double? ratio = GetAllocationRatio(currentMetrics, baselineMetrics);
+            double? invertedRatio = GetAllocationRatio(baselineMetrics, currentMetrics);
+
+            if (ratio == null)
+                return "NA";
+
+            var cultureInfo = summary.GetCultureInfo();
+            var ratioStyle = summary?.Style?.RatioStyle ?? RatioStyle.Value;
+
+            bool advancedPrecision = IsNonBaselinesPrecise(summary, baselineMetrics, benchmarkCase);
+            switch (ratioStyle)
+            {
+                case RatioStyle.Value:
+                    return ratio.Value.ToString(advancedPrecision ? "N3" : "N2", cultureInfo);
+                case RatioStyle.Percentage:
+                    return isBaseline
+                        ? ""
+                        : ratio.Value >= 1.0
+                            ? "+" + ((ratio.Value - 1.0) * 100).ToString(advancedPrecision ? "N1" : "N0", cultureInfo) + "%"
+                            : "-" + ((1.0 - ratio.Value) * 100).ToString(advancedPrecision ? "N1" : "N0", cultureInfo) + "%";
+                case RatioStyle.Trend:
+                    return isBaseline
+                        ? ""
+                        : ratio.Value >= 1.0
+                            ? ratio.Value.ToString(advancedPrecision ? "N3" : "N2", cultureInfo) + "x more"
+                            : invertedRatio == null
+                                ? "NA"
+                                : invertedRatio.Value.ToString(advancedPrecision ? "N3" : "N2", cultureInfo) + "x less";
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(summary), ratioStyle, "RatioStyle is not supported");
+            }
+        }
+
+        private static bool IsNonBaselinesPrecise(Summary summary, IReadOnlyDictionary<string, Metric> baselineMetric, BenchmarkCase benchmarkCase)
+        {
+            string logicalGroupKey = summary.GetLogicalGroupKey(benchmarkCase);
+            var nonBaselines = summary.GetNonBaselines(logicalGroupKey);
+            return nonBaselines.Any(c => GetAllocationRatio(summary[c].Metrics, baselineMetric) < 0.01);
+        }
+
+        private static double? GetAllocationRatio(
+            [CanBeNull] IReadOnlyDictionary<string, Metric> current,
+            [CanBeNull] IReadOnlyDictionary<string, Metric> baseline)
+        {
+            double? currentBytes = GetAllocatedBytes(current);
+            double? baselineBytes = GetAllocatedBytes(baseline);
+
+            if (currentBytes == null || baselineBytes == null)
+                return null;
+
+            if (baselineBytes == 0)
+                return null;
+
+            return currentBytes / baselineBytes;
+        }
+
+        private static double? GetAllocatedBytes([CanBeNull] IReadOnlyDictionary<string, Metric> metrics)
+        {
+            var metric = metrics?.Values.FirstOrDefault(m => m.Descriptor is AllocatedMemoryMetricDescriptor);
+            return metric?.Value;
+        }
+
+        public override ColumnCategory Category => ColumnCategory.Metric; //it should be displayed after Allocated column
+        public override int PriorityInCategory => GC.MaxGeneration + 2; //after AllocatedMemoryMetricDescriptor
+        public override bool IsNumeric => true;
+        public override UnitType UnitType => UnitType.Dimensionless;
+        public override string Legend => "Allocated memory ratio distribution ([Current]/[Baseline])";
+    }
+}
--- a/src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
+++ b/src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
@@ -80,7 +80,7 @@ namespace BenchmarkDotNet.Columns
         }
 
         public override ColumnCategory Category => ColumnCategory.Metric; //it should be displayed after Allocated column
-        public override int PriorityInCategory => AllocatedMemoryMetricDescriptor.Instance.PriorityInCategory + 1; //after AllocatedMemoryMetricDescriptor
+        public override int PriorityInCategory => AllocatedMemoryMetricDescriptor.Instance.PriorityInCategory + 1;
         public override bool IsNumeric => true;
         public override UnitType UnitType => UnitType.Dimensionless;
         public override string Legend => "Allocated memory ratio distribution ([Current]/[Baseline])";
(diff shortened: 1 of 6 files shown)


YegorStepanov:
I have a few questions:

1)

| Error | Ratio | Allocated | Alloc Ratio |
|------:|------:|----------:|------------:|
|    NA |  0.93 |         - |          NA |
|    NA |  1.00 |         - |          NA |

`AllocRatio` should be hidden? 

2)

| Allocated | Alloc Ratio |
|----------:|------------:|
|         - |       0.000 |
|      32 B |       1.000 |

| Allocated | Alloc Ratio |
|----------:|------------:|
|      32 B |       0.000 |
| 1000024 B |       1.000 |

Is ok?
What's desired precision for numbers?

3)

|    Ratio | Allocated | Alloc Ratio |
|---------:|----------:|------------:|
|   -99.7% |         - |     -100.0% |
| baseline |      32 B |             |

|    Ratio | Allocated | Alloc Ratio |
|---------:|----------:|------------:|
|  -100.0% |      32 B |     -100.0% |
| baseline | 1000024 B |             |

Is ok?

4)

| Allocated |      Alloc Ratio |
|----------:|-----------------:|
|      32 B | 31,250.594x less |
| 1000019 B |                  |

Is ok?

src/BenchmarkDotNet/Columns/DefaultColumnProvider.cs
```diff
@@ -63,13 +63,23 @@ public IEnumerable<IColumn> GetColumns(Summary summary)
...
+                    }
                 }
             }
 
             private static bool NeedToShow(Summary summary, Func<Statistics, bool> check)
             {
                 return summary.Reports != null && summary.Reports.Any(r => r.ResultStatistics != null && check(r.ResultStatistics));
             }
+
+            private static bool HasDisassemblyDiagnoser(Summary summary)
```
adamsitnik, line 79: Disassembly -> Memory

```suggestion
                    if (HasMemoryDiagnoser(summary))
                    {
                        yield return new BaselineAllocationRatioColumn();
                    }
                }
            }

            private static bool NeedToShow(Summary summary, Func<Statistics, bool> check)
            {
                return summary.Reports != null && summary.Reports.Any(r => r.ResultStatistics != null && check(r.ResultStatistics));
            }

            private static bool HasMemoryDiagnoser(Summary summary)
```

src/BenchmarkDotNet/Diagnosers/MemoryDiagnoser.cs
```diff
@@ -42,21 +42,7 @@ public IEnumerable<Metric> ProcessResults(DiagnoserResults diagnoserResults)
...
-
-            public string Id => "Allocated Memory";
-            public string DisplayName => "Allocated";
-            public string Legend => "Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)";
-            public string NumberFormat => "0.##";
-            public UnitType UnitType => UnitType.Size;
-            public string Unit => SizeUnit.B.Name;
-            public bool TheGreaterTheBetter => false;
-            public int PriorityInCategory => GC.MaxGeneration + 1;
+            yield return new Metric(new AllocatedMemoryMetricDescriptor(), diagnoserResults.GcStats.GetBytesAllocatedPerOperation(diagnoserResults.BenchmarkCase));
```
adamsitnik, line 45: this type is state-less and I would prefer to to keep reusing the static instance and avoid allocating the memory. Another reason is that in the past we had few places where we were using references to check for equality and I don't remember if metric descriptors don't do that.

```suggestion
            yield return new Metric(AllocatedMemoryMetricDescriptor.Instance, diagnoserResults.GcStats.GetBytesAllocatedPerOperation(diagnoserResults.BenchmarkCase));
```

src/BenchmarkDotNet/Diagnosers/AllocatedMemoryMetricDescriptor.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Reports;
+
+namespace BenchmarkDotNet.Diagnosers
+{
+    internal class AllocatedMemoryMetricDescriptor : IMetricDescriptor
+    {
+        public string Id => "Allocated Memory";
```
adamsitnik, line 9:
```suggestion
        internal static readonly IMetricDescriptor Instance = new AllocatedMemoryMetricDescriptor();

        public string Id => "Allocated Memory";
```

src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        private static double? GetAllocatedBytes([CanBeNull] IReadOnlyDictionary<string, Metric> metrics)
+        {
+            var metric = metrics?.Values.FirstOrDefault(m => m.Descriptor is AllocatedMemoryMetricDescriptor);
+            return metric?.Value;
+        }
+
+        public override ColumnCategory Category => ColumnCategory.Metric; //it should be displayed after Allocated column
+        public override int PriorityInCategory => GC.MaxGeneration + 2; //after AllocatedMemoryMetricDescriptor
```
adamsitnik, line 83:
```suggestion
        public override int PriorityInCategory => AllocatedMemoryMetricDescriptor.Instance.PriorityInCategory + 1; //after AllocatedMemoryMetricDescriptor
```

src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
```diff
@@ -0,0 +1,88 @@
...
+                case RatioStyle.Trend:
+                    return isBaseline
+                        ? ""
+                        : ratio.Value >= 1.0
+                            ? ratio.Value.ToString(advancedPrecision ? "N3" : "N2", cultureInfo) + "x more"
+                            : invertedRatio == null
+                                ? "NA"
+                                : invertedRatio.Value.ToString(advancedPrecision ? "N3" : "N2", cultureInfo) + "x less";
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(summary), ratioStyle, "RatioStyle is not supported");
```
adamsitnik, line 49: could you please refactor this logic to avoid code duplication with `BaselineRatioColumn`?

src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
```diff
@@ -0,0 +1,88 @@
...
+            [CanBeNull] IReadOnlyDictionary<string, Metric> current,
+            [CanBeNull] IReadOnlyDictionary<string, Metric> baseline)
+        {
+            double? currentBytes = GetAllocatedBytes(current);
+            double? baselineBytes = GetAllocatedBytes(baseline);
+
+            if (currentBytes == null || baselineBytes == null)
+                return null;
+
+            if (baselineBytes == 0)
```
adamsitnik, line 70: thanks for improving the code and avoiding the exception rather than catching it!

adamsitnik (commented):
Overall LGTM, I've found only some minor issues. PTAL at my comments.

Thank you @YegorStepanov !

src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
YegorStepanov: @adamsitnik, I think you missed with lines a bit. You highlighted the default case. You meant about `IsNonBaselinesPrecise` or the whole switch or the switch math `(value - 1.0) * 100` reuse?

adamsitnik: @YegorStepanov the whole switch

src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
YegorStepanov: the `double?` math is [safe](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoaosgTgAoAiAAgAYB6dkVtgalZ4YAd1YATCDmAAbGAH5G7AJStOHZQG5aDFnC48+rQY2FjJ0uYv2r1KrTqZsy3XgKGiJU2QsZlbGtpUdE4cnGRuxh7m3lZKAf5BISysAJZ4AGauRoJmXpa+Fj4AdAAKEADOqRipAG4wAJKZ6dUAngEqQdRAA=).

The logic hides special symbols like `inf`, `NaN` for NA for consistency with `Ratio`.

The null checking can be removed too. But it's easy to forget that division for `null` returns `null`.

src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
YegorStepanov: @adamsitnik It's hard.
`Mean` and `Allocation` differs by:
- 'Ratio' type is `Statistics` and `double?`
- one prints `"x slower/faster"`, another prints `x more/less`
- one prints '"baseline"', another prints '""'

I've tried do it thought a class with `FromValue`/`FromPercentage`/`FromTrend` methods (each method has switch for `Mean`/`StdDev`/`Allocation` column). But I can't pass `InvertedRatio', because types are different. (we should check for null, and call `.Value` for both cases.

Also, the code is hard to read.

After I've decided implement `Strategy` pattern. To convert switches to structures I've added `IRatioMetricStrategy`, `RatioMeanStrategy`, `RatioStdDevStrategy`, `RatioAllocationStrategy` (120 lines of code). But this code converts switch to methods only! To decrease duplication, we should add another strategy `IRatioStrategy` with `IMeanStrategy` and `IStdStrategy`. I think I'm wasting time :)

Maybe I don't see a simple solution.

src/BenchmarkDotNet/Reports/SummaryStyle.cs
```diff
@@ -72,7 +72,7 @@ public bool Equals(SummaryStyle other)
                    && PrintUnitsInContent == other.PrintUnitsInContent
                    && PrintZeroValuesInContent == other.PrintZeroValuesInContent
                    && Equals(SizeUnit, other.SizeUnit)
-                   && Equals(CodeSizeUnit, other.CodeSizeUnit)
+                   && Equals(SizeUnit, other.CodeSizeUnit)
```
adamsitnik, line 75:
```suggestion
                   && Equals(CodeSizeUnit, other.CodeSizeUnit)
```

adamsitnik (approved):
LGTM, thank you @YegorStepanov 

I am really looking forward to use it!

src/BenchmarkDotNet/Columns/BaselineAllocationRatioColumn.cs
adamsitnik: @YegorStepanov thanks for taking a look into this. You have convinced me that it's not worth the additional complexity ;)

YegorStepanov:
@adamsitnik If it will need to hide the column with some condition or there will be a bug, tap me.

Issue #1634, opened 2021-01-08
Custom names for argument/parameter values


IdanZel:
Hello,
I recently came across the following issue while writing benchmarks.

Say I have a benchmark that receives one argument of type `List<int>` through an `ArgumentsSource`, for example:
```csharp
public IEnumerable<List<int>> Source => new[]
{
    new List<int> {1, 2},
    new List<int> {1, 3}
};

[Benchmark]
[ArgumentsSource(nameof(Source))]
public void Benchmark(List<int> argument)
{
    // Do stuff
}
```

The summary for this benchmark would look like this (max column width increased, measurement results are random):
```
|    Method |                                        argument |     Mean |     Error |    StdDev |
|---------- |------------------------------------------------ |---------:|----------:|----------:|
| Benchmark | System.Collections.Generic.List`1[System.Int32] | 1.234 ms | 0.0123 ms | 0.0123 ms |
| Benchmark | System.Collections.Generic.List`1[System.Int32] | 5.678 ms | 0.0456 ms | 0.0456 ms |
```

Since the text displayed in the `argument` column is the same for both cases, and only tells of each argument's type, **there's no trivial way to distinguish between them and know which exact value was passed to each case**. This might not be the exact case in this specific example, but this is the general idea. 

**Note that this is relevant not just to `List<T>`, but to any non-primitive type, and also when using `ParamsSource` instead of `ArgumentsSource`.**

I would like to have the aforementioned ability, and one way I thought of doing this is by attaching a custom name to each value in the arguments' source. For example, if I attach the name `"One Two"` to `new List<int> {1, 2}` and `"One Three"` to `new List<int> {1, 3}`, the summary could look like this:
```
|    Method |    argument |     Mean |     Error |    StdDev |
|---------- |------------ |---------:|----------:|----------:|
| Benchmark |     One Two | 1.234 ms | 0.0123 ms | 0.0123 ms |
| Benchmark |   One Three | 5.678 ms | 0.0456 ms | 0.0456 ms |
```

As far as I understood from looking at the source code and existing issues, there's currently no simple way of doing this. 
I could, for example, create a type inheriting from `List<int>` for each value and override `ToString()` with the name I want it to have, but the more values I add the more effort this would take. Furthermore, if I wanted to use a `sealed` type as an argument, this would not be possible.

[...]

PR #1869, opened 2021-12-13
Custom names for argument/parameter values

IdanZel:
Implementation of #1634

I wrote this code quite a while ago and only decided to open a PR now, so let me know if there's anything outdated.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1869) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1869) before we can accept your contribution.<br/><br/>:x: IdanZel [sign now](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1869)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/BenchmarkDotNet?pullRequest=1869) it.</sub>

mawosoft:
I'm not quiet sure yet, but this seems a bit overcomplicated and changes to **SmartParamBuilder.cs** aren't really necessary for this to work. I had written something similar a shorter while ago ([ParamWrapper\<T\>](https://github.com/mawosoft/Mawosoft.Extensions.BenchmarkDotNet/blob/master/src/Mawosoft.Extensions.BenchmarkDotNet/ParamWrapper.cs), here is an [example](https://github.com/mawosoft/Mawosoft.Extensions.BenchmarkDotNet/blob/master/samples/ColumnDisplaySamples/ParamWrapperSample.cs)), although it lacks the fluent-style `WithXxx()` syntax and the Benchmark method expects the wrapped argument or parameter (I simply hadn't thought of adding an conversion).

```cs
public class ParamWrapper<T> : IDisposable
{
    public T Value;
    public string? DisplayText;
    public override string? ToString()
        => DisplayText ?? Value?.ToString() ?? ParameterInstance.NullParameterTextRepresentation;
    public ParamWrapper(T value, string? displayText)
    {
        Value = value;
        DisplayText = displayText;
    }
    public void Dispose() => (Value as IDisposable)?.Dispose();
    // Newly added just now to make the wrapper transparent to the benchmark method.
    public static implicit operator T(ParamWrapper<T> @this) => @this.Value;
}
```
Except for the fluent API, this is really all that's needed, and as far as I can tell, BDN will already do all the necessary type casting.

<details>
<summary>Usage example</summary>

```cs
// Usage 
public IEnumerable<ParamWrapper<MemoryStream>> DataSource_Wrapped()
{
    yield return new ParamWrapper<MemoryStream>(new MemoryStream(50), "small stream");
    yield return new ParamWrapper<MemoryStream>(new MemoryStream(500), "big stream");
}

[ParamsSource(nameof(DataSource_Wrapped))]
public MemoryStream InputStream { get; set; } = null!;

[Benchmark]
[ArgumentsSource(nameof(DataSource_Wrapped))]
public void Method1(MemoryStream input)
{
    input.Seek(0, SeekOrigin.Begin);
    byte[] buffer = new byte[10];
    while (input.Read(buffer, 0, buffer.Length) == buffer.Length) { }
}

[Benchmark]
public void Method2()
{
    InputStream.Seek(0, SeekOrigin.Begin);
    byte[] buffer = new byte[10];
    while (InputStream.Read(buffer, 0, buffer.Length) == buffer.Length) { }
}
```
</details>

AndreyAkinshin:
I like the approach suggested by @mawosoft that doesn't modify `SmartParamBuilder` and his friends. A few comments:
1. Since this wrapper could be applied to both parameters and arguments, I would avoid the `Param` part in the class name. E.g, we can rename it to `NamedValue`.
2. I would also add an implicit cast from `T` to the wrapper class so that we could easily introduce `NamedValue` instances without a specific name.
Thus, the new implementation could look as follows:
```cs
public class NamedValue<T> : IDisposable
{
    public T Value { get; }
    public string? DisplayText { get; }

    public override string? ToString()
        => DisplayText ?? Value?.ToString() ?? ParameterInstance.NullParameterTextRepresentation;

    public NamedValue(T value, string? displayText)
    {
        Value = value;
        DisplayText = displayText;
    }

    public void Dispose() => (Value as IDisposable)?.Dispose();

    public static implicit operator T(NamedValue<T> @this) => @this.Value;
    public static implicit operator NamedValue<T>(T value) => new(value, null);
}
```
P.S. I'm against exposing static extension methods that could be applied to any object (like `static NamedParameter WithName(this object parameter, ...`). It works nice in the suggested example, but it contaminates the list of available object methods in other places.

mawosoft:
@AndreyAkinshin In my own library, Iwas also pondering if the wrapper should implement `Equals()` and how far that should go (i.e. also `IEquatable`, `IStructuralEquatable` in case `T` implements it).

In allmost all cases, BDN uses the text representation to compare parameters and arguments, but `ReturnValueValidator` and `ExecutionValidator`

adamsitnik:
@IdanZel thank you for your contribution!

Since the suggestions have not been implemented for few months, I am closing it as it's stale. Feel free to address them and reopen the PR.

PR #1874, opened 2021-12-20
disable ReSharper's Dynamic Program Analysis by default

### LINKED ISSUE #1871: Detect ReSharper's Dynamic Program Analysis
[Dynamic Program Analysis](https://www.jetbrains.com/help/resharper/Dynamic_Program_Analysis.html) is a feature of ReSharper that can detect memory allocation issues. It's enabled by default when running your project from Visual Studio with ReSharper installed, or from Rider.

 Unfortunately, it also impacts benchmark results.

Here's an example with DPA turned on:

image

And the same benchmark with DPA turned off:

image

Note that while the `PcreRegex` benchmark is barely affected (it does not allocate), the `Regex` benchmark takes a longer time when DPA is turned on. The difference in ratio between the two runs shows this.

I'm not sure if DPA can be detected, but it would be nice to have a large warning/error of the same kind as when you run benchmarks in debug mode, especially since DPA is turned on by default.

adamsitnik:
fixes #1871

### DIFF
--- a/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
@@ -126,6 +126,9 @@ namespace BenchmarkDotNet.Extensions
             if (benchmarkCase.Job.Infrastructure.Toolchain is CoreRunToolchain _)
                 start.SetCoreRunEnvironmentVariables(benchmarkCase, resolver);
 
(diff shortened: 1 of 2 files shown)


adamsitnik:
It was tested in https://github.com/dotnet/BenchmarkDotNet/issues/1871#issuecomment-998055888 so I am just going to merge it.

PR #1880, opened 2021-12-30
make `install-from-source` more macOS friendly

askazakov:
- fix / vs \
- fix `BenchmarkDotNet.Templates::0.0.0-* is not supported`


AndreyAkinshin:
Closed in favor of #2338

PR #1890, opened 2022-01-11
Hiding columns

### LINKED ISSUE #299: Add API to remove columns for baseline comparison
Most of the time I don't want to compare sd and other info against the baseline.

### LINKED ISSUE #1603: Don't display Job and Toolchain column when running benchmarks for multiple runtimes
The `Job` and `Toolchain` columns are redundant, `Runtime` is all we need:

obraz

I see this very often when our users are presenting BDN results at conferences|meetups etc

The `Job` column should remain visible when the user has set the Job id in explicit way.

/cc @stephentoub

### LINKED ISSUE #1775: Please add an easy way to remove columns
For sharing simple results, in a blog post or the like, I don't want to show error, standard deviation, ratio standard deviation, median, nor do I want to show job or toolchain.  Right now I need to manually remove those columns from the resulting data every time I run a benchmark.  I would like to be able to add a one-liner to have the tool simply not output a set of named columns, whereas today it seems I need to instead start with an empty config and manually add all the columns I do care about (which I actually haven't figured out how to comprehensively do yet).

YegorStepanov:
closes #1775 closes #299; partially closes #1603 (there is an open PR for this issue)

It affects Console, Markdowns, Html exporters only.

## API
```C#
[HideColumns(Column.Mean, "Error")] // Hides all columns with such names.
class Benchmarks { ... }

//by config:
DefaultConfig.Instance
    .HideColumns(Column.Mean, "Error")
    .HideColumns(new MyColumn()); //compare by Id, not by name
```

Additionally, you can implement custom rule:
```C#
public class MyRule : IColumnHidingRule
{
    public bool NeedToHide(IColumn column) => column.Category == ColumnCategory.Statistics;
}

DefaultConfig.Instance.HideColumns(new MyRule());
```

When using `--join` these attributes add up:
```c#
//both columns are hidden when --join or RunAllJoined()
[HideColumns("Column1")] class Benchmarks1 {}
[HideColumns("Column2")] class Benchmarks2 {}
```

Instead hiding it moves column value to the rows above the table when:
- one value
- all values are the same

@stephentoub @timcassell @paulomorgado @ltrzesniewski Hello everyone! You have liked the issue. Any thoughts about proposed API? Is it covers all yours cases? Maybe I forgot to include some columns to the enum?

@stephentoub I'm afraid to imagine how much time you spent deleting columns...

### TODO
 - cli support:  `--hide Mean Error`,
 - code completion to the cli (waiting for migration to System.CommandLine)


ltrzesniewski:
Thanks, I like the API :+1:

Just one suggestion: maybe instead of an enum, what would you think about a class with column names as constants?

```C#
public static class ColumnNames
{
    public const string Namespace = "Namespace";
    public const string Type = "Type";
    // ...
}
```

stephentoub:
> @stephentoub I'm afraid to imagine how much time you spent deleting columns...

Yeah, don't, it's depressing :)  Thanks for working on this.

YegorStepanov:
@ltrzesniewski It looks better then enum and no compatibility issues. 

I started with flags enums, confused OR with AND :) and I don't like that they look different:
```C#
[HideColumns("Mean", "Error")]
[HideColumns(Column.Mean | Column.Error)]
```

If we add all columns to the enum/class, we can add [AddColumns] that will implicitly adds diagnostics:
```C#
[AddColumns(Column.P90, Column.Allocated)] //adds [MemoryDiagnostic] and only Allocated, I suppose
```

I'm not sure if someone needs it

YegorStepanov:
I wanted to do something like this:
```C#
public static class Column
{
    public const IColumn Mean;
}

[HideColumns(Column.Mean)]
```
but attribute restrictions are very painful: https://github.com/dotnet/roslyn/issues/16898#issuecomment-277057508

@ltrzesniewski suggestion to compare only by `ColumnName`, so `[HideColumns(ColumnNames.Mean)]` will hide all columns named "Mean".

@adamsitnik @AndreyAkinshin What is your opinion on this?

If enum is the best option, should I add all possible columns? The total number of columns is ~100, but most are useful when they appears.

image

timcassell:
Nice! I sometimes have `UnrollFactor` column appear when I don't care about it.

What about an `IncludeOnlyColumns` option that excludes all others except the ones you specify?

YegorStepanov:
> What about an IncludeOnlyColumns or just OnlyColumns option that excludes all others except the ones you specify?

Maybe add `NoColumns/Empty/BlankConfig` for it? It's like DefaultConfig but without any columns.

> [Edit] I also like @ltrzesniewski's idea for const string names instead of enums, which would play nicely with an OnlyColumns as well when you want to also include custom columns (like the names of params).

It will greatly simplify PR.

timcassell:
> Maybe add `NoColumns/Empty/BlankConfig` for it? It's like DefaultConfig but without any columns. It would duplicate API, some column attributes have additional settings, that can't be set by [OnlyColumns]

I suppose a `HideAllColumns` plus `AddColumns` would do the trick. Perhaps the `HideAllColumns` could have an option to still show the ones with special settings? I'm not really sure how that should work... What's an example of one of the columns when special settings where this wouldn't work?

YegorStepanov:
```C#
ConfidenceIntervalErrorColumnAttribute(ConfidenceLevel level = ConfidenceLevel.L999)
RankColumnAttribute(NumeralSystem system = NumeralSystem.Arabic)
StatisticalTestColumnAttribute(...) // 3 ctors 
```

Almost every in Mutators folder:

image

for example:
```C#
GcForceAttribute(bool value = true)
GcServerAttribute(bool value = false)
InvocationCountAttribute(int invocationCount, int unrollFactor = 1)
MaxRelativeErrorAttribute(double maxRelativeError)
```

mawosoft:
This is how I'm currently doing it for Job columns:
https://mawosoft.github.io/Mawosoft.Extensions.BenchmarkDotNet/api/Mawosoft.Extensions.BenchmarkDotNet.JobColumnSelectionProvider.html

It's not attribute-based, but relatively simple. And it would be easy to add command line support.

YegorStepanov:
Moved to const strings. I updated PR description. I hid tests/docs/samples to increase probability of code review.

Any other suggestions?

@timcassell Did you mean column adding, not showing?

@mawosoft I've read your code before writing this. Thanks :)

timcassell:
> @timcassell Did you mean column adding, not showing?

Yes, that's what I meant. :)

`AddColumns` or `OnlyColumns` isn't as big of a deal as `HideColumns` I think, and they can always be added in a separate PR if there's enough need for it.

YegorStepanov:
@timcassell I extend Summary interface by IColumnHidingRule, so if in the future there is a probability of adding AddColumns, then I should do it in another way.

timcassell:
You could change it to `IColumnVisibilityRule`. I think adding would just be negating whatever you have for hiding anyway, right?

YegorStepanov:
@timcassell IColumnHidingRule.NeedToHide() forces hiding, not On/Off

timcassell:
Oh, so you just force hide, but you can't force show. That makes sense. Yeah, I think that's fine how it is.

YegorStepanov:
One drawback: IntelliSense doesn't suggest `Column.Mean` when writing `Mean`, but did it for enum

YegorStepanov:
Added CLI support: `-h` or `--hide`

The column started with dash(`-`) treated like a new command (the double quotes don't help), but the dash can be used in the middle of name: `my-column-name`

Example: `--hide Mean Error "Method" "Alloc Ratio"`

YegorStepanov:
### Default, no rules:
image

### --hide Method Mean Error "Alloc Ratio"
image

### --hide Method Mean Error "Alloc Ratio" Ratio Allocated
image

YegorStepanov:
Instead of extending IConfig by smth like `IConfig.GetColumnHidingRules()`, it is possible to pass the data through a magic filter (and downcasting it when needed)
```C#
internal class ColumnFilter : IFilter
{
    public bool Predicate(BenchmarkCase benchmarkCase) => true;
    public bool NeedToHide() => ...;
}
```

@adamsitnik, what do you advise to do?

adamsitnik (approved):
LGTM, big thanks for your contribution @YegorStepanov !

PR #1901, opened 2022-01-28
fix Markdown output to escape pipe character

### LINKED ISSUE #1839: Markdown output should escape the output
Various characters (e.g. `|`) if used in a Params will end up invalidating the generated markdown, causing it to render incorrectly.  It'd be helpful if benchmarkdotnet could escape the markdown output, e.g. output `\|` instead of `|`.

bakermo:
Hello! Looking to get my feet wet contributing to open source. I've made changes here that should fix #1839. 

This will escape the pipe `` | `` character in markdown table cells by replacing it with ``\|`` so that the tables render properly.
Example before escaping:
image

And after:
image

### DIFF
--- a/src/BenchmarkDotNet/Exporters/MarkdownExporter.cs
+++ b/src/BenchmarkDotNet/Exporters/MarkdownExporter.cs
@@ -51,7 +51,8 @@ namespace BenchmarkDotNet.Exporters
             CodeBlockStart = "``` ini",
             StartOfGroupHighlightStrategy = MarkdownHighlightStrategy.Bold,
             ColumnsStartWithSeparator = true,
-            EscapeHtml = true
+            EscapeHtml = true,
+            EscapePipe = true
         };
 
         public static readonly IExporter Atlassian = new MarkdownExporter
@@ -71,7 +72,8 @@ namespace BenchmarkDotNet.Exporters
         internal static readonly IExporter Mock = new MarkdownExporter
         {
             Dialect = nameof(Mock),
-            StartOfGroupHighlightStrategy = MarkdownHighlightStrategy.Marker
+            StartOfGroupHighlightStrategy = MarkdownHighlightStrategy.Marker,
+            EscapePipe = true
         };
 
         [PublicAPI] protected string Prefix = string.Empty;
@@ -85,6 +87,7 @@ namespace BenchmarkDotNet.Exporters
         [PublicAPI] protected bool ColumnsStartWithSeparator;
         [PublicAPI] protected string BoldMarkupFormat = "**{0}**";
         [PublicAPI] protected bool EscapeHtml;
+        [PublicAPI] protected bool EscapePipe;
 
         private MarkdownExporter() { }
 
@@ -170,13 +173,13 @@ namespace BenchmarkDotNet.Exporters
             var separatorLine = Enumerable.Range(0, table.ColumnCount).Select(_ => "").ToArray();
             foreach (var line in table.FullContent)
             {
-                if (rowCounter > 0 && table.FullContentStartOfLogicalGroup[rowCounter] && table.SeparateLogicalGroups)
+               if (rowCounter > 0 && table.FullContentStartOfLogicalGroup[rowCounter] && table.SeparateLogicalGroups)
                 {
                     // Print logical separator
                     if (ColumnsStartWithSeparator)
                         logger.WriteStatistic(TableColumnSeparator.TrimStart());
                     table.PrintLine(separatorLine, logger, string.Empty, TableColumnSeparator, highlightRow, false, StartOfGroupHighlightStrategy,
-                        BoldMarkupFormat, false);
+                        BoldMarkupFormat, false, false);
                 }
 
                 // Each time we hit the start of a new group, alternative the color (in the console) or display bold in Markdown
@@ -189,7 +192,7 @@ namespace BenchmarkDotNet.Exporters
                     logger.WriteStatistic(TableColumnSeparator.TrimStart());
 
                 table.PrintLine(line, logger, string.Empty, TableColumnSeparator, highlightRow, table.FullContentStartOfHighlightGroup[rowCounter],
-                    StartOfGroupHighlightStrategy, BoldMarkupFormat, EscapeHtml);
+                    StartOfGroupHighlightStrategy, BoldMarkupFormat, EscapeHtml, EscapePipe);
                 rowCounter++;
             }
         }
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1901) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1901) before we can accept your contribution.<br/><br/>:x: bakermo [sign now](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=1901)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/BenchmarkDotNet?pullRequest=1901) it.</sub>

src/BenchmarkDotNet/Exporters/MarkdownExporter.cs
```diff
@@ -85,6 +87,7 @@ public enum MarkdownHighlightStrategy
         [PublicAPI] protected bool ColumnsStartWithSeparator;
         [PublicAPI] protected string BoldMarkupFormat = "**{0}**";
         [PublicAPI] protected bool EscapeHtml;
+        [PublicAPI] protected bool EscapePipe;
```
adamsitnik, line 90: I can see that you are not setting the value of this property to `true` for `StackOverflow`, but for some reason the SO exporter is escaping it:

```cs
[MarkdownExporterAttribute.Default]
[MarkdownExporterAttribute.GitHub]
[MarkdownExporterAttribute.StackOverflow]
[MarkdownExporterAttribute.Atlassian]
public class Intro_MethodsParamsEscaped
{
    [Params("Should|Escape", "ShouldNotEscape")] public string Param;

    [Benchmark] public bool Method() => Param is null;
}
```

```cmd
dotnet run -c Release -f net5.0 --filter *Intro_MethodsParamsEscaped* --job dry
```

image

Could you please ensure that it does not get escaped for the SO exporter?

src/BenchmarkDotNet/Reports/SummaryTableExtensions.cs
```diff
@@ -57,6 +57,9 @@ public static void PrintLine(this SummaryTable table, string[] line, ILogger log
                     continue;
                 }
 
+                if (escapePipe)
+                    line[columnIndex] = line[columnIndex].Replace("|", "\\|");
```
adamsitnik, line 61: by replacing the original value with the new one the escaping is applied to all other exporters who later use the same data

src/BenchmarkDotNet/Exporters/MarkdownExporter.cs
```diff
@@ -170,13 +173,13 @@ private void PrintTable(SummaryTable table, ILogger logger)
             var separatorLine = Enumerable.Range(0, table.ColumnCount).Select(_ => "").ToArray();
             foreach (var line in table.FullContent)
             {
-                if (rowCounter > 0 && table.FullContentStartOfLogicalGroup[rowCounter] && table.SeparateLogicalGroups)
+               if (rowCounter > 0 && table.FullContentStartOfLogicalGroup[rowCounter] && table.SeparateLogicalGroups)
```
adamsitnik, line 176: nit: we need that space back to keep the code aligned :)

```suggestion
                if (rowCounter > 0 && table.FullContentStartOfLogicalGroup[rowCounter] && table.SeparateLogicalGroups)
```

tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
```diff
@@ -177,6 +177,16 @@ public class MethodBaseline_MethodsParams
                 [Benchmark] public void Bar() { }
             }
 
+            [RankColumn, LogicalGroupColumn, BaselineColumn]
+            public class MethodBaseline_MethodsParamsEscaped
```
adamsitnik, line 181: big thanks for adding a new approval test! 👍

adamsitnik (commented):
@bakermo big thanks for your contribution! PTAL at my comments

adamsitnik:
BTW are there any other characters that should be escaped?

bakermo:
@adamsitnik Appreciate the feedback!  

> BTW are there any other characters that should be escaped?

From what I've seen from testing, the pipe character is the only one that breaks output, but I will make another pass to confirm this after addressing your other comments later today.

src/BenchmarkDotNet/Exporters/MarkdownExporter.cs
bakermo: Ok, this should be fixed now.

Do we want to apply this escaping to console output?  It would render as ```\|``` in the console, which is probably not desired, however there are comments in the code indicating that users often copy-paste from the console into Github:https://github.com/dotnet/BenchmarkDotNet/blob/67bf3c0a170d1c4270420bc122fbf795f1b3546c/src/BenchmarkDotNet/Exporters/MarkdownExporter.cs#L37

YegorStepanov:
@bakermo I think there should be `EscapeSpecialCharacters` too. Is it should escape all characters? `\t` too?

Try run this (`\a` is a sound character, and it makes sounds!):
```C#
public class Benchmarks
{
    [Params("Shou\nld|Esca\ape")] public string Param;

    [Benchmark] public bool Method() => Param is null;
}
````

image

bakermo:
@YegorStepanov Ah, good catch. Yes, I agree.  

Do you think it would be best to have a single flag ```EscapeSpecialCharacters``` that includes the pipe, or keep them separate (```EscapeSpecialCharacters``` and ```EscapePipe```)?

My inclination is to keep them separate, as ```EscapeSpecialCharacters``` could handle all characters that we probably want escaped across multiple (or all) outputs, and leave ```EscapePipe``` specific to the outputs that care about it. That, or each export dialect defines its own character set that it needs escaped.  Thoughts?

YegorStepanov:
@bakermo My opinion is exactly the same as yours. Separate them.

I don't remember any other markdown-specific mistakes in the output, `escapePipe` is enough currently.

By implementing it you will close one or few other issues. You may find them and link to the PR.

YegorStepanov (approved):
👍

Have you corrected all Adam's comments?

It's not so easy to escape `|` in the console, because the table alignment will be broken. To fix it the big refactoring of the markdown exporter is required.

We will need to calculate the max column width and only after it, print the table.

In #2135 I've escaped the control characters (like '\0', '\a' - it's still making sound :))

@adamsitnik

YegorStepanov:
The big refactoring will also fix raw markdown alignment when using bold text:

Current:
image

Should be:
image

PR #1905, opened 2022-01-31
[wasm] fix broken sample

pavelsavara:
Fixes https://github.com/dotnet/runtime/issues/64553

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/Executor.cs
+++ b/src/BenchmarkDotNet/Toolchains/Executor.cs
@@ -116,7 +116,7 @@ namespace BenchmarkDotNet.Toolchains
                 case WasmRuntime wasm:
                     start.FileName = wasm.JavaScriptEngine;
                     start.RedirectStandardInput = false;
-                    start.Arguments = $"{wasm.JavaScriptEngineArguments} main.js -- --run {artifactsPaths.ProgramName}.dll {args} ";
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Toolchains/Executor.cs
```diff
@@ -116,7 +116,7 @@ private ProcessStartInfo CreateStartInfo(BenchmarkCase benchmarkCase, ArtifactsP
                 case WasmRuntime wasm:
                     start.FileName = wasm.JavaScriptEngine;
                     start.RedirectStandardInput = false;
-                    start.Arguments = $"{wasm.JavaScriptEngineArguments} main.js -- --run {artifactsPaths.ProgramName}.dll {args} ";
+                    start.Arguments = $"{wasm.JavaScriptEngineArguments} test-main.js -- --run {artifactsPaths.ProgramName}.dll {args} ";
```
naricc, line 119: I think we can condition this on the RuntimeMoniker to use the right thing if the user explicitly selections WasmNet60 or WasmNet50: https://github.com/dotnet/BenchmarkDotNet/blob/6ca6c6625857aaf08d8441b0ef9c1c613464a469/src/BenchmarkDotNet.Annotations/Jobs/RuntimeMoniker.cs#L151

pavelsavara: There are 2 other files which would have to do that https://github.com/dotnet/BenchmarkDotNet/search?q=test-main

PR #1909, opened 2022-02-02
Log progress and estimated finish time

adamsitnik:
A feature request I got offline from our users who run a LOT of benchmarks (thousands) and would like to know when approximately it's going to finish.

Sample log output:

```log
// ** Remained 17 (94.4%) benchmark(s) to run. Estimated finish 2/2/2022 7:23:32 PM **
```

cc @danmoseley

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -75,7 +75,7 @@ namespace BenchmarkDotNet.Running
                     {
                         var runChronometer = Chronometer.Start();
 
-                        var summary = Run(benchmarkRunInfo, benchmarkToBuildResult, resolver, compositeLogger, artifactsToCleanup, 
+                        var summary = Run(benchmarkRunInfo, benchmarkToBuildResult, resolver, compositeLogger, artifactsToCleanup,
                             resultsFolderPath, logFilePath, totalBenchmarkCount, in globalChronometer, ref runChronometer, ref benchmarksToRunCount);
 
                         if (!benchmarkRunInfo.Config.Options.IsSet(ConfigOptions.JoinSummary))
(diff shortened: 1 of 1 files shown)


danmoseley (approved):

AndreyAkinshin (changes requested):
I like the idea of printing finish estimates, but I have concerns about presenting it in the form of "finish date/time". Typically, I run benchmarks on different cloud machines with different time zones. Using the current approach, I have to be aware of the time zone on a specific machine and make some additional manual calculations transforming the server finish time to my local time zone.

I think it would be better to print just the estimated time to finish in a human-readable format (something like `5 hours 25 minutes`).

adamsitnik:
@AndreyAkinshin I prefer to have the exact finish date, so perhaps we should print both?

```diff
- // ** Remained 17 (94.4%) benchmark(s) to run. Estimated finish 2/2/2022 7:23:32 PM **
+ // ** Remained 17 (94.4%) benchmark(s) to run. Estimated finish 2/2/2022 7:23:32 PM (5h 25m from now) **
```

AndreyAkinshin:
@adamsitnik 
> so perhaps we should print both?
LGTM.
However, could we specify the date format as `yyyy-mm-dd`? Dates like `2/3/2022` are confusing because they could mean February-03 or March-02. Also, it would be nice to have a fixed platform-independent format.

adamsitnik:
@AndreyAkinshin done, PTAL:

```log
// ** Remained 13 (72.2%) benchmark(s) to run. Estimated finish 2022-02-03 15:20 (0h 3m from now) **
```

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -638,5 +653,16 @@ private static void Cleanup(HashSet<string> artifactsToCleanup)
...
         }
+
+        private static void LogProgress(ILogger logger, in StartedClock globalChronometer, int totalBenchmarkCount, int benchmarksToRunCount)
+        {
+            int executedBenchmarkCount = totalBenchmarkCount - benchmarksToRunCount;
+            double avgSecondsPerBenchmark = globalChronometer.GetElapsed().GetTimeSpan().TotalSeconds / executedBenchmarkCount;
+            TimeSpan fromNow = TimeSpan.FromSeconds(avgSecondsPerBenchmark * benchmarksToRunCount);
+            DateTime estimatedEnd = DateTime.Now.Add(fromNow);
+            string message = $"// ** Remained {benchmarksToRunCount} ({(double)benchmarksToRunCount / totalBenchmarkCount:P1}) benchmark(s) to run." +
+                $" Estimated finish {estimatedEnd:yyyy-MM-dd H:mm} ({fromNow:%h}h {fromNow:%m}m from now) **";
```
AndreyAkinshin, line 664: if `fromNow` is 25 hours, `{fromNow:%h}h {fromNow:%m}m` prints `1h 0m`.

AndreyAkinshin (changes requested):

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
adamsitnik: @AndreyAkinshin you are right. I gave up on the smart string formats and used `TotalHours` instead. PTAL

AndreyAkinshin (approved):

Issue #1908, opened 2022-02-02
BenchmarkRunner.RunSource and BenchmarkRunner.RunUrl doesn't work


YegorStepanov:
`'Could not load file or assembly 'System.CodeDom, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
`

### Tested in:
* VS2022, NET Framework 4.8, empty project, BDN v13.1.
* Rider, BDN test project, BDN master.

### Code:
```C#
var s = @"public class Class1
{
    [Benchmark]
    public int Method1() { return 1; }
}";

var error1 = BenchmarkRunner.RunSource(s);
```

In debugger, I cannot navigate to RunSource() in RunUrl(), the same exception is thrown. Although these methods are in the same file.
```C#
var error2 = BenchmarkRunner.RunUrl("https://gist.github.com/lillo42/aa028027af80623cdaf8056a1d9d0728");
```

Maybe this methods should be hidden from IntelliSense? There is no docs explaining supported format, no info in intellisense describing that it's working on netframework only.

PR #1910, opened 2022-02-02
Make FromUrl and FromSource more friendly

YegorStepanov:
### 1st commit
Change unhandled exception with message exception.
Add Unsupported checking directly to the methods, because It fails in .NET core (and they are public).

### 2nd commit
Hide it from IntelliSense cuz it not working and not recommended. Feel free to drop the commit if it isn't desired.

### 3rd commit
Change unhandled exception with message-exception for the case:
```C#
BenchmarkRunner.Run(typeof(GenericBenchmarkType<>));
````

#1908


adamsitnik (approved):

YegorStepanov:
Ah, for some reason I thought that `[EditorBrowsable(EditorBrowsableState.Never)]` works in Rider, but no.

Issue: https://github.com/dotnet/roslyn/issues/4434

PR #1912, opened 2022-02-03
Notified users about private methods with Setup/Cleanup attributes

epeshk:


adamsitnik (commented):
@epeshk thank you for your contribution!

Overall the changes look good, but could you please add a test to `BenchmarkConverterTests`? It can be something simple that just asserts that an `InvalidBenchmarkDeclarationException` is thrown.

epeshk:
I've added test for this

AndreyAkinshin (approved):

PR #1914, opened 2022-02-03
[NO MERGE] don't store all of the generated source code in a single source file

adamsitnik:
tl;dr In dotnet/performance we have more than 4k benchmarks and x86 version of Roslyn is failing to build the auto-generated project (OOM).

I was hoping that it's just related to the fact that BDN generates a single C# file with more than 3 millions lines of code.

That's not the case as Roslyn still fails with OOM.

I am opening the PR as it's still valuable to work with smaller files when repro'ing the bug.

Please ignore it for now and don't merge it.


adamsitnik:
Thanks to https://github.com/dotnet/BenchmarkDotNet/pull/1917 we are not going to need it.

PR #1932, opened 2022-03-02
[wasm] Set the right runtime moniker

radekdoulik:
Fix https://github.com/dotnet/performance/issues/2250

Commit 80f45cedc started using the runtime moniker to decide, which
main JS file to use. We use `test-main.js` in net7.0, while we were
using `main.js` in older versions.

To make that work, let set the right runtime moniker to `WasmRuntime`. So
that later in `Executor::ProcessStartInfo` we select the right file.

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -97,7 +97,8 @@ namespace BenchmarkDotNet.ConsoleArguments
                 settings.MaximumDisplayWidth = Math.Max(MinimumDisplayWidth, GetMaximumDisplayWidth());
             });
 
-        private static bool MonikerIsWasm (RuntimeMoniker moniker)
+        private static bool MonikerIsWasm(RuntimeMoniker moniker)
+
         {
             switch (moniker)
             {
@@ -130,7 +131,8 @@ namespace BenchmarkDotNet.ConsoleArguments
                 {
                      logger.WriteLineError($"The provided {nameof(options.AOTCompilerPath)} \"{ options.AOTCompilerPath }\" does NOT exist. It MUST be provided.");
                 }
-                else if (MonikerIsWasm (runtimeMoniker) && (options.RuntimeSrcDir == null || options.RuntimeSrcDir.IsNotNullButDoesNotExist()))
+                else if (MonikerIsWasm(runtimeMoniker) && (options.RuntimeSrcDir == null || options.RuntimeSrcDir.IsNotNullButDoesNotExist()))
+
                 {
                     logger.WriteLineError($"The provided {nameof(options.RuntimeSrcDir)} \"{options.RuntimeSrcDir}\" does NOT exist. It MUST be provided for wasm-aot.");
                     return false;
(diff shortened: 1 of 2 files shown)


lewing:
cc @adamsitnik

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -97,6 +97,20 @@ private static Parser CreateParser(ILogger logger)
                 settings.MaximumDisplayWidth = Math.Max(MinimumDisplayWidth, GetMaximumDisplayWidth());
             });
 
+        private static bool MonikerIsWasm (RuntimeMoniker moniker)
```
adamsitnik, line 100: nit: extra space

```suggestion
        private static bool MonikerIsWasm(RuntimeMoniker moniker)

```

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -116,7 +130,7 @@ private static bool Validate(CommandLineOptions options, ILogger logger)
                 {
                      logger.WriteLineError($"The provided {nameof(options.AOTCompilerPath)} \"{ options.AOTCompilerPath }\" does NOT exist. It MUST be provided.");
                 }
-                else if (runtimeMoniker == RuntimeMoniker.Wasm && (options.RuntimeSrcDir == null || options.RuntimeSrcDir.IsNotNullButDoesNotExist()))
+                else if (MonikerIsWasm (runtimeMoniker) && (options.RuntimeSrcDir == null || options.RuntimeSrcDir.IsNotNullButDoesNotExist()))
```
adamsitnik, line 133:
```suggestion
                else if (MonikerIsWasm(runtimeMoniker) && (options.RuntimeSrcDir == null || options.RuntimeSrcDir.IsNotNullButDoesNotExist()))

```

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -97,6 +97,20 @@ private static Parser CreateParser(ILogger logger)
...
+            {
+                case RuntimeMoniker.Wasm:
+                case RuntimeMoniker.WasmNet50:
+                case RuntimeMoniker.WasmNet60:
+                case RuntimeMoniker.WasmNet70:
+                    return true;
+                default:
+                    return false;
+            }
+        }
```
adamsitnik, line 112: in the future we might want to move it to [RuntimeMonikerExtensions](https://github.com/dotnet/BenchmarkDotNet/blob/f1aefb59ccd0252c557d195b82ac4d258a7fcd88/src/BenchmarkDotNet/Extensions/RuntimeMonikerExtensions.cs#L7)

adamsitnik (approved):
LGTM, thank you @radekdoulik !

naricc:
Looks good to me.

radekdoulik:
The `build / build-windows` fails with:
```
Standard error: 
  C:\Users\runneradmin\.nuget\packages\microsoft.dotnet.ilcompiler\6.0.0-rc.1.21420.1\build\Microsoft.NETCore.Native.Windows.props(109,5): error : Platform linker not found. To fix this problem, download and install Visual Studio 2019 from http://visualstudio.com./ Make sure to install the Desktop Development for C++ workload. For ARM64 development also install C++ ARM64 build tools. [D:\a\BenchmarkDotNet\BenchmarkDotNet\tests\BenchmarkDotNet.IntegrationTests\bin\Release\net5.0\ddeb76ef-2c31-458a-be4f-b5769587df56\BenchmarkDotNet.Autogenerated.csproj]
```
Is that new fail? From quick look it seems unrelated.

radekdoulik:
Actually I see it fail elsewhere too, so it is not related to this change.

PR #1935, opened 2022-03-09
Fix optimization of action methods for coreclr

### LINKED ISSUE #1934: Ensure WorkloadActionUnroll and similar are optimized if possible
If `{Workload|Overhead}Action{No}Unroll` are not optimized, and the benchmark method allocates a lot, the Tier0 versions of these methods may retain much more heap data than the optimized versions. This can completely change benchmark results.

Current behavior for CLR is that these methods are always eagerly optimized, because they all contain loops and we either do not have tiered compilation (full framework, pre-3.0 core) or have `COMPlus_TC_QuickJitForLoops=1` (core 3.0 - core 6.0). 

For .NET 7 we plan to change the default to `COMPlus_TC_QuickJitForLoops=0` and this leads to these methods not being optimized and significant behavior changes in roughly 50 or so benchmarks on github.com/dotnet/performance.

To preserve current behavior, these methods should always be optimized when running against .NET Core versions that can change tiering strategy. This can be done by adding `[MethodImpl(MethodImplOptions.AggressiveOptimziation)]` which is available in .Net Core 3.0 and up.

Alternatively one could consider restructuring things so that the benchmark delegate is always called from a method that doesn't have loops. This seems more disruptive.

See https://github.com/dotnet/performance/issues/2214#issuecomment-1052585650 and following comments for context.

cc @adamsitnik

AndyAyersMS:
Not sure this is sufficient to catch all the cases where these methods are generated, but this is the rough idea.

Fixes #1934

### DIFF
--- a/src/BenchmarkDotNet/Templates/BenchmarkType.txt
+++ b/src/BenchmarkDotNet/Templates/BenchmarkType.txt
@@ -113,6 +113,9 @@
 
         private BenchmarkDotNet.Engines.Consumer consumer = new BenchmarkDotNet.Engines.Consumer();
 
+#if NETCOREAPP3_0_OR_GREATER
+        [System.Runtime.CompilerServices.MethodImpl(System.Runtime.CompilerServices.MethodImplOptions.AggressiveOptimization)]
+#endif
         private void OverheadActionUnroll(System.Int64 invokeCount)
         {
             $LoadArguments$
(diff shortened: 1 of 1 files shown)


adamsitnik (approved):
LGTM, thank you for your contribution @AndyAyersMS !

timcassell:
Should this `AggressiveOptimization` be included in the InProcess toolchains also?

AndreyAkinshin:
@timcassell yes, it should. A PR is welcome!

PR #1938, opened 2022-03-11
[wasm] Add a `--wasmDataDir` parameter

radical:
In an earlier commit, generated projects got support for importing
props/targets files from the benchmark projects. And that allows
customizing the build.

But there isn't a clean way to inject/add references to the build of the
project, like custom paths, or access to additional files, without
adding new command line parameters.

To solve that, `--wasmDataDir` param is added which is surfaced in the
generated project as the msbuild property - `$(WasmDataDir)`.

An example use case is to have the wasm js file added to this data dir,
and in the props file use:
`<WasmMainJSPath>$(WasmDataDir)test-main.js</WasmMainJSPath>`.

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
@@ -187,6 +187,9 @@ namespace BenchmarkDotNet.ConsoleArguments
         [Option("AOTCompilerMode", Required = false, Default = MonoAotCompilerMode.mini, HelpText = "Mono AOT compiler mode, either 'mini' or 'llvm'")]
         public MonoAotCompilerMode AOTCompilerMode { get; set; }
 
+        [Option("wasmDataDir", Required = false, HelpText = "Wasm data directory")]
+        public DirectoryInfo WasmDataDirectory { get; set; }
+
         internal bool UserProvidedFilters => Filters.Any() || AttributeNames.Any() || AllCategories.Any() || AnyCategories.Any();
 
         [Usage(ApplicationAlias = "")]
--- a/src/BenchmarkDotNet/Environments/Runtimes/WasmRuntime.cs
+++ b/src/BenchmarkDotNet/Environments/Runtimes/WasmRuntime.cs
@@ -16,6 +16,8 @@ namespace BenchmarkDotNet.Environments
 
         public bool Aot { get;  }
 
+        public string WasmDataDir { get; }
+
         /// <summary>
         /// creates new instance of WasmRuntime
         /// </summary>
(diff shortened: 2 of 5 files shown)


radical:
cc @adamsitnik @lewing @radekdoulik

radekdoulik (approved):

src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
```diff
@@ -187,6 +187,9 @@ public class CommandLineOptions
         [Option("AOTCompilerMode", Required = false, Default = MonoAotCompilerMode.mini, HelpText = "Mono AOT compiler mode, either 'mini' or 'llvm'")]
         public MonoAotCompilerMode AOTCompilerMode { get; set; }
 
+        [Option("wasmDataDir", Required = false, HelpText = "Wasm data directory")]
+        public DirectoryInfo WasmDataDirectory { get; set; }
```
adamsitnik, line 191: are there any scenarios in which this argument is mandatory? If so the following method should take that into account:

https://github.com/dotnet/BenchmarkDotNet/blob/5852a1340653250f1fa8801102556026d8f87e45/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs#L100

src/BenchmarkDotNet/Environments/Runtimes/WasmRuntime.cs
```diff
@@ -24,14 +26,16 @@ public class WasmRuntime : Runtime, IEquatable<WasmRuntime>
         /// <param name="msBuildMoniker">moniker, default: "net5.0"</param>
         /// <param name="displayName">default: "Wasm"</param>
         /// <param name="aot">Specifies whether AOT or Interpreter (default) project should be generated.</param>
-        public WasmRuntime(string msBuildMoniker = "net5.0", string displayName = "Wasm", string javaScriptEngine = "v8", string javaScriptEngineArguments = "--expose_wasm", bool aot = false, RuntimeMoniker moniker = RuntimeMoniker.Wasm) : base(moniker, msBuildMoniker, displayName)
+        /// <param name="wasmDataDir">Specifies a wasm data directory surfaced as $(WasmDataDir) for the project</param>
```
adamsitnik, line 29: thank you for adding xml doc comment 👍 (most of our contributors ignore build warnings and don't fix them)

adamsitnik (approved):
LGTM, thank you @radical !

Before I hit the merge button could you please respond to my question?

src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
radical: It's not mandatory.

PR #1941, opened 2022-03-11
Fix ValueTask

### LINKED ISSUE #1595: ValueTask-returning benchmarks are not being called correctly
When a benchmark returns a `ValueTask`, it is simply called with `.GetAwaiter().GetResult()`.

This is illegal `ValueTask` usage: you must not call `GetResult()` until the task has been completed.

`IValueTaskSource`-backed tasks do not need to support blocking on `GetResult()`. The in-box `ManualResetValueTaskSourceCore` does not support blocking `GetResult()`. Nor does the `DOTNET_SYSTEM_THREADING_POOLASYNCVALUETASKS` feature in .NET 5.

Instead, the `ValueTask` should be converted to a `Task` first:

```c#
static void BlockForResult(ValueTask task)
{
    if (task.IsCompleted)
    else
        task.AsTask().GetAwaiter().GetResult();
}
```

(This will be unfortunate for the memory diagnoser, as a bogus `Task` will show up, but the alternative is to just crash...)

### LINKED ISSUE #1738: Async GlobalSetup does not work with InProcess toolchain
https://github.com/dotnet/BenchmarkDotNet/issues/521 added support for async GlobalSetup, but it appears that InProcess toolchain does not wait for the async setup method to finish.  Simple reproducer:

```cs
using System.Threading.Tasks;

namespace AsyncGlobalSetupRepro
{
    [InProcess]
    public class MyBenchmarkClass
    {
        private string _foo;

        [GlobalSetup]
        public async Task MyAsyncSetup()
        {
            await Task.Delay(1000);
            _foo = "hello";
        }

        [Benchmark]
        public int MyBenchmark() => _foo.Length; // NullReferenceException!
    }

    public class Program
    {
        {
            BenchmarkRunner.Run(typeof(MyBenchmarkClass));
        }
    }
}
```

It looks like MyAsyncSetup gets invoked as a regular method here, and then JITting starts immediately without waiting for it to finish: https://github.com/dotnet/BenchmarkDotNet/blob/e7ff4cefcc8d429205a21a76045e82688ee063c5/src/BenchmarkDotNet/Engines/EngineFactory.cs#L28

timcassell:
Fixes #1595 and fixes #1738 

I also added `ValueTask` (non-generic) support for InProcess toolchains that was missing.


timcassell:
Split out to 3 separate PRs for easier review (#2108, #2109, and #2110). cc @YegorStepanov

PR #1950, opened 2022-03-17
DotNetCli*: Allow logging command output

radical:
This is useful when building wasm/aot projects, which can take
10-20mins. Being able to see the progress helps, for example, with
diagnosing issues when the whole thing might timeout, or when running
locally.

This enables it by default for wasm/aot builds.

### DIFF
--- a/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
+++ b/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
@@ -16,7 +16,7 @@ namespace BenchmarkDotNet.Loggers
         private bool logOutput;
         private ILogger logger;
 
-        internal AsyncProcessOutputReader(Process process, bool logOutput=false, ILogger logger=null)
+        internal AsyncProcessOutputReader(Process process, bool logOutput = false, ILogger logger = null)
         {
             if (!process.StartInfo.RedirectStandardOutput)
                 throw new NotSupportedException("set RedirectStandardOutput to true first");
@@ -98,7 +98,7 @@ namespace BenchmarkDotNet.Loggers
             {
                 output.Enqueue(e.Data);
                 if (logOutput)
-                    logger.WriteLineInfo(e.Data);
+                    logger.WriteLine(e.Data);
             }
         }
 
@@ -108,7 +108,7 @@ namespace BenchmarkDotNet.Loggers
             {
                 error.Enqueue(e.Data);
                 if (logOutput)
-                    logger.WriteLineInfo(e.Data);
+                    logger.WriteLineError(e.Data);
             }
         }
 
--- a/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
+++ b/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
@@ -13,18 +13,24 @@ namespace BenchmarkDotNet.Loggers
         private readonly ConcurrentQueue<string> output, error;
 
         private long status;
+        private bool logOutput;
+        private ILogger logger;
 
-        internal AsyncProcessOutputReader(Process process)
+        internal AsyncProcessOutputReader(Process process, bool logOutput=false, ILogger logger=null)
         {
             if (!process.StartInfo.RedirectStandardOutput)
                 throw new NotSupportedException("set RedirectStandardOutput to true first");
             if (!process.StartInfo.RedirectStandardError)
                 throw new NotSupportedException("set RedirectStandardError to true first");
+            if (logOutput && logger == null)
+                throw new ArgumentException($"{nameof(logger)} cannot be null when {nameof(logOutput)} is true");
 
             this.process = process;
             output = new ConcurrentQueue<string>();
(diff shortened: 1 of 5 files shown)


radical:
@adamsitnik @lewing

src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
```diff
@@ -13,18 +13,24 @@ internal class AsyncProcessOutputReader : IDisposable
         private readonly ConcurrentQueue<string> output, error;
 
         private long status;
+        private bool logOutput;
+        private ILogger logger;
 
-        internal AsyncProcessOutputReader(Process process)
+        internal AsyncProcessOutputReader(Process process, bool logOutput=false, ILogger logger=null)
```
adamsitnik, line 19: nit: spacing

```suggestion
        internal AsyncProcessOutputReader(Process process, bool logOutput = false, ILogger logger = null)
```

src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
```diff
@@ -89,13 +95,21 @@ private void Detach()
         private void ProcessOnOutputDataReceived(object sender, DataReceivedEventArgs e)
         {
             if (!string.IsNullOrEmpty(e.Data))
+            {
                 output.Enqueue(e.Data);
+                if (logOutput)
+                    logger.WriteLineInfo(e.Data);
```
adamsitnik, line 101: If we use `WriteLine` we don't get the Info color:

Before:

image

After:

image


```suggestion
                    logger.WriteLine(e.Data);
```

src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
```diff
@@ -89,13 +95,21 @@ private void Detach()
...
+            }
         }
 
         private void ProcessOnErrorDataReceived(object sender, DataReceivedEventArgs e)
         {
             if (!string.IsNullOrEmpty(e.Data))
+            {
                 error.Enqueue(e.Data);
+                if (logOutput)
+                    logger.WriteLineInfo(e.Data);
```
adamsitnik, line 111: for error we can use `WriteLineError` and get red color:

```suggestion
                    logger.WriteLineError(e.Data);
```

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommandExecutor.cs
```diff
@@ -23,13 +23,20 @@ public static class DotNetCliCommandExecutor
...
             using (var process = new Process { StartInfo = BuildStartInfo(parameters.CliPath, parameters.GenerateResult.ArtifactsPaths.BuildArtifactsDirectoryPath, parameters.Arguments, parameters.EnvironmentVariables) })
-            using (var outputReader = new AsyncProcessOutputReader(process))
+            using (var outputReader = new AsyncProcessOutputReader(process, parameters.LogOutput, parameters.Logger))
             using (new ConsoleExitHandler(process, parameters.Logger))
             {
                 parameters.Logger.WriteLineInfo($"// start {parameters.CliPath ?? "dotnet"} {parameters.Arguments} in {parameters.GenerateResult.ArtifactsPaths.BuildArtifactsDirectoryPath}");
 
                 var stopwatch = Stopwatch.StartNew();
 
+                if (parameters.LogOutput && process.StartInfo.EnvironmentVariables.Keys.Count > 0)
```
adamsitnik, line 33: I like the idea of printing env vars, but I think that we should do it only once. With the current proposal we are going to print it for dotnet restore, build and publish.

I think that this logic could be moved to a private method of `DotNetCliCommand` class and invoked at the beginning of:

https://github.com/dotnet/BenchmarkDotNet/blob/937311cb62ac5f3ffab69e243e502d4bc20cd015/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs#L47-L48

https://github.com/dotnet/BenchmarkDotNet/blob/937311cb62ac5f3ffab69e243e502d4bc20cd015/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs#L71-L72

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -38,10 +40,11 @@ public class DotNetCliCommand
             BuildPartition = buildPartition;
             EnvironmentVariables = environmentVariables;
             Timeout = timeout;
+            this.LogOutput = logOutput;
```
adamsitnik, line 43: nit: there is no need for `this` (casing difference)

```suggestion
            LogOutput = logOutput;
```

adamsitnik (approved):
Big thanks for your contribution, I really love this idea!

Since I would like to start my weekend today quite early, I am going to apply my suggestions, push some changes to your fork and merge it so when you start you work day everything will be merged and ready to use. I hope you don't mind.

PR #1952, opened 2022-03-18
Specify output path for "dotnet publish"

Beau-Gosse-dev:
Without this, when benchmarking CoreRT (like from https://github.com/dotnet/performance ) the publish directory ends up being inside of the bin directory of the host console app, and not the sub-directory bin directory of the generated project.

For example when the publish command is this (before this change): ```start dotnet publish -c Release /p:DebugType=portable -bl:benchmarkdotnet.binlog -r win-x64 --no-build --no-restore /p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 /p:Deterministic=true /p:Optimize=true in C:\Code\Local Spikes\performance\performance\artifacts\bin\MicroBenchmarks\Debug\net6.0\Job-PALSMU```

Then it will result in this: ```Executable C:\Code\Local Spikes\performance\performance\artifacts\bin\MicroBenchmarks\Debug\net6.0\Job-PALSMU\bin\Release\net6.0\win-x64\publish\Job-PALSMU.exe not found``` 

because the executable was actually published to ```C:\Code\Local Spikes\performance\performance\artifacts\bin\BenchmarkDotNet.Autogenerated\Release\net6.0\win-x64\publish```


Beau-Gosse-dev:
Replaced with https://github.com/dotnet/BenchmarkDotNet/pull/1955

PR #1953, opened 2022-03-18
add possibility to enable build output logging via command line args

adamsitnik:
In #1950 @radical has added the possibility to log build output to the console. This PR exposes this feature via `--logBuildOutput` command line arg.


src/BenchmarkDotNet/Running/BuildPartition.cs
```diff
@@ -66,6 +67,8 @@ public BuildPartition(BenchmarkBuildInfo[] benchmarks, IResolver resolver)
             ? TimeSpan.FromMinutes(5) // downloading all CoreRT dependencies can take a LOT of time
             : RepresentativeBenchmarkCase.Config.BuildTimeout;
 
+        public bool LogBuildOutput { get; }
```
radical, line 70: I was trying to add command line args too, and added the rest of the changes. But couldn't figure out how to connect from the config, to the DotNetCliCommand! This was the missing piece 👍

PR #1955, opened 2022-03-18
Cleaning up argument spacing in CLI calls. Add output path for publish. CoreRT

Beau-Gosse-dev:
Previous discussion here: https://github.com/dotnet/BenchmarkDotNet/pull/1951 (and https://github.com/dotnet/BenchmarkDotNet/pull/1952)

This was tested locally by manually pointing dotnet/performance to this code and running a CoreRT benchmark against it. 

This fixes 2 different issues:
1. The architecture flag "-r" didn't have a leading space and therefore was combine with the string in front of it, causing it not to be passed to the dotnet restore/build/publish commands. Would look like this: `dotnet restore /p:DebugType=portable -bl:benchmarkdotnet.binlog-r win-x64 /p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 /p:Deterministic=true /p:Optimize=true and dotnet fails to restore/build/publish`

2. The output directory of CoreRT publishes wasn't specified and if running the test project inside another project seemed to output to the first bin directory found from the root of the drive.

### Output path
When benchmarking CoreRT (like from https://github.com/dotnet/performance ) the publish directory ends up being inside of the bin directory of the host console app, and not the sub-directory bin directory of the generated project.

For example when the publish command is this (before this change): start dotnet publish -c Release /p:DebugType=portable -bl:benchmarkdotnet.binlog -r win-x64 --no-build --no-restore /p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 /p:Deterministic=true /p:Optimize=true in C:\Code\Local Spikes\performance\performance\artifacts\bin\MicroBenchmarks\Debug\net6.0\Job-PALSMU

Then it will result in this: Executable C:\Code\Local Spikes\performance\performance\artifacts\bin\MicroBenchmarks\Debug\net6.0\Job-PALSMU\bin\Release\net6.0\win-x64\publish\Job-PALSMU.exe not found

because the executable was actually published to C:\Code\Local Spikes\performance\performance\artifacts\bin\BenchmarkDotNet.Autogenerated\Release\net6.0\win-x64\publish

PR #1960, opened 2022-03-22
Improve support for NativeAOT

adamsitnik:
With the following 3 recent fixes:

* https://github.com/dotnet/runtime/pull/66290 (https://github.com/dotnet/runtime/issues/66028) 
* https://github.com/dotnet/runtime/pull/66650 (https://github.com/dotnet/runtime/issues/66637)
* https://github.com/dotnet/BenchmarkDotNet/pull/1955

It required very little work to improve the NativeAOT support. Main changes:

* CoreRT and NativeAOT are treated as two different runtimes, mostly to be able to run .NET 6 with CoreRT vs .NET 7 with NativeAOT
* `--runtimes` no longer supports `corert70`. The proper name is `nativeaot70` (case and dots are ignored, so you can use `NativeAOT7.0` as well)
* the default feed for the `CoreRtToolchainBuilder` has been upgraded to .NET 7 feed
* the default ILCompiler for the `CoreRtToolchainBuilder` has been upgraded to `7.0.0-*`
* I've renamed `--coreRtVersion` to `--ilCompilerVersion` as it specifies the ILCompiler version. Sample command:
```cmd
dotnet run -c Release -f net5.0 --filter *Basic* --runtimes nativeaot70 --ilcompilerversion 7.0.0-preview.3.22123.2
```

Demo:

```cmd
dotnet run -c Release -f net5.0 --filter *Basic* --runtimes net50 corert60 nativeaot70
```

image

In the meantime I am working on getting all dotnet/performance microbenchmarks running with NativeAOT

cc @jkotas @MichalStrehovsky @hez2010 @Beau-Gosse-dev

### DIFF
--- a/src/BenchmarkDotNet.Annotations/Attributes/GenericTypeArgumentsAttribute.cs
+++ b/src/BenchmarkDotNet.Annotations/Attributes/GenericTypeArgumentsAttribute.cs
@@ -1,5 +1,6 @@
 ﻿using JetBrains.Annotations;
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace BenchmarkDotNet.Attributes
 {
@@ -11,6 +12,6 @@ namespace BenchmarkDotNet.Attributes
         // CLS-Compliant Code requires a constructor without an array in the argument list
         [PublicAPI] public GenericTypeArgumentsAttribute() => GenericTypeArguments = new Type[0];
 
-        public GenericTypeArgumentsAttribute(params Type[] genericTypeArguments) => GenericTypeArguments = genericTypeArguments;
+        public GenericTypeArgumentsAttribute([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] params Type[] genericTypeArguments) => GenericTypeArguments = genericTypeArguments;
     }
 }
--- a/src/BenchmarkDotNet.Annotations/Attributes/DynamicallyAccessedMemberTypes.cs
+++ b/src/BenchmarkDotNet.Annotations/Attributes/DynamicallyAccessedMemberTypes.cs
@@ -1,0 +1,90 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Diagnostics.CodeAnalysis
+{
+    /// <summary>
+    /// Specifies the types of members that are dynamically accessed.
+    ///
+    /// This enumeration has a <see cref="FlagsAttribute"/> attribute that allows a
+    /// bitwise combination of its member values.
+    /// </summary>
+    [Flags]
+    internal enum DynamicallyAccessedMemberTypes
+    {
+        /// <summary>
+        /// Specifies no members.
+        /// </summary>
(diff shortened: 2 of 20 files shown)


src/BenchmarkDotNet.Annotations/Attributes/GenericTypeArgumentsAttribute.cs
```diff
@@ -11,6 +12,6 @@ public class GenericTypeArgumentsAttribute : Attribute
         // CLS-Compliant Code requires a constructor without an array in the argument list
         [PublicAPI] public GenericTypeArgumentsAttribute() => GenericTypeArguments = new Type[0];
 
-        public GenericTypeArgumentsAttribute(params Type[] genericTypeArguments) => GenericTypeArguments = genericTypeArguments;
+        public GenericTypeArgumentsAttribute([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] params Type[] genericTypeArguments) => GenericTypeArguments = genericTypeArguments;
```
MichalStrehovsky, line 15: Unfortunately placing the DAMAttribute on an array is a no-op. Arrays are difficult to model in a static analysis.

This should be generating a warning along the lines of `Trim analysis warning IL2098: GenericTypeArgumentsAttribute..ctor(Type[]): Parameter '#1' of method 'GenericTypeArgumentsAttribute..ctor(Type[])' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to parameters of type 'System.Type' or 'System.String'.`

radical: this actually broke building `dotnet/performance` with `dotnet/runtime`. When building a wasm project, I get - https://gist.github.com/radical/b75b5e4063d97be0adac729b611990ea . Can we revert this change? @adamsitnik

radical: If I remove that attribute from the param, then it works fine.

radical: To be clear, this breaks *when* I update `bdn` reference in `dotnet/performance` to `0.13.1.1740`.

MichalStrehovsky: The attribute is a no-op - it should be removed because even in the absence of the illink bug, it generates a warning. NativeAOT doesn't have the linker bug and that's why it works there. I filed https://github.com/dotnet/linker/issues/2714 on the linker bug - that one should also be fixed independently of this removal.

adamsitnik: I've sent https://github.com/dotnet/BenchmarkDotNet/pull/1973 to address that. @radical is there any chance you could test it to see if it actually helps?

radical: I can build with that change 👍

PR #1963, opened 2022-03-22
Set --llvm in MONO_ENV_OPTIONS for llvm.

naricc:
This sets the MONO_ENV_OPTIONS environment variable to "--llvm" when running the benchmarks. This correctly uses the llvm backend when running.

### DIFF
--- a/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
@@ -10,7 +10,9 @@ using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Portability;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.CoreRun;
+using BenchmarkDotNet.Toolchains.MonoAotLLVM;
 using JetBrains.Annotations;
+
 
 namespace BenchmarkDotNet.Extensions
 {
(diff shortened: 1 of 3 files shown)


naricc:
@adamsitnik

fanyang-mono (approved):

radical:
Should we have some config file to specify extra arguments to be passed to mono when running, or extra build arguments etc? So, the user code can set these without having to make changes in BDN?

src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
```diff
@@ -132,6 +134,17 @@ internal static void SetEnvironmentVariables(this ProcessStartInfo start, Benchm
             if (!benchmarkCase.Job.HasValue(EnvironmentMode.EnvironmentVariablesCharacteristic))
                 return;
 
+            if (benchmarkCase.Job.Infrastructure.Toolchain is MonoAotLLVMToolChain)
```
radical, line 137: I'm new here :) Can we add a way to specify these kinda environment variables, maybe in the toolchain, or the runtime? and `ProcessExtensions` can call a method to get the variables

naricc: I think that is a good idea for an additional feature. But for now I just want to do this, because we pretty much need this environment variable set to actually use llvm, which is the expected behavior here.

naricc:
I am closing this for now, it might not be the issue we actually need to address. Although it does make it fall back to llvm jit.

PR #1972, opened 2022-03-29
[NativeAOT] Set TrimmerDefaultAction to link 

adamsitnik:
To ensure that trimmer only analyzes the parts of the dependencies that are used. With this setting, the build time for our samples project goes from 116 to 15 seconds! I've tested it with dotnet/performance microbenchmarks and everything works fine. In case it does not, users can still disable it:

```cs
builder.SeTrimmerDefaultAction("");
```

cc @MichalStrehovsky

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
+++ b/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
@@ -25,7 +25,7 @@ namespace BenchmarkDotNet.Toolchains.NativeAot
             string runtimeFrameworkVersion, string targetFrameworkMoniker, string cliPath,
             string runtimeIdentifier, IReadOnlyDictionary<string, string> feeds, bool useNuGetClearTag,
             bool useTempFolderForRestore, string packagesRestorePath,
-            bool rootAllApplicationAssemblies, bool ilcGenerateCompleteTypeMetadata, bool ilcGenerateStackTraceData)
+            bool rootAllApplicationAssemblies, bool ilcGenerateCompleteTypeMetadata, bool ilcGenerateStackTraceData, string trimmerDefaultAction)
             : base(targetFrameworkMoniker, cliPath, GetPackagesDirectoryPath(useTempFolderForRestore, packagesRestorePath), runtimeFrameworkVersion)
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -188,9 +190,18 @@ protected override void GenerateProject(BuildPartition buildPartition, Artifacts
...
+
+            sb.AppendLine(rootAllApplicationAssemblies ? "<PublishTrimmed>false</PublishTrimmed>" : "<TrimMode>link</TrimMode>");
+
+            if (!string.IsNullOrEmpty(trimmerDefaultAction))
+            {
+                sb.Append($"<TrimmerDefaultAction>{trimmerDefaultAction}</TrimmerDefaultAction>");
+            }
+
+            return sb.ToString();
+        }
```
MichalStrehovsky, line 204: Nit: these setting got aligned with IL Linker defaults, so it's now:

* PublishTrimmed is implied to be always true when publishing with NativeAOT. At least the framework needs to be trimmed because the experience of not trimming that is just really bad.
* Application assemblies are rooted by default. One can opt them into trimming with `<TrimmerDefaultAction>link</TrimmerDefaultAction>`

So one could actually fold this into the existing `bool rootAllApplicationAssemblies`: if it's true, do nothing. If it's false, add `<TrimmerDefaultAction>link</TrimmerDefaultAction>`.

adamsitnik: It's almost hidden, but `<TrimmerDefaultAction>link</TrimmerDefaultAction>` is our default now.

https://github.com/dotnet/BenchmarkDotNet/blob/09d9396a7980dcb65be97169b8611c208af075a2/src/BenchmarkDotNet/Toolchains/NativeAot/NativeAotToolchainBuilder.cs#L23

adamsitnik: I've sent https://github.com/dotnet/BenchmarkDotNet/pull/1975 to address it

PR #1973, opened 2022-03-29
don't place DynamicallyAccessedMembers on an array

adamsitnik:
it's a nop for NativeAOT and a bug for WASM. Instead of using `params` I've added ctors that accept 1, 2 and 3 types.

https://github.com/dotnet/BenchmarkDotNet/pull/1960#discussion_r832712652

cc @MichalStrehovsky @radical


adamsitnik:
the fix has been verified by @radical : https://github.com/dotnet/BenchmarkDotNet/pull/1960#discussion_r837746729

radical:
The new (`0.13.1.1747`) nuget needs to be published to the public feeds.

src/BenchmarkDotNet.Annotations/Attributes/GenericTypeArgumentsAttribute.cs
```diff
@@ -12,6 +12,18 @@ public class GenericTypeArgumentsAttribute : Attribute
         // CLS-Compliant Code requires a constructor without an array in the argument list
```
MichalStrehovsky, line 12: Nit: from the comment it looks like this constructor no longer needs to exist

Issue #1680, opened 2021-03-29
Cannot override RD.xml for NativeAOT


kant2002:
I would like to be able override RD.xml for the benchmarks,
Right now I do not see how I can do that without manually modifying source code for the BenchmarkDotNet. 
Class `BenchmarkDotNet.Toolchains.CoreRt.Generator` has access only to `BuildPartition` and `ArtifactsPaths` and I cannot detect original project location. Maybe there a way to preserve original rd.xml from the project?

kant2002:
@adamsitnik do you have an idea what would be proper fix? I really like to be able test with NativeAOT. I can implement that myself probably if I see less invasive fix.

adamsitnik:
I was hoping that the `RdXmlFile` properties are transitive, but they are not: https://github.com/dotnet/runtime/issues/66085#issuecomment-1057055013

This should not be hard to implement, I've described how to do it here: https://github.com/dotnet/performance/issues/2338#issue-1186174019

PR #1976, opened 2022-03-30
Add support for Rd.xml file around project file

kant2002:
For now added either `rd.xml` or files which ends with `.rd.xml`
Rename `rd.xml` supplied by BDN to `bdn_generated.rd.xml`

Closes https://github.com/dotnet/BenchmarkDotNet/issues/1680

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
+++ b/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
@@ -20,6 +20,7 @@ namespace BenchmarkDotNet.Toolchains.NativeAot
     public class Generator : CsProjGenerator
     {
         internal const string NativeAotNuGetFeed = "nativeAotNuGetFeed";
+        internal const string GeneratedRdXmlFileName = "bdn_generated.rd.xml";
 
         internal Generator(string ilCompilerVersion, bool useCppCodeGenerator,
             string runtimeFrameworkVersion, string targetFrameworkMoniker, string cliPath,
@@ -150,7 +151,7 @@ $@"<?xml version=""1.0"" encoding=""utf-8""?>
     <ProjectReference Include=""{GetProjectFilePath(buildPartition.RepresentativeBenchmarkCase.Descriptor.Type, logger).FullName}"" />
   </ItemGroup>
   <ItemGroup>
-    <RdXmlFile Include=""rd.xml"" />
+    {GetRdXmlFiles(buildPartition.RepresentativeBenchmarkCase.Descriptor.Type, logger).Select(_ => $"<RdXmlFile Include=\"{_}\" />{Environment.NewLine}")}
   </ItemGroup>
 </Project>";
 
@@ -185,7 +186,7 @@ $@"<?xml version=""1.0"" encoding=""utf-8""?>
     <ProjectReference Include=""{GetProjectFilePath(buildPartition.RepresentativeBenchmarkCase.Descriptor.Type, logger).FullName}"" />
   </ItemGroup>
   <ItemGroup>
-    <RdXmlFile Include=""rd.xml"" />
+    {GetRdXmlFiles(buildPartition.RepresentativeBenchmarkCase.Descriptor.Type, logger).Select(_ => $"<RdXmlFile Include=\"{_}\" />{Environment.NewLine}")}
   </ItemGroup>
 </Project>";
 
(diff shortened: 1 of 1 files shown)


src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -150,7 +151,7 @@ protected override void GenerateProject(BuildPartition buildPartition, Artifacts
     <ProjectReference Include=""{GetProjectFilePath(buildPartition.RepresentativeBenchmarkCase.Descriptor.Type, logger).FullName}"" />
   </ItemGroup>
   <ItemGroup>
-    <RdXmlFile Include=""rd.xml"" />
+    {GetRdXmlFiles(buildPartition.RepresentativeBenchmarkCase.Descriptor.Type, logger).Select(_ => $"<RdXmlFile Include=\"{_}\" />{Environment.NewLine}")}
```
adamsitnik, line 154: nit `_` is typically used when the argument is ignored. In this case, we are using it.

```suggestion
    {GetRdXmlFiles(buildPartition.RepresentativeBenchmarkCase.Descriptor.Type, logger).Select(file => $"<RdXmlFile Include=\"{file}\" />{Environment.NewLine}")}
```

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -203,6 +204,23 @@ private string GetTrimmingSettings()
...
+            var rdXml = Path.Combine(projectFileFolder, "rd.xml");
+            if (File.Exists(rdXml))
+            {
+                yield return rdXml;
+            }
+
+            foreach (var item in Directory.GetFiles(projectFileFolder, "*.rd.xml"))
+            {
+                yield return item;
+            }
```
adamsitnik, line 221: it seems like this could be simplified if we search for `*rd.xm` files:

```suggestion
            yield return GeneratedRdXmlFileName;
            
            var projectFileFolder = GetProjectFilePath(benchmarkTarget, logger).DirectoryName;
            foreach (var rdFile in Directory.GetFiles(projectFileFolder, "*rd.xml"))
            {
                yield return rdFile;
            }
```

adamsitnik (commented):
@kant2002 big thanks for your contribution!

could you PTAL at the CI failures?

```log
/home/runner/work/BenchmarkDotNet/BenchmarkDotNet/tests/BenchmarkDotNet.IntegrationTests/bin/Release/net5.0/0961d928-ba98-4cca-b6af-fc04c1f5807a/BenchmarkDotNet.Autogenerated.csproj(37,3): error MSB4067: The element <#text> beneath element <ItemGroup> is unrecognized.
```

you should be able to repro it locally by doing:

```cmd
dotnet run -c Release -f net5.0 --project .\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj --filter '*Basic*' --job dry --runtimes nativeaot5.0 --keepFiles true
```

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
kant2002: What if somebody have library which ends with `rd`? or just XML files. For example `Thunderbird.xml`, that would be captured.

adamsitnik: I am afraid that neither `*.rd.xml`  and `*rd.xml` are good filters, as users can call their files without rd prefix like `nativeaot_rules.xml`. But since it's a project with BenchmarkDotNet benchmarks we could just tell them that all that we support is `rd.xml`. @kant2002 what do you think?

@MichalStrehovsky what is your opinion on that? Are NativeAOT using names other than `rd.xml` for their `RdXmlFile` files?

kant2002: Based on my experience projects in the wild was using either just `rd.xml` or I recentrly start push for separate files per [library](https://github.com/kant2002/rdxmllibrary), to at least simplify mundane work. Naming convention which I propose is exaclly `nuget_package.rd.xml`. Also seems to be .NET Native Rd.xml also has approximately same convention, so that's why I choose that pattern.

adamsitnik (approved):
LGTM, big thanks @kant2002 !

PR #1978, opened 2022-03-31
Migrate from .NET 5 to .NET 6.0

kant2002:
.NET 5 would be out of support in early May, so maybe it's time to update test infra from it to .NET 6

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,7 @@ skip_tags: true
 #    environment configuration    #
 #---------------------------------#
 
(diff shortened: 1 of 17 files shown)


adamsitnik:
I think it's the right direction. @AndreyAkinshin what do you think?

adamsitnik (approved):
LGTM, thank you @kant2002 !

AndreyAkinshin (approved):
LGTM, it's a good time for this migration.

PR #1981, opened 2022-04-01
Fix for issue NETSDK1150

OlegOLK:
Starting from net5 Microsoft introduces a breaking change - that will prevent the solution from building if 
**executable project references another executable project and the SelfContained values don't match**
https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/referencing-executable-generates-error

It can be fixed by adding 
`<ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>`
But BenchmarkDotNet does not copy it from the original .csproj for some reason. So I've added it directly to the CspProj.txt template. Not sure about the rest of the templates.

To reproduce this issue you can:
1. create the project A - web project/WPF project
2. create the project B - benchmark project and reference project A
2.1 Add     <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
to the project B properties group in .csproj
3. Run benchmark tests.
4. Test is green
5. Benchmark logs 
`
C:\Program Files\dotnet\sdk\6.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(1092,5): error NETSDK1151: The referenced project '<Project A>' is a self-contained executable.  A self-contained executable cannot be referenced by a non self-contained executable.  For more information, see https://aka.ms/netsdk1151 [<project B>\bin\Release\net6.0\24ead923-3f0b-4b04-bdbd-20efe5990d59\BenchmarkDotNet.Autogenerated.csproj]
`

### DIFF
--- a/src/BenchmarkDotNet/Templates/CsProj.txt
+++ b/src/BenchmarkDotNet/Templates/CsProj.txt
@@ -24,6 +24,7 @@
     <!-- we set LangVersion after $COPIEDSETTINGS$ which might contain LangVersion copied from the benchmarks project -->
     <LangVersion Condition="'$(LangVersion)' == '' Or ($([System.Char]::IsDigit('$(LangVersion)', 0)) And '$(LangVersion)' &lt; '7.3')">latest</LangVersion>
     <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
+    <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/BenchmarkDotNet/Templates/CsProj.txt
+++ b/src/BenchmarkDotNet/Templates/CsProj.txt
@@ -24,8 +24,7 @@
     <!-- we set LangVersion after $COPIEDSETTINGS$ which might contain LangVersion copied from the benchmarks project -->
     <LangVersion Condition="'$(LangVersion)' == '' Or ($([System.Char]::IsDigit('$(LangVersion)', 0)) And '$(LangVersion)' &lt; '7.3')">latest</LangVersion>
     <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
-    <!-- fix for NETSDK1150: https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/referencing-executable-generates-error so the tests can be run
-    even if there is a reference to the .exe project -->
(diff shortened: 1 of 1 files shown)


src/BenchmarkDotNet/Templates/CsProj.txt
```diff
@@ -24,6 +24,7 @@
     <!-- we set LangVersion after $COPIEDSETTINGS$ which might contain LangVersion copied from the benchmarks project -->
     <LangVersion Condition="'$(LangVersion)' == '' Or ($([System.Char]::IsDigit('$(LangVersion)', 0)) And '$(LangVersion)' &lt; '7.3')">latest</LangVersion>
     <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
+    <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
```
adamsitnik, line 27: could you please add a comment explaining why it's needed? (so we don't remove it in the future)

OlegOLK: sure

src/BenchmarkDotNet/Templates/CsProj.txt
```diff
@@ -24,6 +24,9 @@
     <!-- we set LangVersion after $COPIEDSETTINGS$ which might contain LangVersion copied from the benchmarks project -->
     <LangVersion Condition="'$(LangVersion)' == '' Or ($([System.Char]::IsDigit('$(LangVersion)', 0)) And '$(LangVersion)' &lt; '7.3')">latest</LangVersion>
     <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
+    <!-- fix for NETSDK1150: https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/referencing-executable-generates-error so the tests can be run
+    even if there is a reference to the .exe project -->
```
adamsitnik, line 28: nit: I think that the link is enough

```suggestion
    <!-- fix for NETSDK1150: https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/referencing-executable-generates-error  -->
```

adamsitnik (approved):
LGTM, thank you @OlegOLK !

src/BenchmarkDotNet/Templates/CsProj.txt
OlegOLK: change it according to your suggestions, thanks!

PR #1982, opened 2022-04-02
DotNetCliCommand: Add new `RetryFailedBuildWithNoDeps` property

radical:
If a build fails, then it is retried with `--no-dependencies`. But this
can be problematic because msbuild can end up skipping some steps,
resulting in an incorrect build. And that can cause inexplicable issues
when running the benchmarks.

This is seen with wasm/aot builds with emcc, or wasm-ld crashes.

The property defaults to true, but is disabled for wasm builds.


radical:
cc @adamsitnik I'm not sure if adding a new `PublicAPI` is the proper way to do this.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliBuilder.cs
```diff
@@ -32,7 +32,8 @@ public BuildResult Build(GenerateResult generateResult, BuildPartition buildPart
                     buildPartition,
                     Array.Empty<EnvironmentVariable>(),
                     buildPartition.Timeout,
-                    logOutput: LogOutput)
+                    logOutput: LogOutput,
+                    retryFailedBuildWithNoDeps: false)
```
adamsitnik, line 36: `DotNetCliBuilder` is used by other toolchains than WASM. Please set the value to `true` by default (to keep current behavior) and to `false` for the WASM toolchain:

https://github.com/dotnet/BenchmarkDotNet/blob/1d34f856a4ba633af9efc63394abbd0311edfbae/src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolChain.cs#L43

adamsitnik (approved):
LGTM, thank you @radical !

PR #1985, opened 2022-04-06
don't emit debug symbols on platforms where it does not work well

adamsitnik:
It's nice to have for EtwProfiler, but not mandatory.

On x86 disabled due to https://github.com/dotnet/roslyn/issues/59240
On ARM and ARM64 disabled due to: https://github.com/dotnet/roslyn/issues/42393


tmat:
Any reason why not always use Portable PDBs? Windows PDBs are less efficient and have the above issues.

src/BenchmarkDotNet/Templates/CsProj.txt
```diff
@@ -11,9 +11,12 @@
     <OutputType>Exe</OutputType>
     <OutputPath>bin\$CONFIGURATIONNAME$</OutputPath>
     <TreatWarningsAsErrors>False</TreatWarningsAsErrors>
-    <!-- nice to have for EtwProfiler, but not mandatory. On x86 and ARM disabled due to bug in Roslyn. -->
-    <DebugType Condition=" '$(PlatformTarget)' != 'x86' And '$(PlatformTarget)' != 'ARM' ">pdbonly</DebugType>
-    <DebugSymbols Condition=" '$(PlatformTarget)' != 'x86' And '$(PlatformTarget)' != 'ARM' ">true</DebugSymbols>
+    <!-- nice to have for EtwProfiler, but not mandatory.
```
tmat, line 14: I'd recommend removing DebugType and DebugSymbols properties altogether and use the defaults (i.e. Portable PDBs)

adamsitnik:
@tmat that is a great question. I'll check if portable pdbs give us what we want (nice symbol information for PerfView and for disassembler that uses ClrMD)

adamsitnik:
With `<DebugType>portable</DebugType>` ClrMD does not provide information about source code line numbers.

Sample output:

## portable

```assembly
; BenchmarkDotNet.Samples.IntroDisassembly.SumLocal()
       7FF7BA4D81A0 mov       rax,[rcx+8]
       7FF7BA4D81A4 xor       edx,edx
       7FF7BA4D81A6 xor       ecx,ecx
       7FF7BA4D81A8 mov       r8d,[rax+8]
       7FF7BA4D81AC test      r8d,r8d
       7FF7BA4D81AF jle       short M00_L01
M00_L00:
       7FF7BA4D81B1 movsxd    r9,ecx
       7FF7BA4D81B4 add       edx,[rax+r9*4+10]
       7FF7BA4D81B9 inc       ecx
       7FF7BA4D81BB cmp       r8d,ecx
       7FF7BA4D81BE jg        short M00_L00
M00_L01:
       7FF7BA4D81C0 mov       eax,edx
       7FF7BA4D81C2 ret
; Total bytes of code 35
```

## pdbonly

```assembly
; BenchmarkDotNet.Samples.IntroDisassembly.SumLocal()
;             var local = field; // we use local variable that points to the field
;             ^^^^^^^^^^^^^^^^^^
       7FF7BAAB81A0 mov       rax,[rcx+8]
;             int sum = 0;
;             ^^^^^^^^^^^^
       7FF7BAAB81A4 xor       edx,edx
;             for (int i = 0; i < local.Length; i++)
;                  ^^^^^^^^^
       7FF7BAAB81A6 xor       ecx,ecx
       7FF7BAAB81A8 mov       r8d,[rax+8]
       7FF7BAAB81AC test      r8d,r8d
       7FF7BAAB81AF jle       short M00_L01
;                 sum += local[i];
;                 ^^^^^^^^^^^^^^^^
M00_L00:
       7FF7BAAB81B1 movsxd    r9,ecx
       7FF7BAAB81B4 add       edx,[rax+r9*4+10]
       7FF7BAAB81B9 inc       ecx
       7FF7BAAB81BB cmp       r8d,ecx
       7FF7BAAB81BE jg        short M00_L00
;             return sum;
;             ^^^^^^^^^^^
M00_L01:
       7FF7BAAB81C0 mov       eax,edx
       7FF7BAAB81C2 ret
; Total bytes of code 35
```

I'll make it conditional then.

adamsitnik:
I've decided to disable it in the project generated by BDN, as it only matters when it's configured in the user project that defines the benchmarks.

leculver:
FYI:  As of 2.0, ClrMD doesn't interact with PDBs at all.

Previous versions of ClrMD tried to be monolithic (take no dependencies and do everything itself) because it was written before we had a healthy nuget ecosystem.  Now we expect folks who want to do PDB things (like resolve down to file and line number) that they will pull in their own nuget packages for native/portable pdbs.

(I have no opinion on changing the symbol types.  Whatever makes sense for this product.)

tmat:
@adamsitnik In your comment above what tool is the output actually produced by? Does that tool now use older ClrMD to read the source line info?

adamsitnik:
> In your comment above what tool is the output actually produced by?

it's BenchmarkDotNet, using ClrMd 1.x (updating to 2.0 is on our TODO list)

https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Disassembler.x64/ClrMdDisassembler.cs

tmat:
I see, so yeah, as @leculver points out once you update to ClrMD 2.0 you'll loose all source line info and will need to implement it yourself. For Portable PDBs you'd use System.Reflection.Metadata .NET APIs and for Windows PDBs Microsoft.DiaSymReader.Native package.

kg:
Is it possible to undo this now? Having symbols for the benchmark binaries would allow me to see line-level CPU usage in Superluminal profiler.

timcassell:
@kg You should be able to do it using MsBuildArguments. `job.WithArguments([ new MsBuildArgument("/p:DebugSymbols=true") ])`

PR #1990, opened 2022-04-14
[Mono] Change AOT mode to Normal AOT with LLVM JIT fall back

fanyang-mono:
Some of the microbenchmarks use generic type, where Mono full AOT would emit a generic function to handle various types, which is very slow and doesn't worth benchmarking. This is one example:
https://github.com/dotnet/performance/blob/main/src/benchmarks/micro/libraries/System.Numerics.Vectors/Perf_VectorOfT.cs

This PR will change the AOT mode to Normal LLVM AOT with LLVM JIT fallback. Because Normal AOT leaves the generic type cases for JIT to handle. That way optimized code will be emitted.

Additionally, Maui Android is using Profiled AOT with JIT fallback, which is similar to Normal AOT with JIT fallback. Ideally, they would like to leverage LLVM too at some point.

cc: @naricc

### DIFF
--- a/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
@@ -137,7 +137,7 @@ namespace BenchmarkDotNet.Extensions
 
                 if (aotruntime.AOTCompilerMode == MonoAotCompilerMode.llvm)
                 {
-                    start.EnvironmentVariables["MONO_ENV_OPTIONS"] = "--full-aot";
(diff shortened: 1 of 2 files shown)


SamMonoRT:
@fanyang-mono  - can you add a comment why we are switching from full to normal aot describing the limitations of these runs in the labs. That way we can discuss with @adamsitnik  and Drew to see if we do something about those.

SamMonoRT (approved):

fanyang-mono:
@adamsitnik Please merge this PR when you get a chance. Thanks in advance.

naricc:
Looks good to me.

PR #1991, opened 2022-04-14
Features snapshot

workgroupengineering:
_This is a preliminary version._

## The purpose
The purpose of this PR is to make it possible to compare benchmarks between two versions or run when nuget packages cannot be used.

## How works

Use the ability to create custom toolchains to inject ExecuteResults previously saved into a store. To associate methods between two executions, hash of BenchmarkCase.Descriptor.  At this time only the json store is implemented.

### Pro

- it is immune to breaking changes
- not breaking changes (add only InternalsVisibleTo to BenchmarkDotNet bcl)
- Plugable store

### Cons

- Using Newtonsoft.Json ( simplejson can not serializze BenchmarkCase.Descriptor )

## Known Issues

- although ExecuteResults is the same as the previous execution it reports slightly different values.

## Example of output

``` ini

BenchmarkDotNet=v0.13.1.20220414-develop, OS=Windows 10 (10.0.19044.1586/21H2/November2021Update)
Intel Core i7-7700K CPU 4.20GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.300-preview.22154.4
  [Host]   : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT DEBUG  [AttachedDebugger]
  Baseline : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  ShortRun : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT


```
|           Method |      Job |          Toolchain | IterationCount | LaunchCount | WarmupCount |     Mean |     Error |   StdDev | Ratio | RatioSD |
|----------------- |--------- |------------------- |--------------- |------------ |------------ |---------:|----------:|---------:|------:|--------:|
|            Sleep | Baseline | JsonShanpshotStore |        Default |     Default |     Default | 15.63 ms |  1.365 ms | 0.075 ms |  1.00 |    0.00 |
|            Sleep | ShortRun |            Default |              3 |           1 |           3 | 16.67 ms |  2.897 ms | 0.159 ms |  1.07 |    0.02 |
|                  |          |                    |                |             |             |          |           |          |       |         |
| Thread.Sleep(15) | Baseline | JsonShanpshotStore |        Default |     Default |     Default | 16.82 ms |  4.494 ms | 0.246 ms |  1.00 |    0.00 |
| Thread.Sleep(15) | ShortRun |            Default |              3 |           1 |           3 | 17.53 ms | 19.891 ms | 1.090 ms |  1.04 |    0.08 |


workgroupengineering:
any suggestion is welcome

workgroupengineering:
@adamsitnik @AndreyAkinshin  do you have any suggestions to improve the quality ad fix issue of this PR?
It was originally intended as a separate repository, but I need to access internal classes.

AndreyAkinshin:
@workgroupengineering unfortunately, we can't reference Newtonsoft.Json because of the following reasons:
* Newtonsoft.Json [uses](https://github.com/JamesNK/Newtonsoft.Json/blob/13.0.1/Src/Newtonsoft.Json.Tests/Newtonsoft.Json.Tests.csproj#L48) BenchmarkDotNet in the tests. If we reference Newtonsoft.Json in BenchmarkDotNet, it would introduce severe problems in Newtonsoft.Json benchmarks.
* Newtonsoft.Json is extremely popular, but different projects use different versions of this library. If we Newtonsoft.Json in BenchmarkDotNet it would introduce a lot of problems for users who use other versions of Newtonsoft.Json.

timcassell:
I find this interesting. I have needed to compare different version of the same library before, and it's quite a pain. If the dependency is the only issue, could a different serialization format be used, like xml?

workgroupengineering:
Hi @AndreyAkinshin, 
Ti is draf, Newtonsoft.Json  is only for tests. 
i have plans for the future to extend simple json and implement other types of serilizer.
I posted the PR for suggestions on how to improve it.

workgroupengineering:
I have  removed Newtonsoft.Json dependecy

PR #1994, opened 2022-04-16
NativeAOT: IlcOptimizationPreference & IlcInstructionSet

adamsitnik:
Changes:

- allow the users to specify `IlcOptimizationPreference`, use `Speed` as default
- allow the users to specify `IlcInstructionSet`, use host process settings as default

Sample benchmark:

```cs
private string a = new string('a', 1000) + "a";
private string b = new string('a', 1000) + "b";

[Benchmark]
public bool Compare() => a.Equals(b, System.StringComparison.Ordinal);
```

```cmd
dotnet run -c Release -f net6.0 --filter *Compare --job short --keepFiles true --runtimes nativeaot7.0
```

## Before

|  Method |       Runtime |          Toolchain |     Mean |    Error |   StdDev |
|-------- |-------------- |------------------- |---------:|---------:|---------:|
| Compare | NativeAOT 7.0 | ILCompiler 7.0.0-* | 66.88 ns | 2.263 ns | 0.124 ns |

## After

|  Method |       Runtime |          Toolchain |     Mean |    Error |   StdDev |
|-------- |-------------- |------------------- |---------:|---------:|---------:|
| Compare | NativeAOT 7.0 | ILCompiler 7.0.0-* | 35.42 ns | 0.539 ns | 0.030 ns |

Sample settings on my x64 AMD machine:

```xml
<ItemGroup>
    <IlcArg Include="--instructionset:base,sse,sse2,sse3,sse4.1,sse4.2,avx,avx2,aes,bmi,bmi2,fma,lzcnt,pclmul,popcnt" />
</ItemGroup>
```

cc @jkotas @MichalStrehovsky

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
+++ b/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
@@ -26,7 +26,8 @@ namespace BenchmarkDotNet.Toolchains.NativeAot
             string runtimeFrameworkVersion, string targetFrameworkMoniker, string cliPath,
             string runtimeIdentifier, IReadOnlyDictionary<string, string> feeds, bool useNuGetClearTag,
             bool useTempFolderForRestore, string packagesRestorePath,
-            bool rootAllApplicationAssemblies, bool ilcGenerateCompleteTypeMetadata, bool ilcGenerateStackTraceData)
+            bool rootAllApplicationAssemblies, bool ilcGenerateCompleteTypeMetadata, bool ilcGenerateStackTraceData,
+            string ilcOptimizationPreference)
             : base(targetFrameworkMoniker, cliPath, GetPackagesDirectoryPath(useTempFolderForRestore, packagesRestorePath), runtimeFrameworkVersion)
         {
             this.ilCompilerVersion = ilCompilerVersion;
@@ -38,6 +39,7 @@ namespace BenchmarkDotNet.Toolchains.NativeAot
             this.rootAllApplicationAssemblies = rootAllApplicationAssemblies;
             this.ilcGenerateCompleteTypeMetadata = ilcGenerateCompleteTypeMetadata;
             this.ilcGenerateStackTraceData = ilcGenerateStackTraceData;
+            this.ilcOptimizationPreference = ilcOptimizationPreference;
         }
 
         private readonly string ilCompilerVersion;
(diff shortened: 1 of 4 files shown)


src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -237,5 +262,55 @@ private void GenerateReflectionFile(ArtifactsPaths artifactsPaths)
...
+                    // todo: handle "lse"
+                    break;
+                default:
+                    yield break;
+            }
+        }
+
+        private static bool GetIsSupported(string typeName)
+        {
+            Type type = Type.GetType(typeName);
```
adamsitnik, line 310: BDN targets `netstandard2.0`, this is why I had to use reflection here ;/

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -237,5 +262,55 @@ private void GenerateReflectionFile(ArtifactsPaths artifactsPaths)
...
+                case Platform.Arm64:
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.ArmBase")) yield return "base";
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.AdvSimd")) yield return "neon";
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.Aes")) yield return "aes";
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.Crc32")) yield return "crc";
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.Dp")) yield return "dotprod";
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.Rdm")) yield return "rdma";
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.Sha1")) yield return "sha1";
+                    if (GetIsSupported("System.Runtime.Intrinsics.Arm.Sha256")) yield return "sha2";
+                    // todo: handle "lse"
```
adamsitnik, line 302: @jkotas when can "lse" be enabled? https://github.com/dotnet/runtime/blob/ce61c09a5f6fc71d8f717d3fc4562f42171869a0/src/coreclr/tools/Common/JitInterface/CorInfoInstructionSet.cs#L741

jkotas: `lse` enables more efficient code generation for `Interlocked` APIs. It is not exposed via hardware intrinics managed APIs. `dczva` and `rcpc` are in the same category.

You would have to replicate what the unmanaged runtime does to auto-detect it.

Issue #1954, opened 2022-02-12
Wishlist of features I'd find useful


AndyAyersMS:

stephentoub:
> iirc @stephentoub asked me for that in the past.

Yup:
https://github.com/dotnet/BenchmarkDotNet/issues/1774

nawfalhasan:
Would like to see this get done: https://github.com/dotnet/BenchmarkDotNet/issues/1634 :(

AndyAyersMS:
I also find myself wishing there was simpler/smoother integration with linux `perf`. Exporting to perfview is ok for CPU samples but for HW counters it's not really viable.

Something along these lines:
* create a `perf` diagnoser that runs the benchmark subprocess under `perf report` or `perf stat`, allowing me to specify the events of interest (often PMU events)
* Enables perf map under the covers
* post processing of the data to inject markers/events for the actual intervals so that later on we can filter the reporting to just those stretches of time (using perhaps the switch on/off capabilities)
* post processing using `perf inject -j` to add in the mappings for jitted code ranges
* continued effort to both describe all the runtime stubs in the perf mappings and (where possible) make unwinding work through all of the stubs somehow so the call stack modes aren't badly broken (would also help perfview)
* Similar injection of runtime events, eg a mashup of `-p EP` with the above

Right now I am running `perf record` over the entire BDN invocation and either boosting the iteration/invocation counts so that the actual intervals clearly dominate everything else, or slicing and looking at only the last 10% (say) of the recorded data.

AndyAyersMS:
> > VTune diagnoser or similar that uses VTune API to mark actual measurement intervals
> 
> this sounds very interesting. Do you have any links to VTune API docs?

https://www.intel.com/content/www/us/en/develop/documentation/vtune-help/top/api-support/instrumentation-and-tracing-technology-apis.html

BruceForstall:
Related: I'd like to see `--join` fixed so multiple filter expression results can all show up in a single results table: https://github.com/dotnet/performance/issues/1855

AndyAyersMS:
It would also be nice to have an integrated diagnoser for ETW that is benchmark interval aware. I have a crude start at this in https://github.com/AndyAyersMS/instructionsretiredexplorer; it can post-process the ETW (actual interval aware) and project onto managed method names & tiering variants. eg

```
[...]

PR #2002, opened 2022-05-05
when user specifies both --runtimes and --corerun, multiple independent jobs should be created

adamsitnik:
Contributes to #1954:

> Mix `--corerun` and `--netX.Y` on one command line

cc @stephentoub @AndyAyersMS

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -78,7 +78,7 @@ namespace BenchmarkDotNet.ConsoleArguments
             {
                 parser
                     .ParseArguments<CommandLineOptions>(args)
-                    .WithParsed(options => result = Validate(options, logger) ? (true, CreateConfig(options, globalConfig), options) : (false, default, default))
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -313,20 +313,45 @@ private static Job GetBaseJob(CommandLineOptions options, IConfig globalConfig)
...
                 yield return CreateCoreJobWithCli(baseJob, options);
+            }
             else
-                foreach (string runtime in options.Runtimes) // known runtimes
-                    yield return CreateJobForGivenRuntime(baseJob, runtime, options);
+            {
+                // in case both --runtimes and --corerun are specified, the first one is returned first and becomes a baseline job
+                string first = args.FirstOrDefault(arg =>
+                    arg.Equals("--runtimes", StringComparison.OrdinalIgnoreCase)
+                    || arg.Equals("--r", StringComparison.OrdinalIgnoreCase)
```
adamsitnik, line 335:
```suggestion
                    || arg.Equals("-r", StringComparison.OrdinalIgnoreCase)

```

Issue #1860, opened 2021-12-04
Provide a way to run benchmarks under Unity Mono and IL2CPP


MichalPetryka:
Currently, there is no official (but there have been attempts it seems, looking at #1838 and #759) way to use BenchmarkDotNet under Unity nor to benchmark code running under Unity. This has resulted in many improper benchmarks with incorrect results being shared around the internet. Providing an option to use Unity Mono and IL2CPP as runtimes would be highly beneficial for game development, since that is an usecase where perfomance often matters.

I thought about how the support should look like and I've came up with a few things:
- BenchmarkDotNet would contain a project template that would contain the code required to run the benchmarks, where the benchmark code would be copied on build.
- Since Unity allows only a single instance of a project to be open at once, BenchmarkDotNet would need to create a separate copy of the project generated from the template or to compile them sequentially.
- The app would be a full Unity app for the Mono runtime, instead of just running the code with their fork of Mono to also measure the engines real impact on performance and also since that would not be possible with IL2CPP.
- The app would use Unitys headless mode to reduce the performance overhead and to allow issue-less measurement under headless server environments.
- Nuget support would have to be realized by downloading the dependencies and copying them into the created project.
- Native libraries referenced in the project would need to be similarly copied.
- Unlike the Mono runtime target, Unity runtimes would be versioned like .Net.
- The engine would need to be installed manually by the user.
- In order to allow benchmarking of the Unity API, BenchmarkDotNet would need to provide reference assemblies allowing access to them. This could be realised by providing a separate BenchmarkDotNet.Unity Nuget package. This would need to be versioned accordingly to Unity
- The IL2CPP target, being a Full AOT runtime like NativeAOT (formerly CoreRT), would face the same restrictions.

Open questions:
- Should BenchmarkDotNet build a code binary with dotnet and then include that in the project or copy the code directly?
- Unlike dotnet, Unity doesn't add itself to PATH nor provide an easy way to access its install location. This could make specifing the Engine path by the user required.
- Should the template project provide optional Unity features like DOTS in order to make it possible to benchmark them?
- I'm not sure if all diagnosers like `MemoryDiagnoser` or `DisassemblyDiagnoser` would be possible to implement.
- AFAIK Unitys Terms of Service wouldn't allow the redistribution of referenece assemblies generated against their code, so this would require some cooperation from Unity and their legal team.

[...]

PR #2018, opened 2022-06-12
Multi-target to .NET Standard 2.1, port JetBrains' nullability annotations et. al.

teo-tsirpanis:
Until now the `BenchmarkDotNet` library exclusively targets .NET Standard 2.0, which means that when installed on newer frameworks it references more packages than actually needed. Worse, some of these packages are of an older version that targeted only .NET Standard 1.x, and significantly bloat BenchmarkDotNet's transitive dependencies.

With this PR, I multi-targeted BenchmarkDotNet to .NET Standard 2.1, saving three packages (and one more from both frameworks), and updated some of them. The package's transitive dependencies have been _drastically_ reduced. To illustrate this, the project's `.deps.json` file was __1921__ lines long, and now has more than halved to __632__ lines on .NET Standard 2.0 and __557__ on .NET Standard 2.1.

An obstacle to this porting was `JetBrains.Annotations`' nullability annotations that predate C# 8.0's nullability annotations. On .NET Standard 2.1 there were two attributes named `NotNullAttribute`, and the compiler did not know which to pick. I enabled `<Nullable>annotations</Nullable>` on the project, removed `[JetBrains.Annotations.NotNullAttribute]`, and replaced `[JetBrains.Annotations.CanBeNull]` with `?`. Entirely annotating the project is a much bigger undertaking which is left for another time.

After this PR gets merged, I plan also multi-targeting for .NET 6, to take advantage of newer framework features. I already did in this PR with https://github.com/dotnet/BenchmarkDotNet/commit/aac2b7efafa54380d06c125053df0fdb60d18893 for example.


teo-tsirpanis:
Yeah, I noticed it after opening this PR, not the first time it happens. 😂

Since mine is bigger and more comprehensive, I recommend to merge it first, and leave the .NET 6 porting work to you.

teo-tsirpanis:
@adamsitnik I synched the PR with the changes #2012 brought. Can you take a look?

teo-tsirpanis:
@adamsitnik friendly reminder.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
```diff
@@ -2,8 +2,9 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <AssemblyTitle>BenchmarkDotNet</AssemblyTitle>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
```
adamsitnik, line 5: In a few weeks from now (Dec 13 2022), .NET Core 3.1 will be out of support and .NET 6 will be the oldest supported .NET Standard 2.1 runtime. Since we already target `net6.0` I don't think it's worth adding `netstandard2.1` to the list.

```suggestion
    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
```

src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/BenchmarkActionFactory_Base.cs
```diff
@@ -38,10 +36,10 @@ protected static TDelegate CreateWorkload<TDelegate>([CanBeNull] object targetIn
...
 
             protected static TDelegate CreateWorkloadOrOverhead<TDelegate>(
-                [CanBeNull] object targetInstance,
-                [CanBeNull] MethodInfo workloadMethod,
-                [NotNull] TDelegate overheadStaticCallback,
-                [NotNull] TDelegate overheadInstanceCallback)
+                object? targetInstance,
+                MethodInfo? workloadMethod,
+                TDelegate overheadStaticCallback,
+                TDelegate overheadInstanceCallback) where TDelegate : notnull
```
adamsitnik, line 42: nice, I was not aware of the existence of the `notnull` generic constraint 👍

adamsitnik (commented):
The nullability annotations look very good, I wonder how many warnings we get right now? cc @AndreyAkinshin who introduced the JetBrains annotations

When it comes to adding .NET Standard 2.1 target, I am against as in 3 weeks `net6.0` will the oldest officially supported .NET Standard 2.1 runtime and we already target `net6.0`.

Thank you for your contribution @teo-tsirpanis !

src/BenchmarkDotNet/BenchmarkDotNet.csproj
teo-tsirpanis: There's also Xamarin and Unity which are still supported.

AndreyAkinshin (approved):
I like it!

@adamsitnik once we agree on the set of target frameworks, we can merge it.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
adamsitnik: We don't support Unity (it would require a lot of extra work) and Xamarin works just fine with .NET Standard 2.0.

teo-tsirpanis: 2.1 would remove a couple more dependencies and avoid reflection in `IsAot`.

In my projects I usually target .NET Standard 2.0 for compatibility, .NET 6 for new features and .NET Standard 2.1 for everything in between, but I understand you not wanting this (the most obvious reason being that the 2.1 binaries won't be tested).

teo-tsirpanis:
Please wait before merging, I have to change the title and description, and add a couple more changes I thought.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
MichalPetryka: > We don't support Unity (it would require a lot of extra work) and Xamarin works just fine with .NET Standard 2.0.

#1860 is an open issue for that.

YegorStepanov:
`EnableNETAnalyzers` adds many new warnings, shouldn't the build fail because of `TreatWarningsAsErrors`?

__common.props__:
```xml
<EnableNETAnalyzers>true</EnableNETAnalyzers>
<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```

teo-tsirpanis:
@YegorStepanov these warnings were already enabled but with the obsolete `Microsoft.CodeAnalysis.FxCopAnalyzers` package. This PR does it the right way.

AndreyAkinshin:
@teo-tsirpanis @adamsitnik all of our CI builds are red now. Here is the error:

```
/home/runner/work/BenchmarkDotNet/BenchmarkDotNet/tests/BenchmarkDotNet.IntegrationTests/BenchmarkDotNet.IntegrationTests.csproj : error NU1504: Duplicate 'PackageReference' items found. Remove the duplicate items or use the Update functionality to ensure a consistent restore behavior. The duplicate 'PackageReference' items are: System.Runtime.CompilerServices.Unsafe 6.0.0, System.Runtime.CompilerServices.Unsafe 6.0.0. [TargetFramework=net462]
```

adamsitnik:
@AndreyAkinshin @YegorStepanov appologies for breaking the build, thanks for fixing it!

teo-tsirpanis:
I set it to `annotations` to minimize the work I had to do. Feel free to change it to `enable` @YegorStepanov, I don't have time for it.

PR #2033, opened 2022-07-06
Bugfix MetricColumn: Respect unit when formatting values.

mawosoft:
The previous implementation would only use the given unit if SummaryStyle.PrintUnitsInContent was true, otherwise it would fallback to the default unit, regardless whether SummaryStyle.PrintUnitsInHeader was true or not. This fix takes both SummaryStyle.PrintUnitsIn* into account.

It should be noted that other columns than the MetricColumn do **not** take the visibility of units into account and always use the given unit, be it explicitly selected by the user, or calculated via GetBest*Unit().


AndreyAkinshin:
@mawosoft thanks for the PR! FYI: I have added a test for your fix: https://github.com/dotnet/BenchmarkDotNet/commit/9216a7727cf01ed61674f302932c49b708de1a54

Issue #2009, opened 2022-05-16
Cleanup the dependencies


adamsitnik:
In https://github.com/dotnet/msbuild/commit/20cdf6ff1f617fc6d486c170dff95df0300676e3#r73773876 @ericstj pointed that we could do the following things to cleanup BDN dependencies:

> There are references to a lot of packages which are out of date and not used on later frameworks.

> In this particular case, the references to System.Reflection.Emit 4.3.x from Benchmark.NET are causing .NETCore 1.1 packages to be downloaded and never used.  Referencing the newer version of System.Reflection.Emit would avoid this, so that's the most important thing to fix these Component Governance warnings.

> I took a few minutes to review the dependencies of BDN and here's what I would recommend:
> 1. In the netstandard2.0 configuration:
    - remove System.ValueTuple - it's part of netstandard2.0
    - update System.Reflection.Emit* to 4.7.0
    - update Microsoft.DotNet.PlatformAbstractions to 3.1.6
    - update System.Threading.Tasks.Extensions to 4.5.4
    - update Microsoft.Win32.Registry to 5.0.0
    - update System.Management to 6.0.0

> 2. Consider a `netstandard2.1` target and do the following (in addition to 1)
    - Remove System.Reflection.Emit*
    - Remove System.Threading.Tasks.Extensions
    - Request that Perfolizer produce a version that targets this framework and removes System.Memory reference.

> 3. Consider a `net6.0` or newer target.
    - Remove PlatformAbstractions and use [RuntimeInformation](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.runtimeidentifier?view=net-6.0)
    - Remove Microsoft.Win32.Registry

We should definitely address the `nestandard2.0` issues. We can also consider adding the `net6.0` target.

I am marking this issue as up-for-grabs. The person who is willing to work on it should test following scenarios:

1. "Modern" project files - you can create a new .NET 6.0+ console app, install BDN and run a simple benchmark
2. "Old" project files - you can create a new .NET 4.6.1 console app, install BDN and run a simple benchmark

To test local packages, you need to use the following command:

```cmd
dotnet pack -c Release .\BenchmarkDotNet.sln
```

And add the folders that contain produced NuGet packages to nuget.config file (you can run `dotnet new nugetconfig` to create the file):

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
[...]

PR #2043, opened 2022-07-18
restore Microsoft.DotNet.PlatformAbstraction dependency

adamsitnik:
I think that we misunderstood @ericstj suggestion from https://github.com/dotnet/BenchmarkDotNet/issues/2009#issue-1237469322 

By:

> Remove PlatformAbstractions and use [RuntimeInformation](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.runtimeidentifier?view=net-6.0)

@ericstj most likely meant stop using the old APIs and use the new ones **if they meet your requirements**.

But as @martincostello has discovered in https://github.com/dotnet/BenchmarkDotNet/pull/2012#issuecomment-1139662288 the way the OS name and version is represented has changed and we cannot accept that. Some people (including .NET Perf Team) might be using the string to identify the OS.

Example: the link to the Windows results contains OS version number in certain format: https://pvscmdupload.blob.core.windows.net/reports/allTestHistory/refs/heads/main_x64_Windows%2010.0.18362/AllTestindex.html 

Since it's OK to keep using the old package: https://github.com/dotnet/docs/issues/18438

> You can continue to use older versions of the library if they meet your requirements.

I am restoring this dependency and removing the copy of `PlatformApi` code.


adamsitnik:
CI failures are unrelated (https://github.com/dotnet/runtime/pull/72325)

PR #2020, opened 2022-06-16
Opt out of metadata trimming

MichalStrehovsky:
Benchmark.NET opts itself into trimming, but it's trim compatible only by accident ~~(it's reflecting on things that are statically referenced and therefore would be kept by IL-based trimming solutions). NativeAOT is able to trim metadata more aggressively - e.g. an accessed field might still have its name and metadata removed and the only thing that is kept is a location in memory that can be read and written.~~

~~I'm now seeing Benchmark.NET failing with latest ILCompiler packages after https://github.com/dotnet/runtime/pull/70201 that enabled more aggressive trimming.~~

It's reflecting on a field that is not used in any other way than by unanalyzable reflection. We would previously keep all fields, irrespective of whether they're used on reflected on https://github.com/dotnet/runtime/pull/70546.

Benchmark.NET should ideally `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` in the BenchmarkDotNet.csproj project and fix all the warnings and the problem will go away. But enabling compat mode should help too.

PR #2046, opened 2022-07-20
Make a bit of BenchmarkDotNet trimmable

MichalStrehovsky:
Another attempt at fixing the problem I tried to fix in #2020.

In that pull request, I reflection-rooted all of BenchmarkDotNet, but that looks like it increases the size of the closure too much (https://github.com/dotnet/performance/pull/2532). So here I'm rolling that part back and annotating enough of the `Characteristic` APIs to make it so that we shouldn't run into the original problem.

I basically added `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` to the BechmarkDotNet project, looked at the trimming warning in https://github.com/dotnet/BenchmarkDotNet/blob/63e28c100a42a6492d09a0b93a8a4c141061bb0d/src/BenchmarkDotNet/Characteristics/CharacteristicHelper.cs#L49-L68 where we were getting the crash and kept adding annotations until it stopped generating new warnings.

We now have an annotation in a spot that should make sure the `Job` class has all fields and properties kept (it's used with one of the annotated `Create` methods).

There are more trimming warnings that I didn't try to solve because they're unrelated to the problem at hand.

Cc @adamsitnik 
Cc @dotnet/ilc-contrib FYI

### DIFF
--- a/src/BenchmarkDotNet.Annotations/Attributes/DynamicallyAccessedMembersAttribute.cs
+++ b/src/BenchmarkDotNet.Annotations/Attributes/DynamicallyAccessedMembersAttribute.cs
@@ -23,7 +23,7 @@ namespace System.Diagnostics.CodeAnalysis
     /// </remarks>
     [AttributeUsage(
         AttributeTargets.Field | AttributeTargets.ReturnValue | AttributeTargets.GenericParameter |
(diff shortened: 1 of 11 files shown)


adamsitnik (approved):
Looks good to me! Thanks a lot @MichalStrehovsky !

PR #2049, opened 2022-07-22
Added the kernel stacks keyword as a parameter to the ETWConfig so that call stacks for specific kernel keywords can be collected. 

mrsharm:
Use case, there are times where the trace file needs to contain specialized keywords for stacks. Right now, the value is hardcoded to ``Profile`` and this is useful in many cases but to get the flexibility of analyzing cases such as Page Faults via the ``ReferenceSet`` keyword additional stacks need to be made available for more wholistic analysis.


adamsitnik (approved):
LGTM, thank you @mrsharm !

PR #2051, opened 2022-07-25
extend printed Runtime Info with simple Hardware Intrinsics information

adamsitnik:
Today I was running benchmarks for https://github.com/dotnet/runtime/pull/72788 on a remote ARM machine and I had no idea whether I was comparing apples to apples, because I did not know whether this machine supported `AdvSimd` or not.

I propose that we extend the info printed by BDN with simple Hardware Intrinsics information

```cmd
dotnet run -c Release -f net6.0 -- --info
```

New information is printed after the architecture and JIT:

```ini
BenchmarkDotNet=v0.13.1.20220725-develop, OS=Windows 11 (10.0.22000.795/21H2)
AMD Ryzen Threadripper PRO 3945WX 12-Cores, 1 CPU, 24 logical and 12 physical cores
.NET SDK=7.0.100-preview.6.22352.1
  [Host] : .NET 6.0.7 (6.0.722.32202), X64 RyuJIT AVX2
```

@tannergooding @egorbo @stephentoub please feel free to suggest improvements


adamsitnik:
I am wondering whether Vector Size should also be printed?

```ini
  [Host] : .NET 6.0.7 (6.0.722.32202), X64 RyuJIT AVX2 VectorSize=256
```

EgorBo:
> I am wondering whether Vector Size should also be printed?

I'd say that it's redundant, it's only ever 256 with AVX 🙂

kunalspathak:
I love the idea of printing as much hardware information as possible. When we file issues, I also expect to have this information printed in some fashion in the issues. cc: @Drewscoggins

DrewScoggins:
Yeah, once this goes in I will add it to the information that we collect as per run.

tannergooding:
> I'd say that it's redundant, it's only ever 256 with AVX 🙂

It's 256 with `AVX2`, it's 128 with `AVX`. In the future, it may be dynamic or support 512; so I think its worth printing.

It's worth noting that we likely want:
* Pick highest of `SSE`, `SSE2` (baseline, but users may disable ISAs with `COMPlus_Enable*=0`), `SSE3`, `SSSE3`, `SSE41`, `SSE42`, `AVX`, `AVX2`
* We may also want each of these: `POPCNT`, `BMI1`, `BMI2` (not an extension of `BMI1`), `FMA`, `LZCNT`, `MOVBE`
* In the future we may also have others to consider like `AVX512F` and its extensions
* There is also `AES`, `PCLMUL`, `AvxVnni`, and `X86Serialize` but those are "less important"

For Arm64:
* `AdvSimd` (baseline, but users may disable ISAs with `COMPlus_Enable*=0`)
* We may also want each of these `Dp`, `Rdm`
* There is also `Aes`, `Crc32`, `Sha1`, and `Sha256` but those are "less important"

If we want to keep it "simple", I'd suggest printing the highest hierarchical "SIMD" ISA and then having a verbose output which prints "everything".

MichalPetryka:
We might wanna print `Vector<T>.IsHardwareAccelerated` on older runtimes without `System.Runtime.Intrinsics`.

adamsitnik:
The way I implemented it right now:

The Summary printed at the end contains only the highest of AVX/SSE:

image

But in the log for every benchmark you can find the full list:

image

And exported JSON file has it as well:

image

Since the intrinsics can be disabled using env vars, this is part of the benchmar info (unique to every benchmark) rather than host info (common for all benchmarks)

```cmd
--envVars COMPlus_EnableAVX:0
```

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
```diff
@@ -9,6 +12,83 @@ namespace BenchmarkDotNet.Portability.Cpu
...
+        {
+            if (IsX86Avx2Supported)
+                return "AVX2";
+            else if (IsX86AvxSupported)
+                return "AVX";
+            else if (IsX86Sse42Supported)
+                return "SSE4.2";
+            else if (IsX86Sse41Supported)
+                return "SSE4.1";
+            else if (IsX86Sse3Supported)
```
tannergooding, line 34: Ssse3 should be between sse3 and sse41

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
```diff
@@ -9,6 +12,83 @@ namespace BenchmarkDotNet.Portability.Cpu
...
+                return "SSE3";
+            else if (IsX86Sse2Supported)
+                return "SSE2";
+            else if (IsX86SseSupported)
+                return "SSE";
+            else if (IsArmAdvSimdSupported)
+                return "AdvSIMD";
+            else if (IsArmBaseSupported)
```
tannergooding, line 42: Base should likely be returned rather than empty string or it should also check for x86base

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
adamsitnik: great catch @tannergooding !  👍

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
```diff
@@ -0,0 +1,272 @@
...
+namespace BenchmarkDotNet.Portability.Cpu
+{
+    internal static class HardwareIntrinsics
+    {
+        internal static string GetVectorSize()
+        {
+#if NET6_0_OR_GREATER
+            if (Vector.IsHardwareAccelerated)
+                return $"VectorSize={Vector<byte>.Count * 8}";
+#endif
```
MichalPetryka, line 20: Why is this .Net 6+? Those methods also existed on .Net Framework and such.

adamsitnik: @MichalPetryka thanks, I was not aware of that!

MichalPetryka: Just do note that there you have to install a nuget package for it.

PR #2052, opened 2022-07-26
Address code review feedback

adamsitnik:
@tannergooding PTAL

### DIFF
--- a/src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
+++ b/src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
@@ -1,9 +1,9 @@
 ﻿using System.Collections.Generic;
+using System.Numerics;
 using BenchmarkDotNet.Environments;
 #if NET6_0_OR_GREATER
 using System.Runtime.Intrinsics.X86;
 using System.Runtime.Intrinsics.Arm;
-using System.Numerics;
 #elif NETSTANDARD2_0_OR_GREATER
 using System;
 #endif
@@ -12,14 +12,7 @@ namespace BenchmarkDotNet.Portability.Cpu
 {
     internal static class HardwareIntrinsics
     {
-        internal static string GetVectorSize()
-        {
-#if NET6_0_OR_GREATER
-            if (Vector.IsHardwareAccelerated)
-                return $"VectorSize={Vector<byte>.Count * 8}";
-#endif
-            return string.Empty;
-        }
+        internal static string GetVectorSize() => Vector.IsHardwareAccelerated ? $"VectorSize={Vector<byte>.Count * 8}" : string.Empty;
 
         internal static string GetShortInfo()
         {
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
```diff
@@ -12,14 +12,7 @@ namespace BenchmarkDotNet.Portability.Cpu
...
     {
-        internal static string GetVectorSize()
-        {
-#if NET6_0_OR_GREATER
-            if (Vector.IsHardwareAccelerated)
-                return $"VectorSize={Vector<byte>.Count * 8}";
-#endif
-            return string.Empty;
-        }
+        internal static string GetVectorSize() => Vector.IsHardwareAccelerated ? $"VectorSize={Vector<byte>.Count * 8}" : string.Empty;
```
adamsitnik, line 15: cc @MichalPetryka

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
```diff
@@ -117,6 +117,13 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
             GetIsSupported("System.Runtime.Intrinsics.X86.Sse3");
 #endif
 
+        internal static bool IsX86Ssse3Supported =>
+#if NET6_0_OR_GREATER
+            Ssse3.IsSupported;
+#elif NETSTANDARD
+            GetIsSupported("System.Runtime.Intrinsics.X86.Ssse3");
+#endif
+
```
MichalPetryka, line 126: Does the `#if` here really help with performance in a noticeable way? Changing all of those to just this would make everything much shorter:
```suggestion
        internal static bool IsX86Ssse3Supported { get; } =
            GetIsSupported("System.Runtime.Intrinsics.X86.Ssse3");
```

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
```diff
@@ -75,6 +74,7 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
                         break;
```
MichalPetryka, line 74: [SERIALIZE](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics.x86.x86serialize?view=net-7.0) is missing here.

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
adamsitnik: > Does the #if here really help with performance in a noticeable way? Changing all of those to just this would make everything much shorter:

This code is going to be executed by NativeAOT which would require some extra ceremony to get reflection working.

tannergooding (approved):

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
adamsitnik: @MichalPetryka another great catch, thank you!

MichalPetryka: What about adding it to NativeAOT, seems like it's [serialize](https://github.com/dotnet/runtime/blob/main/src/coreclr/tools/Common/JitInterface/ThunkGenerator/InstructionSetDesc.txt#L48) there.

PR #2056, opened 2022-07-28
Bugfix copied project settings

### LINKED ISSUE #2055: Comment after <LangVersion> breaks generated project
Adding an XML comment after `<LangVersion>` in your project results in the comment being copied verbatim into the generated project's comment, leading to comment-in-comment error:

```xml
<LangVersion>9.0</LangVersion><!-- Some comment -->
```

produces:

```xml
<!-- we set LangVersion after <LangVersion>9.0</LangVersion><!-- Some comment --> which might contain LangVersion copied from the benchmarks project -->
```

which is invalid due to the embedded `--` from `<!-- Some` inside the current comment.

mawosoft:
Fixes #2055


adamsitnik (approved):
Nice! Thank you very much for your contribution @mawosoft !

djee-ms:
Thanks!

PR #2067, opened 2022-08-02
fix .NET 4.8.1 detection

### LINKED ISSUE #2059: BenchmarkDotNet misreports .Net Framework as 4.8.1
AFAIK .Net Framework 4.8.1 was only released on ARM PCs and Windows 11 Insider program, yet with BDN `0.13.1.1835` from the nightlies my benchmark says:
```
Runtime=.NET Framework 4.8.1 (4.8.4526.0), X64 RyuJIT
```
on a Windows 8.1 x64 PC, which seems incorrect since like I said I think it's not available there yet.

adamsitnik:
fixes #2059


adamsitnik:
cc @MichalPetryka

adamsitnik:
I've tested it using Windows 11 insider build and it works as expected. Merging

PR #2073, opened 2022-08-17
add two fallbacks to CoreRun copying

adamsitnik:
@kunalspathak has tried to run BDN benchmarks with a config where CoreRun is stored in a readonly folder (by the Helix, a system used by the .NET Team to run tests on multiple machines. That has failed with an exception.

As solution we can do the following:
- when creating the destination folder fails, try to store it in TEMP
- when storing in TEMP fails as well, don't perform the copy at all

This should be quite bulletproof.


adamsitnik:
The failures are unrelated, merging

PR #2086, opened 2022-08-26
Tell the user about the asterisk

YegorStepanov:
Is this written somewhere? Looks pretty used

https://github.com/dotnet/BenchmarkDotNet/blob/f17453cef3a91ac90540dec1dcdee21725def6f8/src/BenchmarkDotNet/Running/UserInteraction.cs#L97-L103

FYI: 
It's not `string.Contains` but `IEnumerable.Contains`, so the input "a*b" or "**" doesn't work (as excepted).


AndreyAkinshin:
@YegorStepanov thanks!

PR #2087, opened 2022-08-26
Nit: fix broken link

YegorStepanov:
[Docs](https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html?tabs=tabid-1%2Ctabid-a#:~:text=Do%20not%20use%20the%20%40uid%20link%20in%20brackets%20(like%20this%3A%20(%40uid)).%20DocFX%20cannot%20parse%20this%20link.%20The%20%40uid%20link%20should%20be%20separated%20with%20white%20spaces.%20If%20you%20need%20to%20add%20a%20link%20in%20brackets%2C%20use%20%5B%5D(xref%3Auid).):
> Do not use the @uid link in brackets (like this: (@uid)). DocFX cannot parse this link. The @uid link should be separated with white spaces. If you need to add a link in brackets, use [](xref:uid).

---

Master:
image

PR:
image

### DIFF
--- a/docs/articles/guides/how-to-run.md
+++ b/docs/articles/guides/how-to-run.md
@@ -28,7 +28,7 @@ static void Main(string[] args) => BenchmarkSwitcher.FromAssembly(typeof(Program
 Also you can use the config command style to specify some config from command line (more [](xref:docs.console-args)):
 
 ```log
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
@YegorStepanov thanks!

Issue #784, opened 2018-06-07
Using a benchmark target's return value in a column


Warpten:
I'm writing a deserialization library that is basically taking a file stream and producing a `List<T>`:

```cs
        [Benchmark(Description = "Achievement (WDC1)")]
        public StorageList<WDC1.AchievementEntry> WDC1()
        {
            using (var fs = OpenFile("Achievement.WDC1.db2"))
                return new StorageList<WDC1.AchievementEntry>(fs);
        }
```

And it goes on for a couple versions.

However, the execution times don't make much sense since they are a measurement of the total amount of entries per file, which changes per version (I have no control over the data). It would make a lot more sense to display the average time needed to load one single element of each version.

I'd love to be able to have an implementation of `IColumn` that is able to obtain the return value from a benchmark target and work on it.

I have a local project that is basically a dumbed down performance meter but it has its limitations, namely in regards to results output - producing histograms is a chore.

Is there a way to do that, or am I out of luck?

adamsitnik:
hi @Warpten 

I am aware of this limitation. The problem is that we run the benchmark in a separate process, so passing any data from one to another is not trivial.

I know that @krzysztofcwalina has faced this issue in ML.NET

@Warpten @krzysztofcwalina would it be enough if I would add a mechanism to return a string and print it in a dedicated column? Sth like:

```cs
[ExtraData]
public string Size() => Serializer(sth).Size.ToString();
````

The method would be executed just once.

**Comment by Warpten** [comment]
Maybe exposing a method's result as an `object` would be less limiting? It would be great to have an extra column where I can just do

```cs
return (benchmark.Target.Result as IList).Count
```

Or even more complex things like

```cs 
// Suppose this is a variation of MeanColumn
return /* mean time here */ / (benchmark.Target.Result as IList).Count;
```

adamsitnik:
> Maybe exposing a method's result as an object would be less limiting? 

but how should I then serialize it and pass from one process to another without introducing any dependencies to BenchmarkDotNet?

gulbanana:
i think the string is fine, we could implement serialisation on top of that ourselves for advanced cases

Warpten:
Coming back at this, I didn't know at the time that BenchmarkDotNet was spawning subprocesses. So I guess strings are the best way out.

Konard:
Well, something like this is possible even now. Look at `Config` class, `SQLiteOutput` and `DoubletsOutput` methods.
``` C#
[...]

Issue #190, opened 2016-05-24
Feature request: quiet mode


jskeet:
It would be really nice to be able to start a long run, and instead of getting a long scrolling list where most of the time I can't see what's executing, instead have an in-place display, updated appropriately and consisting of:

- Possibly the options involved (may be too long-winded)
- Current / total test count, e.g. "Running test 32/152"
- Name of currently running test
- Current execution time and estimated total time (or remaining time)

I don't know enough about the execution timing to know whether estimating the total remaining time is viable, but I *suspect* it is... we have a target time for each test, right?

jskeet:
I realize that a multi-line updating status box may be tricky - a weaker form of this feature request would be a mode which just says which test it's executing each time it starts a new one, along with the count, e.g. output of

```text
[1/250] FooBenchmark
[2/250] BarBenchmark
[...]
[250/250] BazBenchmark
Tests complete. Total time: ...
Output in ...
```

Also, "quiet" may not be an ideal option name for this, as that would probably suggest even quieter output...

ig-sinicyn:
I've recently done something like this. There's HostLogger class that filters the output and prints only error messages and messages that starts with `// !` or `// ?`.
Also, all output between `// !<---` and `// !-->` lines is not filtered at all.

[Code](https://github.com/rsdn/CodeJam/blob/bf022133c55468f2ecb145c7a7ca7d86c3d83ff3/PerfTests/src/%5BL3_CompetitionsCore%5D/Loggers/HostLogger.cs)
[Tests](https://github.com/rsdn/CodeJam/blob/bf022133c55468f2ecb145c7a7ca7d86c3d83ff3/PerfTests/tests/%5BL3_CompetitionsCore%5D/HostLoggerTests.cs)

This approach works well for us, for example, output for the test simplifies up to:
```

Test Name:	RunSensitivityPerfTests
Test Outcome:	Skipped
Result Message:	
Test completed with warnings, details below.
Warnings:
    * Run #1: Job X64_Jit-RyuJit_Warmup3_Target10_Process1_IterationTime10, property Platform: The current process is not run as x64.
    * Run #1: Job X64_Jit-RyuJit_Warmup3_Target10_Process1_IterationTime10, property Jit: The current setup does not support RyuJit.
Diagnostic messages:
    * Run #1: Analyser CompetitionLimitsAnnotateAnalyser: no warnings.
Result StandardOutput:	
BenchmarkDotNet=v0.9.6.0
[...]

Issue #1661, opened 2021-03-05
Stop gracefully when benchmark fails with an exception that can not be caught by a catch block


adamsitnik:
The original issue was reported in https://github.com/dotnet/performance/issues/1701:

```log
// Benchmark Process Environment Information:
// Runtime=.NET 6.0.0 (6.0.21.11801), X64 RyuJIT
// GC=Concurrent Workstation
// Job: Job-AQTFSE(PowerPlanMode=00000000-0000-0000-0000-000000000000, IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, WarmupCount=1)

OverheadJitting  1: 1 op, 537000.00 ns, 537.0000 us/op
WorkloadJitting  1: 1 op, 6767100.00 ns, 6.7671 ms/op

OverheadJitting  2: 16 op, 731100.00 ns, 45.6938 us/op
WorkloadJitting  2: 16 op, 2527400.00 ns, 157.9625 us/op

WorkloadPilot    1: 16 op, 7800.00 ns, 487.5000 ns/op
// Benchmark Process 3336 has exited with code -1073741819
Unhandled exception. System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.ThrowHelper.ThrowNoMatchException()
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Execute(ILogger logger, BenchmarkCase benchmarkCase, BenchmarkId benchmarkId, IToolchain toolchain, BuildResult buildResult, IResolver resolver)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.RunCore(BenchmarkCase benchmarkCase, BenchmarkId benchmarkId, ILogger logger, IResolver resolver, BuildResult buildResult)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Run(BenchmarkRunInfo benchmarkRunInfo, Dictionary`2 buildResults, IResolver resolver, ILogger logger, List`1 artifactsToCleanup, String resultsFolderPath, String logFilePath, StartedClock& runChronometer)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Run(BenchmarkRunInfo[] benchmarkRunInfos)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.RunWithDirtyAssemblyResolveHelper(String[] args, IConfig config)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.Run(String[] args, IConfig config)
[...]

PR #2092, opened 2022-08-28
Use Pipes for host and benchmark process communication

### LINKED ISSUE #2070: BenchmarkDotNet crashing on Linux with DisassemblyDiagnoser
Opening as per https://github.com/dotnet/BenchmarkDotNet/issues/2000#issuecomment-1203883513.

My configuration is:
Kubuntu 20.04
and the provided commands return:
```
xterm-256color


pl_PL.UTF-8
```
(the middle ones don't return anything)

adamsitnik:
I definitely need to do more testing (Unix, Mono, WASM), but to tell the long story short redirecting STDIN & OUT and using it for process communication has historically proven to be unreliable. We can just use Anonymous Pipes and avoid issues like #2070.

Moreover, by using Pipes we could allow for:
- quiet mode: https://github.com/dotnet/BenchmarkDotNet/issues/190 (the processes can talk, but just don't print stuff to output)
- serializing custom types and doing things like https://github.com/dotnet/BenchmarkDotNet/issues/784 (this would require exposing new API, but that is fine)

Until I am done with everything this PR is a draft

fixes #2070


adamsitnik:
The tests are failing because they are relying on writing some custom messages to standard ouput. Since it's not redirected anymore and we don't parse it, they are failing. I'll just rewrite them to write to specific file instead.

adamsitnik:
@EgorBo and @stephentoub reported offline an issue where having the std out redirected and BDN reading from it can cause hangs when `DOTNET_JitDisasmSummary` is enabled.

Repro:

```cs
[2:19 PM] Egor Bogatov
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using System;
[DisassemblyDiagnoser]
[HideColumns("Error", "StdDev", "Median", "RatioSD")]
public partial class Program
{
    static void Main(string[] args) => BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);
    private static readonly int s_year = DateTime.UtcNow.Year;
    [Benchmark]
    public int Count()
    {
        int count = 0;
        for (int i = 0; i < 1_000_000; i++)
        {
            if (s_year == 2021)
            {
                count++;
            }
        }
        return count;
    }
}
```

```cmd
$env:DOTNET_JitDisasmSummary=1
dotnet run -c Release -f net7.0 --filter *
```

Once I get back to working on this PR I am going to ensure that this scenario works as expected.

adamsitnik:
I was able to confirm that this PR solves the problem reported by @stephentoub and @EgorBo

Previously it was hanging:

image

Now it works fine:

image

adamsitnik:
I've tested locally .NET Framework, .NET (Core), Mono (the old one) and WASM (this was not easy due to missing docs). Everything works fine, merging.

YegorStepanov:
The  deadlock on an unhandled exception occurs [here](https://github.com/dotnet/BenchmarkDotNet/pull/2092/files#diff-2a7da1208e90b7fc444cb8e20e424674d94b66aaac41ab6c5b35449f0b79d4fbR43):
```C#
[Benchmark]
public unsafe void Benchmark_UnhandledException()
{
    *(byte*)65536 = 100;
}
```
---

Slightly relevant here:

To proper detect the unhandled exception (#1661) and display its message we should set `process.RedirectStandardError=true` and read from **StandardError**, but it will another deadlock (#1631)

PR #2095, opened 2022-09-02
[NativeAOT] use PublishAot and don't reference ILCompiler in explicit way

adamsitnik:
After recent SDK changes the NativeAOT support got broken again. This time the failure was silent: since we were not setting `PublishAot` to true, the published app was regular .NET app, not NativeAOT.

This PR sets  `PublishAot` to `true` and removes the explicit reference to `ILCompiler` if it's specific version was not provided.

I've tested:
- support for the old CoreRT bits (we are going to remove it once .NET 7 is released and BDN repo can move to .NET 7 SDK)
- support for the latests NativeAOT bits
- support for benchmarking specific NativeAOT version

@AndreyAkinshin we might need to release a new BDN version soon (preferably with #2092 merged as well)

cc @MichalStrehovsky


MichalStrehovsky (approved):
Sorry for the shuffles.

PR #2096, opened 2022-09-07
ExecuteResult: Surface a Data property so the full output can be acce…

radical:
…ssed.

This would help with exporters that need to access the process output, which might have error details.

cc @adamsitnik


adamsitnik:
@radical I am currently working on switching the process communication from standard in/out to anonymous pipes: https://github.com/dotnet/BenchmarkDotNet/pull/2092 and I was actually hoping that I'll be able to remove the output parsing/reading. I assume that here you would like to read standard output as it might contain error details? How important is this scenario?

radical:
Sometimes the benchmark runs crash with exceptions, or js errors etc, so we definitely need that output.

adamsitnik (approved):
LGTM, but I may rename it in #2092 to sth like `ParsedStandardOutput`

PR #2101, opened 2022-09-11
Added a command line arg for not inducing any GCs while running Benchmarks

mrsharm:
Added a new commandline argument: ``--noForcedGCs`` that doesn't invoke any induced GCs. This is an important consideration for GC based microbenchmarks where we want to not force any GCs so that we can compare multiple runs in an equitable way. 

PS: I am not sure where to have added documentation for this.

### DIFF
--- a/tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
+++ b/tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
@@ -569,7 +569,7 @@ namespace BenchmarkDotNet.Tests
 
             foreach (var job in parsedConfiguration.config.GetJobs())
             {
-                Assert.True(job.Environment.Gc.Force == false);
+                Assert.False(job.Environment.Gc.Force);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
```diff
@@ -560,5 +560,17 @@ public void InvalidEnvVarAreRecognized()
...
+
+        [Fact]
+        public void UserCanSpecifyNoForceGCs()
+        {
+            var parsedConfiguration = ConfigParser.Parse(new[] { "--noForcedGCs" }, new OutputLogger(Output));
+            Assert.True(parsedConfiguration.isSuccess);
+
+            foreach (var job in parsedConfiguration.config.GetJobs())
+            {
+                Assert.True(job.Environment.Gc.Force == false);
```
adamsitnik, line 572: nit: you could simplify that

```suggestion
                Assert.False(job.Environment.Gc.Force);
```

adamsitnik (approved):
LGTM, thank you @mrsharm !

adamsitnik:
The CI failure is unrelated, merging

PR #2106, opened 2022-09-18
If a metric value could not be calculated, display `?` instead of `-`.

### LINKED ISSUE #1783: MemoryDiagnoser table output invalid symbol
When a benchmark fails or is ran on a runtime that doesn't support memory measurements (mono), the table displays `-` which is the same as it displays for actual 0 measurements. Can that be changed to display `?` to easily discern at a glance that it's not actually 0?

timcassell:
Resolves #1783.

Master:

|      Method |  Runtime |       Mean |   Gen0 | Allocated |
|------------ |--------- |-----------:|-------:|----------:|
| AllocObject | .NET 6.0 |  8.6188 ns | 0.0076 |      24 B |
| AllocObject |     Mono | 12.8588 ns | 0.0038 |         - |
|             |          |            |        |           |
|   ZeroAlloc | .NET 6.0 |  0.0059 ns |      - |         - |
|   ZeroAlloc |     Mono |  0.0878 ns |      - |         - |
|             |          |            |        |           |
|      Throws | .NET 6.0 |         NA |      - |         - |
|      Throws |     Mono |         NA |      - |         - |

This PR:

|      Method |  Runtime |       Mean |   Gen0 | Allocated |
|------------ |--------- |-----------:|-------:|----------:|
| AllocObject | .NET 6.0 |  8.4596 ns | 0.0076 |      24 B |
| AllocObject |     Mono | 13.4370 ns | 0.0038 |         ? |
|             |          |            |        |           |
|   ZeroAlloc | .NET 6.0 |  0.0061 ns |      - |         - |
|   ZeroAlloc |     Mono |  0.1116 ns |      - |         ? |
|             |          |            |        |           |
|      Throws | .NET 6.0 |         NA |     NA |        NA |
|      Throws |     Mono |         NA |     NA |        NA |

```cs
public class AllocBenchmark
{
    [Benchmark]
    public object AllocObject()
    {
        return new object();
    }

    [Benchmark]
    public void ZeroAlloc()
    {

    }

    [Benchmark]
    public void Throws()
    {
        throw new Exception();
    }
}
```


src/BenchmarkDotNet/Reports/Metric.cs
```diff
@@ -6,11 +6,11 @@ namespace BenchmarkDotNet.Reports
 {
     public class Metric
     {
-        public double Value { get; }
+        public double? Value { get; }
 
         public IMetricDescriptor Descriptor { get; }
 
-        public Metric(IMetricDescriptor descriptor, double value)
+        public Metric(IMetricDescriptor descriptor, double? value)
```
timcassell, line 13: This is a breaking change in the public API. I thought about using `NaN` instead of nullable, but I'm not sure if that is already a valid metric value.

timcassell: Or another option is to revert the `Metric` changes, and just don't add the metric if the value couldn't be calculated (instead of adding null), but that may omit the column entirely if none of the benchmarks can calculate it, and I don't think that's desirable (probably better to see a column of all `?` rather than no column at all).

timcassell: I opted for the latter option. I reverted `Metric.cs` to not break the public API, don't add the metric if the allocation could not be calculated, and force the allocated column to show if the memory diagnoser is attached (in case no allocations were measured for any benchmark).

timcassell:
@AndreyAkinshin Could you take a look?

src/BenchmarkDotNet/Columns/MetricColumn.cs
```diff
@@ -23,13 +23,28 @@ public class MetricColumn : IColumn
...
-        public bool IsAvailable(Summary summary) => summary.Reports.Any(report => report.Metrics.ContainsKey(descriptor.Id));
+        public bool IsAvailable(Summary summary) => descriptor is Diagnosers.AllocatedMemoryMetricDescriptor || summary.Reports.Any(IsAvailable);
+
+        private bool IsAvailable(BenchmarkReport report)
+        {
+            if (!report.Metrics.TryGetValue(descriptor.Id, out var metric))
+            {
+                return false;
+            }
+            if (metric.Descriptor is Diagnosers.MemoryDiagnoser.GarbageCollectionsMetricDescriptor)
```
AndreyAkinshin, line 34: I don't like these explicit dependencies on specific diagnosers in the base `MetricColumn` class. Please, introduce a corresponding property in `IMetricDescriptor` that controls this behavior.

src/BenchmarkDotNet/Columns/MetricColumn.cs
```diff
@@ -23,13 +23,28 @@ public class MetricColumn : IColumn
...
+            return true;
+        }
 
         public string GetValue(Summary summary, BenchmarkCase benchmarkCase) => GetValue(summary, benchmarkCase, SummaryStyle.Default);
 
         public string GetValue(Summary summary, BenchmarkCase benchmarkCase, SummaryStyle style)
         {
-            if (!summary.HasReport(benchmarkCase) || !summary[benchmarkCase].Metrics.TryGetValue(descriptor.Id, out Metric metric) || (metric.Value == 0.0 && !style.PrintZeroValuesInContent))
+            if (!summary.HasReport(benchmarkCase) || !summary[benchmarkCase].Metrics.TryGetValue(descriptor.Id, out Metric metric))
+                return "?";
```
AndreyAkinshin, line 46: Since `"?"` is reused in multiple places (e.g., e.g., `TryParse` from `GcStats`; `CsvMeasurementsExporter`), it will be better to extract it to a named constant so that we can easily change it in the future.

AndreyAkinshin (changes requested):

src/BenchmarkDotNet/Columns/MetricColumn.cs
timcassell: Yeah, I thought that would be cleaner, but I didn't want to break the interface since it's public (and netstandard2.0 does not support DIM). I refactored it to work in a more abstract way without changing the interfaces, but if you don't mind breaking changes, I can update the interfaces instead.

AndreyAkinshin: @timcassell the new approach with a hidden *static* state looks too hacky and may lead to unpleasant side effects. It's OK to extend the interface.

timcassell: Ok, I updated the interfaces. Since that's a breaking change, I think this should go into 0.14.0 instead of 0.13.4. What do you think?

AndreyAkinshin (changes requested):
@timcassell thanks for the update, now it looks much better. However, the current solution still looks overcomplicated for such a use case.

I have another proposal. What if we use `double.NaN` for an unknown `Metric.Value`? It will allow us to avoid introducing `IDiagnoser.GetForceShowColumns`, `IMetricDescriptor.GetIsAvailable`, and the second argument in `MetricColumn.ctor`: we should just implement handling such values in a few places. What do you think?

We already have too many tricky properties in public interfaces. Ultimately, I want to simplify the APIs and reduce the number of properties as much as possible.

timcassell:
> I have another proposal. What if we use `double.NaN` for an unknown `Metric.Value`?

@AndreyAkinshin  I also had that idea (mentioned in a resolved comment earlier), but I wasn't sure if it was already a valid value being used somewhere. Is it safe to use for this purpose?

Also, that only solves the `IMetricDescriptor.GetIsAvailable` case, it doesn't solve the `IDiagnoser.GetForceShowColumns` case, which I am doing to show the column if a metric value was never added at all (so `float.NaN` wouldn't exist). I also mentioned in that resolved conversation that we could just keep the current behavior where the column doesn't show if no measurements are made, but I think it's useful to see the column if a specific diagnoser is used. But maybe that doesn't really matter if no results were gathered anyway and the benchmarks need to be fixed.

AndreyAkinshin:
> I also mentioned in that resolved conversation that we could just keep the current behavior where the column doesn't show if no measurements are made, but I think it's useful to see the column if a specific diagnoser is used. But maybe that doesn't really matter if no results were gathered anyway and the benchmarks need to be fixed.

What if we keep the existing behavior (the column without non-`?` value is not shown) and add a validation warning like "The column <ColumnName> is not shown because ...". It should solve all of our problems:
* The result will be informative enough because we explain why the column is not shown (I think, it is even better than just printing the column full of `?`). Avoiding extra columns is (almost) always good because it increases the chances that the table is fit within the terminal width.
* We have to need to introduce advanced API or use hacks with double.NaN.

timcassell:
> * We have to need to introduce advanced API or use hacks with double.NaN.

AFAIK allocated is the only column that can have an unknown value with a real result (the benchmark isn't broken). My ideas were either don't add the metric (current behavior in this PR), change the `double` to `double?` in the Metric, use `double.NaN` or add a separate flag on the metric.

All other metrics will only be `?` of the benchmark failed.

AndreyAkinshin:
@timcassell I need some time to think. I will back to you tomorrow with updated thoughts on this problem.

timcassell:
> @timcassell I need some time to think. I will back to you tomorrow with updated thoughts on this problem.

Sure. Food for thought, this is my idea for the cleanest solution without breaking public APIs. Change `Metric` to this:

```cs
public class Metric
{
    private double? value;

    public double Value => value ?? double.NaN;

    public IMetricDescriptor Descriptor { get; }

    internal bool HasValue => value.HasValue;

    internal bool IsAvailable { get; }

    public Metric(IMetricDescriptor descriptor, double value) : this(descriptor, value, true) { }

    internal Metric(IMetricDescriptor descriptor, double? value, bool isAvailable)
    {
        Descriptor = descriptor;
        this.value = value;
        IsAvailable = isAvailable;
    }
}
```

And just leave columns as not showing if no benchmarks completed successfully (i.e. the current behavior).

timcassell:
@AndreyAkinshin I made those changes. It looks much cleaner to me and no breaking changes. What do you think?

timcassell:
@AndreyAkinshin Friendly reminder.

AndreyAkinshin:
@timcassell sorry for the late reply. This problem looks simple, but it is not easy to find the right solution. The new version looks good, but the existence of both `Metric.HasValue` and `Metric.IsAvailable` feels confusing. I played with different implementation options a little bit and came up with the following design idea:

We distinguish three kinds of situations:
* A metric is not reported (e.g., due to an error) -> display `?`
* A metric is reported
  * A valid value is reported -> display the reported value
  * An "invalid"/"non-applicable" state is reported (metric cannot be evaluated) -> display `NA`

 With such a design, we don't need additional fields in the `Metric` class (non-applicable state can be expressed with `double.NaN`).

timcassell:
> but the existence of both `Metric.HasValue` and `Metric.IsAvailable` feels confusing.
>

\<snip\>

> 
> We distinguish three kinds of situations:
> 
> * A metric is not reported (e.g., due to an error) -> display `?`
> * A metric is reported
>   
>   * A valid value is reported -> display the reported value
>   * An "invalid"/"non-applicable" state is reported (metric cannot be evaluated) -> display `NA`
> 
> With such a design, we don't need additional fields in the `Metric` class (non-applicable state can be expressed with `double.NaN`).

I think that's fine for output. But we still need `IsAvailable` to preserve the behavior of `gen0/1/2` columns that don't show if all rows are `0`. Otherwise, if we don't report them as `0` and "unavailable", it will show as `?` which is incorrect.

I used `HasValue` because I'm not 100% sure that `double.NaN` is not a valid value. But if you're okay with that, I can change it to use `double.NaN` instead.

AndreyAkinshin:
> But we still need IsAvailable to preserve the behavior of gen0/1/2 columns that don't show if all rows are 0

It looks like a display feature, not a reporting feature. The metric itself should not be "responsible" for the logic that defines how we build the summary table, it's a task for `IMetricDescriptor` and `SummaryStyle`.

> I'm not 100% sure that double.NaN is not a valid value.

I don't know any cases in which `double.NaN` is a valid value. And I doubt that we will have such cases in the future. At the moment, we can freely use it to represent the "non-applicable" state.

timcassell:
@AndreyAkinshin I made those changes, except I swapped `NA` and `?`. I think it makes more sense that way.

|      Method |  Runtime |       Mean |   Gen0 | Allocated |
|------------ |--------- |-----------:|-------:|----------:|
| AllocObject | .NET 6.0 |  8.4596 ns | 0.0076 |      24 B |
| AllocObject |     Mono | 13.4370 ns | 0.0038 |         ? |
|             |          |            |        |           |
|   ZeroAlloc | .NET 6.0 |  0.0061 ns |      - |         - |
|   ZeroAlloc |     Mono |  0.1116 ns |      - |         ? |
|             |          |            |        |           |
|      Throws | .NET 6.0 |         NA |     NA |        NA |
|      Throws |     Mono |         NA |     NA |        NA |

AndreyAkinshin:
> except I swapped NA and ?. I think it makes more sense that way.

timcassell:
@AndreyAkinshin @adamsitnik Can this be merged?

AndreyAkinshin:
@timcassell I guess we can merge it.
Could you please merge the latest master so that we can check the status of the CI builds?

AndreyAkinshin (approved):

AndreyAkinshin:
@timcassell thanks!

PR #1123, opened 2019-03-30
+ InProcessNoEmitToolchain (obsoletes the InProcessToolchain)

ig-sinicyn:
As suggested [here](https://github.com/dotnet/BenchmarkDotNet/pull/1093#issuecomment-470422460).

I've added a copy of `InProcessToolchain`, `InProcessNoEmitToolchain`, that has no emit-related code.
Origin classes marked obsolete, tests and usages updated.

Any suggestions are welcome, I'm not going to insist on this implementation.

PR #2108, opened 2022-09-20
Fix consume ValueTask backed by IValueTaskSource

### LINKED ISSUE #1595: ValueTask-returning benchmarks are not being called correctly

timcassell:
Fixes #1595.

This was originally #1941, but I split it out to just include the `ValueTask` fixes (and I added more tests).


timcassell:
I'm pretty sure the test failures are unrelated. Other PRs are failing on the same tests.

timcassell:
@adamsitnik Can you take a look?

timcassell:
@AndreyAkinshin `InProcessToolchain` has been deprecated in favor of `InProcessNoEmitToolchain` since #1123 (almost 4 years ago). I updated it to match the changes I made here and in my other followup PRs, but I wonder if it can just be removed? Or should I just not add the updates and leave it in its old state? It's not good to maintain both.

AndreyAkinshin:
I have no objection to removing `InProcessToolchain`. @adamsitnik what do you think?

timcassell:
@ig-sinicyn I touched the InProcess toolchains here, can you take a look?

ig-sinicyn (approved):

ig-sinicyn:
@timcassell, great job!

There should be a diff test that proves that emitted IL matches with compiled c# code. If it passes, all should be fine.

AndreyAkinshin (approved):
No concerns from my side. If @adamsitnik has no objections, we can merge it.

PR #2109, opened 2022-09-20
Fix async GlobalSetup/Cleanup with InProcessEmit toolchain

### LINKED ISSUE #1738: Async GlobalSetup does not work with InProcess toolchain

### LINKED ISSUE #2236: Benchmarks sporadically failing on Executable <GUID>Emitted.dll not found
hi
I am running multiple benchmarks with InProcessEmitToolchain running in net7.0 framework
and I'm sporadically (different benchmark every time) getting error:

// Runtime=.NET 7.0.1 (7.0.122.56804), X64 RyuJIT AVX2
// GC=Concurrent Workstation
// HardwareIntrinsics=AVX2,AES,BMI1,BMI2,FMA,LZCNT,PCLMUL,POPCNT,AvxVnni,SERIALIZE VectorSize=256
// Job: Job-RFHJAM(Platform=X64, Runtime=.NET 7.0, Toolchain=InProcessEmitToolchain, IterationCount=3, LaunchCount=1, WarmupCount=3)

OverheadJitting  1: 1 op, 373900.00 ns, 373.9000 us/op

System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.SqlServer.SqlThreatProtectionService.TdService.Benchmarks.DetectionStorageBenchmarks.ObfuscatedQueryStorageOuterStatementFirstBenchmark.GetObfuscatedQuery() in C:\dev\ads-tpscore\src\TdService\Benchmarks\DetectionStorageBenchmarks\ObfuscatedQueryStorageOuterStatementFirstBenchmark.cs:line 90
   at BenchmarkDotNet.Autogenerated.Runnable_4.WorkloadActionNoUnroll(Int64 invokeCount)
   at BenchmarkDotNet.Engines.Engine.RunIteration(IterationData data)
   at BenchmarkDotNet.Engines.EngineFactory.Jit(Engine engine, Int32 jitIndex, Int32 invokeCount, Int32 unrollFactor)
   at BenchmarkDotNet.Engines.EngineFactory.CreateReadyToRun(EngineParameters engineParameters)
   at BenchmarkDotNet.Autogenerated.Runnable_4.Run(BenchmarkCase benchmarkCase, IHost host)
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at BenchmarkDotNet.Toolchains.InProcess.Emit.Implementation.RunnableProgram.Run(BenchmarkId benchmarkId, Assembly partitionAssembly, BenchmarkCase benchmarkCase, IHost host)
Executable C:\dev\ads-tpscore\bin\release\TdService.Benchmarks\net7.0\win-x64\bbdca8b1-3e76-45d9-b910-fa15822c521bEmitted.dll not found

what can cause the "Emitted.dll" to not be present some of the times?
the same benchmark sometimes succeeds and sometimes fails on this error
any help would be appreciated

timcassell:
Fixes #1738 and fixes #2236.

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/InProcessEmitTest.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/InProcessEmitTest.cs
@@ -239,35 +239,125 @@ namespace BenchmarkDotNet.IntegrationTests
             }
         }
 
-        [Fact]
-        public void InProcessEmitToolchainSupportsIterationSetupAndCleanup()
+        [Theory]
+        [InlineData(typeof(IterationSetupCleanup))]
+        [InlineData(typeof(GlobalSetupCleanupTask))]
+        [InlineData(typeof(GlobalSetupCleanupValueTask))]
+        [InlineData(typeof(GlobalSetupCleanupValueTaskSource))]
+        public void InProcessEmitToolchainSupportsSetupAndCleanup(Type benchmarkType)
         {
             var logger = new OutputLogger(Output);
             var config = CreateInProcessConfig(logger);
 
-            WithIterationSetupAndCleanup.SetupCounter = 0;
-            WithIterationSetupAndCleanup.BenchmarkCounter = 0;
-            WithIterationSetupAndCleanup.CleanupCounter = 0;
+            Counters.SetupCounter = 0;
+            Counters.BenchmarkCounter = 0;
+            Counters.CleanupCounter = 0;
 
-            var summary = CanExecute<WithIterationSetupAndCleanup>(config);
+            var summary = CanExecute(benchmarkType, config);
 
-            Assert.Equal(1, WithIterationSetupAndCleanup.SetupCounter);
-            Assert.Equal(16, WithIterationSetupAndCleanup.BenchmarkCounter);
-            Assert.Equal(1, WithIterationSetupAndCleanup.CleanupCounter);
+            Assert.Equal(1, Counters.SetupCounter);
+            Assert.Equal(16, Counters.BenchmarkCounter);
+            Assert.Equal(1, Counters.CleanupCounter);
         }
 
-        public class WithIterationSetupAndCleanup
+        private static class Counters
         {
             public static int SetupCounter, BenchmarkCounter, CleanupCounter;
+        }
 
+        public class IterationSetupCleanup
+        {
             [IterationSetup]
-            public void Setup() => Interlocked.Increment(ref SetupCounter);
+            public void Setup() => Interlocked.Increment(ref Counters.SetupCounter);
 
             [Benchmark]
-            public void Benchmark() => Interlocked.Increment(ref BenchmarkCounter);
+            public void Benchmark() => Interlocked.Increment(ref Counters.BenchmarkCounter);
 
             [IterationCleanup]
-            public void Cleanup() => Interlocked.Increment(ref CleanupCounter);
+            public void Cleanup() => Interlocked.Increment(ref Counters.CleanupCounter);
+        }
+
+        public class GlobalSetupCleanupTask
+        {
+            [GlobalSetup]
+            public static async Task GlobalSetup()
+            {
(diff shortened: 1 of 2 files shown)


tests/BenchmarkDotNet.IntegrationTests/InProcessEmitTest.cs
```diff
@@ -239,35 +239,125 @@ public static ValueTask<decimal> InvokeOnceStaticValueTaskOfT()
...
         }
 
-        [Fact]
-        public void InProcessEmitToolchainSupportsIterationSetupAndCleanup()
+        [Theory]
+        [InlineData(typeof(IterationSetupCleanup))]
+        [InlineData(typeof(GlobalSetupCleanupTask))]
+        [InlineData(typeof(GlobalSetupCleanupValueTask))]
+        [InlineData(typeof(GlobalSetupCleanupValueTaskSource))]
+        public void InProcessEmitToolchainSupportsSetupAndCleanup(Type benchmarkType)
```
adamsitnik, line 247: So far this test was a `[Fact]` (having a single test case), so we did not have to worry about multiple tests accessing the same static field at the same time.

As long as we keep the test parallelization disabled here:
https://github.com/dotnet/BenchmarkDotNet/blob/c7ed714a34a4de04f7320b6d971a51b04545bb1c/tests/BenchmarkDotNet.IntegrationTests/xunit.runner.json#L6-L7

It's OK. But if we ever change our policy on that, multiple tests may update `Counters` fields at the same time and make this test flaky. 

There is no need to change anything as of now,  just something to keep in mind when writing more tests in the future.

adamsitnik (approved):
LGTM, thank you for providing the fix @timcassell and apologies for not noticing the PR (due to the fact that I work on the .NET Team I get few dozens of GH notifications a day).

tests/BenchmarkDotNet.IntegrationTests/InProcessEmitTest.cs
timcassell: According to the [documentation](https://xunit.net/docs/running-tests-in-parallel.html),

> By default, each test class is a unique test collection. Tests within the same test class will not run in parallel against each other.

So we should be good, regardless of that setting.

Issue #2099, opened 2022-09-10
WASM is recognized as NativeAOT


adamsitnik:
When running WASM benchmarks the benchmark process is incorrectly recognized as NativeAOT:

![image](https://user-images.githubusercontent.com/6011991/189481579-9a38a67a-f815-4070-81d0-b4486c474b8c.png)

Most likely the following code is true also for WASM:

https://github.com/dotnet/BenchmarkDotNet/blob/32ddeb56bd38631e052350cecfbf9e86df1b2e7c/src/BenchmarkDotNet/Portability/RuntimeInformation.cs#L44-L46

cc @radical @naricc

PR #2112, opened 2022-09-20
Disambiguate NativeAOT, and Wasm identification in BenchmarkDotNet.Po…

radical:
…rtability.RuntimeInformation

Fixes https://github.com/dotnet/BenchmarkDotNet/issues/2099 .

cc @adamsitnik @lewing

### DIFF
--- a/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
@@ -43,7 +43,8 @@ namespace BenchmarkDotNet.Portability
 
         public static bool IsNativeAOT
             => Environment.Version.Major >= 5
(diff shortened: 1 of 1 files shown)


lewing (commented):
I'm curious what case NativeAOT is trying to distinguish here?

adamsitnik (approved):
LGTM, thank you @radical !

lewing:
I think this has broken mono aot reporting

adamsitnik:
> there doesn't seem to be a wasm report with a label indicating it was aot'd since 9/1

@lewing  The performance repo has not updated BDN since August, so the reason for that is most likely something different.

https://github.com/dotnet/performance/blob/06fee6199750607af3484816a0456d757a4cead9/src/harness/BenchmarkDotNet.Extensions/BenchmarkDotNet.Extensions.csproj#L10-L11

PR #2114, opened 2022-09-21
Rework ReturnValueValidator and ExecutionValidator

### LINKED ISSUE #848: ParamsSource attribute doesn't work as expected
The properties marked with the ParamsSource attribute aren't set in the ExecutionValidatorBase.
This causes issues when this validator calls the GlobalSetup. 
The property marked as this does work as expected (gets inited to the first listed value before calling the GlobalSetup method):
```
        [Params(
            typeof(BasicTestSet),
            typeof(UnsafeTestSet),
            typeof(SlowDeferredLinqTestSet),
            typeof(FastLinqTestSet),
            typeof(UnsafeLinqTestSet),
            typeof(UnsafeParallelTestSet)
        )]
        public Type FilterType { get; set; }

```
The property marked as this doesn't work as expected - it remains null upon the GlobalSetup method call:
```
        [ParamsSource(nameof(GetTestTypes))]
        public Type FilterType { get; set; }
        public static IEnumerable<Type> GetTestTypes() => new[] {typeof(BasicTestSet),
            typeof(UnsafeTestSet),
            typeof(SlowDeferredLinqTestSet),
            typeof(FastLinqTestSet),
            typeof(UnsafeLinqTestSet),
            typeof(UnsafeParallelTestSet)};
```

The root cause is in the [ExecutionValidatorBase.TryToSetParamsProperties](https://github.com/dotnet/BenchmarkDotNet/blob/559773646b56a0dbf9887a90ffad373a904d8bc5/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs#L160) code. It just scans the properties for the Params attribute, ignoring the ParamsSource attribute:
```
var paramProperties = benchmarkTypeInstance
                .GetType()
                .GetAllProperties()
                .Where(propertyInfo => propertyInfo.GetCustomAttributes(false).OfType<ParamsAttribute>().Any())
.ToArray();
```
Instead it should rely on a common init code, like ReturnValueValidator does: 
` InProcessRunner.FillMembers(benchmarkTypeInstance, benchmark);`

### LINKED ISSUE #2083: `ReturnValueValidator` does not run `GlobalSetup` methods for each `Params`
When running this benchmark:
```cs
using System;
using System.Linq;
using Ardalis.Extensions.Enumerable;
using System.Collections.Generic;

namespace Ardalis.Extensions.Benchmarks.Enumerable;

[MemoryDiagnoser]
[ReturnValueValidator(failOnError: true)]
public class ProductBenchmarks
{
  [Params(4, 5, 6, 7)]
  public int N { get; set; }

  private int[] _numbers;

  [GlobalSetup]
  public void GlobalSetup()
  {
    _numbers = System.Linq.Enumerable.Range(1, N).ToArray();
    Console.WriteLine($"N = {N} and _numbers has a length of {_numbers.Length}");
  }


  [Benchmark(Baseline = true)]
  public long ProductAggregate()
  {
    return _numbers.ProductAggregate();
  }

  [Benchmark]
  public int ProductForEach()
  {
    return _numbers.ProductForEach();
  }

  [Benchmark]
  public int Product()
  {
    return _numbers.Product();
  }
}

static class ProductBenchmarksExtensions
{
  public static int ProductAggregate(this IEnumerable<int> numbers)
  {
    return numbers.Aggregate(1, (acc, x) => acc * x);
  }

  public static int ProductForEach(this IEnumerable<int> numbers)
  {
    int product = 2; // <- Set to 2 so that we have different return values and make validation fail
    foreach (var x in numbers)
    {
      product *= x;
    }
    return product;
  }
}
```

I get the following output:
```terminal
// Validating benchmarks:
N = 4 and numbers has a length of 4
Inconsistent benchmark return values in ProductBenchmarks: ProductAggregate: 24, ProductForEach: 48, Product: 24 [N=4]
Inconsistent benchmark return values in ProductBenchmarks: ProductAggregate: 24, ProductForEach: 48, Product: 24 [N=5]
Inconsistent benchmark return values in ProductBenchmarks: ProductAggregate: 24, ProductForEach: 48, Product: 24 [N=6]
Inconsistent benchmark return values in ProductBenchmarks: ProductAggregate: 24, ProductForEach: 48, Product: 24 [N=7]
```

The methods being benchmarked are returning the factorial of N. As you can see from the return values being identical, the `GlobalSetup` method that updates the List of integers to multiply doesn't seem to be ran before validating the return values for the subsequent `N`s.

I can confirm that the `GlobalSetup` method runs before each benchmark for each N when there is not failing validation and we proceed with benchmarking.

Feel free to clone this repo, here is the exact commit to reproduce the above state: https://github.com/snowfrogdev/Ardalis.Extensions/tree/b17763f52b890e25098557095c1ec954168cafe4

YegorStepanov:
- Add `ParamsSource` support
- Add `Arguments`/`ArgumentsSource` support
- Fill `Params`/`ParamsSource` before `GlobalSetup` (fixes #2083 fixes #848)
- Await async benchmarks (previously, only `GlobalSetup`/`GlobalCleanup` were awaited)
- Call `IterationSetup`/`IterationCleanup`
- Show error for the async iteration methods

Each benchmark is run on the new instance.

**ReturnValueValidator**:
- Compare `Task` value instead task objects itself

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/InProcessNoEmitRunner.cs
+++ b/src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/InProcessNoEmitRunner.cs
@@ -4,6 +4,7 @@ using BenchmarkDotNet.Engines;
 using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Exporters;
 using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Parameters;
 using BenchmarkDotNet.Running;
 
 using JetBrains.Annotations;
@@ -61,39 +62,43 @@ namespace BenchmarkDotNet.Toolchains.InProcess.NoEmit
             }
         }
 
-        /// <summary>Fills the properties of the instance of the object used to run the benchmark.</summary>
-        /// <param name="instance">The instance.</param>
-        /// <param name="benchmarkCase">The benchmark.</param>
+        /// <summary>Fills the properties of the instance of the object used to run the benchmark. Arguments are not supported</summary>
         internal static void FillMembers(object instance, BenchmarkCase benchmarkCase)
         {
             foreach (var parameter in benchmarkCase.Parameters.Items)
             {
-                var flags = BindingFlags.Public;
-                flags |= parameter.IsStatic ? BindingFlags.Static : BindingFlags.Instance;
+                FillMember(instance, benchmarkCase, parameter);
+            }
+        }
 
-                var targetType = benchmarkCase.Descriptor.Type;
-                var paramProperty = targetType.GetProperty(parameter.Name, flags);
+        /// <summary>Fills the property of the instance of the object used to run the benchmark. Arguments are not supported</summary>
+        internal static void FillMember(object instance, BenchmarkCase benchmarkCase, ParameterInstance parameter)
+        {
+            var flags = BindingFlags.Public;
+            flags |= parameter.IsStatic ? BindingFlags.Static : BindingFlags.Instance;
 
-                if (paramProperty == null)
-                {
-                    var paramField = targetType.GetField(parameter.Name, flags);
-                    if (paramField == null)
-                        throw new InvalidOperationException(
-                            $"Type {targetType.FullName}: no property or field {parameter.Name} found.");
+            var targetType = benchmarkCase.Descriptor.Type;
+            var paramProperty = targetType.GetProperty(parameter.Name, flags);
 
-                    var callInstance = paramField.IsStatic ? null : instance;
-                    paramField.SetValue(callInstance, parameter.Value);
-                }
-                else
-                {
-                    var setter = paramProperty.GetSetMethod();
-                    if (setter == null)
-                        throw new InvalidOperationException(
-                            $"Type {targetType.FullName}: no settable property {parameter.Name} found.");
+            if (paramProperty == null)
+            {
+                var paramField = targetType.GetField(parameter.Name, flags);
+                if (paramField == null)
+                    throw new InvalidOperationException(
+                        $"Type {targetType.FullName}: no property or field {parameter.Name} found.");
 
-                    var callInstance = setter.IsStatic ? null : instance;
-                    setter.Invoke(callInstance, new[] { parameter.Value });
-                }
+                var callInstance = paramField.IsStatic ? null : instance;
+                paramField.SetValue(callInstance, parameter.Value);
+            }
+            else
+            {
+                var setter = paramProperty.GetSetMethod();
+                if (setter == null)
+                    throw new InvalidOperationException(
+                        $"Type {targetType.FullName}: no settable property {parameter.Name} found.");
+
+                var callInstance = setter.IsStatic ? null : instance;
+                setter.Invoke(callInstance, new[] { parameter.Value });
             }
         }
 
--- a/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
+++ b/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
@@ -121,9 +121,9 @@ namespace BenchmarkDotNet.Validators
             return true;
         }
 
-        private string GetAttributeName(Type type) => type.Name.Replace("Attribute", string.Empty);
+        private static string GetAttributeName(Type type) => type.Name.Replace("Attribute", string.Empty);
 
-        private void TryToGetTaskResult(object result)
+        private static void TryToGetTaskResult(object result)
         {
             if (result == null)
             {
--- a/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
+++ b/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Extensions;
 using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.InProcess.NoEmit;
 
 namespace BenchmarkDotNet.Validators
 {
@@ -24,29 +25,34 @@ namespace BenchmarkDotNet.Validators
 
             foreach (var typeGroup in validationParameters.Benchmarks.GroupBy(benchmark => benchmark.Descriptor.Type))
             {
-                if (!TryCreateBenchmarkTypeInstance(typeGroup.Key, errors, out var benchmarkTypeInstance))
+                var executors = new List<BenchmarkExecutor>();
+
+                foreach (var benchmark in typeGroup)
                 {
-                    continue;
+                    if (!TryCreateBenchmarkTypeInstance(typeGroup.Key, errors, out var benchmarkTypeInstance))
+                        continue;
+
+                    if (!TryToFillParameters(benchmark, benchmarkTypeInstance, errors))
+                        continue;
+
+                    if (!TryToCallGlobalSetup(benchmarkTypeInstance, errors))
+                        continue;
+
+                    if (!TryToCallIterationSetup(benchmarkTypeInstance, errors))
+                        continue;
+
+                    executors.Add(new BenchmarkExecutor(benchmarkTypeInstance, benchmark));
                 }
 
-                if (!TryToSetParamsFields(benchmarkTypeInstance, errors))
+                ExecuteBenchmarks(executors, errors);
+
+                foreach (var executor in executors)
                 {
-                    continue;
+                    if (!TryToCallIterationCleanup(executor.Instance, errors))
+                        continue;
+
+                    TryToCallGlobalCleanup(executor.Instance, errors);
                 }
-
-                if (!TryToSetParamsProperties(benchmarkTypeInstance, errors))
-                {
-                    continue;
-                }
-
-                if (!TryToCallGlobalSetup(benchmarkTypeInstance, errors))
-                {
-                    continue;
-                }
-
-                ExecuteBenchmarks(benchmarkTypeInstance, typeGroup, errors);
-
-                TryToCallGlobalCleanup(benchmarkTypeInstance, errors);
             }
 
             return errors;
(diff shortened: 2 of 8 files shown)


YegorStepanov:
Deadlock in tests. I cannot reproduce it on local machine. I ran them few times, the deadlock happens every time.

**Github Actions**: deadlock on **Windows**/**Linux**. **MacOS** is ok
**Azure Pipelines**: deadlock on **Linux**. **Window**/**MacOS** is ok
**AppVeyor**: deadlock


I need to execute a method via Reflection, await it (if the method is async: **Task** or **ValueTask**) and return result (if the **Task** is not void).

```C#
MethodInfo benchmarkMethod = ...
var result = benchmarkMethod.Invoke(benchmarkClassInstance, arguments); // (arguments.Any() ? arguments : null) not helped

if (TryAwaitTask(result, out var taskResult))
    result = taskResult;

// returns true if the method is async
private static bool TryAwaitTask(object task, out object result)
{
    result = null;

    if (task is null)
    {
        return false;
    }
    
    // ValueTask<T>
    var returnType = task.GetType();
    if (returnType.IsGenericType && returnType.GetGenericTypeDefinition() == typeof(ValueTask<>))
    {
        var asTaskMethod = task.GetType().GetMethod("AsTask");
        task = asTaskMethod.Invoke(task, null);
    }

    if (task is ValueTask valueTask)
        task = valueTask.AsTask();
    
    // Task or Task<T>
    if (task is Task t)
    {
        if (TryGetTaskResult(t, out var taskResult))
            result = taskResult;

        return true;
    }

    return false;
}

// https://stackoverflow.com/a/52500763
private static bool TryGetTaskResult(Task task, out object result)
{
    result = null;

    var voidTaskType = typeof(Task<>).MakeGenericType(Type.GetType("System.Threading.Tasks.VoidTaskResult"));
    if (voidTaskType.IsInstanceOfType(task))
    {
        return false;
    }

    var property = task.GetType().GetProperty("Result", BindingFlags.Public | BindingFlags.Instance);
    if (property is null)
    {
        return false;
    }

    result = property.GetValue(task);
    return true;
}
```


The deadlock occurs on executing these async methods in parallel
```C#
public class AsyncGenericValueTaskGlobalSetup
{
    [GlobalSetup]
    public async ValueTask<int> GlobalSetup()
    {
        await Task.Delay(1);
        return 42;
    }
}

public class AsyncSetupIsSupportedClass
{
    [Benchmark]
    public async Task<int> Foo()
    {
        await Task.Delay(1);
        return 1;
    }
}
```

No one else uses the task. 

`valueTask.AsTask()` creates a new normal task? `valueTask.AsTask().Result` is a correct way to sync wait? 

@stephentoub, only You know what's going on here. Maybe it's a BLC bug?

timcassell:
I don't see anything wrong with that code. But I just updated my `AwaitHelper` in #2108 to work for multiple threads, and I wonder if you tried the same thing if it would make a difference.

YegorStepanov:
Hey, Tim! I tried your solution, deadlock still exists.

Changing `ValueTask` to `Task` doesn't help.
I tried to `lock` invocation, but it didn't help. Although a new object is created on each `ExecutionValidatorBase.Validate` call. And each test method is isolated.

The deadlock occurs on execution `ExecutionValidatorTests` and `ReturnValueValidatorTests` in parallel. If I remove all async methods from any of these files - deadlock is done.

I deleted all other tests except these and all sync methods in these tests - the situation is the same.

I think the problem is due to the version of .NET Framework (net461) or with XUnit.

YegorStepanov:
What I did...

### Summary:
- Change `ValueTask` to `Task`
- Add `lock` to both validators and to their base class
- Bumped .Net Framework from `4.61`/`4.62` to `4.8` in solution
- Bumped .Net Core from `netcoreapp2.1`/`net6.0` to `6.0` in solution
- Updated xUnit packages
- Updated Microsoft.NET.Test.Sdk, Microsoft.NETCore.Platforms, Microsoft.NETFramework.ReferenceAssemblies
- Updated VM Ubuntu-20.04 to Ubuntu-latest
- Updated VM Windows-20.04 to Windows-latest

Still deadlock on Windows/Linux

What works:
- Remove all "async" tests from any of these classes
- Add `[Collection("Disable parallelism")]` to both test classes to disable parallelism.

I hope it's a xUnit issue. I will send it to them.

src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
```diff
@@ -24,29 +25,34 @@ public IEnumerable<ValidationError> Validate(ValidationParameters validationPara
...
-                if (!TryToCallGlobalSetup(benchmarkTypeInstance, errors))
-                {
-                    continue;
+                    if (!TryToCallGlobalSetup(benchmarkTypeInstance, errors))
+                        continue;
+
+                    if (!TryToCallIterationSetup(benchmarkTypeInstance, errors))
+                        continue;
+
+                    executors.Add(new BenchmarkExecutor(benchmarkTypeInstance, benchmark));
```
timcassell, line 44: You added an internal foreach loop, so these `continue;` aren't continuing on the outer foreach loop that they were before. I wonder if that's contributing to the deadlock you are seeing.

timcassell:
I'm curious why the deadlock only happens with your changes, and not in the existing tests that do the same blocking wait (just without getting the result). By deduction, it must be some other change causing it, no?

YegorStepanov:
Currently, these validators don't await async methods. They simple call `Invoke`:
```C#
MethodInfo benchmarkMethod = ...;
benchmarkMethod.Invoke(null);
``` 

Yep, it's a xUnit bug, msTest and nUnit are completed successfully. 

I'll send a bug in a few hours.

YegorStepanov:
The bug is unlikely to be fixed anytime soon.

timcassell:
You added validation here to make sure IterationSetup/Cleanup are not async. I actually added async support for them in #2111, and this seems likely to get merged first, so I'll have to remember to update the validator there.

timcassell:
I wonder if that xUnit bug is the cause of the `AllSetupAndCleanupTest` flakiness. :thinking:

YegorStepanov:
>You added validation here to make sure IterationSetup/Cleanup are not async. I actually added async support for them in https://github.com/dotnet/BenchmarkDotNet/pull/2111, and this seems likely to get merged first, so I'll have to remember to update the validator there.

Okay, I'll keep an eye on it too. 
Also you can ping me because I want to move more checks from `BenchmarkConverter` to the validators and remove duplicate checks from these PR's classes. 

I hope your PRs will be accepted before the Unity moves to .NET Core. Just two years left to wait 😭

YegorStepanov:
> I wonder if that xUnit bug is the cause of the AllSetupAndCleanupTest flakiness. 🤔

I didn't even think why some tests are flakiness.
These tests never fail locally, only on the VM?

Hmm, it is a good idea to check it, it not hard (I don't want to do it 😄):
1) add new project
2) copy flakiness tests
3) update xUnit assertion to NUnit/MsUnit
4) Start spamming in the fork

timcassell:
> I hope your PRs will be accepted before the Unity moves to .NET Core. Just two years left to wait 😭

Yeah it would be nice to unblock async engine support, but what does it have to do with Core in Unity?

adamsitnik:
I've observed similar CI issues in the past in other projects. What is unique about most CI machines is that they have only one or two cores, so getting into a deadlock is way easier. You could try to reproduce it locally by running the tests affinized to 1 or 2 cores.

src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/InProcessNoEmitRunner.cs
```diff
@@ -61,39 +62,43 @@ public static int Run(IHost host, BenchmarkCase benchmarkCase)
...
+            {
+                var setter = paramProperty.GetSetMethod();
+                if (setter == null)
+                    throw new InvalidOperationException(
+                        $"Type {targetType.FullName}: no settable property {parameter.Name} found.");
+
+                var callInstance = setter.IsStatic ? null : instance;
+                setter.Invoke(callInstance, new[] { parameter.Value });
             }
         }
```
timcassell, line 104: Do these changes need to be made in `InProcess/InProcessRunner.cs` also? It's deprecated, but I'm not sure if we want to keep it in sync.

src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
```diff
@@ -121,135 +146,115 @@ private bool TryToCallGlobalMethod<T>(object benchmarkTypeInstance, List<Validat
...
+
+            bool hasError = false;
 
-            foreach (var paramProperty in paramProperties)
+            foreach (var parameter in benchmark.Parameters.Items)
             {
-                var setter = paramProperty.SetMethod;
-                if (setter == null || !setter.IsPublic)
+                // InProcessNoEmitToolchain does not support arguments
+                if (!parameter.IsArgument)
```
timcassell, line 215: Can that be added as a `ValidationError` if the toolchain is `InProcessNoEmitToolchain`?

YegorStepanov: I did not understand your question.
`ReturnValueValidator`/`ExecutionValidator` works with `InProcessToolchain` only under the hood.

src/BenchmarkDotNet/Toolchains/InProcess.NoEmit/InProcessNoEmitRunner.cs
YegorStepanov: I don't know. I simply extract the method

YegorStepanov:
> > I hope your PRs will be accepted before the Unity moves to .NET Core. Just two years left to wait 😭
> 
> Yeah it would be nice to unblock async engine support, but what does it have to do with Core in Unity?

Since you are writing a library for Unity, I thought it was related to Unity.

Anyway, we need to support taskable object to support `UniTask` and `UnityEngine.Awaitable/UnityTask` without converting them to `Task`.

To do this, we need to slightly modify your `AwaitHelper`

src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs
timcassell: I mean, you're checking for that condition already, so if the user is actually using `InProcessNoEmitToolchain`, that should add the `ValidationError` there instead of continuing. Otherwise I guess it will fail downstream when the `BenchmarkActionFactory` actually builds the actions.

YegorStepanov: BDN treats parameters and arguments as one entity. So, we should split them:

I only pass parameters to `InProcessNoEmitRunner.FillMember` because arguments are not supported.
I'm passing arguments here (they are already filled with values):
https://github.com/dotnet/BenchmarkDotNet/blob/ab54ca823490b49d42999f4a701f4c630e68770e/src/BenchmarkDotNet/Validators/ExecutionValidatorBase.cs#L286

Now I understand you correctly? 😊

timcassell: Oh I guess this makes more sense to do in the `BenchmarkExecutor` than here. Something like this:

```cs
if (arguments.Length > 0)
{
    foreach (var job in benchmark.Config.GetJobs())
    {
#pragma warning disable CS0618 // Type or member is obsolete
        if (job.Infrastructure.Toolchain is InProcessNoEmitToolchain || job.Infrastructure.Toolchain is InProcessToolchain)
#pragma warning restore CS0618 // Type or member is obsolete
        {
            errors.Add(new ValidationError(
                true,
                $"Failed to execute benchmark '{benchmark.DisplayInfo}' with {job.Infrastructure.Toolchain.GetType().Name}, this toolchain does not support arguments.",
                benchmark));
        }
    }
}
```

YegorStepanov: I'm not sure that those optional validators are supposed to validate toolchains.

I think this is a job for the new mandatory `ToolchainValidator`.

snowfrogdev:
@YegorStepanov any plans for this?

timcassell:
@snowfrogdev Yegor is MIA, feel free to take over here.

timcassell:
Superseded by #2958 and #3147

PR #2115, opened 2022-09-21
Fix stopping the collection

janvorli:
The problem was that the parameters.Process passed in for the "stop" command was null. I've moved the try / finally one call level down to a place where we still have the process.
With this change, the traces were successfully generated.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliExecutor.cs
+++ b/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliExecutor.cs
@@ -34,24 +34,15 @@ namespace BenchmarkDotNet.Toolchains.DotNetCli
                 return ExecuteResult.CreateFailed();
             }
 
-            try
-            {
-                return Execute(
-                    executeParameters.BenchmarkCase,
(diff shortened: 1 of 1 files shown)


adamsitnik:
@janvorli could you please share a sample trace file produced by perfcollect? sometimes it seems to work, but it's empty

janvorli:
Hmm, interesting. The zip was generated, it contains various files, but it doesn't seem to contain the perf.data file.

janvorli:
But I wonder - are we expecting the perf.data to be there? It seems that the perfcollect script sets usePerf to 0 when using the "start" / "stop" method and only allows perf to be used in the mode when the perfcollect is kept running and canceled via ctrl+c. 
https://github.com/microsoft/perfview/blob/main/src/perfcollect/perfcollect#L2025-L2032
Even with usePerf set to 0, we should get events though.

adamsitnik:
> "start" / "stop" method and only allows perf to be used in the mode when the perfcollect is kept running and canceled via ctrl+c.

I was not aware of that when I was implementing the prototype. In such case we need to perform collect + Ctrl+C.

PR #2116, opened 2022-09-21
Implement apples to apples comparison mode

adamsitnik:
Sample:

```cs
public class IntroApplesToApples
{
    private readonly string path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
    private readonly string content = new string('a', 100_000);

    [Benchmark]
    public void WriteAllText() => File.WriteAllText(path, content);

    [GlobalCleanup]
    public void Delete() => File.Delete(path);
}
```

```cmd
dotnet run -c Release -f net6.0 --filter *IntroApplesToApples* --apples --runtimes net6.0 net7.0 --iterationCount 20 --statisticalTest 5%
```

It runs every benchmark for just one iteration, just so the Pilot Stage can estimate invocation count:

image

Then for every job defined in the config, it runs the benchmarks again, but using exact same invocation count:

image

image

Users can at the same time use one of the built-in profilers or provide their own or don't use any profiler at all.

cc @mrsharm @Maoni0

### DIFF
--- a/src/BenchmarkDotNet/Configs/ConfigOptions.cs
+++ b/src/BenchmarkDotNet/Configs/ConfigOptions.cs
@@ -40,7 +40,11 @@ namespace BenchmarkDotNet.Configs
         /// <summary>
         /// Determines whether to generate msbuild binlogs
         /// </summary>
-        GenerateMSBuildBinLog = 1 << 7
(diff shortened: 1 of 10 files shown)


adamsitnik:
I am going to merge it now and later improve it and add docs, once I verify that this solves the problem we have.

PR #2118, opened 2022-09-23
Resolve indirect addresses in disassembly

adamsitnik:


adamsitnik:
I am going to merge it right away and continue from there.

PR #2120, opened 2022-09-24
Roslyn Toolchain does not support .NET Core

YegorStepanov:
Currently, it writes the error and does not return to console.


adamsitnik (approved):

PR #1621, opened 2020-12-17
Hide columns for multiple runtime

marcnet80:
https://github.com/dotnet/BenchmarkDotNet/issues/1603

PR #2121, opened 2022-09-24
Show toolchain column when multiple toolchains are used

YegorStepanov:
Since https://github.com/dotnet/BenchmarkDotNet/pull/1621 the toolchain column is never shown.

There was no way to compare `InProcess` with `Core` toolchain.

It will always show toolchain on comparing `Core` with `NativeAOT` since there are the different toolchains, but it easy to use `[HideColumn(Column.Toolchain)`

I made `IsMultipleRuntimes` internal because it is used only in one place. Later I will try inline it.


adamsitnik (commented):
I am afraid this is trickier that it looks like. 

Currently when I run

```cmd
cd samples\BenchmarkDotNet.Samples
dotnet run -c Release -f net6.0 --filter *IntroBasic.Sleep --join --runtimes net462 net6.0 --job dry
```

I get:

image

With your changes the toolchain column becomes visible again:

image

We want the toolchain column to be visible only when there are at least two benchmarks that use the same Runtime, but different Toolchain.

YegorStepanov:
It looks that we should never display the `Id` column:
1) All BDN attributes (like SimpleJob) set `JobId`, so we can't check `IdCharacterisctic`. 
2) We should not show `Id` if there is only one job, because it's already displayed under `[Host]`.
3) When `Jobs` have a significant difference (e.g. different `InvocationCount`), then the `InvocationCount` column should be displayed

I think we should disallow setting `JobId` by user, because now it only shows up in the logs. The random `Id` is enough.

PS I found a bug, do not review

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -444,6 +446,7 @@ private static Job CreateAotJob(Job baseJob, CommandLineOptions options, Runtime
                 builder.UseNuGet(ilCompilerVersion, nuGetFeedUrl);
 
             var runtime = runtimeMoniker.GetRuntime();
+            builder.DisplayName(runtime.Name);
             builder.TargetFrameworkMoniker(runtime.MsBuildMoniker);
```
YegorStepanov, line 450: This line changed `ILCompiler 6.0.0-*` to `NativeAot6.0`. Revert it


```suggestion
            var runtime = runtimeMoniker.GetRuntime();
            builder.TargetFrameworkMoniker(runtime.MsBuildMoniker);
```

tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
```diff
@@ -370,22 +370,53 @@ public void NetFrameworkMonikerParsedCorrectly(string tfm)
...
+            Assert.Single(config.GetJobs());
+            var job = config.GetJobs().Single();
+            Assert.NotNull(job);
+            Assert.Equal(expected, job.Environment.Runtime.Name);
+            Assert.Equal(expected, job.GetToolchain().Name);
+        }
+
+        [Theory]
+        [InlineData("nativeaot6.0", "NativeAOT 6.0", "ILCompiler 6.0.0-*")]
+        [InlineData("nativeaot7.0", "NativeAOT 7.0", "Latest ILCompiler")]
```
YegorStepanov, line 403: Is `Latest ILCompiler` the desired toolchain name? Should it be `ILCompiler 7.0.0-*`?

tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
```diff
@@ -370,22 +370,53 @@ public void NetFrameworkMonikerParsedCorrectly(string tfm)
...
+        public void NativeAotMonikersHaveNiceNames(string input, string runtimeName, string toolchainName)
+        {
+            // CoreRuntime.
+            var config = ConfigParser.Parse(new[] { "-r", input }, new OutputLogger(Output)).config;
```
YegorStepanov, line 407:
```suggestion
        {
            var config = ConfigParser.Parse(new[] { "-r", input }, new OutputLogger(Output)).config;
```

YegorStepanov:
Ready for review.

The only problem I found is:
https://github.com/dotnet/BenchmarkDotNet/blob/aa8ef4321888dde0af6a93c8e0c55f872f11297c/tests/BenchmarkDotNet.Tests/Columns/JobColumnsApprovalTests.cs#L92-L97 

I didn't dare to change a default toolchain.

YegorStepanov:
From #1603: "The Job column should remain visible when the user has set the Job id in explicit way."
**Done**

2) Changed runtime names when called from cmd:
`net50` -> `.NET 5.0`
3) Changed tfm name when called from cmd:
`net50` -> `net5.0` (to the correct tfm)
4) Simplified name:
`InProcessEmitToolchain` -> `InProcessEmit` (like other toolchains)

YegorStepanov:
### Some possible improvements:
#### 1) The `[Host]` section displays runtime info for each job Id.

This means that the `[Host]` section will only show the first job if multiple jobs have the same ID:
```C#
DefaultConfig.Instance
    .AddJob(Job.Dry)
    .AddJob(Job.Dry.WithRuntime(...)); // not displayed in the [Host] section
```

We can display the runtime info for each job.


#### 2) When `Id` is not set, a random `Id` is used.
In such cases we can set the `Id` to be the difference between the jobs.

This is similar to how `id` is set using attributes. In this case, we do not need `implicitId`.

```C#
DefaultConfig.Instance
    .AddJob(Job.Dry)                     // id=Dry
    .AddJob(Job.Dry.WithRuntime(Net6.0)) // id=Dry-NET6.0

DefaultConfig.Instance
    .AddJob(Job.Dry.WithRuntime(Net6.0).WithInvocationCount(10)) // id=Dry-InvocationCount=10
    .AddJob(Job.Dry.WithRuntime(Net6.0)) // id=Dry
```

Cons: 
When a new job is added, it can change another job Id (currently, the random Id is persistent)

#### 3) Drop a point from runtime names
`.NET 6.0` -> `NET 6.0`
`.NET Framework 4.8.1` -> `NET Framework 4.8.1`

Look at the `Host` section and the `Runtime` column [here](https://github.com/dotnet/BenchmarkDotNet/blob/2c99d26ec4cd36ad33cd4c1c00a2a4b87068684e/tests/BenchmarkDotNet.Tests/Columns/ApprovedFiles/JobColumnsApprovalTests.MultipleInputJobsAreDiscovered.approved.txt)

PR #2123, opened 2022-09-26
Added other arm64 constant form extraction plus other changes

janvorli:
* Added some cleanup
* arm64 JIT in .NET 7 can generate a new form of constant loading using `adrp` / `add` instruction pair, so add handling that case
* Reordered address checks in `TryTranslateAddressToName` to first check for method addresses and only when it fails to look for `MethodDesc` / `MethodTable`. The `MethodDesc` is checked on arm64 only if the `i`sAddressPrecodeMD` is passed in, the `MethodTable` check is never made. The reason is that on arm64, we only try translate call / jump targets.

### DIFF
--- a/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
+++ b/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
@@ -199,34 +199,31 @@ namespace BenchmarkDotNet.Disassemblers
                     ulong address = 0;
                     if (TryGetReferencedAddress(instruction, accumulator, (uint)state.Runtime.DataTarget.DataReader.PointerSize, out address, out isIndirect))
                     {
-                        if (isIndirect)
+                        if (isIndirect && state.IsNet7)
                         {
-                            if (state.IsNet7)
-                            {
                             // Check if the target is a known stub
                             // The stubs are allocated in interleaved code / data pages in memory. The data part of the stub
                             // is at an address one memory page higher than the code.
-                                byte[] buffer = new byte[12];
+                            byte[] buffer = new byte[12];
 
-                                if (state.Runtime.DataTarget.DataReader.Read(address, buffer) == buffer.Length)
+                            if (state.Runtime.DataTarget.DataReader.Read(address, buffer) == buffer.Length)
+                            {
+                                if (buffer.SequenceEqual(callCountingStubTemplate))
                                 {
-                                    if (buffer.SequenceEqual(callCountingStubTemplate))
-                                    {
-                                        const ulong TargetMethodAddressSlotOffset = 8;
-                                        address = state.Runtime.DataTarget.DataReader.ReadPointer(address + (ulong)Environment.SystemPageSize + TargetMethodAddressSlotOffset);
-                                    }
-                                    else if (buffer.SequenceEqual(stubPrecodeTemplate))
-                                    {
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
```diff
@@ -57,43 +66,55 @@ public void Feed(Arm64Instruction instruction)
...
+                        details.Operands[2].Type == Arm64OperandType.Immediate)
+                    {
+                        _value = _value | instruction.Details.Operands[2].Immediate;
+                        _state = State.LookingForPossibleLdr;
+                    }
+                    break;
                 case State.LookingForPossibleLdr:
                     if (instruction.Id == Arm64InstructionId.ARM64_INS_LDR &&
-                        _registerId == details.Operands[0].Register.Id && // Target of the LDR is the register we are tracking
+                        //_registerId == details.Operands[0].Register.Id && // Target of the LDR is the register we are tracking
```
adamsitnik, line 81: should the commented code be removed?

```suggestion
```

src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
```diff
@@ -137,13 +194,44 @@ protected override IEnumerable<Asm> Decode(byte[] code, ulong startAddress, Stat
...
+                                    }
+                                    else if (buffer.SequenceEqual(fixupPrecodeTemplate))
+                                    {
+                                        const ulong MethodDescSlotOffset = 8;
+                                        address = state.Runtime.DataTarget.DataReader.ReadPointer(address + (ulong)Environment.SystemPageSize + MethodDescSlotOffset);
+                                        isPrestubMD = true;
+                                    }
+                                }
+                            }
+                        }
```
adamsitnik, line 233: nit: we could reduce the nesting depth:

```suggestion
                        if (isIndirect && state.IsNet7)
                        {
                            // Check if the target is a known stub
                            // The stubs are allocated in interleaved code / data pages in memory. The data part of the stub
                            // is at an address one memory page higher than the code.
                            byte[] buffer = new byte[12];

                            if (state.Runtime.DataTarget.DataReader.Read(address, buffer) == buffer.Length)
                            {
                                if (buffer.SequenceEqual(callCountingStubTemplate))
                                {
                                    const ulong TargetMethodAddressSlotOffset = 8;
                                    address = state.Runtime.DataTarget.DataReader.ReadPointer(address + (ulong)Environment.SystemPageSize + TargetMethodAddressSlotOffset);
                                }
                                else if (buffer.SequenceEqual(stubPrecodeTemplate))
                                {
                                    const ulong MethodDescSlotOffset = 0;
                                    address = state.Runtime.DataTarget.DataReader.ReadPointer(address + (ulong)Environment.SystemPageSize + MethodDescSlotOffset);
                                    isPrestubMD = true;
                                }
                                else if (buffer.SequenceEqual(fixupPrecodeTemplate))
                                {
                                    const ulong MethodDescSlotOffset = 8;
                                    address = state.Runtime.DataTarget.DataReader.ReadPointer(address + (ulong)Environment.SystemPageSize + MethodDescSlotOffset);
                                    isPrestubMD = true;
                                }
                            }
                        }
```

adamsitnik (approved):
@janvorli it looks great to me, big thanks for your amazing contribution! and thank you for adding all the comments!

PR #2124, opened 2022-09-26
Fix crash when using HardwareCounter.NotSet

### LINKED ISSUE #1856: Not Set Hardware Counter throws Argument Null Exception
Hi, 

First, thank you for your effort. BenchmarkDotNet is a great tool. 

**Issue:**

I have encountered an interesting exception. 

When you set 

`... AddHardwareCounters(HardwareCounter.NotSet) ...` on configs,

.net throws

`System.ArgumentNullException: 'Value cannot be null. (Parameter 'key')`

on  https://github.com/dotnet/BenchmarkDotNet/blob/34817bfdd8909711b4802415808d26849aaed98e/src/BenchmarkDotNet.Diagnostics.Windows/HardwareCounters.cs#L56

Since `HardwareCounter.NotSet` is enum, there could be a proper way to handle/prevent it at  
https://github.com/dotnet/BenchmarkDotNet/blob/34817bfdd8909711b4802415808d26849aaed98e/src/BenchmarkDotNet.Diagnostics.Windows/HardwareCounters.cs#L53

I believe.

If you consider to fix it, I may create a PR.

Regards,
Emre

YegorStepanov:
Fixes #1856

### DIFF
--- a/src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
+++ b/src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
@@ -38,7 +38,7 @@ namespace BenchmarkDotNet.Configs
             var uniqueLoggers = source.GetLoggers().ToImmutableHashSet();
             var configAnalyse = new List<Conclusion>();
 
-            var uniqueHardwareCounters = source.GetHardwareCounters().ToImmutableHashSet();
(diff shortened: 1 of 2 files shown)


AndreyAkinshin (approved):

AndreyAkinshin:
@YegorStepanov thanks!

PR #2132, opened 2022-10-04
fix: include argument and param names in --filter

### LINKED ISSUE #2131: --filter should include argument/params names
Currently the filter provided with `--filter` is applied only to `namespace.typeName.methodName`. It should be applied to full benchmark ID, which includes argument and parameter values, as it's currently impossible to run a single benchmark when it has multiple test cases.

Requirements:
- to remain backward compatibility, using `namespace.typeName.methodName` without an asterisk at the end should still run all benchmarks for given method (treated as `namespace.typeName.methodName*`)
- hints that are printed when filter returns no benchmarks to run should be fixed as well, the person who is willing to work on it can take a look at https://github.com/dotnet/BenchmarkDotNet/pull/1644 to see what was needed in the past to fix the suggestions

Some hints:
- the code lives here: https://github.com/dotnet/BenchmarkDotNet/blob/a78c2e6a6e3db79069fb5bbbd6da6e5cbea8c029/src/BenchmarkDotNet/Filters/GlobFilter.cs#L19
- it should be using https://github.com/dotnet/BenchmarkDotNet/blob/a78c2e6a6e3db79069fb5bbbd6da6e5cbea8c029/src/BenchmarkDotNet/Exporters/FullNameProvider.cs#L50

blouflashdb:
Closes #2131

### DIFF
--- a/src/BenchmarkDotNet/Filters/GlobFilter.cs
+++ b/src/BenchmarkDotNet/Filters/GlobFilter.cs
@@ -1,5 +1,7 @@
+using System;
 using System.Linq;
 using System.Text.RegularExpressions;
+using BenchmarkDotNet.Exporters;
 using BenchmarkDotNet.Running;
 
 namespace BenchmarkDotNet.Filters
@@ -11,15 +13,18 @@ namespace BenchmarkDotNet.Filters
     {
         private readonly Regex[] patterns;
 
-        public GlobFilter(string[] patterns) => this.patterns = ToRegex(patterns);
+        public GlobFilter(string[] patterns) => this.patterns = ToRegex(AddWildcardCharacterToEnd(patterns));
 
         public bool Predicate(BenchmarkCase benchmarkCase)
         {
             var benchmark = benchmarkCase.Descriptor.WorkloadMethod;
-            string fullBenchmarkName = benchmarkCase.Descriptor.GetFilterName();
+            string fullBenchmarkName = FullNameProvider.GetBenchmarkName(benchmarkCase);
 
             return patterns.Any(pattern => pattern.IsMatch(fullBenchmarkName));
         }
+
+        internal static string[] AddWildcardCharacterToEnd(string[] patterns)
+            => Array.ConvertAll(patterns, pattern => pattern.EndsWith("*") ? pattern : $"{pattern}*");
 
         internal static Regex[] ToRegex(string[] patterns)
             => patterns.Select(pattern => new Regex(WildcardToRegex(pattern), RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)).ToArray();
--- a/tests/BenchmarkDotNet.Tests/GlobFilterTests.cs
+++ b/tests/BenchmarkDotNet.Tests/GlobFilterTests.cs
@@ -1,4 +1,6 @@
 ﻿using System.Linq;
+using System.Threading;
+using ApprovalUtilities.Utilities;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Filters;
 using BenchmarkDotNet.Running;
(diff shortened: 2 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2132) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2132) before we can accept your contribution.<br/><br/>:x: blouflashdb [sign now](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2132)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/BenchmarkDotNet?pullRequest=2132) it.</sub>

blouflashdb:
Not sure why some checks are failing but I think it has nothing to do with my changes.

src/BenchmarkDotNet/Filters/GlobFilter.cs
```diff
@@ -11,16 +13,19 @@ public class GlobFilter : IFilter
...
         {
             var benchmark = benchmarkCase.Descriptor.WorkloadMethod;
-            string fullBenchmarkName = benchmarkCase.Descriptor.GetFilterName();
+            string fullBenchmarkName = FullNameProvider.GetBenchmarkName(benchmarkCase);
 
             return patterns.Any(pattern => pattern.IsMatch(fullBenchmarkName));
         }
 
+        internal static string[] AddWildcardCharacterToEnd(string[] patterns)
+            => Array.ConvertAll(patterns, pattern => pattern.EndsWith("*") ? pattern : $"{pattern}*");
```
adamsitnik, line 27: This might have some unintended side effects. Example: let's say that the user provides `--filter $namespace` without asterisk at the end. This filter is wrong, and it should return 0 benchmarks, but by adding `*` here we make it "correct".

To remain the backward compatibility, we could do something like this:

```cs
string nameWithoutArgs = benchmarkCase.Descriptor.GetFilterName();
string fullBenchmarkName = FullNameProvider.GetBenchmarkName(benchmarkCase)

return patterns.Any(pattern => pattern.IsMatch(fullBenchmarkName) || pattern.IsMatch(nameWithoutArgs));
```

@blouflashdb what do you think?

tests/BenchmarkDotNet.Tests/GlobFilterTests.cs
```diff
@@ -27,9 +29,38 @@ public void TheFilterIsCaseInsensitive(string pattern, bool expected)
...
+        [InlineData("TYPEWITHBENCHMARKSANDPARAMS", 0)] // type name uppercase
+        [InlineData("*TypeWithBenchmarksAndParams*", 2)] // regular expression
+        [InlineData("*typewithbenchmarksandparams*", 2)] // regular expression lowercase
+        [InlineData("*TYPEWITHBENCHMARKSANDPARAMS*", 2)] // regular expression uppercase
+        [InlineData("*", 2)]
+        [InlineData("WRONG", 0)]
+        [InlineData("*stillWRONG*", 0)]
+        [InlineData("BenchmarkDotNet.Tests.TypeWithBenchmarksAndParams", 2)]
+        [InlineData("BenchmarkDotNet.Tests.TypeWithBenchmarksAndParams.TheBenchmark(A: 100)", 1)]
+        public void TheFilterWorksWithParams(string pattern, int expectedBenchmarks)
```
adamsitnik, line 44: big thanks for adding great tests! 👍

adamsitnik (commented):
Overall, it looks good, but I think that we should address one concern before merging. Please take a look at my comment.

Thank you for your contribution @blouflashdb !

src/BenchmarkDotNet/Filters/GlobFilter.cs
blouflashdb: Ah yes, I was not thinking about that sideeffect. I like your idea to remain backward compatibility. Going to change it. Thanks for the review.

AndyAyersMS:
Didn't realize this had changed away from the `--parameter-filter` in the perf repo. 

At any rate some more detailed usage examples would be useful, eg it took me a while to figure out how to properly quote things from windows batch scripts, eg:
```
--filter "System.Tests.Perf_Int32.Parse(value: ""-2147483648"")" 
```

adamsitnik:
> At any rate some more detailed usage examples would be useful

@AndyAyersMS Would you like to send a PR? You would need to add a new example here:

https://github.com/dotnet/BenchmarkDotNet/blob/276f1add338e6a4c49053dc4809474d9ad9b91f8/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs#L246

PR #2133, opened 2022-10-05
ensure CompositeLogger is synchronized

### LINKED ISSUE #2125: Tests Just Stop Running During Run
This issue seems to be back where while running a benchmark test, my console app just disappears. No error, nothing even in the BDN log (see below). It seems to be happening randomly, meaning different classes and different methods. Just a week or so ago, they were working fine.

I do have a try/catch set up in the main method. I do have it configured too:

config.WithOption(ConfigOptions.DisableOptimizationsValidator, true)
  .WithOption(ConfigOptions.StopOnFirstError, true);

But I am not seeing anything. The project targets .NET 6 & 7 and I am using BenchmarkDotNet version 0.13.2.1913.

I did find two events in the Event Log that I have attached below.

[dotNetTips.CodePerf.Example.App.BenchmarkTests.CollectionLinqBenchmark-20220927-183242.log](https://github.com/dotnet/BenchmarkDotNet/files/9668922/dotNetTips.CodePerf.Example.App.BenchmarkTests.CollectionLinqBenchmark-20220927-183242.log)

[BenchmarkDotNet.zip](https://github.com/dotnet/BenchmarkDotNet/files/9668970/BenchmarkDotNet.zip)

adamsitnik:
fixes #2125

### DIFF
--- a/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
+++ b/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
@@ -106,8 +106,14 @@ namespace BenchmarkDotNet.Loggers
             if (!string.IsNullOrEmpty(e.Data))
             {
                 output.Enqueue(e.Data);
+
                 if (logOutput)
-                    logger.WriteLine(e.Data);
+                {
+                    lock (this) // #2125
+                    {
+                        logger.WriteLine(e.Data);
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


timcassell:
I know the class is internal, but wouldn't it be safer to lock on one of the readonly fields than `this`?

adamsitnik:
> I know the class is internal, but wouldn't it be safer to lock on `logger` than `this`?

`logger` is provided via `ctor` and in theory it can be locked by some other components as well.

My initial idea was to make entire `CompositeLogger` synchronized, but later I got to the conclusion that we don't need to lock all its usages, just the one from async stream output reader

PR #2135, opened 2022-10-05
Escape Param data for the exporters

### LINKED ISSUE #1839: Markdown output should escape the output

### LINKED ISSUE #1867: Trailing newline characters in input value break summary table
Version: 0.13.1

When a parameterised input value has trailing newline characters, they are rendered in the summary table on the console, breaking the table.

```c#
public class Repro
{
    [ParamsSource(nameof(Data))]
    public string Input { get; set; }

    [Benchmark]
    public void A()
    {
        _ = Input.Length;
    }

    public static IEnumerable<string> Data()
    {
        yield return "Input without trailing newline character";
        yield return "Input with trailing newline characters\r\n";
    }
}
```

Screenshot of the console window (notice the line break in the first data row):
grafik

YegorStepanov:
This PR adds escaping the `Param` data for the exporters only. 
Currently, you can only pass `\t`, `\n`, other control characters throw an exception in the generated project). 

Try `Param["\0"] string Param` (In the linked PR below I fix it).

---

`Microsoft.CodeAnalysis.CSharp.SymbolDisplay.FormatLiteral(str, quote);` 
the naming is 10/10 😄. It escapes the non-displayed characters and emojis.

---

I don't escape the `column.ColumnName`, because that would require calling `.EscapeSpecialCharacter()` after every 'column.ColumName' (~20 times).

Fixes #1867
Fixes #1839

### DIFF
--- a/src/BenchmarkDotNet/Extensions/StringAndTextExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/StringAndTextExtensions.cs
@@ -28,6 +28,16 @@ namespace BenchmarkDotNet.Extensions
         }
 
         /// <summary>
+        /// Escapes UNICODE control characters
+        /// </summary>
+        /// <param name="str">string to escape</param>
+        /// <param name="quote">True to put (double) quotes around the string literal</param>
+        internal static string EscapeSpecialCharacters(this string str, bool quote)
+        {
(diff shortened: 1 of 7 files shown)


adamsitnik (approved):

PR #2140, opened 2022-10-06
Update console title with benchmark information

franciscomoloureiro:


src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -614,6 +620,15 @@ private static void Cleanup(HashSet<string> artifactsToCleanup)
             }
         }
 
+        private static void UpdateTitle(in StartedClock runsChronometer, int totalBenchmarkCount, int benchmarksToRunCount)
+        {
+            int executedBenchmarkCount = totalBenchmarkCount - benchmarksToRunCount;
+            double avgSecondsPerBenchmark = runsChronometer.GetElapsed().GetTimeSpan().TotalSeconds / executedBenchmarkCount;
+            TimeSpan fromNow = TimeSpan.FromSeconds(avgSecondsPerBenchmark * benchmarksToRunCount);
+
+            Console.Title = $"{executedBenchmarkCount}/{totalBenchmarkCount} Benchmarks runned - Estimated {(int)fromNow.TotalHours}h {fromNow.Minutes}m to finish";
```
timcassell, line 629: Seems kind of long (and "runned" is not a word).

How about `1/16 - 0h 32m - Benchmarks`?

I think it should be # currently executing instead of # complete, but that's just my opinion.

Also, the console title should be reverted after the benchmarks are complete.

franciscomoloureiro: Sorry for bad english.

I have changed the title a bit, number of remaining benchmarks + time to finish

Also refactored small code duplication

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -614,15 +622,38 @@ private static void Cleanup(HashSet<string> artifactsToCleanup)
...
+                int executedBenchmarkCount = totalBenchmarkCount - benchmarksToRunCount;
+                TimeSpan fromNow = GetEstimatedFinishTime(runsChronometer, benchmarksToRunCount, executedBenchmarkCount);
+                string toFinish = string.Empty;
+
+                if (executedBenchmarkCount > 0)
+                {
+                    toFinish = $"- {(int)fromNow.TotalHours}h {fromNow.Minutes}m to finish";
+                }
+
+                Console.Title = $"{benchmarksToRunCount} Remaining {toFinish}";
```
timcassell, line 638: That doesn't really tell you the progress. You need a % or fraction like you had before.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -218,6 +221,11 @@ internal static Summary[] Run(BenchmarkRunInfo[] benchmarkRunInfos)
                 }
             }
 
+            if (RuntimeInformation.IsWindows())
+            {
+                Console.Title = string.Empty;
```
timcassell, line 226: Instead of setting the title to empty, you should cache the old title before the run, then reset it back to the old title after the run.

```cs
var oldTitle = Console.Title;
...
Console.Title = oldTitle;
```

franciscomoloureiro: I actualy thought about this, I don't think we set the a title before runing the benchmark, or after, that is why I did not cache it.

But this is in fact more future proof solution, I will change it

timcassell: It's because other applications can run in the same console window.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
timcassell:
```cs
Console.Title = $"{benchmarksToRunCount}/{totalBenchmarkCount} Remaining {toFinish}";
```

adamsitnik (commented):
@franciscomoloureiro big thanks for your contribution! PTAL at my comments.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -164,6 +166,8 @@ internal static Summary[] Run(BenchmarkRunInfo[] benchmarkRunInfos)
                     var info = buildResults[benchmark];
                     var buildResult = info.buildResult;
 
+                    UpdateTitle(in runsChronometer, totalBenchmarkCount, benchmarksToRunCount);
```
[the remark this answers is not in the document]
timcassell: I think we want at least the number of benchmarks to run in the title at the start, even if we don't yet have a time estimate. I think that could be set once at the start, then use `LogProgress` to update it with the time.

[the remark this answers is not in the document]
franciscomoloureiro: Updated this, now we update the title before the loop with the amount of benchmarks to run and then each iteration with the time to finish info

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -614,15 +623,38 @@ private static void Cleanup(HashSet<string> artifactsToCleanup)
             }
         }
 
+        private static void UpdateTitle(in StartedClock runsChronometer, int totalBenchmarkCount, int benchmarksToRunCount)
+        {
+            if (RuntimeInformation.IsWindows())
```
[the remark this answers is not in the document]
franciscomoloureiro: updated if statment. I guess we cannot cache the previous console title for linux and os then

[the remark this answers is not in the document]
adamsitnik: > I guess we cannot cache the previous console title for linux and os then

that is correct

adamsitnik (approved):
LGTM, thank you for your contribution @franciscomoloureiro !

Issue #2064, opened 2022-08-02
Running with .Net 6/7 Mono JIT


MichalPetryka:
BenchmarkDotNet seems to have no easy way to run using the dotnet/runtime Mono JIT, which would be useful for comparing it against CoreCLR.

This can be achieved with `UseMonoRuntime` property in a csproj.

adamsitnik:
Hi @MichalPetryka 

You can provide MSBuild arguments for each job. Sth like:

```cs
using BenchmarkDotNet.Configs;
using BenchmarkDotNet.Environments;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Running;

namespace BenchmarkDotNet.Samples
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Job net70job = Job.Default.WithRuntime(CoreRuntime.Core70);

            BenchmarkSwitcher
               .FromAssembly(typeof(Program).Assembly)
               .Run(args,
                   DefaultConfig.Instance
                        .AddJob(net70job.WithId("CoreCLR"))
                        .AddJob(net70job.WithArguments(new Argument[] { new MsBuildArgument("/p:UseMonoRuntime=true") }).WithId("Mono")));
        }
    }
}
```

adamsitnik:
I agree that it would be nice to expose `UseMonoRuntime` as built-in BDN feature. We could add `mono7.0` moniker and support it with `--runtimes` so `--runtimes net7.0 mono7.0` would compare both runtimes running as .NET 7.

@marek-safar who would be the best person to describe all `UseMonoRuntime` possibilities and limitations? If this is just a matter of setting one property it should be fairly simple to implement. But we would also need to implement CLR vs Mono Runtime detection. The current one is probably outdated:

https://github.com/dotnet/BenchmarkDotNet/blob/32ddeb56bd38631e052350cecfbf9e86df1b2e7c/src/BenchmarkDotNet/Portability/RuntimeInformation.cs#L28

Serg046:
I would like to work on this

adamsitnik:
@Serg046 great! I've assigned you. Please let me know if you need some help

marek-safar:
https://github.com/dotnet/runtime/blob/main/src/tests/Common/CoreCLRTestLibrary/Utilities.cs#L95 is what we use recently for MonoVM

PR #2079, opened 2022-08-24
fix the CI

adamsitnik:
The builds has recently started to fail with:

```log
 C:\Program Files\dotnet\sdk\7.0.100-rc.1.22423.16\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.ConflictResolution.targets(112,5): error NETSDK1152: Found multiple publish output files with the same relative path: D:\projects\BenchmarkDotNet\tests\BenchmarkDotNet.IntegrationTests\xunit.runner.json, D:\projects\BenchmarkDotNet\tests\BenchmarkDotNet.Tests\xunit.runner.json. [D:\projects\BenchmarkDotNet\tests\BenchmarkDotNet.IntegrationTests\bin\Release\net6.0\7fafdcdc-4f53-49fa-afb4-3e835b41b979\BenchmarkDotNet.Autogenerated.csproj]
```

Since it's just a complain about test project file settings, we can ignore it.

Update: and #2080

PR #2142, opened 2022-10-07
Issue 2064: Mono70 moniker

Serg046:
Not sure I understand the task properly, please take a look
#2064

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -17,13 +17,13 @@ using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Portability;
 using BenchmarkDotNet.Reports;
-using BenchmarkDotNet.Toolchains.NativeAot;
 using BenchmarkDotNet.Toolchains.CoreRun;
 using BenchmarkDotNet.Toolchains.CsProj;
 using BenchmarkDotNet.Toolchains.DotNetCli;
 using BenchmarkDotNet.Toolchains.InProcess.Emit;
-using BenchmarkDotNet.Toolchains.MonoWasm;
 using BenchmarkDotNet.Toolchains.MonoAotLLVM;
+using BenchmarkDotNet.Toolchains.MonoWasm;
+using BenchmarkDotNet.Toolchains.NativeAot;
 using CommandLine;
 using Perfolizer.Horology;
 using Perfolizer.Mathematics.OutlierDetection;
@@ -377,6 +377,7 @@ namespace BenchmarkDotNet.ConsoleArguments
                     return baseJob
                         .WithRuntime(runtimeMoniker.GetRuntime())
                         .WithToolchain(CsProjClassicNetToolchain.From(runtimeId, options.RestorePath?.FullName));
+
                 case RuntimeMoniker.NetCoreApp20:
                 case RuntimeMoniker.NetCoreApp21:
                 case RuntimeMoniker.NetCoreApp22:
@@ -391,30 +392,43 @@ namespace BenchmarkDotNet.ConsoleArguments
                     return baseJob
                         .WithRuntime(runtimeMoniker.GetRuntime())
                         .WithToolchain(CsProjCoreToolchain.From(new NetCoreAppSettings(runtimeId, null, runtimeId, options.CliPath?.FullName, options.RestorePath?.FullName)));
+
                 case RuntimeMoniker.Mono:
                     return baseJob.WithRuntime(new MonoRuntime("Mono", options.MonoPath?.FullName));
+
                 case RuntimeMoniker.NativeAot60:
                     return CreateAotJob(baseJob, options, runtimeMoniker, "6.0.0-*", "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json");
+
                 case RuntimeMoniker.NativeAot70:
                     return CreateAotJob(baseJob, options, runtimeMoniker, "", "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7/nuget/v3/index.json");
+
                 case RuntimeMoniker.Wasm:
                     return MakeWasmJob(baseJob, options, RuntimeInformation.IsNetCore ? CoreRuntime.GetCurrentVersion().MsBuildMoniker : "net5.0", runtimeMoniker);
+
                 case RuntimeMoniker.WasmNet50:
                     return MakeWasmJob(baseJob, options, "net5.0", runtimeMoniker);
+
                 case RuntimeMoniker.WasmNet60:
                     return MakeWasmJob(baseJob, options, "net6.0", runtimeMoniker);
+
                 case RuntimeMoniker.WasmNet70:
                     return MakeWasmJob(baseJob, options, "net7.0", runtimeMoniker);
+
                 case RuntimeMoniker.MonoAOTLLVM:
                     return MakeMonoAOTLLVMJob(baseJob, options, RuntimeInformation.IsNetCore ? CoreRuntime.GetCurrentVersion().MsBuildMoniker : "net6.0");
+
                 case RuntimeMoniker.MonoAOTLLVMNet60:
                     return MakeMonoAOTLLVMJob(baseJob, options, "net6.0");
+
                 case RuntimeMoniker.MonoAOTLLVMNet70:
                     return MakeMonoAOTLLVMJob(baseJob, options, "net7.0");
+
                 case RuntimeMoniker.Mono60:
                     return MakeMonoJob(baseJob, options, MonoRuntime.Mono70, "net6.0");
+
                 case RuntimeMoniker.Mono70:
                     return MakeMonoJob(baseJob, options, MonoRuntime.Mono70, "net7.0");
+
                 default:
                     throw new NotSupportedException($"Runtime {runtimeId} is not supported");
             }
@@ -479,7 +493,7 @@ namespace BenchmarkDotNet.ConsoleArguments
                 wasmDataDir: options.WasmDataDirectory?.FullName,
                 moniker: moniker);
 
-            var toolChain = WasmToolChain.From(new NetCoreAppSettings(
+            var toolChain = WasmToolchain.From(new NetCoreAppSettings(
                 targetFrameworkMoniker: wasmRuntime.MsBuildMoniker,
                 runtimeFrameworkVersion: null,
                 name: wasmRuntime.Name,
@@ -566,7 +580,6 @@ namespace BenchmarkDotNet.ConsoleArguments
                         break;
                     }
             }
-
 
             if (commonLongestPrefixIndex <= 1)
                 return coreRunPath.FullName;
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -460,7 +460,13 @@ namespace BenchmarkDotNet.ConsoleArguments
         {
             return baseJob
                 .WithRuntime(runtime)
-                .WithToolchain(MonoToolchain.From(new NetCoreAppSettings(runtime.MsBuildMoniker, null, runtime.Name, options.CliPath?.FullName, options.RestorePath?.FullName)));
+                .WithToolchain(MonoToolchain.From(
+                    new NetCoreAppSettings(
+                        targetFrameworkMoniker: runtime.MsBuildMoniker,
+                        runtimeFrameworkVersion: null,
+                        name: runtime.Name,
+                        customDotNetCliPath: options.CliPath?.FullName,
+                        packagesPath: options.RestorePath?.FullName)));
         }
 
         private static Job MakeMonoAOTLLVMJob(Job baseJob, CommandLineOptions options, string msBuildMoniker)
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -387,9 +387,7 @@ namespace BenchmarkDotNet.ConsoleArguments
                 case RuntimeMoniker.Net50:
                 case RuntimeMoniker.Net60:
                 case RuntimeMoniker.Net70:
-                    return baseJob
-                        .WithRuntime(runtimeMoniker.GetRuntime())
-                        .WithToolchain(CsProjCoreToolchain.From(new NetCoreAppSettings(runtimeId, null, runtimeId, options.CliPath?.FullName, options.RestorePath?.FullName)));
+                    return CreateCoreJob(baseJob, runtimeId, runtimeId, options, runtimeMoniker.GetRuntime());
                 case RuntimeMoniker.Mono:
                     return baseJob.WithRuntime(new MonoRuntime("Mono", options.MonoPath?.FullName));
                 case RuntimeMoniker.NativeAot60:
@@ -410,9 +408,19 @@ namespace BenchmarkDotNet.ConsoleArguments
                     return MakeMonoAOTLLVMJob(baseJob, options, "net6.0");
                 case RuntimeMoniker.MonoAOTLLVMNet70:
                     return MakeMonoAOTLLVMJob(baseJob, options, "net7.0");
+                case RuntimeMoniker.Mono70:
+                    return CreateCoreJob(baseJob, "net7.0", "Mono with .NET 7.0", options, CoreRuntime.Core70)
+                        .WithArguments(new Argument[] { new MsBuildArgument("/p:UseMonoRuntime=true") });
                 default:
                     throw new NotSupportedException($"Runtime {runtimeId} is not supported");
             }
+        }
+
+        private static Job CreateCoreJob(Job baseJob, string runtimeId, string runtimeName, CommandLineOptions options, Runtime runtime)
+        {
+            return baseJob
+                .WithRuntime(runtime)
+                .WithToolchain(CsProjCoreToolchain.From(new NetCoreAppSettings(runtimeId, null, runtimeName, options.CliPath?.FullName, options.RestorePath?.FullName)));
         }
 
         private static Job CreateAotJob(Job baseJob, CommandLineOptions options, RuntimeMoniker runtimeMoniker, string ilCompilerVersion, string nuGetFeedUrl)
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -424,10 +424,10 @@ namespace BenchmarkDotNet.ConsoleArguments
                     return MakeMonoAOTLLVMJob(baseJob, options, "net7.0");
 
                 case RuntimeMoniker.Mono60:
-                    return MakeMonoJob(baseJob, options, MonoRuntime.Mono70, "net6.0");
+                    return MakeMonoJob(baseJob, options, MonoRuntime.Mono60);
 
                 case RuntimeMoniker.Mono70:
-                    return MakeMonoJob(baseJob, options, MonoRuntime.Mono70, "net7.0");
+                    return MakeMonoJob(baseJob, options, MonoRuntime.Mono70);
 
                 default:
                     throw new NotSupportedException($"Runtime {runtimeId} is not supported");
@@ -456,11 +456,11 @@ namespace BenchmarkDotNet.ConsoleArguments
             return baseJob.WithRuntime(runtime).WithToolchain(builder.ToToolchain());
         }
 
-        private static Job MakeMonoJob(Job baseJob, CommandLineOptions options, MonoRuntime runtime, string msBuildMoniker)
+        private static Job MakeMonoJob(Job baseJob, CommandLineOptions options, MonoRuntime runtime)
         {
             return baseJob
                 .WithRuntime(runtime)
-                .WithToolchain(MonoToolchain.From(new NetCoreAppSettings(msBuildMoniker, null, runtime.Name, options.CliPath?.FullName, options.RestorePath?.FullName)));
+                .WithToolchain(MonoToolchain.From(new NetCoreAppSettings(runtime.MsBuildMoniker, null, runtime.Name, options.CliPath?.FullName, options.RestorePath?.FullName)));
         }
 
         private static Job MakeMonoAOTLLVMJob(Job baseJob, CommandLineOptions options, string msBuildMoniker)
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -29,6 +29,7 @@ using Perfolizer.Horology;
 using Perfolizer.Mathematics.OutlierDetection;
 using Perfolizer.Mathematics.SignificanceTesting;
 using Perfolizer.Mathematics.Thresholds;
+using BenchmarkDotNet.Toolchains.Mono;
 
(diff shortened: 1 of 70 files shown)


src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs
```diff
@@ -19,6 +19,7 @@ public class CoreRuntime : Runtime
         public static readonly CoreRuntime Core50 = new CoreRuntime(RuntimeMoniker.Net50, "net5.0", ".NET 5.0");
         public static readonly CoreRuntime Core60 = new CoreRuntime(RuntimeMoniker.Net60, "net6.0", ".NET 6.0");
         public static readonly CoreRuntime Core70 = new CoreRuntime(RuntimeMoniker.Net70, "net7.0", ".NET 7.0");
+        public static readonly CoreRuntime Mono70 = new CoreRuntime(RuntimeMoniker.Mono70, "mono7.0", "Mono with .NET 7.0");
```
Serg046, line 22: Not sure if this should be here or in `MonoRuntime`

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -25,7 +25,7 @@ internal static class RuntimeInformation
         internal const string ReleaseConfigurationName = "RELEASE";
         internal const string Unknown = "?";
 
-        public static bool IsMono { get; } = Type.GetType("Mono.Runtime") != null; // it allocates a lot of memory, we need to check it once in order to keep Engine non-allocating!
+        public static bool IsMono { get; } = Type.GetType("Mono.RuntimeStructs") != null; // it allocates a lot of memory, we need to check it once in order to keep Engine non-allocating!
```
timcassell, line 28: Does this still return true on the old Mono runtime?

Mono checks are performed elsewhere in the library like in the memory diagnoser (`GcStats`) since it did not support `AppDomain.MonitoringIsEnabled`. Is that still the case with the new Mono runtime?

It's also checked in the `MonoDisassembler`, and I have no idea if that's relevant here.

And other places in RuntimeInformation.cs, all of which should be verified.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -410,11 +408,21 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
                     return MakeMonoAOTLLVMJob(baseJob, options, "net6.0");
                 case RuntimeMoniker.MonoAOTLLVMNet70:
                     return MakeMonoAOTLLVMJob(baseJob, options, "net7.0");
+                case RuntimeMoniker.Mono70:
+                    return CreateCoreJob(baseJob, "net7.0", "Mono with .NET 7.0", options, CoreRuntime.Core70)
+                        .WithArguments(new Argument[] { new MsBuildArgument("/p:UseMonoRuntime=true") });
```
adamsitnik, line 413: I am afraid that this might not be enough. I've failed to find any official documentation (this is sad), but according to https://twitter.com/lambdageek/status/1457784508371980291 we need to specify `/p:UseMonoRuntime=true` and publish a self-contained application for specific runtime. 

Here is a small repro:

```cs
Console.WriteLine(Type.GetType("Mono.RuntimeStructs") != null);
```

```log
PS D:\projects\repros\consoleBdn> dotnet publish -c Release -f net7.0 --self-contained -r win-x64 /p:UseMonoRuntime=true
MSBuild version 17.4.0-preview-22428-01+14c24b2d3 for .NET
  Determining projects to restore...
  Restored D:\projects\repros\consoleBdn\consoleBdn.csproj (in 11 ms).
C:\Program Files\dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(219,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-p
olicy [D:\projects\repros\consoleBdn\consoleBdn.csproj]
  consoleBdn -> D:\projects\repros\consoleBdn\bin\Release\net7.0\win-x64\consoleBdn.dll
  consoleBdn -> D:\projects\repros\consoleBdn\bin\Release\net7.0\win-x64\publish\
PS D:\projects\repros\consoleBdn> D:\projects\repros\consoleBdn\bin\Release\net7.0\win-x64\publish\consoleBdn.exe
True
PS D:\projects\repros\consoleBdn> dotnet build -c Release -f net7.0 /p:UseMonoRuntime=true
MSBuild version 17.4.0-preview-22428-01+14c24b2d3 for .NET
  Determining projects to restore...
  Restored D:\projects\repros\consoleBdn\consoleBdn.csproj (in 5 ms).
C:\Program Files\dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(219,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-p
olicy [D:\projects\repros\consoleBdn\consoleBdn.csproj]
  consoleBdn -> D:\projects\repros\consoleBdn\bin\Release\net7.0\consoleBdn.dll

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.29
PS D:\projects\repros\consoleBdn> dotnet D:\projects\repros\consoleBdn\bin\Release\net7.0\consoleBdn.dll
False
```

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -410,11 +408,21 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
...
                 default:
                     throw new NotSupportedException($"Runtime {runtimeId} is not supported");
             }
         }
 
+        private static Job CreateCoreJob(Job baseJob, string runtimeId, string runtimeName, CommandLineOptions options, Runtime runtime)
+        {
+            return baseJob
+                .WithRuntime(runtime)
+                .WithToolchain(CsProjCoreToolchain.From(new NetCoreAppSettings(runtimeId, null, runtimeName, options.CliPath?.FullName, options.RestorePath?.FullName)));
```
adamsitnik, line 423: `CsProjCoreToolchain` uses `DotNetCliBuilder` for building:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs#L38

which performs `dotnet build` to build the benchmark:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliBuilder.cs#L39

and then `DotNetCliExecutor` to run the executable:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs#L39

the executor basically performs `dotnet $pathToLibrary.dll` to run the benchmarks:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliExecutor.cs#L72

You most likely need to define a new toolchain that consists of:
* `CsProjGenerator` same way as `CsProjCoreToolchain` does: https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs#L37
* `DotNetCliPublisher` with `extraArguments` set to `-r $runtimeId`: https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliPublisher.cs#L11
You can take a look at `NatvieAOTToolchain` to see how it provides the runtime id: https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/NativeAot/NativeAotToolchain.cs#L46
* `Executor` that runs provided `.exe` rather than `dotnet $pathTo.dll`: https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/Executor.cs

The tricky part will be to make sure that `ExecutablePath` provided by `CsProjGenerator` is the one that points to the executable build by `dotnet publish`, not the `dll` build by `dotnet build`.

It's used here by the executor:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/Executor.cs#L27

It comes from here:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/GeneratorBase.cs#L114

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs#L52

You might need to introduce a new type that derives from `CsProjGenerator` and does this:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs#L66-L67

src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs
adamsitnik: I agree, `MonoRuntime` would most likely be a better place.

tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
```diff
@@ -383,6 +383,18 @@ public void NetMonikersAreRecognizedAsNetCoreMonikers(string tfm)
...
+        public void MonoMonikersAreRecognizedAsNetCoreMonikers(string tfm, string actualRuntimeId)
+        {
+            var config = ConfigParser.Parse(new[] { "-r", tfm }, new OutputLogger(Output)).config;
+
+            Assert.Single(config.GetJobs());
+            CsProjCoreToolchain toolchain = config.GetJobs().Single().GetToolchain() as CsProjCoreToolchain;
+            Assert.NotNull(toolchain);
+            Assert.Equal(actualRuntimeId, ((DotNetCliGenerator)toolchain.Generator).TargetFrameworkMoniker);
+        }
+
```
adamsitnik, line 397: Please add tests similar to [NativeAotTests](https://github.com/dotnet/BenchmarkDotNet/blob/master/tests/BenchmarkDotNet.IntegrationTests/NativeAotTests.cs) where you ask to run a `Mono .NET 6` benchmark (our CI uses .NET 6 SDK) and the benchmark just throws when it does not run on Mono. Like here where we do that for NativeAOT:

https://github.com/dotnet/BenchmarkDotNet/blob/1f5637c784ad9887a2d5abdba54d1337655281b1/tests/BenchmarkDotNet.IntegrationTests/NativeAotTests.cs#L55-L57

adamsitnik (commented):
@Serg046 I am impressed that you have selected such ambitious task for first contribution!

It looks promising, but we need a new toolchain to make sure it actually runs Mono. PTAL at my comments. Thank you!

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
Serg046: Thanks for very well-detailed explanation. Everything is clear, just few notes you might want to comment on:
> DotNetCliPublisher with extraArguments set to -r $runtimeId

I will do this because I see there is a way to obtain it automatically anyway (so that it is still user-friendly). But your `small repro` returns `true`  with `dotnet publish -c Release -f net6.0 --self-contained -r win-x64 /p:UseMonoRuntime=true` as well as with `dotnet publish -c Release -f net6.0 --self-contained /p:UseMonoRuntime=true` on my local Windows 10.

> Executor that runs provided .exe rather than dotnet $pathTo.dll:

Same thing here. Both `.exe` execution and `dotnet $pathTo.dll` works fine locally.
However, I am not sure what happens on Linux and Mac. I can check on Linux using Docker or WSL if you are interested in, otherwise I will just do what you suggested (just ignore these notes then).

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
timcassell: Ok, I just tested (using Adam's example), and it seems monitoring is still not available on the new Core/Mono runtime. I would assume everything else behaves the same, too (`MonoDisassembler`), but I didn't test.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
adamsitnik: > dotnet publish -c Release -f net6.0 --self-contained /p:UseMonoRuntime=true

This is something new to me, I remember dotnet SDK always complaining about lack of runtime id when running `--self-contained`. But the produced exe is put to the runtime-specific folder, so we might need it anyway:

```log
PS D:\projects\repros\consoleBdn> dotnet publish -c Release -f net7.0 --self-contained /p:UseMonoRuntime=true
MSBuild version 17.4.0-preview-22428-01+14c24b2d3 for .NET
  Determining projects to restore...
  All projects are up-to-date for restore.
C:\Program Files\dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(219,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-p
olicy [D:\projects\repros\consoleBdn\consoleBdn.csproj]
  consoleBdn -> D:\projects\repros\consoleBdn\bin\Release\net7.0\win-x64\consoleBdn.dll
  consoleBdn -> D:\projects\repros\consoleBdn\bin\Release\net7.0\win-x64\publish\
```

> However, I am not sure what happens on Linux and Mac

If you add a test similar to what I've described in https://github.com/dotnet/BenchmarkDotNet/pull/2142#discussion_r992741339 and it works on your Windows machine it should work on Linux and Mac too and it's enough to just use CI to verify that.

src/BenchmarkDotNet/Toolchains/Mono/MonoGenerator.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class MonoGenerator : CsProjGenerator
+    {
+        public MonoGenerator(string targetFrameworkMoniker, string cliPath, string packagesPath, string runtimeFrameworkVersion, bool isNetCore = true) : base(targetFrameworkMoniker, cliPath, packagesPath, runtimeFrameworkVersion, isNetCore)
+        {
+        }
+
+        protected override string GetRuntimeSettings(GcMode gcMode, IResolver resolver)
+        {
+            // Workaround for 'Found multiple publish output files with the same relative path' error
+            return base.GetRuntimeSettings(gcMode, resolver) + "<PropertyGroup><ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles></PropertyGroup>";
```
Serg046, line 16: Same issue https://github.com/dotnet/BenchmarkDotNet/pull/2079

src/BenchmarkDotNet/Toolchains/Mono/MonoPublisher.cs
```diff
@@ -0,0 +1,36 @@
...
+        public BuildResult Build(GenerateResult generateResult, BuildPartition buildPartition, ILogger logger)
+            => new DotNetCliCommand(
+                    CustomDotNetCliPath,
+                    ExtraArguments,
+                    generateResult,
+                    logger,
+                    buildPartition,
+                    EnvironmentVariables,
+                    buildPartition.Timeout)
+                .Publish().ToBuildResult(generateResult);
```
Serg046, line 34: The thing is that `--self-contained` is not supported for restore/build operations that's why I introduce a new publisher and call `publish` only here.

src/BenchmarkDotNet/Toolchains/Mono/MonoToolchain.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        [PublicAPI]
+        public static new IToolchain From(NetCoreAppSettings settings)
+        {
+            var runtimeIdentifier = CustomDotNetCliToolchainBuilder.GetPortableRuntimeIdentifier();
+            return new MonoToolchain(settings.Name,
+                        new MonoGenerator(settings.TargetFrameworkMoniker, settings.CustomDotNetCliPath, settings.PackagesPath, settings.RuntimeFrameworkVersion),
+                        new MonoPublisher(
+                            settings.CustomDotNetCliPath,
+                            $"--self-contained -r {runtimeIdentifier} /p:UseMonoRuntime=true /p:RuntimeIdentifiers={runtimeIdentifier}"),
```
Serg046, line 27: I set RuntimeIdentifiers explicitly here because I had the issue describe [here](https://github.com/dotnet/sdk/issues/10566).

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
Serg046: Thank you @adamsitnik, this is done. Could you please take a look?

tests/BenchmarkDotNet.IntegrationTests/MonoTests.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace BenchmarkDotNet.IntegrationTests
+{
+    public class MonoTests : BenchmarkTestExecutor
+    {
+        [FactDotNetCoreOnly("UseMonoRuntime option is available in .NET Core only starting from .NET 6")]
+        public void Mono60IsSupported()
+        {
+            var config = ManualConfig.CreateEmpty().AddJob(Job.Dry.WithRuntime(MonoRuntime.Mono60));
+            CanExecute<MonoBenchmark>(config);
+        }
```
Serg046, line 17: Not sure if there is a need to add something like
```csharp
if (ContinuousIntegration.IsAppVeyorOnWindows()) // too time consuming for AppVeyor (1h limit)
                return;
```
https://github.com/dotnet/BenchmarkDotNet/blob/master/tests/BenchmarkDotNet.IntegrationTests/NativeAotTests.cs#L24

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -449,6 +456,13 @@ private static Job CreateAotJob(Job baseJob, CommandLineOptions options, Runtime
             return baseJob.WithRuntime(runtime).WithToolchain(builder.ToToolchain());
         }
 
+        private static Job MakeMonoJob(Job baseJob, CommandLineOptions options, MonoRuntime runtime)
+        {
+            return baseJob
+                .WithRuntime(runtime)
+                .WithToolchain(MonoToolchain.From(new NetCoreAppSettings(runtime.MsBuildMoniker, null, runtime.Name, options.CliPath?.FullName, options.RestorePath?.FullName)));
```
adamsitnik, line 463: nit: it would be nice to use named parameters so it's obvious that all strings are passed as right arguments (when they are multiple arguments of the same type it's easy to provide them in wrong order).

You could refactor following logic to a separate method: https://github.com/dotnet/BenchmarkDotNet/blob/c02c3d82bbefa6236660a7b7405f38e739679daf/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs#L534-L539

src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs
```diff
@@ -24,7 +24,7 @@ public class CsProjCoreToolchain : Toolchain, IEquatable<CsProjCoreToolchain>
         [PublicAPI] public static readonly IToolchain NetCoreApp60 = From(NetCoreAppSettings.NetCoreApp60);
         [PublicAPI] public static readonly IToolchain NetCoreApp70 = From(NetCoreAppSettings.NetCoreApp70);
 
-        private CsProjCoreToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath)
+        internal CsProjCoreToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath)
```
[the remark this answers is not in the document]
adamsitnik: @Serg046 making it internal was the right choice! 👍

src/BenchmarkDotNet/Toolchains/Mono/MonoToolchain.cs
adamsitnik: For that it would be great to add a comment to the code and provide the link to the issue (so in the future we don't remove it by accident).

I also wonder if this logic should be moved to `MonoPublisher`, as it's specific for this publisher and it won't ever be reused by other toolchains.

src/BenchmarkDotNet/Toolchains/Mono/MonoGenerator.cs
```diff
@@ -0,0 +1,19 @@
+﻿using BenchmarkDotNet.Characteristics;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Toolchains.CsProj;
+
+namespace BenchmarkDotNet.Toolchains.Mono
+{
+    public class MonoGenerator : CsProjGenerator
+    {
+        public MonoGenerator(string targetFrameworkMoniker, string cliPath, string packagesPath, string runtimeFrameworkVersion, bool isNetCore = true) : base(targetFrameworkMoniker, cliPath, packagesPath, runtimeFrameworkVersion, isNetCore)
```
adamsitnik, line 9: nit: `isNetCore` will always be true for this generator

```suggestion
        public MonoGenerator(string targetFrameworkMoniker, string cliPath, string packagesPath, string runtimeFrameworkVersion) : base(targetFrameworkMoniker, cliPath, packagesPath, runtimeFrameworkVersion, true)
```

src/BenchmarkDotNet/Environments/Runtimes/MonoRuntime.cs
```diff
@@ -15,8 +17,15 @@ public class MonoRuntime : Runtime, IEquatable<MonoRuntime>
 
         public string MonoBclPath { get; }
 
+        public bool IsDotNetBuiltIn { get; set; }
```
adamsitnik, line 20: I am not a big fan of this name, on the other hand I can't suggest a better name... could you please make it internal for now, so we can rename it easily in the future?

```suggestion
        internal bool IsDotNetBuiltIn { get; }
```

tests/BenchmarkDotNet.IntegrationTests/MonoTests.cs
adamsitnik: If AppVeyor CI builds are passing, we don't need it, we can always add it in the future when it becomes a problem,

adamsitnik (commented):
It looks very, very good!

I have one last request: could you please fix the runtime name displayed for the "new" Mono:

image

The code can be found here: https://github.com/dotnet/BenchmarkDotNet/blob/c02c3d82bbefa6236660a7b7405f38e739679daf/src/BenchmarkDotNet/Portability/RuntimeInformation.cs#L164-L166

Thank you for your contribution!

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
Serg046: Done https://github.com/dotnet/BenchmarkDotNet/pull/2142/commits/14a256ef608bbbb228d728747fa50ffa0deb64da

adamsitnik (approved):
Looks great to me @Serg046 ! Very impressive PR, thank you!

PR #2145, opened 2022-10-11
Handle addresses larger than 2GB for 32-bit benchmarks #1469

### LINKED ISSUE #1469: Host exe marked /largeaddressaware
Couldn't find this in the docs

1) Is it possible to have the benchmark host app (the one running my benchmark code) be marked /LargeAddressAware on Windows? I have some benchmarks that process large dump files in a 32 bit space and it would be REALLY great if I could mark the host app as /LargeAddressAware so it wouldn't OOM if I hit the 2 gig cap.

To be clear this isn't an OOM in that my benchmark code is leaking, it is legitimately trying to process a 3 gig dump file, and while it doesn't keep it all resident at once I am trying to benchmark my cache at various cap levels and OOMing if I set it too high, even though such a high setting works fine in my real program.

leonvandermeer:
Fix #1469

It seems that BenchmarkDotNet does not support `Job.Dry.WithRuntime(CoreRuntime.Core60).WithPlatform(Platform.X86)`, so the feature becomes .Net Framework only. See also [ProcessorArchitectureTest](https://github.com/dotnet/BenchmarkDotNet/blob/master/tests/BenchmarkDotNet.IntegrationTests/ProcessorArchitectureTest.cs#L22) which skips this too.

### DIFF
--- a/docs/articles/configs/jobs.md
+++ b/docs/articles/configs/jobs.md
@@ -33,9 +33,9 @@ It's a single string characteristic. It allows to name your job. This name will 
   * `CpuGroups`:  Specifies whether garbage collection supports multiple CPU groups
   * `Force`: Specifies whether the BenchmarkDotNet's benchmark runner forces full garbage collection after each benchmark invocation
   * `AllowVeryLargeObjects`:  On 64-bit platforms, enables arrays that are greater than 2 gigabytes (GB) in total size
-* `LargeAddressAware`: Specifies that BenchmarkDotNet's benchmark runner can handle addresses larger than 2 gigabytes. See also: @BenchmarkDotNet.Samples.IntroLargeAddressAware
-  * `false`: Benchmark runner uses the default (64-bit: enabled; 32-bit:disabled).
-  * `true`: Explicitly specify that BenchmarkDotNet's benchmark runner can handle addresses larger than 2 gigabytes.
+* `LargeAddressAware`: Specifies that benchmark can handle addresses larger than 2 gigabytes. See also: @BenchmarkDotNet.Samples.IntroLargeAddressAware and [LARGEADDRESSAWARE](https://learn.microsoft.com/cpp/build/reference/largeaddressaware-handle-large-addresses)
+  * `false`: Benchmark uses the defaults (64-bit: enabled; 32-bit: disabled).
+  * `true`: Explicitly specify that benchmark can handle addresses larger than 2 gigabytes.
 * `EnvironmentVariables`: customized environment variables for target benchmark. See also: @BenchmarkDotNet.Samples.IntroEnvVars
 
 BenchmarkDotNet will use host process environment characteristics for non specified values.
--- a/docs/articles/configs/jobs.md
+++ b/docs/articles/configs/jobs.md
@@ -33,6 +33,9 @@ It's a single string characteristic. It allows to name your job. This name will 
   * `CpuGroups`:  Specifies whether garbage collection supports multiple CPU groups
   * `Force`: Specifies whether the BenchmarkDotNet's benchmark runner forces full garbage collection after each benchmark invocation
   * `AllowVeryLargeObjects`:  On 64-bit platforms, enables arrays that are greater than 2 gigabytes (GB) in total size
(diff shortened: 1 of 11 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2145) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2145) before we can accept your contribution.<br/><br/>:x: leonvandermeer [sign now](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2145)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/BenchmarkDotNet?pullRequest=2145) it.</sub>

src/BenchmarkDotNet/Jobs/EnvironmentMode.cs
```diff
@@ -96,6 +97,17 @@ public IReadOnlyList<EnvironmentVariable> EnvironmentVariables
...
 
+        /// <summary>
+        /// Specifies that BenchmarkDotNet's benchmark runner can handle addresses larger than 2 gigabytes.
+        /// <value>false: Benchmark runner uses the default (64-bit: enabled; 32-bit:disabled). This is the default.</value>
+        /// <value>true: Explicitly specify that BenchmarkDotNet's benchmark runner can handle addresses larger than 2 gigabytes.</value>
+        /// </summary>
+        public bool LargeAddressAware
+        {
+            get => LargeAddressAwareCharacteristic[this];
+            set => LargeAddressAwareCharacteristic[this] = value;
```
adamsitnik, line 108: This concept is Windows-specific, we should add a guard here.

```suggestion
            set
            {
                if (!RuntimeInformation.IsWindows())
                {
                    throw new NotSupportedException("LargeAddressAware is a Windows-specific concept, it must not be set on other OSes.");
                }

                LargeAddressAwareCharacteristic[this] = value;
            }
```

docs/articles/configs/jobs.md
```diff
@@ -33,6 +33,9 @@ It's a single string characteristic. It allows to name your job. This name will
   * `CpuGroups`:  Specifies whether garbage collection supports multiple CPU groups
   * `Force`: Specifies whether the BenchmarkDotNet's benchmark runner forces full garbage collection after each benchmark invocation
   * `AllowVeryLargeObjects`:  On 64-bit platforms, enables arrays that are greater than 2 gigabytes (GB) in total size
+* `LargeAddressAware`: Specifies that BenchmarkDotNet's benchmark runner can handle addresses larger than 2 gigabytes. See also: @BenchmarkDotNet.Samples.IntroLargeAddressAware
+  * `false`: Benchmark runner uses the default (64-bit: enabled; 32-bit:disabled).
+  * `true`: Explicitly specify that BenchmarkDotNet's benchmark runner can handle addresses larger than 2 gigabytes.
```
adamsitnik, line 38: since this is BenchmarkDotNet docs, there is no need to specify project name or the runner.

```suggestion
* `LargeAddressAware`: Specifies that benchmark can handle addresses larger than 2 gigabytes. See also: @BenchmarkDotNet.Samples.IntroLargeAddressAware and [LARGEADDRESSAWARE](https://learn.microsoft.com/cpp/build/reference/largeaddressaware-handle-large-addresses)
  * `false`: Benchmark uses the defaults (64-bit: enabled; 32-bit: disabled).
  * `true`: Explicitly specify that benchmark can handle addresses larger than 2 gigabytes.
```

Please apply similar changes to the xml doc comments

adamsitnik:
> It seems that BenchmarkDotNet does not support Job.Dry.WithRuntime(CoreRuntime.Core60).WithPlatform(Platform.X86), so the feature becomes .Net Framework only

To run x86 benchmarks using .NET Core, you need to install x86 SDK and pass the path to it in explicit way via `--cli` command line argument, or move it in your $PATH above x64 SDK path.

leonvandermeer:
I finished the rework. I don't know why the AllSetupAndCleanupMethodRunsAsyncGenericTaskSetupTest fails on linux. It seems to be a flaky test.
Please have another look.

adamsitnik (approved):
LGTM, thank you @leonvandermeer !

PR #2148, opened 2022-10-13
adding validation errors when the benchmarks are unsupported

### LINKED ISSUE #989: [Suggestion] add API for detecting cases where toolchain is not supported.
There's no way to detect run failures such as incompatible target platform.

Use case:
```cs
		[Fact]
		public void InProcessCannotRun()
		{
			var otherPlatform = IntPtr.Size == 8
				? Platform.X86
				: Platform.X64;

			var otherPlatformConfig = new ManualConfig()
				.With(Job.Dry.With(InProcessToolchain.Instance).WithInvocationCount(UnrollFactor).WithUnrollFactor(UnrollFactor).With(otherPlatform))
				.With(new OutputLogger(Output))
				.With(DefaultColumnProviders.Instance);

			var runInfo = BenchmarkConverter.TypeToBenchmarks(typeof(BenchmarkAllCases), otherPlatformConfig);
			var summary = BenchmarkRunner.Run(runInfo);

			// summary is empty; no way to detect from code what vent wrong;
			// output is:
			/*
				// Benchmark BenchmarkAllCases.InvokeOnceVoid: Dry(Platform=X86, Toolchain=InProcessToolchain, InvocationCount=16, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=16, WarmupCount=1)
				// cannot be run in-process. Validation errors:
				//    * Job Dry, EnvironmentMode.Platform was run as X64 (X86 expected). Fix your test runner options.
			 */

			// Workaround (not reliable)
			Assert.Empty(summary.BenchmarksCases);

			return;
		}
```

Suggestions:
* call the `IToolchain.IsSupported` method from validator and add the result into summary.ValidationErrors
* add a property that indicates that benchmark completed without errors, `summary.CompletedWithoutErrors` or something like this.

emanuel-v-r:
Aims to fix #989, added a similar test to the one that is described in the issue.
After this PR, the summary will include validation errors and the output will look like this:
```
  Standard Output: 
// Benchmark BenchmarkAllCases.InvokeOnceVoid: Dry(Platform=X86, Toolchain=InProcessToolchain, InvocationCount=16, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=16, WarmupCount=1)
// cannot be run in-process. Validation errors:
//    * Job Dry, EnvironmentMode.Platform was run as X64 (X86 expected). Fix your test runner options.

// Benchmark BenchmarkAllCases.InvokeOnceTaskAsync: Dry(Platform=X86, Toolchain=InProcessToolchain, InvocationCount=16, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=16, WarmupCount=1)
// cannot be run in-process. Validation errors:
//    * Job Dry, EnvironmentMode.Platform was run as X64 (X86 expected). Fix your test runner options.

// Benchmark BenchmarkAllCases.InvokeOnceRefType: Dry(Platform=X86, Toolchain=InProcessToolchain, InvocationCount=16, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=16, WarmupCount=1)
// cannot be run in-process. Validation errors:
//    * Job Dry, EnvironmentMode.Platform was run as X64 (X86 expected). Fix your test runner options.

// Benchmark BenchmarkAllCases.InvokeOnceValueType: Dry(Platform=X86, Toolchain=InProcessToolchain, InvocationCount=16, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=16, WarmupCount=1)
// cannot be run in-process. Validation errors:
//    * Job Dry, EnvironmentMode.Platform was run as X64 (X86 expected). Fix your test runner options.

// Benchmark BenchmarkAllCases.InvokeOnceTaskOfTAsync: Dry(Platform=X86, Toolchain=InProcessToolchain, InvocationCount=16, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=16, WarmupCount=1)
// cannot be run in-process. Validation errors:
//    * Job Dry, EnvironmentMode.Platform was run as X64 (X86 expected). Fix your test runner options.

// Benchmark BenchmarkAllCases.InvokeOnceValueTaskOfT: Dry(Platform=X86, Toolchain=InProcessToolchain, InvocationCount=16, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=16, WarmupCount=1)
// cannot be run in-process. Validation errors:
//    * Job Dry, EnvironmentMode.Platform was run as X64 (X86 expected). Fix your test runner options.

No suitable benchmarks were found
```
Although I believe it fits the purpose of what is described in the original issue, the error is not that specific, if we want to have something more specific we will have to introduce a new method in the IToolChain interface as the existing one returns only a bool https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Toolchains/IToolchain.cs#L16

### DIFF
--- a/src/BenchmarkDotNet/Loggers/LoggerExtensions.cs
+++ b/src/BenchmarkDotNet/Loggers/LoggerExtensions.cs
@@ -20,7 +20,7 @@ namespace BenchmarkDotNet.Loggers
 
         public static void WriteLineHint(this ILogger logger, string text) => logger.WriteLine(LogKind.Hint, text);
 
-        public static void Write(this ILogger logger, string text) =>  logger.Write(LogKind.Default, text);
+        public static void Write(this ILogger logger, string text) => logger.Write(LogKind.Default, text);
 
         [PublicAPI]
         public static void WriteHelp(this ILogger logger, string text) => logger.Write(LogKind.Help, text);
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -47,13 +47,19 @@ namespace BenchmarkDotNet.Running
             {
                 var compositeLogger = CreateCompositeLogger(benchmarkRunInfos, streamLogger);
 
-                var supportedBenchmarks = GetSupportedBenchmarks(benchmarkRunInfos, compositeLogger, resolver);
-                if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))
-                    return new[] { Summary.NothingToRun(title, resultsFolderPath, logFilePath) };
+                compositeLogger.WriteLineInfo("// Validating benchmarks:");
 
-                var validationErrors = Validate(supportedBenchmarks, compositeLogger);
+                var (supportedBenchmarks, validationErrors) = GetSupportedBenchmarks(benchmarkRunInfos, compositeLogger, resolver);
+
+                validationErrors.AddRange(Validate(supportedBenchmarks));
+
+                PrintValidationErrors(compositeLogger, validationErrors);
+
                 if (validationErrors.Any(validationError => validationError.IsCritical))
-                    return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors) };
+                    return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors.ToImmutableArray()) };
+
+                if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))
+                    return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath) };
 
                 int totalBenchmarkCount = supportedBenchmarks.Sum(benchmarkInfo => benchmarkInfo.BenchmarksCases.Length);
                 int benchmarksToRunCount = totalBenchmarkCount;
@@ -145,7 +151,7 @@ namespace BenchmarkDotNet.Running
             var cultureInfo = config.CultureInfo ?? DefaultCultureInfo.Instance;
             var reports = new List<BenchmarkReport>();
             string title = GetTitle(new[] { benchmarkRunInfo });
-            var consoleTitle =  RuntimeInformation.IsWindows() ?  Console.Title : string.Empty;
+            var consoleTitle = RuntimeInformation.IsWindows() ? Console.Title : string.Empty;
 
             logger.WriteLineInfo($"// Found {benchmarks.Length} benchmarks:");
             foreach (var benchmark in benchmarks)
@@ -236,7 +242,7 @@ namespace BenchmarkDotNet.Running
                 logFilePath,
                 runEnd.GetTimeSpan() - runStart.GetTimeSpan(),
                 cultureInfo,
-                Validate(new[] { benchmarkRunInfo }, NullLogger.Instance), // validate them once again, but don't print the output
+                Validate(benchmarkRunInfo), // validate them once again, but don't print the output
                 config.GetColumnHidingRules().ToImmutableArray());
         }
 
@@ -306,10 +312,8 @@ namespace BenchmarkDotNet.Running
             logger.WriteLineHeader("// ***** BenchmarkRunner: End *****");
         }
 
-        private static ImmutableArray<ValidationError> Validate(BenchmarkRunInfo[] benchmarks, ILogger logger)
+        private static ImmutableArray<ValidationError> Validate(params BenchmarkRunInfo[] benchmarks)
         {
-            logger.WriteLineInfo("// Validating benchmarks:");
-
             var validationErrors = new List<ValidationError>();
 
             if (benchmarks.Any(b => b.Config.Options.IsSet(ConfigOptions.JoinSummary)))
@@ -325,9 +329,6 @@ namespace BenchmarkDotNet.Running
 
             foreach (var benchmarkRunInfo in benchmarks)
                 validationErrors.AddRange(benchmarkRunInfo.Config.GetCompositeValidator().Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config)));
-
-            foreach (var validationError in validationErrors.Distinct())
-                logger.WriteLineError(validationError.Message);
 
             return validationErrors.ToImmutableArray();
         }
@@ -531,13 +532,26 @@ namespace BenchmarkDotNet.Running
         private static void LogTotalTime(ILogger logger, TimeSpan time, int executedBenchmarksCount, string message = "Total time")
             => logger.WriteLineStatistic($"{message}: {time.ToFormattedTotalTime(DefaultCultureInfo.Instance)}, executed benchmarks: {executedBenchmarksCount}");
 
-        private static BenchmarkRunInfo[] GetSupportedBenchmarks(BenchmarkRunInfo[] benchmarkRunInfos, ILogger logger, IResolver resolver)
-            => benchmarkRunInfos.Select(info => new BenchmarkRunInfo(
-                    info.BenchmarksCases.Where(benchmark => benchmark.GetToolchain().IsSupported(benchmark, logger, resolver)).ToArray(),
+        private static (BenchmarkRunInfo[], List<ValidationError>) GetSupportedBenchmarks(BenchmarkRunInfo[] benchmarkRunInfos, ILogger logger, IResolver resolver)
+        {
+            List<ValidationError> validationErrors = new ();
+
+            var benchmarksRunInfo = benchmarkRunInfos.Select(info => new BenchmarkRunInfo(
+                    info.BenchmarksCases.Where(benchmark =>
+                    {
+                        var benchmarkValidationErrors = benchmark.GetToolchain().Validate(benchmark, resolver);
+
+                        validationErrors.AddRange(benchmarkValidationErrors);
+
+                        return !benchmarkValidationErrors.Any();
+                    }).ToArray(),
                     info.Type,
                     info.Config))
                 .Where(infos => infos.BenchmarksCases.Any())
                 .ToArray();
+
+            return (benchmarkRunInfos, validationErrors);
+        }
 
         private static string GetRootArtifactsFolderPath(BenchmarkRunInfo[] benchmarkRunInfos)
         {
@@ -651,6 +665,20 @@ namespace BenchmarkDotNet.Running
             double avgSecondsPerBenchmark = executedBenchmarkCount > 0 ? runsChronometer.GetElapsed().GetTimeSpan().TotalSeconds / executedBenchmarkCount : 0;
             TimeSpan fromNow = TimeSpan.FromSeconds(avgSecondsPerBenchmark * benchmarksToRunCount);
             return fromNow;
+        }
+
+        private static void PrintValidationErrors(ILogger logger, IEnumerable<ValidationError> validationErrors)
+        {
+            foreach (var validationError in validationErrors.Distinct())
+            {
+                if (validationError.BenchmarkCase != null)
+                {
+                    logger.WriteLineInfo($"// Benchmark {validationError.BenchmarkCase.DisplayInfo}");
+                }
+
+                logger.WriteLineError($"//    * {validationError.Message}");
+                logger.WriteLine();
+            }
         }
     }
 }
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -539,11 +539,9 @@ namespace BenchmarkDotNet.Running
             var benchmarksRunInfo = benchmarkRunInfos.Select(info => new BenchmarkRunInfo(
                     info.BenchmarksCases.Where(benchmark =>
                     {
-                        var benchmarkValidationErrors = benchmark.GetToolchain().Validate(benchmark, resolver);
-
-                        validationErrors.AddRange(benchmarkValidationErrors);
-
-                        return !benchmarkValidationErrors.Any();
+                        var errors = benchmark.GetToolchain().Validate(benchmark, resolver).ToArray();
+                        validationErrors.AddRange(errors);
+                        return !errors.Any();
                     }).ToArray(),
                     info.Type,
                     info.Config))
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
@@ -136,7 +136,7 @@ namespace BenchmarkDotNet.Running
             catch (InvalidBenchmarkDeclarationException e)
             {
                 ConsoleLogger.Default.WriteLineError(e.Message);
-                return Summary.NothingToRun(e.Message, string.Empty, string.Empty);
+                return Summary.ValidationFailed(e.Message, string.Empty, string.Empty);
             }
         }
 
@@ -149,7 +149,7 @@ namespace BenchmarkDotNet.Running
             catch (InvalidBenchmarkDeclarationException e)
             {
                 ConsoleLogger.Default.WriteLineError(e.Message);
-                return new[] { Summary.NothingToRun(e.Message, string.Empty, string.Empty) };
+                return new[] { Summary.ValidationFailed(e.Message, string.Empty, string.Empty) };
             }
         }
     }
--- a/src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
@@ -1,9 +1,10 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using BenchmarkDotNet.Characteristics;
-using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.DotNetCli;
+using BenchmarkDotNet.Validators;
 
 namespace BenchmarkDotNet.Toolchains.CoreRun
 {
@@ -57,18 +58,17 @@ namespace BenchmarkDotNet.Toolchains.CoreRun
 
         public override string ToString() => Name;
 
-        public bool IsSupported(BenchmarkCase benchmark, ILogger logger, IResolver resolver)
+        public IEnumerable<ValidationError> Validate(BenchmarkCase benchmark, IResolver resolver)
         {
             if (!SourceCoreRun.Exists)
             {
-                logger.WriteLineError($"Provided CoreRun path does not exist, benchmark '{benchmark.DisplayInfo}' will not be executed. Please remember that BDN expects path to CoreRun.exe (corerun on Unix), not to Core_Root folder.");
-                return false;
+                yield return
+                    new ValidationError(true, $"Provided CoreRun path does not exist, benchmark '{benchmark.DisplayInfo}' will not be executed. Please remember that BDN expects path to CoreRun.exe (corerun on Unix), not to Core_Root folder.", benchmark);
             }
-
-            if (Toolchain.InvalidCliPath(CustomDotNetCliPath?.FullName, benchmark, logger))
-                return false;
-
-            return true;
+            else if (Toolchain.IsCliPathInvalid(CustomDotNetCliPath?.FullName, benchmark, out var invalidCliError))
+            {
+                yield return invalidCliError;
+            }
         }
 
         private static FileInfo GetShadowCopyPath(FileInfo coreRunPath)
--- a/src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
@@ -62,8 +62,9 @@ namespace BenchmarkDotNet.Toolchains.CoreRun
         {
             if (!SourceCoreRun.Exists)
             {
-                yield return
-                    new ValidationError(true, $"Provided CoreRun path does not exist, benchmark '{benchmark.DisplayInfo}' will not be executed. Please remember that BDN expects path to CoreRun.exe (corerun on Unix), not to Core_Root folder.", benchmark);
+                yield return new ValidationError(true,
+                    $"Provided CoreRun path does not exist, benchmark '{benchmark.DisplayInfo}' will not be executed. Please remember that BDN expects path to CoreRun.exe (corerun on Unix), not to Core_Root folder.",
+                    benchmark);
             }
             else if (Toolchain.IsCliPathInvalid(CustomDotNetCliPath?.FullName, benchmark, out var invalidCliError))
             {
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
@@ -43,8 +43,9 @@ namespace BenchmarkDotNet.Toolchains.CsProj
 
             if (!RuntimeInformation.IsWindows())
             {
-                yield return
-                    new ValidationError(true, $"Classic .NET toolchain is supported only for Windows, benchmark '{benchmarkCase.DisplayInfo}' will not be executed", benchmarkCase);
+                yield return new ValidationError(true,
+                    $"Classic .NET toolchain is supported only for Windows, benchmark '{benchmarkCase.DisplayInfo}' will not be executed",
+                    benchmarkCase);
             }
             else if (IsCliPathInvalid(customDotNetCliPath: null, benchmarkCase, out var invalidCliError))
             {
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
@@ -1,8 +1,9 @@
-﻿using BenchmarkDotNet.Characteristics;
-using BenchmarkDotNet.Loggers;
+﻿using System.Collections.Generic;
+using BenchmarkDotNet.Characteristics;
 using BenchmarkDotNet.Portability;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.DotNetCli;
+using BenchmarkDotNet.Validators;
 using JetBrains.Annotations;
 
 namespace BenchmarkDotNet.Toolchains.CsProj
@@ -33,21 +34,22 @@ namespace BenchmarkDotNet.Toolchains.CsProj
         public static IToolchain From(string targetFrameworkMoniker, string packagesPath = null)
             => new CsProjClassicNetToolchain(targetFrameworkMoniker, targetFrameworkMoniker, packagesPath);
 
-        public override bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver)
+        public override IEnumerable<ValidationError> Validate(BenchmarkCase benchmarkCase, IResolver resolver)
         {
-            if (!base.IsSupported(benchmarkCase, logger, resolver))
-                return false;
+            foreach (var validationError in base.Validate(benchmarkCase, resolver))
+            {
+                yield return validationError;
+            }
 
             if (!RuntimeInformation.IsWindows())
             {
-                logger.WriteLineError($"Classic .NET toolchain is supported only for Windows, benchmark '{benchmarkCase.DisplayInfo}' will not be executed");
-                return false;
+                yield return
+                    new ValidationError(true, $"Classic .NET toolchain is supported only for Windows, benchmark '{benchmarkCase.DisplayInfo}' will not be executed", benchmarkCase);
             }
-
-            if (InvalidCliPath(customDotNetCliPath: null, benchmarkCase, logger))
-                return false;
-
-            return true;
+            else if (IsCliPathInvalid(customDotNetCliPath: null, benchmarkCase, out var invalidCliError))
+            {
+                yield return invalidCliError;
+            }
         }
     }
 }
--- a/src/BenchmarkDotNet/Toolchains/IToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/IToolchain.cs
@@ -1,6 +1,7 @@
-﻿using BenchmarkDotNet.Characteristics;
-using BenchmarkDotNet.Loggers;
+﻿using System.Collections.Generic;
+using BenchmarkDotNet.Characteristics;
 using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Validators;
 using JetBrains.Annotations;
 
 namespace BenchmarkDotNet.Toolchains
@@ -13,6 +14,6 @@ namespace BenchmarkDotNet.Toolchains
         IExecutor Executor { get; }
         bool IsInProcess { get; }
 
-        bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
+        IEnumerable<ValidationError> Validate(BenchmarkCase benchmarkCase, IResolver resolver);
     }
 }
--- a/src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
@@ -26,28 +26,33 @@ namespace BenchmarkDotNet.Toolchains.Mono
                 yield return validationError;
             }
 
-            if (!benchmarkCase.Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic) || !(benchmarkCase.Job.Environment.Runtime is MonoRuntime))
+            if (!benchmarkCase.Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic) || benchmarkCase.Job.Environment.Runtime is not MonoRuntime)
             {
-                yield return
-                    new ValidationError(true, "The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class", benchmarkCase);
+                yield return new ValidationError(true,
+                    "The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class",
+                    benchmarkCase);
             }
 
             if ((benchmarkCase.Job.Environment.Runtime is MonoRuntime monoRuntime) && !string.IsNullOrEmpty(monoRuntime.MonoBclPath) && !Directory.Exists(monoRuntime.MonoBclPath))
             {
-                yield return
-                    new ValidationError(true, $"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.", benchmarkCase);
+                yield return new ValidationError(true,
+                    $"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.",
+                    benchmarkCase);
             }
 
             if (benchmarkCase.Job.HasValue(InfrastructureMode.BuildConfigurationCharacteristic)
                 && benchmarkCase.Job.ResolveValue(InfrastructureMode.BuildConfigurationCharacteristic, resolver) != InfrastructureMode.ReleaseConfigurationName)
             {
-                yield return
-                    new ValidationError(true, "The MonoAOT toolchain does not allow to rebuild source project, so defining custom build configuration makes no sense", benchmarkCase);
+                yield return new ValidationError(true,
+                    "The MonoAOT toolchain does not allow to rebuild source project, so defining custom build configuration makes no sense",
+                    benchmarkCase);
             }
 
             if (benchmarkCase.Job.HasValue(InfrastructureMode.NuGetReferencesCharacteristic))
             {
-                yield return new ValidationError(true, "The MonoAOT toolchain does not allow specifying NuGet package dependencies", benchmarkCase);
+                yield return new ValidationError(true,
+                    "The MonoAOT toolchain does not allow specifying NuGet package dependencies",
+                    benchmarkCase);
             }
         }
     }
--- a/src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
@@ -1,11 +1,12 @@
-﻿using BenchmarkDotNet.Characteristics;
+﻿using System.Collections.Generic;
+using System.IO;
+using BenchmarkDotNet.Characteristics;
 using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Jobs;
-using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.Roslyn;
+using BenchmarkDotNet.Validators;
 using JetBrains.Annotations;
-using System.IO;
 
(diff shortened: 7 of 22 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2148) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2148) before we can accept your contribution.<br/><br/>:x: emanuel-v-r [sign now](https://cla.dotnetfoundation.org/dotnet/BenchmarkDotNet?pullRequest=2148)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/BenchmarkDotNet?pullRequest=2148) it.</sub>

YegorStepanov:
Yep, we need a specific error message in `summary.ValidationErrors`.

The problem is that the benchmarks are filtered by `Toolchain.IsSupported` before the validation executes.

```C#
var supportedBenchmarks = GetSupportedBenchmarks(benchmarkRunInfos, compositeLogger, resolver); //filtering by Toolchain.IsSupported
if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))
    return new[] { Summary.NothingToRun(title, resultsFolderPath, logFilePath) };

var validationErrors = Validate(supportedBenchmarks, compositeLogger);
if (validationErrors.Any(validationError => validationError.IsCritical))
    return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors) };
```

---

> if we want to have something more specific we will have to introduce a new method in the IToolChain interface as the existing one returns only a bool https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Toolchains/IToolchain.cs#L16

If we replace `ILogger logger` with `List<ValidationError> validationErrors`, we can do smth like this:
```C#
var supportedBenchmarks = GetSupportedBenchmarks(benchmarkRunInfos, resolver, out List<ValidationErrors> validationErrors);
validationErrors.AddRange(Validate(supportedBenchmarks, compositeLogger));

if (validationErrors.Any(validationError => validationError.IsCritical))
    return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors) };

if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))  // looks redundant, validators should check it
    return new[] { Summary.NothingToRun(title, resultsFolderPath, logFilePath) };
```

---

For the maintainers: 
Validators are pretty bad at the moment, they need to be improved in some places. Someday I will write a "plan" issue. But the task is quite big.

emanuel-v-r:
> Yep, we need a specific error message in `summary.ValidationErrors`.
> 
> The problem is that the benchmarks are filtered by `Toolchain.IsSupported` before the validation executes.
> 
> ```cs
> var supportedBenchmarks = GetSupportedBenchmarks(benchmarkRunInfos, compositeLogger, resolver); //filtering by Toolchain.IsSupported
> if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))
>     return new[] { Summary.NothingToRun(title, resultsFolderPath, logFilePath) };
> 
> var validationErrors = Validate(supportedBenchmarks, compositeLogger);
> if (validationErrors.Any(validationError => validationError.IsCritical))
>     return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors) };
> ```
> 
> > if we want to have something more specific we will have to introduce a new method in the IToolChain interface as the existing one returns only a bool https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Toolchains/IToolchain.cs#L16
> 
> If we replace `ILogger logger` with `List<ValidationError> validationErrors`, we can do smth like this:
> 
> ```cs
> var supportedBenchmarks = GetSupportedBenchmarks(benchmarkRunInfos, resolver, out List<ValidationErrors> validationErrors);
> validationErrors.AddRange(Validate(supportedBenchmarks, compositeLogger));
> 
> if (validationErrors.Any(validationError => validationError.IsCritical))
>     return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors) };
> 
> if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))  // looks redundant, validators should check it
>     return new[] { Summary.NothingToRun(title, resultsFolderPath, logFilePath) };
> ```
> 
> For the maintainers: Validators are pretty bad at the moment, they need to be improved in some places. Someday I will write a "plan" issue. But the task is quite big.



> Yep, we need a specific error message in `summary.ValidationErrors`.
> 
> The problem is that the benchmarks are filtered by `Toolchain.IsSupported` before the validation executes.
> 
> ```cs
> var supportedBenchmarks = GetSupportedBenchmarks(benchmarkRunInfos, compositeLogger, resolver); //filtering by Toolchain.IsSupported
> if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))
>     return new[] { Summary.NothingToRun(title, resultsFolderPath, logFilePath) };
> 
> var validationErrors = Validate(supportedBenchmarks, compositeLogger);
> if (validationErrors.Any(validationError => validationError.IsCritical))
>     return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors) };
> ```
> 
> > if we want to have something more specific we will have to introduce a new method in the IToolChain interface as the existing one returns only a bool https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Toolchains/IToolchain.cs#L16
> 
> If we replace `ILogger logger` with `List<ValidationError> validationErrors`, we can do smth like this:
> 
> ```cs
> var supportedBenchmarks = GetSupportedBenchmarks(benchmarkRunInfos, resolver, out List<ValidationErrors> validationErrors);
> validationErrors.AddRange(Validate(supportedBenchmarks, compositeLogger));
> 
> if (validationErrors.Any(validationError => validationError.IsCritical))
>     return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath, validationErrors) };
> 
> if (!supportedBenchmarks.Any(benchmarks => benchmarks.BenchmarksCases.Any()))  // looks redundant, validators should check it
>     return new[] { Summary.NothingToRun(title, resultsFolderPath, logFilePath) };
> ```
> 
> For the maintainers: Validators are pretty bad at the moment, they need to be improved in some places. Someday I will write a "plan" issue. But the task is quite big.

Thanks for the feedback, changed accordingly here https://github.com/dotnet/BenchmarkDotNet/pull/2148/commits/524dbf9f4ab3ae90a725001d9598f78301e2c258.
Notes:
* I haven't replaced logger because it's being used int the toolchain methods (ex: https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Toolchains/InProcess/InProcessValidator.cs#L107), seems that the lib needs some more generic approach for validations/errors, so that we don't need to span "writeline" all over the place
* I have kept the original "IsSupported" untouched, so that we don't break any contract

Please let me know if that works for you

tests/BenchmarkDotNet.IntegrationTests/InProcessTest.cs
```diff
@@ -56,10 +57,28 @@ public InProcessTest(ITestOutputHelper output) : base(output)
         [Fact]
         public void BenchmarkActionValueTaskOfTSupported() => TestInvoke(x => x.InvokeOnceValueTaskOfT(), UnrollFactor, DecimalResult);
 
+        [Fact]
+        public void BenchmarkDifferentPlatformReturnsValidationError()
+        {
+            var otherPlatform = IntPtr.Size == 8
+                ? Platform.X86
+                : Platform.X64;
```
adamsitnik, line 65: this is really clever, thank you for adding the test! 👍

tests/BenchmarkDotNet.IntegrationTests/InProcessTest.cs
```diff
@@ -56,10 +57,28 @@ public InProcessTest(ITestOutputHelper output) : base(output)
...
+
+            var otherPlatformConfig = new ManualConfig()
+                .With(Job.Dry.With(InProcessToolchain.Instance).WithInvocationCount(UnrollFactor).WithUnrollFactor(UnrollFactor).With(otherPlatform))
```
adamsitnik, line 68: nit: `Job.Dry` already defines invocation count and unroll factor (1)

```suggestion
                .With(Job.Dry.With(InProcessToolchain.Instance).With(otherPlatform))
```

src/BenchmarkDotNet/Toolchains/IToolchain.cs
```diff
@@ -14,5 +17,6 @@ public interface IToolchain
         bool IsInProcess { get; }
 
         bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
+        IEnumerable<ValidationError> CheckSupport(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
```
adamsitnik, line 20: since this method returns `IEnumerable<ValidationError>`, it  could be called `Validate`. Moreover since we should change its responsibility to just return the errors (not report them), and don't pass `ILogger` instance to it:


```suggestion
        IEnumerable<ValidationError> Validate(BenchmarkCase benchmarkCase, IResolver resolver);
```

src/BenchmarkDotNet/Toolchains/IToolchain.cs
```diff
@@ -14,5 +17,6 @@ public interface IToolchain
         bool IsInProcess { get; }
 
         bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
```
adamsitnik, line 19: Is `IsSupported` still needed? We should consider removing it.

src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
```diff
@@ -58,17 +61,27 @@ public class CoreRunToolchain : IToolchain
...
         {
             if (!SourceCoreRun.Exists)
             {
-                logger.WriteLineError($"Provided CoreRun path does not exist, benchmark '{benchmark.DisplayInfo}' will not be executed. Please remember that BDN expects path to CoreRun.exe (corerun on Unix), not to Core_Root folder.");
-                return false;
+                var validationError = new ValidationError(true, $"Provided CoreRun path does not exist, benchmark '{benchmark.DisplayInfo}' will not be executed. Please remember that BDN expects path to CoreRun.exe (corerun on Unix), not to Core_Root folder.", benchmark);
+
+                logger.WriteLineError(validationError.Message);
+
+                return new List<ValidationError> { validationError };
```
adamsitnik, line 74: we can take advantage of yield return/break and simplify the code a bit:

```suggestion
                yield return new ValidationError(true, $"Provided CoreRun path does not exist, benchmark '{benchmark.DisplayInfo}' will not be executed. Please remember that BDN expects path to CoreRun.exe (corerun on Unix), not to Core_Root folder.", benchmark);
                yield break;
```

src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
```diff
@@ -58,17 +61,27 @@ public class CoreRunToolchain : IToolchain
...
             }
 
             if (Toolchain.InvalidCliPath(CustomDotNetCliPath?.FullName, benchmark, logger))
-                return false;
+            {
+                var validationError = new ValidationError(true, "Invalid Cli Path", benchmark);
+
+                return new List<ValidationError> { validationError };
```
adamsitnik, line 81:
```suggestion
                yield return new ValidationError(true, "Invalid Cli Path", benchmark);
```

src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
```diff
@@ -58,17 +61,27 @@ public class CoreRunToolchain : IToolchain
...
+            }
 
-            return true;
+            return Enumerable.Empty<ValidationError>();
```
adamsitnik, line 84:
```suggestion
```

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -531,13 +532,26 @@ private static BenchmarkReport RunCore(BenchmarkCase benchmarkCase, BenchmarkId
         private static void LogTotalTime(ILogger logger, TimeSpan time, int executedBenchmarksCount, string message = "Total time")
             => logger.WriteLineStatistic($"{message}: {time.ToFormattedTotalTime(DefaultCultureInfo.Instance)}, executed benchmarks: {executedBenchmarksCount}");
 
-        private static BenchmarkRunInfo[] GetSupportedBenchmarks(BenchmarkRunInfo[] benchmarkRunInfos, ILogger logger, IResolver resolver)
-            => benchmarkRunInfos.Select(info => new BenchmarkRunInfo(
-                    info.BenchmarksCases.Where(benchmark => benchmark.GetToolchain().IsSupported(benchmark, logger, resolver)).ToArray(),
+        private static BenchmarkRunInfo[] GetSupportedBenchmarks(BenchmarkRunInfo[] benchmarkRunInfos, ILogger logger, IResolver resolver, out List<ValidationError> validationErrors)
```
adamsitnik, line 535: inside `BenchmarkRunnerClean` it's OK to use `ValueTuples`, we could use them here as well:

```suggestion
        private static (BenchmarkRunInfo[] benchmarks, IReadOnlyList<ValidationError> validationErrors) GetSupportedBenchmarks(BenchmarkRunInfo[] benchmarkRunInfos, ILogger logger, IResolver resolver)
```

adamsitnik (commented):
@emanuel-v-r big thanks for your contribution! PTAL at my comments

YegorStepanov:
I guess this test won't contain `validationError`:
```C#
[Fact]
public void BenchmarkDifferentPlatformReturnsValidationError()
{
    var сonfig = new ManualConfig()
        .With(Job.Dry.With(InProcessToolchain.Instance).With(Platform.X86))
        .With(Job.Dry.With(InProcessToolchain.Instance).With(Platform.X64))
        .With(new OutputLogger(Output))
        .With(DefaultColumnProviders.Instance);

    var runInfo = BenchmarkConverter.TypeToBenchmarks(typeof(BenchmarkAllCases), сonfig);
    var summary = BenchmarkRunner.Run(runInfo);

    Assert.NotEmpty(summary.ValidationErrors);
}
```
Also, we need to add a specific reason to the validation errors.

To fix it we need change interface to it 
```diff
public interface IToolchain                                                            
{                                                                                     
    [PublicAPI] string Name { get; }                                                   
    IGenerator Generator { get; }                                                      
    IBuilder Builder { get; }                                                          
    IExecutor Executor { get; }                                                        
    bool IsInProcess { get; }                                                          
                                                                                      
-   bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
+   bool IsSupported(BenchmarkCase benchmarkCase, IResolver resolver, List<ValidationError> validationErrors);
}                                                                                     
```

and do it for each toolchain:
```diff
- logger.WriteLineError("The Roslyn toolchain is only supported on .NET Framework");
+ validationErrors.Add(new ValidationError(false, "The Roslyn toolchain is only supported on .NET Framework"));
```

But:
1) We need approval from the maintainers to do it.
2) Refactoring is desirable here, but it requires an understanding of the project's code
2.1) For example: change `List<ValidationError>` to the add-only collection `ValidationErrors` in the whole project.

src/BenchmarkDotNet/Toolchains/IToolchain.cs
emanuel-v-r: Actually the responsibility remains the same, but it returns an IEnumerable<ValidationError> instead of just a bool, so that the client knows specifically what happened.
Removing the logger means that we would have to create some generic approach for "printing" toolchain information, as the logger is used inside the toolchain implementations, in places like this https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Toolchains/InProcess/InProcessValidator.cs#L107.
Renamed to Validate

src/BenchmarkDotNet/Toolchains/IToolchain.cs
emanuel-v-r: Yes, still used

src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
emanuel-v-r: Do you think it's worth it? Seems that it requires the same amount of code. I tend to avoid the yield keyword for finite collections, as it uses a state machine under the hood.

emanuel-v-r:
> I guess this test won't contain `validationError`:
> 
> ```cs
> [Fact]
> public void BenchmarkDifferentPlatformReturnsValidationError()
> {
>     var сonfig = new ManualConfig()
>         .With(Job.Dry.With(InProcessToolchain.Instance).With(Platform.X86))
>         .With(Job.Dry.With(InProcessToolchain.Instance).With(Platform.X64))
>         .With(new OutputLogger(Output))
>         .With(DefaultColumnProviders.Instance);
> 
>     var runInfo = BenchmarkConverter.TypeToBenchmarks(typeof(BenchmarkAllCases), сonfig);
>     var summary = BenchmarkRunner.Run(runInfo);
> 
>     Assert.NotEmpty(summary.ValidationErrors);
> }
> ```
> 
> Also, we need to add a specific reason to the validation errors.
> 
> To fix it we need change interface to it
> 
> ```diff
> public interface IToolchain                                                            
> {                                                                                     
>     [PublicAPI] string Name { get; }                                                   
>     IGenerator Generator { get; }                                                      
>     IBuilder Builder { get; }                                                          
>     IExecutor Executor { get; }                                                        
>     bool IsInProcess { get; }                                                          
>                                                                                       
> -   bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
> +   bool IsSupported(BenchmarkCase benchmarkCase, IResolver resolver, List<ValidationError> validationErrors);
> }                                                                                     
> ```
> 
> and do it for each toolchain:
> 
> ```diff
> - logger.WriteLineError("The Roslyn toolchain is only supported on .NET Framework");
> + validationErrors.Add(new ValidationError(false, "The Roslyn toolchain is only supported on .NET Framework"));
> ```
> 
> But:
> 
> 1. We need approval from the maintainers to do it.
> 2. Refactoring is desirable here, but it requires an understanding of the project's code
>    2.1) For example: change `List<ValidationError>` to the add-only collection `ValidationErrors` in the whole project.

I was not able to replicate the issue using your test but I do understand what the issue is, nice catch thank you!
This commit should do the trick https://github.com/dotnet/BenchmarkDotNet/pull/2148/commits/c45077d5b04a0751f990559c5f39ee2166166dff.
Also @adamsitnik thanks for your feedback, I addressed most of your comments, but I kept open some of them that I still have some doubts.
More specifically this one https://github.com/dotnet/BenchmarkDotNet/pull/2148#discussion_r995790537.
Seems that we all agree that this validate method should not log/print anything, and therefore the logger should be removed.
The problem is that inside the toolchain implementations, we do have some logging/printing, and there's no existing generic approach for such error handling.
My suggestion is that we create a separated issue for handling this, as it could take some effort.
Anyway the changes that I have done in this PR will not affect such behavior.

src/BenchmarkDotNet/Loggers/LoggerExtensions.cs
```diff
@@ -20,7 +21,9 @@ public static class LoggerExtensions
 
         public static void WriteLineHint(this ILogger logger, string text) => logger.WriteLine(LogKind.Hint, text);
 
-        public static void Write(this ILogger logger, string text) =>  logger.Write(LogKind.Default, text);
+        public static void Write(this ILogger logger, string text) => logger.Write(LogKind.Default, text);
+
+        public static void WriteLineValidationError(this ILogger logger, ValidationError validationError) => logger.WriteLineError(validationError.Message);
```
YegorStepanov, line 26: It's redundant since it will only be used in one place (when you remove `Logger` from `Validate`), it's best not to duplicate methods.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
```diff
@@ -33,21 +36,34 @@ private CsProjClassicNetToolchain(string targetFrameworkMoniker, string name, st
...
+
+                logger.WriteLineValidationError(validationError);
+
+                return validationError.AsEnumerable();
             }
 
             if (InvalidCliPath(customDotNetCliPath: null, benchmarkCase, logger))
-                return false;
+            {
+                var validationError = new ValidationError(true, $"InvalidCliPath, benchmark '{benchmarkCase.DisplayInfo}' will not be executed", benchmarkCase);
```
YegorStepanov, line 59: Do not use fast exit, collect all errors and return them.

It's better to display as much as possible errors.
If the user is not on windows AND the cli path is invalid, it displays only the first message.

src/BenchmarkDotNet/Validators/ValidationError.cs
```diff
@@ -50,6 +51,8 @@ public override int GetHashCode()
             }
         }
 
+        internal IEnumerable<ValidationError> AsEnumerable() => new List<ValidationError>() { this };
```
YegorStepanov, line 54: If you remove "fast-exit" (as I suggested above), it will be redundant.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
```diff
@@ -33,21 +36,34 @@ private CsProjClassicNetToolchain(string targetFrameworkMoniker, string name, st
...
-                logger.WriteLineError($"Classic .NET toolchain is supported only for Windows, benchmark '{benchmarkCase.DisplayInfo}' will not be executed");
-                return false;
+                var validationError = new ValidationError(true, $"Classic .NET toolchain is supported only for Windows, benchmark '{benchmarkCase.DisplayInfo}' will not be executed", benchmarkCase);
+
+                logger.WriteLineValidationError(validationError);
+
+                return validationError.AsEnumerable();
             }
 
             if (InvalidCliPath(customDotNetCliPath: null, benchmarkCase, logger))
```
YegorStepanov, line 57: You should refactor InvalidCLIPath and return validation errors from it. After it you will can delete logger

src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
emanuel-v-r: I am not sure if we want to check all the errors, or just fail fast here, anyway I kep the existing behavior, not sure if we want to change that.

YegorStepanov: You can open the validators folder, all of them may display multiple errors.

src/BenchmarkDotNet/Loggers/LoggerExtensions.cs
YegorStepanov: @emanuel-v-r When the `Toolchain.InvalidCliPath(...) == false` you don't return actual error, so it's in the scope 😊

The new version of BDN should be out soon (1-2 weeks).
But it looks unfinished for the release (IMO)

```С#
public interface IToolchain
{
    bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
    IEnumerable<ValidationError> Validate(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver);
}
```

```C#
var validationError = new ValidationError(true, "...", benchmarkCase);
logger.WriteLineValidationError(validationError);
return validationError.AsEnumerable();
```

And it would be cool, if you add a few tests that hits, for example, `Toolchain.InvalidCliPath` path.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
emanuel-v-r: > You can open the validators folder, all of them may display multiple errors.

I mean toolchains errors, currently it only prints one error and exits as you can see in the previous code for "IsSupported" method
I am ok to change it as well, but I am trying to be objective here, and change only the necessary for what is described in the issue, and try not have side effects
Maybe @adamsitnik can provide his opinion here.

src/BenchmarkDotNet/Loggers/LoggerExtensions.cs
emanuel-v-r: I am talking about removing logger dependencies from the `Toolchain` implementations.
As you mentioned `Toolchain.InvalidCliPath` is in scope, and it was addressed in scope of your other comment https://github.com/dotnet/BenchmarkDotNet/pull/2148#discussion_r995925947.
Here's the commit https://github.com/dotnet/BenchmarkDotNet/pull/2148/commits/40ae83c481bca11bc5855852ee7e203eb87f7c7a

src/BenchmarkDotNet/Validators/ValidationError.cs
adamsitnik, line 54: this method should not be needed as we can just use `yield` everywhere. Also, if we really needed that it should be returning an array (with the list the caller can cast it to list and add some elements) + array is cheaper than list `new [1] { this }` 

```suggestion
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +21,45 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
         {
         }
 
-        public override bool IsSupported(BenchmarkCase benchmarkCase, ILogger logger, IResolver resolver)
+        public override IEnumerable<ValidationError> Validate(BenchmarkCase benchmarkCase, IResolver resolver)
         {
-            if (!base.IsSupported(benchmarkCase, logger, resolver))
+            var baseValidations = base.Validate(benchmarkCase, resolver);
+
+            if (baseValidations.Any())
```
adamsitnik, line 28:
```suggestion
            foreach (var validationError in base.Validate(benchmarkCase, resolver))
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +21,45 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
...
             {
-                return false;
+                return baseValidations;
```
adamsitnik, line 30:
```suggestion
                yield return validationError;
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +21,45 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
...
             }
 
             if (!benchmarkCase.Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic) || !(benchmarkCase.Job.Environment.Runtime is MonoRuntime))
             {
-                logger.WriteLineError("The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class");
-                return false;
+                return
+                    new ValidationError(true, "The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class", benchmarkCase)
+                    .AsEnumerable();
```
adamsitnik, line 37:
```suggestion
                yield return
                    new ValidationError(true, "The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class", benchmarkCase);
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +21,45 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
...
             }
 
             if ((benchmarkCase.Job.Environment.Runtime is MonoRuntime monoRuntime) && !string.IsNullOrEmpty(monoRuntime.MonoBclPath) && !Directory.Exists(monoRuntime.MonoBclPath))
             {
-                logger.WriteLineError($"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.");
-                return false;
+                return
+                    new ValidationError(true, $"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.", benchmarkCase)
+                    .AsEnumerable();
```
adamsitnik, line 44:
```suggestion
                yield return
                    new ValidationError(true, $"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.", benchmarkCase);
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +21,45 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
...
             }
 
             if (benchmarkCase.Job.HasValue(InfrastructureMode.BuildConfigurationCharacteristic)
                 && benchmarkCase.Job.ResolveValue(InfrastructureMode.BuildConfigurationCharacteristic, resolver) != InfrastructureMode.ReleaseConfigurationName)
             {
-                logger.WriteLineError("The MonoAOT toolchain does not allow to rebuild source project, so defining custom build configuration makes no sense");
-                return false;
+                return
+                    new ValidationError(true, "The MonoAOT toolchain does not allow to rebuild source project, so defining custom build configuration makes no sense", benchmarkCase)
+                    .AsEnumerable();
```
adamsitnik, line 52:
```suggestion
                yield return
                    new ValidationError(true, "The MonoAOT toolchain does not allow to rebuild source project, so defining custom build configuration makes no sense", benchmarkCase);
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +21,45 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
...
             }
 
             if (benchmarkCase.Job.HasValue(InfrastructureMode.NuGetReferencesCharacteristic))
             {
-                logger.WriteLineError("The MonoAOT toolchain does not allow specifying NuGet package dependencies");
-                return false;
+                var validationError = new ValidationError(true, "The MonoAOT toolchain does not allow specifying NuGet package dependencies", benchmarkCase);
+
+                return validationError.AsEnumerable();
```
adamsitnik, line 59:
```suggestion
                yield return new ValidationError(true, "The MonoAOT toolchain does not allow specifying NuGet package dependencies", benchmarkCase);
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +21,45 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
...
             }
 
-            return true;
+            return Enumerable.Empty<ValidationError>();
```
adamsitnik, line 62:
```suggestion
```

src/BenchmarkDotNet/Toolchains/CoreRun/CoreRunToolchain.cs
adamsitnik: I do. It simplifies the code + by returning `List` we return a mutable collection, which can be misused by some callers:

```cs
IEnumerable<ValidationError> collection = Validate(benchmark, resolver);
((List<ValidationError>)collection).Add(new ...)
```

In places where we know the size of the collection and it's fixed we can use arrays instead of `yield` or even `ImmutableArray` to make sure that the content never changes.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjClassicNetToolchain.cs
adamsitnik: I like the idea of returning all errors instead the first one. 👍 

Currently if there are two or more errors:

1. The users tries to run the benchmarks, gets a single error and fixes it.
2. The users tries to run the benchmarks, gets another error and fixes it.
3. The user actually runs the benchmarks.

If we provide all errors at once we can reduce the number of steps needed and hence improve the UX.

adamsitnik (commented):
Overall, it looks very good, but it would be great if you could change the validator behavior to return all known errors when possible. I know that the original issue did not mention it, but since you are improving the error handling, why not make it even better when we can?

Thank you @emanuel-v-r !

src/BenchmarkDotNet/Toolchains/Toolchain.cs
```diff
@@ -29,39 +29,34 @@ public Toolchain(string name, IGenerator generator, IBuilder builder, IExecutor
...
+                    yield return
+                        new ValidationError(true, $"We could not find Mono in provided path ({mono.CustomPath}), benchmark '{benchmarkCase.DisplayInfo}' will not be executed", benchmarkCase);
                 }
             }
 
-            return Enumerable.Empty<ValidationError>();
         }
 
-        internal static bool InvalidCliPath(string customDotNetCliPath, BenchmarkCase benchmarkCase, out ValidationError validationError)
+        internal static bool IsCliPathInvalid(string customDotNetCliPath, BenchmarkCase benchmarkCase, out ValidationError validationError)
```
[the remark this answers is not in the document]
YegorStepanov: Hmm, `IsCliPathInvalid` returns one of two validation errors always. You can leave the current version.

emanuel-v-r:
> Overall, it looks very good, but it would be great if you could change the validator behavior to return all known errors when possible. I know that the original issue did not mention it, but since you are improving the error handling, why not make it even better when we can?
> 
> Thank you @emanuel-v-r !

Thank you Adam, applied your suggestions, and also changed the validate method in the other toolchains so that it returns multiple erros.

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
```diff
@@ -136,7 +136,7 @@ private static Summary RunWithExceptionHandling(Func<Summary> run)
             catch (InvalidBenchmarkDeclarationException e)
             {
                 ConsoleLogger.Default.WriteLineError(e.Message);
-                return Summary.NothingToRun(e.Message, string.Empty, string.Empty);
+                return Summary.ValidationFailed(e.Message, string.Empty, string.Empty);
```
YegorStepanov, line 139: Looks a little weird here. Maybe find a better name?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -531,14 +532,27 @@ private static BenchmarkReport RunCore(BenchmarkCase benchmarkCase, BenchmarkId
...
+            List<ValidationError> validationErrors = new ();
+
+            var benchmarksRunInfo = benchmarkRunInfos.Select(info => new BenchmarkRunInfo(
+                    info.BenchmarksCases.Where(benchmark =>
+                    {
+                        var benchmarkValidationErrors = benchmark.GetToolchain().Validate(benchmark, resolver);
+
+                        validationErrors.AddRange(benchmarkValidationErrors);
+
+                        return !benchmarkValidationErrors.Any();
```
YegorStepanov, line 546: "the more nested the type, the shorter its name" 

Bug: it will execute `Validation` twice.


```suggestion
                        var errors = benchmark.GetToolchain().Validate(benchmark, resolver).ToArray();
                        validationErrors.AddRange(errors);
                        return !errors.Any();
```

src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
emanuel-v-r: Any suggestion? Maybe just Summary.Failed?

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -18,39 +19,36 @@ public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(
...
+                yield return
+                    new ValidationError(true, "The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class", benchmarkCase);
             }
 
             if ((benchmarkCase.Job.Environment.Runtime is MonoRuntime monoRuntime) && !string.IsNullOrEmpty(monoRuntime.MonoBclPath) && !Directory.Exists(monoRuntime.MonoBclPath))
             {
-                logger.WriteLineError($"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.");
-                return false;
+                yield return
+                    new ValidationError(true, $"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.", benchmarkCase);
```
YegorStepanov, line 38:
```suggestion
                yield return new ValidationError(true, $"The MonoBclPath provided for MonoAOT toolchain: {monoRuntime.MonoBclPath} does NOT exist.", benchmarkCase);
```

emanuel-v-r: It was actually me, I do prefer this way, to avoid such a long line.
My IDE does not complain about it as well...

YegorStepanov: It not critical, but the project uses 2 types of formatting for validation errors:
```C#
//#1
yield return new ValidationError(true, "The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class", benchmarkCase);
//#2
yield return new ValidationError(true,
    "The MonoAOT toolchain requires the Runtime property to be configured explicitly to an instance of MonoRuntime class", 
    benchmarkCase);
```

The perfectionist is crying 😊
image

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolchain.cs
```diff
@@ -1,40 +1,46 @@
...
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.DotNetCli;
+using BenchmarkDotNet.Validators;
+using JetBrains.Annotations;
 
 namespace BenchmarkDotNet.Toolchains.MonoWasm
 {
     [PublicAPI]
-    public class WasmToolChain : Toolchain
+    public class WasmToolchain : Toolchain
```
adamsitnik, line 12: Renaming public types is considered to be a breaking change and we avoid doing that if we can. However, this particular type is most likely not used directly by anyone, so it's OK-ish ;)

src/BenchmarkDotNet/Validators/ValidationError.cs
```diff
@@ -50,6 +51,7 @@ public override int GetHashCode()
             }
         }
 
+
```
adamsitnik, line 54: nit: redundant empty lone

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
YegorStepanov: > Than this: image

@emanuel-v-r FYI:
Visual Studio split it by 2 lines due to [soft wrap mode](https://learn.microsoft.com/en-gb/visualstudio/ide/reference/how-to-manage-word-wrap-in-the-editor?view=vs-2022
). You can toggle it

PR #2151, opened 2022-10-16
Fix user input matching

### LINKED ISSUE #1709: BenchmarkSwitcher executes all benchmarks that share a base class
I have a large set of benchmarks that have common initialization. To avoid having duplicate code I created a base class with the BenchmarkDotNet initialization. Here's a small example that replicates my scenario:

``` csharp
using System.Linq;

namespace TestBenchmarkDotNet
{
    public abstract class TestsBase
    {
        protected int[] source;
        
        [Params(10)]
        public int Count { get; set; }
        
        [GlobalSetup]
        public void GlobalSetup()
            => source = Enumerable.Range(0, Count).ToArray();
    }
}
```

In each test I simply have to inherit from this class. I have many like these:

``` csharp
using System.Linq;

namespace TestBenchmarkDotNet
{
    public class Test01 : TestsBase
    {
        [Benchmark(Baseline = true)]
        public int BaselineTest()
            => source.Sum();

        [Benchmark]
        public int AnotherTest()
            => source.Sum();
    }
}
```

In `Program.cs`, I setup a `BenchmarkSwitcher`:

``` csharp

namespace TestBenchmarkDotNet
{
    class Program
    {
        {
        }
    }
}
```

My issue is that, when I execute and choose an option, it performs all benchmarks that derive from `TestsBase`. Not just the one I chose. Is this an expected behavior?

To easily replicate, here's a minimal example in a repository: https://github.com/aalmada/TestBenchmarkDotNet

YegorStepanov:
Fixes #1709

### Problem
```
Available Benchmarks:
  #0 Test01
  #1 Test02
```
if you write `0` it matches:
* Test01 (by benchmark class index)
* Test01 (class name contains `0`)
* Test02 (class name contains `0`)

### Note

This only applies to interactive mode, the `--filter` cmd option does not work with benchmark class index (`--filter 0` or `--filter #0`)

### DIFF
--- a/src/BenchmarkDotNet/Running/UserInteraction.cs
+++ b/src/BenchmarkDotNet/Running/UserInteraction.cs
@@ -90,18 +90,23 @@ namespace BenchmarkDotNet.Running
             if (userInput.IsEmpty())
                 yield break;
 
+            var integerInput = userInput.Where(arg => IsInteger(arg)).ToArray();
+            var stringInput = userInput.Where(arg => !IsInteger(arg)).ToArray();
+
             for (int i = 0; i < allTypes.Count; i++)
             {
                 var type = allTypes[i];
 
-                if (userInput.Any(arg => type.GetDisplayName().ContainsWithIgnoreCase(arg))
-                    || userInput.Contains($"#{i}")
-                    || userInput.Contains(i.ToString())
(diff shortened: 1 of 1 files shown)


adamsitnik (approved):

PR #2152, opened 2022-10-16
Make ParamsAllValues validator mandatory

YegorStepanov:
The author forgot to add this validator to the list of required validators in #908.

### Example
```C#
[DryJob]
public class MyClass
{
    [ParamsAllValues] public int MyParam;
    [Benchmark] public void MyMethod(){ }
}
```

The first command has `DefaultConfig`.
The second command has `ManualConfig.CreateMinimumViable()`.
image


adamsitnik (approved):

PR #2153, opened 2022-10-16
Fix array comparison in params

### LINKED ISSUE #791: Benchmark(Baseline = true) doesn't work when using byte[] arrays in Params()
```csharp
    public class ToBase64UrlEncodedString
    {
        [Params(new byte[] { 57, 48, 0, 0, 1, 39, 2, 0, 0, }, new byte[] { 57, 48, 0, 0, 0, 0, 0, 0, 0, })]
        public byte[] From { get; set; }

        [Benchmark(Baseline = true)]
        public string Convoluted()
        {
            return ImageUrlEncryptor.ToBase64UrlEncodedString(From);
        }

        [Benchmark]
        public string Microsoft_IdentityModel_Tokens()
        {
            return ImageUrlSafe.ToBase64UrlEncodedString(From);
        }
    }
```

Results in:

```
C:\Working\Console2\Console2\bin\Release> .\Console2.exe
// ***** BenchmarkRunner: Start   *****
// Found benchmarks:
//   ToBase64UrlEncodedString.Convoluted: InProcess(Toolchain=InProcessToolchain [From=System.Byte[]]
//   ToBase64UrlEncodedString.Convoluted: InProcess(Toolchain=InProcessToolchain) [From=System.Byte[]]
//   ToBase64UrlEncodedString.Microsoft_IdentityModel_Tokens: InProcess(Toolchain=InProcessToolchain) [From=System.Byte[]]
//   ToBase64UrlEncodedString.Microsoft_IdentityModel_Tokens: InProcess(Toolchain=InProcessToolchain) [From=System.Byte[]]

// Validating benchmarks:
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group InProcess(Toolchain=InProcessToolchain)-[From=System.Byte[]] in class ToBase64UrlEncodedString has 2
// * Artifacts cleanup *
```

Is this a supported scenario?

YegorStepanov:
fixes #791

Look at the third commit to see what the PR changes.

### DIFF
--- a/src/BenchmarkDotNet/Parameters/ParameterComparer.cs
+++ b/src/BenchmarkDotNet/Parameters/ParameterComparer.cs
@@ -24,11 +24,33 @@ namespace BenchmarkDotNet.Parameters
             if (x == null) return -1;
             for (int i = 0; i < Math.Min(x.Count, y.Count); i++)
             {
+                //todo: compare non-primitive types too
                 int compareTo = PrimitiveComparer.CompareTo(x[i]?.Value, y[i]?.Value);
                 if (compareTo != 0)
                     return compareTo;
+
+                int arrayCompareTo = CompareArrays(x[i]?.Value, y[i]?.Value);
+                if (arrayCompareTo != 0)
+                    return arrayCompareTo;
             }
             return string.CompareOrdinal(x.DisplayInfo, y.DisplayInfo);
+        }
+
+        private static int CompareArrays(object x, object y)
+        {
+            if (x is Array xArray && y is Array yArray)
+            {
+                for (int i = 0; i < Math.Min(xArray.Length, yArray.Length); i++)
+                {
+                    //todo: compare non-primitive types too
+                    int compareTo = PrimitiveComparer.CompareTo(xArray.GetValue(i), yArray.GetValue(i));
+                    if (compareTo != 0)
+                        return compareTo;
+                }
+                if (xArray.Length != yArray.Length)
+                    return xArray.Length.CompareTo(yArray.Length);
+            }
+            return 0;
         }
 
         private class Comparer
--- a/src/BenchmarkDotNet/Parameters/ParameterInstances.cs
+++ b/src/BenchmarkDotNet/Parameters/ParameterInstances.cs
@@ -31,9 +31,25 @@ namespace BenchmarkDotNet.Parameters
 
         public string DisplayInfo =>  Items.Any() ? "[" + string.Join(", ", Items.Select(p => $"{p.Name}={p.ToDisplayText()}")) + "]" : "";
 
-        public string ValueInfo => Items.Any() ? "[" + string.Join(", ", Items.Select(p => $"{p.Name}={p.Value?.ToString() ?? ParameterInstance.NullParameterTextRepresentation}")) + "]" : "";
+        public string ValueInfo => Items.Any() ? "[" + string.Join(", ", Items.Select(p => $"{p.Name}={GetValueInfo(p.Value)}")) + "]" : "";
 
         public string PrintInfo => printInfo ?? (printInfo = string.Join("&", Items.Select(p => $"{p.Name}={p.ToDisplayText()}")));
+
+        private static string GetValueInfo(object value)
+        {
+            if (value is not Array array)
+                return value?.ToString() ?? "null";
+
+            //why it passes array values to cmd?
+            var strings = new List<string>(array.Length); //test array of array
+            for (int i = 0; i < array.Length; i++)
+            {
+                string str = GetValueInfo(array.GetValue(i));
+                strings.Add(str);
+            }
+
+            return string.Join(",", strings);
+        }
 
         public ParameterInstance GetArgument(string name) => Items.Single(parameter => parameter.IsArgument && parameter.Name == name);
 
--- a/tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
+++ b/tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Linq;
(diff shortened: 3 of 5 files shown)


tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
```diff
@@ -262,10 +263,41 @@ public class Escape_ParamsAndArguments
...
+                [Benchmark] public void Bar() { }
+            }
+
+            /* Param Arrays */
+
+            [BaselineColumn]
+            [LogicalGroupColumn]
+            public class ParamArray
+            {
+                [Params(new byte[] { 0 }, new byte[] { 1 }, new byte[] { 0 })]
```
adamsitnik, line 277: what do we gain by having duplicate params (`new byte[] { 0 }`)? one day we might move https://github.com/dotnet/performance/blob/main/src/harness/BenchmarkDotNet.Extensions/UniqueArgumentsValidator.cs into BDN and this would be considered as a bug

```suggestion
                [Params(new byte[] { 0 }, new byte[] { 1 })]
```

tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
```diff
@@ -262,10 +263,41 @@ public class Escape_ParamsAndArguments
...
+            public class ParamArrayOfArrays
+            {
+                [ParamsSource(nameof(GetValues))]
+                public int[][] Param;
+
+                public IEnumerable<int[][]> GetValues()
+                {
+                    yield return new int[][] { new[] { 0 }, new[] { 0 }, };
+                    yield return new int[][] { new[] { 0 }, new[] { 1 }, };
+                    yield return new int[][] { new[] { 0 }, new[] { 0 }, };
```
adamsitnik, line 295: same as above

```suggestion
```

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -24,13 +24,35 @@ public int Compare(ParameterInstances x, ParameterInstances y)
             if (x == null) return -1;
             for (int i = 0; i < Math.Min(x.Count, y.Count); i++)
             {
+                //todo: compare non-primitive types too
                 int compareTo = PrimitiveComparer.CompareTo(x[i]?.Value, y[i]?.Value);
                 if (compareTo != 0)
                     return compareTo;
+
+                int arrayCompareTo = CompareArrays(x[i]?.Value, y[i]?.Value);
```
adamsitnik, line 32: I was surprised to see that it works, then I've checked and found that we are relying on the `.ToString` comparison:

https://github.com/dotnet/BenchmarkDotNet/blob/58d4bae8996b713f21d6472702d95b21a77f79c9/src/BenchmarkDotNet/Parameters/ParameterComparer.cs#L49

It would be more future proof to perform an explicit check: 

```cs
int compareTo = x[i]?.Value is Array xArray && y[i]?.Value is Array yArray
   ? CompareArrays(xArra, yArray)
   : PrimitiveComparer.CompareTo(x[i]?.Value, y[i]?.Value);
```

src/BenchmarkDotNet/Parameters/ParameterInstances.cs
```diff
@@ -31,10 +31,26 @@ public void Dispose()
...
+        public string ValueInfo => Items.Any() ? "[" + string.Join(", ", Items.Select(p => $"{p.Name}={GetValueInfo(p.Value)}")) + "]" : "";
 
         public string PrintInfo => printInfo ?? (printInfo = string.Join("&", Items.Select(p => $"{p.Name}={p.ToDisplayText()}")));
 
+        private static string GetValueInfo(object value)
+        {
+            if (value is not Array array)
+                return value?.ToString() ?? "null";
+
+            //why it passes array values to cmd?
```
adamsitnik, line 43: There are two scenarios when we need it:

- when attaching a debugger to a running process to know what we are attaching to (imagine 2/3 params work fine and we want to attach debugger to only the last one that fails)
image

- when analyzing trace files produced by `EtwProfiler` and searching for the process we care about in PerfView or other tool:
image

```suggestion
```

adamsitnik (commented):
It's almost ready. PTAL at my comments and again thank you @YegorStepanov !

tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
YegorStepanov: #### You want to remove the purpose of these tests and PR :)
These tests check that arrays can be grouped if they have the same primitive values (but they are actually different arrays).

We can disallow duplicate arguments for `[Params]` because `Attribute` supports only the built-in types. (IMO it should do Roslyn Analyzer only, not a validator)

#### But how can we disable it for [ParamSource]?
If the values are not primitive, it fallbacks to `.ToString`, which means that __most of__ user classes are equal:
```C#
x.ToString().CompareTo(y.String()) == 0 //always if .ToString is not overloaded
```
#### UniqueArgumentsValidator
If I remember correctly, you cannot making a difference between [Params] and [ParamSource] values, so `UniqueArgumentsValidator` falls for:
```C#
[ArgumentSource("GetValues")]
[Benchmark]
public void Benchmark(MyType instance) { }

public IEnumerable<MyType> GetValues()
{
    // fails because both ToString() are equal:
    yield return new MyType { Value = 10 };
    yield return new MyType { Value = 100 };
}

public class MyType { public int Value; }
```

tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
YegorStepanov: answered above

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
YegorStepanov: But what the code will do if the type is neither an array nor a primitive? :)

The code works like this:
```
1) if type is primitive:
1.1) if (x > y or x < y) : return 1 or -1
2) if type is array:
1.1) if (x > y or x < y) : return 1 or -1
3) compare by string and return
```

So, if the `x=0 and y=0`, it exits on the last line:
```
return string.CompareOrdinal(0?.ToString(), 0?.ToString()); 
```

tests/BenchmarkDotNet.Tests/Exporters/MarkdownExporterApprovalTests.cs
YegorStepanov: If the type implements `IComparable` then we shouldn't jump to ToString() (because it can be overloaded).
So the code will look like:
```C#
if(IComparableComparer.TryCompareTo(x, y, var userCompareTo))
    return userCompareTo;
```
But it works for primitives/arrays.

Are you satisfied with the current implementation, or should I remake it?

kapilepatel:
Hi, is this still active/needed ?
Can I work on it?

timcassell:
@kapilepatel Yes, feel free to work on this.

timcassell:
Done by #2887

PR #2154, opened 2022-10-17
remove dependency to Mono.Posix.NETStandard to unblock WASM benchmarks runs

### LINKED ISSUE #2146: Build warning MSB3245: Could not locate the assembly `Mono.Posix`
When targeting `.NET Framework`, I'm getting the above build warning due to (@adamsitnik): https://github.com/dotnet/BenchmarkDotNet/blob/db8f8d85db44f35af97e763b90f2ffd256eb99e0/src/BenchmarkDotNet/BenchmarkDotNet.csproj#L24
This is a problem with `Mono.Posix.NETStandard` (it has a separate reference assembly for `net40` with type forwards to `Mono.Posix` while its `netstandard2.0` version doesn't). I don't think there is anything to be done about it on BDN's side and I know how to force my project to not use the `net40` ref assembly, but I'm wondering:

Why doesn't the warning appear in the CI build with BDN tests targeting net462 on windows-latest?

I'm getting the warning when I build the BDN repo via build.bat locally and when I build my own project on GitHub's windows-latest.

adamsitnik:
fixes #2146

reported by @radical and @lewing in https://github.com/dotnet/performance/pull/2639#issuecomment-1279537243

### DIFF
--- a/samples/BenchmarkDotNet.Samples/IntroLargeAddressAware.cs
+++ b/samples/BenchmarkDotNet.Samples/IntroLargeAddressAware.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Runtime.InteropServices;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Environments;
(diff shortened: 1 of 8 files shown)


adamsitnik:
I've verified that it works with CLR x64 and arm64, merging.

PR #1796, opened 2021-09-02
fixes(configuration): Not unique exporter for exporter type 

workgroupengineering:
fixes #1700

alternative version of PR #1702

PR #2155, opened 2022-10-17
Rework "Not unique exporter for exporter type" PR

YegorStepanov:
### Problem
Starting with #1796 PR (#1700 issue), only one exporter per type can be.

#### This means that the following cases stopped working in 0.13.2:
```C#
// DefaultConfig has MarkdownExporterAttribute.Default already
// It prints a warning and Atlassian file is not created.
[MarkdownExporterAttribute.Atlassian] 
public class Klass
{
    [Benchmark] public void M() { }
}
```
Also, it means that it will never works:
```C#
// not working for ManualConfig too
[XmlExporterAttribute.Full] // it uses XmlExporter.Full under the hood
[XmlExporterAttribute.Brief] // it uses XmlExporter.Brief under the hood
```

### Solution
1) revert #1796
2) compare by `IExporter.Name`
3) rename `IExporter.Name` -> `IExporter.Id` because `Name` is misleading, the other entities in BDN uses `Id` for it:
```
Characteristic.Id
IAnalyser.Id
ILogger.Id
IDiagnoser.Ids
```
4) Add `RPlotValidator` to check that user do not override `csv` separator (Currently, all version of csv generates file with same name)

### DIFF
--- a/src/BenchmarkDotNet/Configs/DebugConfig.cs
+++ b/src/BenchmarkDotNet/Configs/DebugConfig.cs
@@ -52,7 +52,6 @@ namespace BenchmarkDotNet.Configs
 
     public abstract class DebugConfig : IConfig
     {
-        private readonly static Conclusion[] emptyConclusion = Array.Empty<Conclusion>();
         public abstract IEnumerable<Job> GetJobs();
 
         public IEnumerable<IValidator> GetValidators() => Array.Empty<IValidator>();
@@ -85,7 +84,5 @@ namespace BenchmarkDotNet.Configs
         public IEnumerable<BenchmarkLogicalGroupRule> GetLogicalGroupRules() => Array.Empty<BenchmarkLogicalGroupRule>();
 
         public ConfigOptions Options => ConfigOptions.KeepBenchmarkFiles | ConfigOptions.DisableOptimizationsValidator;
-
-        public IReadOnlyList<Conclusion> ConfigAnalysisConclusion => emptyConclusion;
     }
 }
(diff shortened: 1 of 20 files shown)


YegorStepanov:
I'm not at all sure about renaming (`IExporter.Name -> IExporter.Id`)

I don't think anyone uses it so it one line change for customers.

We will never be able to use "C# default interface implementation" due to .NET Framework?

workgroupengineering:
I am not very convinced of the solution. Because if you have developed custom exporters, no anomalies will be reported unless a validator is written.

YegorStepanov:
> I am not very convinced of the solution. Because if you have developed custom exporters, no anomalies will be reported unless a validator is written.

@workgroupengineering Could you write an example please? It covers your issue (#1700)

If you mean about 'IExporterDependencies':
1) only `RPlotExporter` uses it. 
2) it's internal interface.

If the plots are rewritten in C#, `IExporterDependencies` will be deleted.

YegorStepanov:
#### Instead of making `IExporterDependencies` public, what do you think about the next one:

The naming of Json/Xml exporters is depending on "exporter style":
```
XmlExporter.Full -> creates "NAME-full.xml"
XmlExporter.Brief -> creates "NAME-brief.xml"
JsonExporter.Full -> creates "NAME-full.json"
JsonExporter.Brief -> creates "NAME-brief.json"
```

Maybe we should do the same for `csv`?
```
CsvExporter(CsvSeparator.Comma) -> creates "NAME-comma.csv"
CsvExporter(CsvSeparator.Semicolon) -> creates "NAME-semicolon.csv"
CsvExporter(CsvSeparator.CurrentCulture) -> creates "NAME-comma.csv" or creates "NAME-semicolon.csv"
```

`CsvSeparator.CurrentCulture` looks dangerous, but as I understand it, R graphs depend on this

delixfe:
@YegorStepanov, please don't forget individual names for the Markdown exporters.

Issue #2102, opened 2022-09-12
Add benchmarking progress to console title


Khaos66:
Benchmarks run quite long to get accurate results. This time multiplies by params and benchmarks defined.
It would be nice to have a progress indicator for the running benchmarking process.

As most benchmarks run in a console app, it is possible to use the consoles title bar as a place to display this progress.

At a minimum it should display: `Running 5/10`
Or maybe: `Running: 5/10 - MyMethod(Param1,Param2) - ... configs like Framework,Runtime and such`

What do you think?

Khaos66:
I only just realized that there is this block in the output.
![image](https://user-images.githubusercontent.com/4013009/189618758-af45d922-3e9b-4566-90bd-8fc989f78c1a.png)

So, why not display some of this information in the console title?

adamsitnik:
Sure, we can add that. My requirements:

- the title needs to be short, so the most important information is always visible
- it should never throw:
   - if some platforms don't support it, we should not call this [property](https://learn.microsoft.com/en-us/dotnet/api/system.console.title?view=net-6.0) for them
   - same for the redirected output

franciscomoloureiro:
I will work on this issue!

adamsitnik:
@franciscomoloureiro to check the current platform you can use helper methods exposed by `RuntimeInformation` class:

https://github.com/dotnet/BenchmarkDotNet/blob/a78c2e6a6e3db79069fb5bbbd6da6e5cbea8c029/src/BenchmarkDotNet/Portability/RuntimeInformation.cs#L80

The current progress logging happens here:

https://github.com/dotnet/BenchmarkDotNet/blob/8ec00dd53ff1ed6d8f9a8720857599474c8c36c1/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L617

adamsitnik:
Fixed by #2140

timcassell:
As a followup, it might be nice to display the progress in the taskbar, but I only see the API available in .Net Framework. Not sure if there's an equivalent in .Net Core/Standard. https://learn.microsoft.com/en-us/dotnet/api/system.windows.shell.taskbariteminfo.progressstate?view=windowsdesktop-7.0

Khaos66:
Funny. I did this in another project some time ago. 
You don't need another NuGet. Just some good old PInvoke ;)
```C#
public static class TaskbarProgress
    {
        public enum TaskbarStates
        {
[...]

PR #2158, opened 2022-10-18
Add taskbar progress

timcassell:
I initially tried to use a [nuget package](https://www.nuget.org/packages/Microsoft-WindowsAPICodePack-Shell), but it does not support netstandard2.0. So I copied the relevant code from https://github.com/contre/Windows-API-Code-Pack-1.1 to `Helpers/Taskbar` to get this working (thanks @contre and @aybe).

Also thanks to [ivan_pozdeev on StackOverflow](https://stackoverflow.com/a/28616832/5703407) for grabbing the proper console window handle.

image


src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -33,6 +34,14 @@ internal static class BenchmarkRunnerClean
 
         internal static readonly IResolver DefaultResolver = new CompositeResolver(EnvironmentResolver.Instance, InfrastructureResolver.Instance);
 
+        private static IntPtr ConsoleWindowHandle;
+        private static ConsoleCancelEventHandler OnConsoleCancelEvent = delegate {
+            TaskbarManager.Instance.SetProgressState(TaskbarProgressBarState.NoProgress, ConsoleWindowHandle);
+        };
+
+        [System.Runtime.InteropServices.DllImport("kernel32.dll")]
+        private static extern IntPtr GetConsoleWindow();
```
adamsitnik, line 43: I would prefer to avoid adding any OS-specific sys-calls to this type (`BenchmarkRunner` is already way too big and too ugly ;) ). Instead, we could create a new `IDisposable` helper type that would do the following:

- Do nothing on non-Windows OS-es
- On Windows:
   - read console handle in the `ctor` and sign up for the Ctrl+C just once
   - unregister from the event in Dispose
   - expose two methods: `ResetProgress` and `SetProgress`

and use the helper from runner

adamsitnik (commented):
The screen shot that you have provided looks very promising, but how about the license? Can we just copy paste the code?

@timcassell thank you for your contribution!

timcassell:
@adamsitnik The license: https://github.com/contre/Windows-API-Code-Pack-1.1/blob/master/LICENSE

I was reading about it, apparently it was originally written by Microsoft a while ago and then they disappeared it, and other people picked up the code and uploaded to nuget ([Source](https://stackoverflow.com/questions/24081665/windows-api-code-pack-where-is-it#:~:text=packages%2FWindowsAPICodePack%2DSensors-,https%3A%2F%2Fwww.nuget.org%2Fpackages%2FWindowsAPICodePack%2D,org%2Fpackages%2FWindowsAPICodePack%2DShellExtensions)). So I don't think the license is an issue, but I'm no legal expert. ;)

Also, your change request makes sense, I'll do that.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
timcassell: Done

timcassell:
In the future we could even use this to alert warnings or errors if we want.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -33,6 +33,8 @@ internal static class BenchmarkRunnerClean
 
         internal static readonly IResolver DefaultResolver = new CompositeResolver(EnvironmentResolver.Instance, InfrastructureResolver.Instance);
 
+        private static TaskbarProgress s_taskbarProgress;
```
adamsitnik, line 36: We should avoid having disposable static instances, please just pass the instance to the right method.

timcassell: Of course, much cleaner!

Khaos66:
I just commented on #2102. You don't need another NuGet. This is a simple ComIport. 
P.S. feel free to copy my code

timcassell:
Thanks @Khaos66! I implemented your code, it's so much less than I had before!

adamsitnik (commented):
The code looks good to me, but I've tested it and it does not work on my Windows 10 and 11 machines, both using `cmd` and Windows Terminal:

image

timcassell:
@adamsitnik I pushed a change to get the console window's ancestor. Unfortunately, I only have a Win7 machine to test on, so could you let me know if that fixes it?

timcassell:
@adamsitnik I got the same error when I was testing initially, which was why I found it necessary to call the `GetConsoleWindow()`. This was the code I had working:

```cs
using Microsoft.WindowsAPICodePack.Taskbar;
using System;
using System.Runtime.InteropServices;
using System.Threading;

namespace TaskbarProgress
{
    internal class Program
    {
        [DllImport("kernel32.dll")]
        private static extern IntPtr GetConsoleWindow();

        {
            if (TaskbarManager.IsPlatformSupported)
            {
                IntPtr consoleWindowHandle = GetConsoleWindow();
                TaskbarManager tbManager = TaskbarManager.Instance;

                for (int i = 0; i < 100; i++)
                {
                    tbManager.SetProgressValue(i, 100, consoleWindowHandle);

                    Thread.Sleep(100);
                }
                tbManager.SetProgressState(TaskbarProgressBarState.NoProgress, consoleWindowHandle);
            }
        }
    }
}
```

> I am sorry but I don't have more time to spend on this.

Understandable. I might be able to get my hands on a win 11 machine in a couple days.

Khaos66:
Strange. I used this code on all of the windowses. Win 7, 8, 10, 11. 
You need the right window handle however. I've not used it with a console app before. Sorry

timcassell:
I was able to see it working in both cmd and PowerShell on Win11, but it doesn't work with Terminal. Not sure why, I even tried walking the `GetAncestor` with `Parent` flag until I hit the Desktop window, but still no luck. We might be able to get it to work by pushing an actual Terminal command for it, but not sure if we want to do that (or if we can even detect that we're running in Terminal). https://github.com/microsoft/terminal/issues/3004

image
image
image

Also, Win11 is weird, when I report `0` progress, it fills the whole bar instead of having an empty bar (Win7 doesn't do this). So I left the state as `Indeterminate` until the first benchmark is done.

Khaos66:
Yes win11 does that so you know that the window is open. Because no bar unter the icon means the app is pinned to the taskbar but not running ;)

I guess that Windows Terminal uses sub-processes instead of embedded windows. It could just redirect all inputs/outputs for you. 
You might want to check the parent process instead of the parent window?

timcassell:
From reading further into issues on the Windows Terminal repo, it seems they are aware of this limitation (and do not currently have a workaround).

@adamsitnik Can this be merged as-is? It works in other shells and has no negative effects where it doesn't work. If/when Windows Terminal adds a mechanism to show the progress from a console app at some point in the future, that can be added to BDN at that time.

timcassell:
Thanks to help from folks on the Windows Terminal repo, I was able to make it work. It now works for both old conhost hosted shells and Windows Terminal hosted shells.

adamsitnik (approved):
It looks and works great, thank you @timcassell !

You should consider writing a blog post about hot to get it working for Windows 7 + 11 since we were both unable to find any solution on the internet.

PR #2160, opened 2022-10-18
Corrected logic to restore foreground color in ConsoleLogger.cs

farQtech:


adamsitnik (approved):
LGTM (I've tested it manually), thank you for your contribution @farQtech !

PR #2161, opened 2022-10-18
add quiet logger

franciscomoloureiro:
closes [#190](https://github.com/dotnet/BenchmarkDotNet/issues/190)

### DIFF
--- a/src/BenchmarkDotNet/Attributes/QuietModeAttribute.cs
+++ b/src/BenchmarkDotNet/Attributes/QuietModeAttribute.cs
@@ -1,0 +1,21 @@
+﻿using BenchmarkDotNet.Configs;
+using JetBrains.Annotations;
+using System;
+
+namespace BenchmarkDotNet.Attributes
+{
+    /// <summary>
+    /// determines if all auto-generated files should be kept or removed after running the benchmarks
+    /// </summary>
+    [PublicAPI]
+    [AttributeUsage(AttributeTargets.Class)]
+    public class QuietModeAttribute : Attribute, IConfigSource
+    {
+        public IConfig Config { get; }
+
+        public QuietModeAttribute(bool value = false)
+        {
+            Config = ManualConfig.CreateEmpty().WithOption(ConfigOptions.QuietMode, value);
+        }
+    }
+}
--- a/src/BenchmarkDotNet/Configs/ConfigExtensions.cs
+++ b/src/BenchmarkDotNet/Configs/ConfigExtensions.cs
@@ -76,6 +76,11 @@ namespace BenchmarkDotNet.Configs
         [PublicAPI] public static ManualConfig WithCultureInfo(this IConfig config, CultureInfo cultureInfo) => config.With(m => m.CultureInfo = cultureInfo);
 
         /// <summary>
+        /// Run benchmars in quiet mode.
+        /// </summary>
+        [PublicAPI] public static IConfig QuietMode(this IConfig config, bool value = true) => config.WithOption(ConfigOptions.QuietMode, value);
+
+        /// <summary>
         /// determines if all auto-generated files should be kept or removed after running the benchmarks
         /// </summary>
         [PublicAPI] public static IConfig KeepBenchmarkFiles(this IConfig config, bool value = true) => config.WithOption(ConfigOptions.KeepBenchmarkFiles, value);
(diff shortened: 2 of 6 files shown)


adamsitnik:
@franciscomoloureiro could you please solve the merge conflict?

src/BenchmarkDotNet/Attributes/QuietModeAttribute.cs
```diff
@@ -0,0 +1,21 @@
+﻿using BenchmarkDotNet.Configs;
+using JetBrains.Annotations;
+using System;
+
+namespace BenchmarkDotNet.Attributes
+{
+    /// <summary>
+    /// determines if all auto-generated files should be kept or removed after running the benchmarks
+    /// </summary>
```
YegorStepanov, line 9: Wrong documentation?

src/BenchmarkDotNet/Attributes/QuietModeAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+    [PublicAPI]
+    [AttributeUsage(AttributeTargets.Class)]
+    public class QuietModeAttribute : Attribute, IConfigSource
```
YegorStepanov, line 12: IMO there should be 1 attribute for all verbosity options and 1 cmd option instead of a separate attribute/option for each.

YegorStepanov:
What the PR implement? `[QuietMode]` could mean the following cases:
1) from #2092
> quiet mode: https://github.com/dotnet/BenchmarkDotNet/issues/190 (the processes can talk, but just don't print stuff to output) 
2) [from #190 Adam ](https://github.com/dotnet/BenchmarkDotNet/issues/190#issuecomment-1279299365)
3) Completely no output (from some comment in #190)

franciscomoloureiro:
> What the PR implement? `[QuietMode]` could mean the following cases:
> 
> 1. from [Use Pipes for host and benchmark process communication #2092](https://github.com/dotnet/BenchmarkDotNet/pull/2092)
> 
> > quiet mode: #190 (the processes can talk, but just don't print stuff to output)
> 
> 2. [from #190 Adam ](https://github.com/dotnet/BenchmarkDotNet/issues/190#issuecomment-1279299365)
> 3. Completely no output (from some comment in [Feature request: quiet mode #190](https://github.com/dotnet/BenchmarkDotNet/issues/190))

It is impementing 2. Running the benchmark with less output

src/BenchmarkDotNet/Attributes/QuietModeAttribute.cs
franciscomoloureiro: I am not sure I understand this, the attribute/options are just different ways of configuring the same mode, with quiet=false you get the same output as before with quiet=true you only get the more relevant output

YegorStepanov: Sorry, I'm mean about it: https://github.com/dotnet/BenchmarkDotNet/issues/190#issuecomment-1281672728
The PR should ideally implement all options, because it's not easy to add logging levels to the BDN.

It would be intuitive to see the following api:
```C#
[ConsoleVerbosity(Verbosity.Diagnostic, writeOutputFromGeneratedProject: true/false)]
[ConsoleVerbosity(Verbosity.Detailed)] // your PR does this level
[ConsoleVerbosity(Verbosity.Normal)]
[ConsoleVerbosity(Verbosity.Minimal)]
[ConsoleVerbosity(Verbosity.Quiet)]
```
cmd
```
dotnet run -c Release -- --verbosity diagnostic
dotnet run -c Release -- --verbosity detailed
dotnet run -c Release -- --verbosity normal
dotnet run -c Release -- --verbosity minimal
dotnet run -c Release -- --verbosity quiet
//short form:
dotnet run -c Release -- -v diag
dotnet run -c Release -- -v d
dotnet run -c Release -- -v n
dotnet run -c Release -- -v m
dotnet run -c Release -- -v q
```

If we implement your PR, it would be hard to do the standard .NET `--verbosity` option.

And the naming `[QuietMode]` is super confusing. 
Below I have written 3 possible options. By my opinion it should mean 1) or 3).

franciscomoloureiro: I understand now the full proposal, I had not seen last Adam's message on the issue, I will work on this during the next couple of days!

YegorStepanov: It seems to me that this can be quite complicated.

My proposal is not final.
It's worth to post verbosity output to #190 (as mine) to start a discussion (because it affects the BDN a lot)

PS. If you don't do it, I'll get to it in a month. Any progress that you make will be very helpful!!

PR #2168, opened 2022-10-24
Fix bugs and typos

YegorStepanov:


src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -548,7 +548,7 @@ private static (BenchmarkRunInfo[], List<ValidationError>) GetSupportedBenchmark
                 .Where(infos => infos.BenchmarksCases.Any())
                 .ToArray();
 
-            return (benchmarkRunInfos, validationErrors);
+            return (runInfos, validationErrors);
```
YegorStepanov, line 551: @emanuel-v-r, is it a bug? 
It is strange that the IDE did not highlight you that the variable `benchmarksRunInfo` is not used

emanuel-v-r: Yes, looks like it's a bug, thank you for fixing it.
Seems that I've missed it.

We could add an .editorconfig to the repo, and start treating such scenarios as an "error" so that we don't miss it.

emanuel-v-r (approved):

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
YegorStepanov: Did you use Visual Studio and it didn't show that the variable was not in use?

Today, I tried migrating from `ruleset` to `.editorconfig` and got 700 warnings (currently they treated as errors). 
It is definitely necessary to migrate to it, but this will require cleaning the solution - most PRs will be "broken"

emanuel-v-r: Not sure, If I had noticed I would obviously have it fixed.
Probably we should try some rule by rule approach, instead of having it all at once

YegorStepanov: *I mean migration to dotnet/runtime `.editorconfig` with some BenchmarkDotNet modification.

You are right, it is better to do it in several stages.

Сurrently, most of the enabled analyzers are for `Rider/Resharper` only, they don't work in VS.

PS this will finally fix that one user adding an empty last line and in the next PR another user removes it.

adamsitnik (approved):

PR #2171, opened 2022-10-26
Issue #1024: Calculate baseline by the fastest benchmark

### LINKED ISSUE #1024: Calculate baseline by fastest with config
Hi,
I'm implementing a `: ManualConfig` and I want to have the "Ratio" column.
Adding it with `Add(BaselineRatioColumn.RatioMean);` works fine.

But I always want the baseline to be the fastest run instead of having to manually (predetermine) what benchmark should be considered the baseline.

Is that possible somehow? If so, how?

Serg046:
Fixes #1024 
The idea is inside `Summary.cs`

### DIFF
--- a/src/BenchmarkDotNet/Reports/Summary.cs
+++ b/src/BenchmarkDotNet/Reports/Summary.cs
@@ -37,6 +37,7 @@ namespace BenchmarkDotNet.Reports
         private ImmutableDictionary<BenchmarkCase, BenchmarkReport> ReportMap { get; }
         private BaseliningStrategy BaseliningStrategy { get; }
         private bool? isMultipleRuntimes;
+        private readonly BenchmarkCase inferredBaselineBenchmarkCase;
 
         public Summary(
             string title,
@@ -62,11 +63,30 @@ namespace BenchmarkDotNet.Reports
             DisplayPrecisionManager = new DisplayPrecisionManager(this);
             Orderer = GetConfiguredOrdererOrDefaultOne(reports.Select(report => report.BenchmarkCase.Config));
             BenchmarksCases = Orderer.GetSummaryOrder(reports.Select(report => report.BenchmarkCase).ToImmutableArray(), this).ToImmutableArray(); // we sort it first
+            inferredBaselineBenchmarkCase = GetFastestBenchmarkCase(reports);
             Reports = BenchmarksCases.Select(b => ReportMap[b]).ToImmutableArray(); // we use sorted collection to re-create reports list
             BaseliningStrategy = BaseliningStrategy.Create(BenchmarksCases);
             Style = GetConfiguredSummaryStyleOrDefaultOne(BenchmarksCases).WithCultureInfo(cultureInfo);
             Table = GetTable(Style);
             AllRuntimes = BuildAllRuntimes(HostEnvironmentInfo, Reports);
+        }
+
+        private static BenchmarkCase GetFastestBenchmarkCase(ImmutableArray<BenchmarkReport> reports)
+        {
+            if (reports.Any() && reports.All(r => r.BenchmarkCase.Config.AutomaticBaselineMode == AutomaticBaselineMode.Fastest))
+            {
+                var fastestReport = reports.First();
+                foreach (var report in reports.Skip(1))
+                {
+                    if (report.ResultStatistics.Mean < fastestReport.ResultStatistics.Mean)
+                    {
+                        fastestReport = report;
+                    }
+                }
+                return fastestReport.BenchmarkCase;
+            }
+
+            return null;
         }
 
         [PublicAPI] public bool HasReport(BenchmarkCase benchmarkCase) => ReportMap.ContainsKey(benchmarkCase);
@@ -133,7 +153,11 @@ namespace BenchmarkDotNet.Reports
             => Orderer.GetLogicalGroupKey(BenchmarksCases, benchmarkCase);
 
         public bool IsBaseline(BenchmarkCase benchmarkCase)
-            => BaseliningStrategy.IsBaseline(benchmarkCase);
+        {
+            return inferredBaselineBenchmarkCase != null
+                ? inferredBaselineBenchmarkCase == benchmarkCase
+                : BaseliningStrategy.IsBaseline(benchmarkCase);
+        }
 
         [CanBeNull]
         public BenchmarkCase GetBaseline(string logicalGroupKey)
--- a/src/BenchmarkDotNet/Attributes/AutomaticBaselineAttribute.cs
+++ b/src/BenchmarkDotNet/Attributes/AutomaticBaselineAttribute.cs
@@ -1,0 +1,13 @@
+﻿using BenchmarkDotNet.Configs;
+using System;
+
+namespace BenchmarkDotNet.Attributes
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class AutomaticBaselineAttribute : Attribute, IConfigSource
+    {
+        public IConfig Config { get; }
+
+        public AutomaticBaselineAttribute(AutomaticBaselineMode mode) => Config = ManualConfig.CreateEmpty().WithAutomaticBaseline(mode);
+    }
+}
--- a/src/BenchmarkDotNet/Configs/AutomaticBaselineMode.cs
+++ b/src/BenchmarkDotNet/Configs/AutomaticBaselineMode.cs
@@ -1,0 +1,8 @@
+﻿namespace BenchmarkDotNet.Configs
+{
+    public enum AutomaticBaselineMode
+    {
+        None,
+        Fastest
+    }
+}
--- a/src/BenchmarkDotNet/Configs/DebugConfig.cs
+++ b/src/BenchmarkDotNet/Configs/DebugConfig.cs
@@ -87,5 +87,7 @@ namespace BenchmarkDotNet.Configs
         public ConfigOptions Options => ConfigOptions.KeepBenchmarkFiles | ConfigOptions.DisableOptimizationsValidator;
 
         public IReadOnlyList<Conclusion> ConfigAnalysisConclusion => emptyConclusion;
+
+        public AutomaticBaselineMode AutomaticBaselineMode { get; } = AutomaticBaselineMode.None;
     }
 }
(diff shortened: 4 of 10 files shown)


src/BenchmarkDotNet/Reports/Summary.cs
```diff
@@ -62,13 +63,32 @@ public class Summary
...
             Reports = BenchmarksCases.Select(b => ReportMap[b]).ToImmutableArray(); // we use sorted collection to re-create reports list
             BaseliningStrategy = BaseliningStrategy.Create(BenchmarksCases);
             Style = GetConfiguredSummaryStyleOrDefaultOne(BenchmarksCases).WithCultureInfo(cultureInfo);
             Table = GetTable(Style);
             AllRuntimes = BuildAllRuntimes(HostEnvironmentInfo, Reports);
         }
 
+        private static BenchmarkCase GetFastestBenchmarkCase(ImmutableArray<BenchmarkReport> reports)
+        {
+            if (reports.Any() && reports.All(r => r.BenchmarkCase.Config.AutomaticBaselineMode == AutomaticBaselineMode.Fastest))
```
YegorStepanov, line 76:
```
reports.All(r => r.BenchmarkCase.Config.AutomaticBaselineMode == AutomaticBaselineMode.Fastest))
```

Is it correctly works when 2 benchmark classes have different `AutomaticBaselineMode`?
```C#
[AutomaticBaseline(AutomaticBaselineMode.Fastest)]
public class C1 { }

[AutomaticBaseline(AutomaticBaselineMode.None)]
public class C2 { }
```

To join summaries, use:
1) `--join` (for cmd)
2) `BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).RunAllJoined()`

Serg046: The idea itself works because the benchmarks are already grouped by the type so that Summary contains the cases of either C1 or C2 only. However, this particular example doesn't print the ratio for all the methods. I guess BDN trims columns that don't exist in all results (C1's benchmarks contain ration but C2's ones don't).

src/BenchmarkDotNet/Reports/Summary.cs
```diff
@@ -62,13 +63,32 @@ public class Summary
...
+            {
+                var fastestReport = reports.First();
+                foreach (var report in reports.Skip(1))
+                {
+                    if (report.ResultStatistics.Mean < fastestReport.ResultStatistics.Mean)
```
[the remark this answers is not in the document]
Serg046: > instead of the failed validation:

I actually thought about it but decided to keep a new feature overriding more local configurations. Added a validation.

> maybe we should add "Fastest"/"Slowest" for some other columns, like Allocation?

Could make sense but not to me. Ratio column is enough for me, the fastest is just a baseline (ratio == 1)

> ResultStatistics can be null.

Fixed

Thanks for your review Yegor 👍

src/BenchmarkDotNet/Reports/Summary.cs
YegorStepanov: `ManualConfig.Union(...)` resets the `AutomaticBaselineMode`:

```
config = ManualConfig.Union(config, configFromAttribute.Config) // C1, set the value to "Fastest"
config = ManualConfig.Union(config, configFromAttribute.Config) // C2, set the value to "None" (even if the `[AutomaticBaseline(None)]` is not used)
```

The problem is that the final config has only one value of `AutomaticBaselineMode`, but it should have a different value for each benchmark class to allow:

```C#
[AutomaticBaseline(AutomaticBaselineMode.Fastest)] public class B1 { }
public class B2 { } //implicit AutomaticBaselineMode.None
[AutomaticBaseline(AutomaticBaselineMode.Slowest)] public class B3 { }
```

Serg046: No, it is not like that, the configs are different. I see that this [inferredBaselineBenchmarkCase](https://github.com/dotnet/BenchmarkDotNet/blob/fd9f34fc837d3bebfeca888506e4c6a5fcbec232/src/BenchmarkDotNet/Reports/Summary.cs#L66) is calculated properly. There is probably a need to add a column or something like that, I will try to identify the problem.

adamsitnik:
I am not very familiar with this part of BDN and if possible I would prefer @AndreyAkinshin to review it. Thanks!

AndreyAkinshin (changes requested):
I love this feature, it looks cool! Thanks for the contribution!

My primary concern is about the tests. Thread.Sleep-based integration tests are too fragile and too slow, we should avoid them whenever it's possible. We need fast regular unit tests to cover this feature. In order to do it, we can use `MockFactory` to create an artifical summary based on the given type and check the column values manually. It would be a fast, reliable and isolated testing approach.

src/BenchmarkDotNet/Reports/Summary.cs
Serg046: Yep, you are right. I didn't notice that Summary is built the third time (where you have all 4 methods). I was checking first two instances (first two and second two). Thank you!

Serg046: So, this is what I suggest we do https://github.com/dotnet/BenchmarkDotNet/pull/2171/commits/d20c117b567f0ddc9effd56eeda9c299d97b8ece

1. We skip all defaut (`None`) values to build configs.
2. We take the latest non-default value to determine whether we need to override baseline functionality or we don't.

Serg046:
I tried to fix all the notes we have here, please let me know if there is anything else to deal with @YegorStepanov @AndreyAkinshin @adamsitnik

src/BenchmarkDotNet/Reports/Summary.cs
```diff
@@ -73,7 +73,8 @@ public class Summary
 
         private static BenchmarkCase GetFastestBenchmarkCase(ImmutableArray<BenchmarkReport> reports)
         {
-            if (reports.Any() && reports.All(r => r.BenchmarkCase.Config.AutomaticBaselineMode == AutomaticBaselineMode.Fastest))
+            var baselineReport = reports.Where(r => r.BenchmarkCase.Config.AutomaticBaselineMode != AutomaticBaselineMode.None).LastOrDefault();
+            if (baselineReport != null && baselineReport.BenchmarkCase.Config.AutomaticBaselineMode == AutomaticBaselineMode.Fastest)
```
YegorStepanov, line 77:
```suggestion
            if (reports.Any(r => r.BenchmarkCase.Config.AutomaticBaselineMode == AutomaticBaselineMode.Fastest)
```

You can simplify it.

YegorStepanov:
## Cons of the current implementation:
We cannot extend the enum, because attribute actually set a boolean flag in config. 
It make sense to drop `AutomaticBaselineMode` completely. (Attribute will set `IConfig.IsAutomaticBaseline = true`).

I like the current attribute API but it will need to rename the attribute, if we drop the enum.

## I'm thinking of a more powerful solution:

Not for all users `Mean` is the most interesting column:
- In the real-time apps such as games, allocation is more painful than mean time.
- Andrey will like this: `Mean` is **never** interesting. Compare methods by percentage (like `P90`) or at least by `Median`. 
- Indeed, all the math columns can be interesting here.

#### API:
```C#
[AutomaticBaseline(string columnName, Strategy strategy = Maximum)]

enum Strategy { Minimum, Maximum } // AutomaticBaselineStrategy or AutomaticStrategy or BaselineStrategy?
```
#### how to use:
```C#
[AutomaticBaseline("P90")]
[AutomaticBaseline(Column.P90))]
```

We can compare values from the `string IColumn.GetValue(...)` method (if `IColumn.IsNumeric == true` than we do `TryParse` and compare them by number, not by string).
Some columns such as `Runtime` need special handling.

### Questions:
Does it make sense to add support for `assembly:` to attribute?
```C#
[assembly: [AutomaticBaseline("Mean")]]

public Benchmarks
{
    // currently, it would be an error: "Do not use both AutomaticBaseline and [Benchmark(Baseline = true)]"
    [Benchmark(Baseline = true)] public void B1() { }
    [Benchmark] public void B1() => Thread.Sleep(100);
}

```
How can this be resolved:
1) make `assembly:` attribute overridable by `[Benchmark(Baseline = true)]` and `[class: AutomaticBaseline("another-column")]`
2) add parameter: `[AutomaticBaseLine(..., bool overridable = false)]`

If [Serg046](https://github.com/Serg046) does not have so much time, then I can do it within a few months.

src/BenchmarkDotNet/Reports/Summary.cs
Serg046: Yes, I just wanted to highlight the idea that we find the latest value and then use it. Since Fastest is the only mode for now, then you suggestion makes sense for sure.

YegorStepanov:
Are both solutions really needed? They look very similar: 
```C#
[AutomaticBaseline(AutomaticBaselineMode.Fastest)]
[AutomaticBaseline("Mean")] // how to name it?
```

#### AutomaticBaselineMode.Fastest
As I wrote above, it is not clear what the `Fastest` means.

Why is it compared by the `Mean` and not the `Median`? 
Maybe BDN calculates `Fastest` benchmark with complex formulas based on benchmark iterations?

If we change it to the `Median`, users may think that is a bug, because by default the `Median` column is not shown, only the `Mean` column.

Serg046:
> Are both solutions really needed?

No, I meant it would be great to suport it for manual/explicit mode as well as for the new AutomaticBaseline attribute. Something like that:
```csharp
public Benchmarks
{
    // Existing functionality with a new feature
    [Benchmark(Baseline = true, BaselineMetric = "Mean")] public void B1() { }
}
```
> Why is it compared by the Mean and not the Median?

Just my choice during development. I am fine to change this default to something else, e.g. `Median` like you say.

> If we change it to the Median, users may think that is a bug, because by default the Median column is not shown, only the Mean column.

Exactly

AndreyAkinshin:
The `Ratio` column is not just the ratio of means (and it shouldn't be a ratio of other statistics). The current implementation can be found in `BenchmarkDotNet.Mathematics.Statistics.Divide`, but I'm not happy with it. I'm working on a new approach based on [ratio functions](https://aakinshin.net/posts/shift-and-ratio-functions/) (which is a part of a new summary table with a better statistical engine; the research notes can be found in my [blog](https://aakinshin.net/tags/statistics/)).
tl;dr: avoid using specific metrics in the baseline API.

PR #2172, opened 2022-10-26
Cleanup NuGet.config

Youssef1313:


NuGet.Config
```diff
@@ -8,11 +8,5 @@
     <clear />
 
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
-    <add key="perfolizer-nightly" value="https://www.myget.org/F/perfolizer/api/v3/index.json" />
-    <!-- reuquired to run Mono Wasm benchmarks -->
```
adamsitnik, line 12: these comments make it clear why we need these feeds or I am missing something?

adamsitnik (changes requested):

NuGet.Config
Youssef1313: I thought if everything restores correctly then it should be fine?

My assumption when I opened this PR was that they were needed at that time, but now the needed packages were pushed to NuGet.org.

Youssef1313: At the very least, perfolizer feed can be removed since the version used is already on NuGet.org

https://github.com/dotnet/BenchmarkDotNet/blob/64c3a3cb4ad4057b6065a2783a4b333206f8b5e5/src/BenchmarkDotNet/BenchmarkDotNet.csproj#L23

AndreyAkinshin: @Youssef1313 you are right, perfolizer-nightly can be safely removed.

adamsitnik: > I thought if everything restores correctly then it should be fine?

BenchmarkDotNet builds code when running benchmarks, so you don't need Mono AOT feeds until you run Mono AOT benchmarks. You would need to cleanup your local nuget cache and run Mono AOT benchmarks for .NET 6 and 7 to ensure that everything is fine when the feeds are gone..

Youssef1313: Thanks for clarifying @AndreyAkinshin @adamsitnik 

I sent you an email to discuss this further

adamsitnik (changes requested):
It's OK to remove `perfolizer-nightly` and `dotnet5` feeds, but we need to keep `dotnet6` and `dotnet7` feeds. Please update the PR accordingly.

adamsitnik (approved):
LGTM, thank you for your contribution @Youssef1313 !

PR #2173, opened 2022-10-27
Don't store references in Consumer

### LINKED ISSUE #1942: Consider changing `Consume` to not hold onto references for very long
Currently some of the `Consume` methods stash the reference passed to them in a local field, keeping the object alive as long as the consume object itself remains live.

https://github.com/dotnet/BenchmarkDotNet/blob/b78ec96631c5065895255165b7113bf5062aa407/src/BenchmarkDotNet/Engines/Consumer.cs#L98-L109

If that object roots a large object graph or large array allocation it can alter GC behavior and impact benchmark timing.

Consider either immediately over-writing the local field with null or nulling out this field at the end of measurement intervals.

timcassell:
Fixes #1942

Also fixes boxing nullable value types (`default(T) == null` also returns true for nullable value types, causing the box to occur).

### DIFF
--- a/src/BenchmarkDotNet/Engines/Consumer.cs
+++ b/src/BenchmarkDotNet/Engines/Consumer.cs
@@ -30,8 +30,6 @@ namespace BenchmarkDotNet.Engines
         private double doubleHolder;
         private long longHolder;
         private ulong ulongHolder;
-        private string stringHolder;
-        private object objectHolder;
         private IntPtr ptrHolder;
         private UIntPtr uptrHolder;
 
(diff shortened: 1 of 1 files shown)


adamsitnik (commented):
This code is so old that some parts of it "predates me" (word I learned at MS). If I remember correctly, we were doing the volatile writes for some good reasons, I just don't remember why exactly (to prevent from Out-of-order execution?).

@AndreyAkinshin could you please take a look?

timcassell:
From its usage, it looks like this is just for preventing dead code elimination, and a convenient method for consuming iterables (like the `IntroDeferredExecution` sample).

AndreyAkinshin:
As Adam well noted, we use volatile writes to prevent out-of-order execution. The suggested change can have unpredictable effects on the measurements in nanobenchmarking. Another severe issue is inlining: Consumer methods are aggressively inlined so that the instruction pointer stays inside the scope of the current method. `DeadCodeEliminationHelper` has NoInlining policy, which forces the instruction pointer to jump to a random place. It also can be harmful for nanobenchmarks.

As far as I know, we use `DeadCodeEliminationHelper` only in the jitting phase (we want to make sure that all methods are fully jitted, while we don't care about the accuracy of the measurements). It should not be used in actual measurements.

timcassell:
@AndreyAkinshin That consume method is repeated in the overhead measurement, so I don't see how it will negatively affect nano benchmarks.

The other option would be to immediately overwrite the field with null, which I thought was less desirable than calling `DeadCodeEliminationHelper`.

And the third option is to null the field after the loop, but that could still affect the GC during the loop.

How do you suggest we fix the issue?

AndreyAkinshin:
> The other option would be to immediately overwrite the field with null

I like this option. @adamsitnik what do you think?

timcassell:
#2191

PR #2174, opened 2022-10-27
Added the ability to not run with Evaluation Overhead

mrsharm:
For the GC team, doing an equitable comparison without evaluation overhead is important as we do want an equal number of GCs amongst comparands and the baseline we use to run our microbenchmarks with. 

This PR allows us to specify the flag to not run with evaluation overhead, a feature previously only available with the fluent API.

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
@@ -218,6 +218,9 @@ namespace BenchmarkDotNet.ConsoleArguments
         [Option("noForcedGCs", Required = false, HelpText = "Specifying would not forcefully induce any GCs.")]
         public bool NoForcedGCs { get; set; }
 
+        [Option("noEvaluationOverhead", Required = false, HelpText = "Specifying would not run the evaluation overhead iterations.")]
+        public bool NoEvaluationOverhead { get; set; }
+
         [Option("resume", Required = false, Default = false, HelpText = "Continue the execution if the last run was stopped.")]
         public bool Resume { get; set; }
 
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
```diff
@@ -218,6 +218,9 @@ public bool UseDisassemblyDiagnoser
         [Option("noForcedGCs", Required = false, HelpText = "Specifying would not forcefully induce any GCs.")]
         public bool NoForcedGCs { get; set; }
 
+        [Option("noEvaluationOverhead", Required = false, HelpText = "Specifying would not run the evaluation overhead iterations.")]
```
adamsitnik, line 221: subjective opinion: `noOverheadEvaluation` sounds more intuitive than `noEvaluationOverhead`. What do you think?

```suggestion
        [Option("noOverheadEvaluation", Required = false, HelpText = "Specifying would not run the evaluation overhead iterations.")]
```

adamsitnik (commented):
Overall, it looks good, but I would like to discuss the name of the argument before I hit the merge button. Thank you @mrsharm !

tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
```diff
@@ -572,5 +572,17 @@ public void UserCanSpecifyNoForceGCs()
                 Assert.False(job.Environment.Gc.Force);
             }
         }
+
+        [Fact]
+        public void UsersCanSpecifyWithoutOverheadEvalution()
+        {
+            var parsedConfiguration = ConfigParser.Parse(new[] { "--noEvaluationOverhead" }, new OutputLogger(Output));
```
adamsitnik, line 579:
```suggestion
            var parsedConfiguration = ConfigParser.Parse(new[] { "--noOverheadEvaluation" }, new OutputLogger(Output));
```

adamsitnik (approved):
LGTM, thank you @mrsharm !

PR #2176, opened 2022-10-28
Fix #1521

### LINKED ISSUE #1521: Iteration setup and cleanup causes job baseline error with multiple runtimes passed to BenchmarkSwitcher
Consider this benchmark class...
```c#
public class Benchmarks
{
    [GlobalSetup()]
    public void GlobalSetup() { }

    [IterationSetup()]
    public void IterationSetup() { }

    [IterationCleanup()]
    public void IterationCleanup() { }

    [GlobalCleanup()]
    public void GlobalCleanup() { }

    [Benchmark(Baseline = true)]
    public void Benchmark1() => Thread.Sleep(10);

    [Benchmark()]
    public void Benchmark2() => Thread.Sleep(20);

    [Benchmark()]
    public void Benchmark3() => Thread.Sleep(30);
}
```
With v0.12.1, if I use `BenchmarkRunner`...
```c#
BenchmarkRunner.Run<Benchmarks>(
    DefaultConfig.Instance
        .AddJob(Job.ShortRun.WithRuntime(CoreRuntime.Core31))
        .AddJob(Job.ShortRun.WithRuntime(ClrRuntime.Net48))
);
```
...launched with this command...
<pre>dotnet run --configuration release --framework netcoreapp3.1</pre>
...benchmarking succeeds.

If I try to do the same thing using `BenchmarkSwitcher`...
```c#
```
...launched with this command...
<pre>dotnet run --configuration release --framework netcoreapp3.1 -- --filter * --job Short --runtimes netcoreapp3.1 net48</pre>
...it fails with...

> // Validating benchmarks:
> Only 1 job in a group can have "Baseline = true" applied to it, group * in class Benchmarks has 3

I found that benchmarking succeeds if I do any of the following:
 - Remove **both** `[IterationSetup()]` and `[IterationCleanup()]`
 - Change `--job Short` to `--job Dry`
 - Pass only one runtime to `--runtimes`

Removing one or both of `[GlobalSetup()]` and `[GlobalCleanup()]` makes no difference.

I would also note that in the `BenchmarkRunner` results table I get two per-runtime groups of three methods with `Benchmark1` for each runtime being the baseline, whereas in the `BenchmarkSwitcher` results I get one group of six methods with `Benchmark1` for `netcoreapp3.1` being the baseline; it appears the latter is limited to using only the first runtime specified as a baseline.  I tried working around that with...
```c#
BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(
    args,
    DefaultConfig.Instance.AddJob(
        Job.Default.AsBaseline().AsDefault()
    )
);
```
...but I get...

> Only 1 job in a group can have "Baseline = true" applied to it, group * in class Benchmarks has 6

...with `[IterationSetup()]` and/or `[IterationCleanup()]`, and...

> Only 1 job in a group can have "Baseline = true" applied to it, group * in class Benchmarks has 2

...without.

YegorStepanov:


adamsitnik (approved):

PR #2178, opened 2022-10-29
Introduce MockToolchain

AndreyAkinshin:
This new toolchain allows writing simple unit tests for summary tables without a need to actually run benchmarks. The usage of the new API is shown in `RatioColumnTest`.


AndreyAkinshin:

tests/BenchmarkDotNet.Tests/Mocks/MockHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using Xunit.Abstractions;
+
+namespace BenchmarkDotNet.Tests.Mocks
+{
+    public static class MockHelper
+    {
+        public static Summary Run<T>(ITestOutputHelper output)
```
adamsitnik, line 12: We could make this method accept a mandatory `Func<BenchmarkCase, List<Measurement>>` argument, pass it to `MockToolchain` instance and remove the need of having a custom attribute, interface and reflection.

adamsitnik (commented):
Overall it LGTM, but I believe that we could simplify it.

tests/BenchmarkDotNet.Tests/Mocks/MockHelper.cs
AndreyAkinshin: @adamsitnik good idea, I like it! I rewrote the implementation, could you please check out the new version?

AndreyAkinshin:
@adamsitnik I just merged #2268 to `mocktoolchain`, thanks for the edits!
Can I merge this PR now?

adamsitnik (approved):
LGTM, and great idea BTW!

PR #2179, opened 2022-10-29
Improve appearance of validation output

YegorStepanov:
Before #2148, like v0.13.2:
v013
Master:
Master
PR:
PR

## no validation errors:
Master:
image

PR:
image

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -52,7 +52,7 @@ namespace BenchmarkDotNet.Running
             {
                 var compositeLogger = CreateCompositeLogger(benchmarkRunInfos, streamLogger);
 
-                compositeLogger.WriteLineInfo("// Validating benchmarks:");
+                compositeLogger.WriteLineHeader("// ***** Validating benchmarks *****");
 
                 var (supportedBenchmarks, validationErrors) = GetSupportedBenchmarks(benchmarkRunInfos, resolver);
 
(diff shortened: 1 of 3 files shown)


YegorStepanov:
What is the purpose of the double slash? Are they meant to show a newline in the console? 
For example:
image

AndreyAkinshin:
We use double slashes to distinguish user messages from the payload (the results of the measurements).
@adamsitnik can we avoid using slashes with the help of our new [Pipe approach](https://github.com/dotnet/BenchmarkDotNet/pull/2092)?

YegorStepanov:
>can we avoid using slashes with the help of our new https://github.com/dotnet/BenchmarkDotNet/pull/2092?

Yes, if we remove the first line or pass the `NullLogger` to the broker, the `measurements` and input from generated project will not be printed.

https://github.com/dotnet/BenchmarkDotNet/blob/3906381eaa7a7a28dc8c68eb7582626c858b68e4/src/BenchmarkDotNet/Loggers/Broker.cs#L45-L51

#### But how can we distinguish user input from `measurements` (it requires to implement silent mode)?

```C#
[Benchmark]
// I haven't tried this, but it looks like we can break the engine.
public void M() => Console.WriteLine("OverheadJitting  100500: 99 op, 24 h, 0 us/op ");
```

#### Because it's just parsing the input:
https://github.com/dotnet/BenchmarkDotNet/blob/32bb2deea8771fe5aadc38063b8a3fe1495ccf81/src/BenchmarkDotNet/Reports/Measurement.cs#L105-L124

PR #2180, opened 2022-10-29
Fix race in AsyncProcessOutputReader

AndreyAkinshin:
The process.WaitForExit call doesn't guarantee that the last calls of ProcessOnOutputDataReceived and ProcessOnErrorDataReceived are processed. The subsequent call of reader.StopRead detaches the *DataReceived events, which leaves the output/error queues incomplete.
This race leads to flakiness of various tests (e.g., AllSetupAndCleanupMethodRunsForSpecificBenchmark).

### DIFF
--- a/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
+++ b/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
@@ -10,9 +10,12 @@ namespace BenchmarkDotNet.Loggers
     internal class AsyncProcessOutputReader : IDisposable
     {
         private readonly Process process;
-        private readonly ConcurrentQueue<string> output, error;
-        private readonly bool logOutput, readStandardError;
         private readonly ILogger logger;
+        private readonly bool logOutput, readStandardError;
+
+        private static readonly TimeSpan FinishEventTimeout = TimeSpan.FromMilliseconds(500);
+        private readonly AutoResetEvent outputFinishEvent, errorFinishEvent;
+        private readonly ConcurrentQueue<string> output, error;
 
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
It seems that this race is responsible for most of our build failures.

src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
```diff
@@ -103,34 +115,44 @@ private void Detach()
...
                     {
-                        logger.WriteLine(e.Data);
+                        lock (this) // #2125
+                        {
+                            logger.WriteLine(e.Data);
+                        }
                     }
                 }
             }
+            else
```
adamsitnik, line 133: nit: it would be nice to add a comment that `null` means EOF

adamsitnik (approved):
LGTM, big thanks for solving the test flakiness!!

src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs
AndreyAkinshin: Done.

AndreyAkinshin:
Now the situation is better, but the race is not fully fixed: https://github.com/dotnet/BenchmarkDotNet/actions/runs/3393415684/jobs/5640726686#step:5:220
@adamsitnik do you any ideas how we can improve the current solution?

adamsitnik:
@AndreyAkinshin I don't have any good solutions to suggest at the moment. But as a workaround, we could change these benchmarks to write to a file rather than print to the console. But I know it's far from ideal.

PR #2183, opened 2022-11-01
Remove allowMultiple=true from column attributes

YegorStepanov:
We can remove attributes because **ColumnConfigBaseAttribute** already has
```C#
[AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly)]
```


AndreyAkinshin (approved):

AndreyAkinshin:
@YegorStepanov thanks!

PR #2184, opened 2022-11-01
Fix #1715

### LINKED ISSUE #1715: Description with spaces and MethodTitle quotes
I've noticed that having a white space in the Description of a benchmark causes additional quotes to be added to the MethodTitle (all exporters).

Method | Mean | Error | StdErr | Min | Max | Median | Ratio | MannWhitney(5%) | Gen 0 | Gen 1 | Gen 2 | Allocated
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
'new (value, unit)' | 743.6 ns | NA | 0.0 ns | 743.6 ns | 743.6 ns | 743.6 ns | 1.00 | Base | - | - | - | -
FromUnit(quantityValue) | 1,395.2 ns | NA | 0.0 ns | 1,395.2 ns | 1,395.2 ns | 1,395.2 ns | 1.88 | ? | - | - | - | -
Quantity.From(value,unit) | 3,785.3 ns | NA | 0.0 ns | 3,785.3 ns | 3,785.3 ns | 3,785.3 ns | 5.09 | ? | - | - | - | 32 B
'Quantity&lt;Q,U&gt;.From(value, unit)' | 5,232.6 ns | NA | 0.0 ns | 5,232.6 ns | 5,232.6 ns | 5,232.6 ns | 7.04 | ? | - | - | - | 56 B
'Quantity.From(value, randomUnit)' | 8,620.2 ns | NA | 0.0 ns | 8,620.2 ns | 8,620.2 ns | 8,620.2 ns | 11.59 | ? | - | - | - | 32 B

This was also the case in v0.12 I just _really_ noticed it now when I tried to 'echo' a json string

YegorStepanov:
fix #1715

### DIFF
--- a/src/BenchmarkDotNet/Running/Descriptor.cs
+++ b/src/BenchmarkDotNet/Running/Descriptor.cs
@@ -51,21 +51,13 @@ namespace BenchmarkDotNet.Running
             IterationCleanupMethod = iterationCleanupMethod;
             OperationsPerInvoke = operationsPerInvoke;
             AdditionalLogic = additionalLogic ?? string.Empty;
-            WorkloadMethodDisplayInfo = FormatDescription(description) ?? workloadMethod?.Name ?? "Untitled";
+            WorkloadMethodDisplayInfo = description ?? workloadMethod?.Name ?? "Untitled";
             Baseline = baseline;
             Categories = categories ?? Array.Empty<string>();
             MethodIndex = methodIndex;
         }
 
         public override string ToString() => DisplayInfo;
-
-        private static string FormatDescription([CanBeNull] string description)
-        {
-            var specialSymbols = new[] { ' ', '\'', '[', ']' };
-            return description != null && specialSymbols.Any(description.Contains)
-                ? "'" + description + "'"
-                : description;
-        }
 
         public bool HasCategory(string category) => Categories.Any(c => c.EqualsWithIgnoreCase(category));
 


AndreyAkinshin:
I'm OK with removing quotes, but it's a breaking change that can cause problems for BenchmarkDotNet users who parse the output of Exporters. If we introduce this change, we should provide a way to rollback to the previous behavior.

YegorStepanov:
Done.

Instead of extending `SummaryStyle`, I was thinking of adding a global attribute:
```C#
[assembly: RevertOldBehaviourForBenchmarkName]
```

The attribute seems more elegant for this.

YegorStepanov:
`[assembly: RevertOldBehaviourForBenchmarkName]`
## Option 1
it will work like this:
```C#
//BenchmarkConverter.cs
//type is benchmark type
BenchmarkRunInfo MethodsToBenchmarksWithFullConfig(Type type, ...)
{
    bool revertOldBehaviour = type.Assembly.GetCustomAttributes().OfType<RevertOldBehaviourForBenchmarkName>();
    var targets = GetTargets(..., revertOldBehaviour);
}
```

##### cons:
- benchmarks can be defined in different assemblies.
- hard to test it

## Option 2
Instead, we can make `SummaryStyle.OldBehaviorForBenchmarkName` internal and only allow it to set via an assembly attribute.
##### cons
- hard to test (we can simply remove `RevertOldBehaviorTest`)


# Changelog
Maybe it makes sense to add a "breaking change" paragraph to the changelog? 

Users will notice this breaking change when they break their CI or erase benchmark history (because the name is changed).

The `BreakingChange.md` will also be useful. I can help with it (also for few older version).

AndreyAkinshin:
1. It would be better to use a more meaningful option name instead of `oldBehaviorForBenchmarkName` (you never know how many behavior generations we will have in the future). E.g., `QuoteBenchmarkNamesWithSpecialChars`.
2. We need a flexible way to control this behavior. The approach with assembly attributes is not flexible enough, it can be inconvenient for some users. Meanwhile, I don't really want to extend the Config API (which is already quite extensive) with additional temporary properties.

I was thinking about an approach inspired by the .NET Runtime knobs. We can introduce `string? IConfig.GetKnob(string key)` that returns the value of the given knob (additionally, we can introduce extension methods like `bool? IConfig.GetKnobBool(string key)` which implements the bool parsing logic). The Knob can be specified via `ManualConfig` (e.g., `ManualConfig.WithKnob(string key, string value)`), an attribute (`[Knob(string key, string value)]`), or via environment variables (e.g., `BDN_<KnobKey>`).

@adamsitnik @YegorStepanov what do you think about such a design?

YegorStepanov:
I like it. Will be done.

Does it make sense to add the `Knob` class?
```C#
public class Knob
{
     public const string QuoteBenchmarkNamesWithSpecialChars = "...";
}

[Knob(Knob.QuoteBenchmarkNamesWithSpecialChars, true)]
```

AndreyAkinshin:
@YegorStepanov, I was thinking about something like this:
```cs
public abstract class Knob<T>
{
    public string Key { get; }
    public T DefaultValue { get; }

    protected Knob(string key, T defaultValue)
    {
        Key = key;
        DefaultValue = defaultValue;
    }

    protected abstract T Parse(string value);

    public T GetValue(IConfig config)
    {
        string value = config.GetKnobValue(Key) ?? Environment.GetEnvironmentVariable("BDN_" + Key);
        return Parse(value);
    }
}

public class BoolKnob : Knob<bool>
{
    public BoolKnob(string key, bool defaultValue) : base(key, defaultValue) { }

    protected override bool Parse(string value) =>
        value?.ToLowerInvariant() switch
        {
            "0" or "f" or "false" => false,
            "1" or "t" or "true" => true,
            _ => DefaultValue
        };
}

public class Knobs
{
    public const string QuoteBenchmarkNamesWithSpecialCharsKey = "QuoteBenchmarkNamesWithSpecialChars";

    public BoolKnob QuoteBenchmarkNamesWithSpecialChars = new (QuoteBenchmarkNamesWithSpecialCharsKey, false);
}
```

P.S. Let's wait for an opinion by @adamsitnik before we start to implement this.

PR #2191, opened 2022-11-08
Don't hold onto references in Consumer

timcassell:
Fixes https://github.com/dotnet/BenchmarkDotNet/issues/1942

Also fixes boxing nullable value types.


AndreyAkinshin (approved):
It seems that everything is OK (at least, I wasn't able to find any regressions).

PR #2202, opened 2022-11-20
Fix job filtering

YegorStepanov:
Currently, the jobs with same reference are excluded, but if you use any `.With*` method, a new job will be created.

```C#
DefaultConfig.Instance
    .AddJob(Job.Dry)
    .AddJob(Job.Dry) // not added
    .AddJob(Job.Dry.WithRuntime(CoreRuntime.Core70))
    .AddJob(Job.Dry.WithRuntime(CoreRuntime.Core70));
```

```C#
// 1 job added
[DryJob]
[DryJob]
// 2 jobs added
[DryJob(RuntimeMoniker.Net70)]
[DryJob(RuntimeMoniker.Net70)]
```

#### Maybe we shouldn't distinct jobs at all?
https://github.com/dotnet/BenchmarkDotNet/blob/1fb1015556039664451dd101b4e2ef9501e03805/src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs#L213-L218

### DIFF
--- a/src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
+++ b/src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs
@@ -215,7 +215,7 @@ namespace BenchmarkDotNet.Configs
         /// </summary>
         private static IReadOnlyList<Job> GetRunnableJobs(IEnumerable<Job> jobs)
         {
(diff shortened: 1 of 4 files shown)


adamsitnik (approved):
It looks good to me. To be honest I am suprised that the fix was so simple!

Thank you @YegorStepanov !

PR #2206, opened 2022-11-28
Add single quote when use pattern with filters

erlangxk:


AndreyAkinshin (changes requested):
@erlangxk, you have used the wrong type of quotes (`‘’` instead of `''`)

erlangxk:
very weried, I don't know what is that single quote I used in the first commit. maybe I use some non-English input method that time.

AndreyAkinshin (approved):

AndreyAkinshin:
@erlangxk thanks!

AndreyAkinshin:
@erlangxk I just realized that this PR was to the `docs-stable` branch. According to the current workflow, all the documentation changes should be merged to the `master` branch and cherry-picked to the `docs-stable`. Once another release version is published, `docs-stable` is moving to the latest `master`. This allows us to maintain a nicer git history and track changes in a more transparent way. Sorry, but I lost your commit in `v0.13.3`. Just cherry-picked it to `master`, it will be included in `v0.13.4`.

PR #2207, opened 2022-11-28
Avoid hangs when starting benchmark process fails

### LINKED ISSUE #2189: broker.ProcessData() hangs if something wrong is the spawned process
Let us a consider a job with invalid configuration on the current platform. For example, let us run the following benchmark on mac M1:

```cs
public class Bench
{
    [Benchmark] public void Foo() => Thread.Sleep(10);
}

BenchmarkRunner.Run<Bench>(DefaultConfig.Instance.AddJob(Job.Default.WithPlatform(Platform.X64)));
```

In this case, BenchmarkDotNet hangs in `BenchmarkDotNet.Toolchains.DotNetCli.DotNetCliExecutor.Execute` on `broker.ProcessData()`.

@adamsitnik do you have any thoughs on solving this problem?

adamsitnik:
fixes #2189

### DIFF
--- a/src/BenchmarkDotNet/Loggers/Broker.cs
+++ b/src/BenchmarkDotNet/Loggers/Broker.cs
@@ -1,10 +1,12 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Pipes;
+using System.Threading;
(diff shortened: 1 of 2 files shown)


AndreyAkinshin (approved):

AndreyAkinshin:
@adamsitnik thanks, now it works without hangs.

PR #2221, opened 2022-12-08
NuGet.org should be the default feed for NativeAOT 7.0 ILCompiler

adamsitnik:
cc @MichalStrehovsky


MichalStrehovsky:
Ideally, we should just switch to `<PublishAot>true</PublishAot>` and stop referencing the package altogether. If a specific version is desired `<RuntimeFrameworkVersion>` should be able to control the version of ILC that is used, same as for CoreCLR-JIT.

PR #2224, opened 2022-12-08
[DRAFT] Public Tool Chains

naricc:
I am experimenting with making enough things in BenchmarkDotnet public so that users of BDN can make their own tool chains, which would allow maximum flexibility for BDN users to control how benchmarks are built an executed. 

Experimenting with the WasmToolChain, this seems to be the minimal set of changes that need to be made public for that to happen. However, I don't know if we want to just make these things public or if we should hide them behind new interfaces, and what kind of interface would be best if so.  Thus I have created a draft PR to get some feedback.

### DIFF
--- a/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
@@ -121,7 +121,7 @@ namespace BenchmarkDotNet.Extensions
             }
         }
 
-        internal static void SetEnvironmentVariables(this ProcessStartInfo start, BenchmarkCase benchmarkCase, IResolver resolver)
+        public static void SetEnvironmentVariables(this ProcessStartInfo start, BenchmarkCase benchmarkCase, IResolver resolver)
         {
             if (benchmarkCase.Job.Environment.Runtime is ClrRuntime clrRuntime && !string.IsNullOrEmpty(clrRuntime.Version))
                 start.EnvironmentVariables["COMPLUS_Version"] = clrRuntime.Version;
--- a/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
@@ -121,7 +121,7 @@ namespace BenchmarkDotNet.Extensions
             }
         }
 
-        public static void SetEnvironmentVariables(this ProcessStartInfo start, BenchmarkCase benchmarkCase, IResolver resolver)
+        internal static void SetEnvironmentVariables(this ProcessStartInfo start, BenchmarkCase benchmarkCase, IResolver resolver)
         {
             if (benchmarkCase.Job.Environment.Runtime is ClrRuntime clrRuntime && !string.IsNullOrEmpty(clrRuntime.Version))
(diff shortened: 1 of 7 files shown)


timcassell:
I'm not fond of exposing internal types like this, it makes it harder to change implementation details. An interface is better, but still seems like it would tie the hands of the implementation for these particular methods you exposed. Should the toolchain implementer really care about the `AsyncProcessOutputReader`? Can these calls be moved a level above the `IToolchain` without the toolchain using them directly?

I haven't looked deeply at how creating new toolchains works, but this looks pretty complex. I would be interested in making a Unity toolchain (#1860) if this is possible.

naricc:
@timcassell I'm definitely not planning to expose all these by justing making them public types. But these are basically the things I found that need to be available some how to make tool chains implementable outside BDN. The point of this draft PR is to have something concrete to point at to solicit feedback on how to design this.

Basically, the problem we keep having on the mono team is that we need to either do an experiment where we build benchmarks in some unusual way, or else just update how they are built. And if every time we do that it requires a PR against BDN, then one on dotnet/performance to pull the new version, it takes a lot longer.

So, for these things, I can see basically three categories:

1. Just make them public (probably acceptable for some things).
2. Hide them behind some kind of interface.
3. Just say that tool chain implementors need to implement their own.

YegorStepanov:
`ProcessExtensions.SetEnvironmentVariables` is an extension method. BDN do not have **public** extension methods to BCL types.

Maybe it's easier to "duplicate" the interesting internal classes in your repo, and call the implementation through reflection? For stability, you only need to write unit tests to check that BDN hasn't changed the signatures.

For example you need `ResourceHelper.LoadTemplate`:
```C#
//It makes sense to write a reflection helper to call the implementation in one line

//Not sure what the signature of `InvokeMethod` should be to be able to have any parameters and any return type
//Made like `Func`: 
//InvokeMethod<Param1, Param2, ..., ReturnType>(string methodName, Param1, Param2, ...) 
public static class ResourceHelper
{
    public static string LoadTemplate(string name) => 
        BenchmarkReflection
            .FindType(nameof(ResourceHelper))
            .InvokeMethod<string, string>(nameof(LoadTemplate), name);
}
```

naricc:
@YegorStepanov  Is there some reason you think it would be better to use reflection instead of say, passing a function to BDN?

naricc:
I am closing this for now. I think I got some ideas how to do this correctly, but if you have any more comments feel free.

YegorStepanov:
@naricc I misread the PR description. I was thinking about being able to implement custom toolchains (for example Unity/SourceGenerator) without forking a BDN. 

I'm not sure if it's possible to make a user-friendly API because a lot of internal extension methods/helpers/classes/ should be changed to public. 

> Basically, the problem we keep having on the mono team is that we need to either do an experiment where we build benchmarks in some unusual way, or else just update how they are built. And if every time we do that it requires a PR against BDN, then one on dotnet/performance to pull the new version, it takes a lot longer.

If you need access to a few internal helpers or extension methods, I think getting them though reflection is the fastest and easiest way. Because different mono experiments require different internal types (I just guessing).

FYI: If you need a lot of internal logic there is [IgnoresAccessChecksToGenerator](https://github.com/aelij/IgnoresAccessChecksToGenerator) package which gives access to all internal types (I haven't tested it)

PR #2230, opened 2022-12-14
Add support for .NET SDK that uses Mono instead CLR as a default VM

adamsitnik:
fixes https://github.com/dotnet/performance/issues/2787


adamsitnik:
The failures are unrelated, I was able to reproduce them on master branch:

image

PR #2243, opened 2023-01-11
JitStatsDiagnoser

adamsitnik:
I'm currently working on improving System.CommandLine (mainly design, but also perf).

One of the things unique about S.CL is that it's typically used just once, during process startup time. So what matters is the number of JIT compilations. I've implemented this simple diagnoser that prints simple JIT stats.

Sample usage:

```cmd
dotnet run -c Release -f net7.0 --filter *IntroBasic.Sleep --profiler jit
```

| Method |     Mean |    Error |   StdDev | Methods JITted | Methods Tiered | JIT allocated memory |
|------- |---------:|---------:|---------:|---------------:|---------------:|---------------------:|
|  Sleep | 15.53 ms | 0.034 ms | 0.032 ms |          1,102 |             15 |            221,736 B |

@AndyAyersMS @kunalspathak @stephentoub @EgorBo @kouvel I am opened to extending it, but I need your feedback before I invest my time. I know that PerfView offers more:

image

### DIFF
--- a/src/BenchmarkDotNet.Diagnostics.Windows/InliningDiagnoser.cs
+++ b/src/BenchmarkDotNet.Diagnostics.Windows/InliningDiagnoser.cs
@@ -1,12 +1,13 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using BenchmarkDotNet.Diagnosers;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Running;
 using Microsoft.Diagnostics.Tracing.Session;
 
 namespace BenchmarkDotNet.Diagnostics.Windows
 {
-    public class InliningDiagnoser : JitDiagnoser<object>
+    public class InliningDiagnoser : JitDiagnoser<object>, IProfiler
     {
         private static readonly string LogSeparator = new string('-', 20);
 
@@ -42,6 +43,8 @@ namespace BenchmarkDotNet.Diagnostics.Windows
         }
 
         public override IEnumerable<string> Ids => new[] { nameof(InliningDiagnoser) };
+
+        public string ShortName => "inlining";
 
         protected override void AttachToEvents(TraceEventSession session, BenchmarkCase benchmarkCase)
         {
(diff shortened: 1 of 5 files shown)


adamsitnik:
Also, I am not sure if `TieredCompilationBackgroundJitStart` is just "methods tiered" or "the number of times tiered JIT kicked in and started working"

EgorBo:
@adamsitnik nice! `Methods JITted` includes both tier0 and tier1, right?

adamsitnik:
> Methods JITted includes both tier0 and tier1, right?

This is the number of `MethodJittingStarted` CLR events (that is all I know right now).

EgorBo:
Ah, right.
Also, it'd be nice for the diagnoser to generate a `MethodsJitted.md` report that will list all the methods jitted 🙂 can probably contribute that myself if you think it makes sense

adamsitnik:
> think it makes sense

I do, please go ahead and send a PR.

BTW I was also wondering whether we should filter the methods by namespace and for example remove all BenchmarkDotNet* methods from the report.

kouvel:
> Also, I am not sure if `TieredCompilationBackgroundJitStart` is just "methods tiered" or "the number of times tiered JIT kicked in and started working"

This event is emitted when a background thread starts jitting to tier up and includes info about how many methods were pending jitting at that time. The corresponding `Stop` event is emitted when the background thread stops jitting, and includes info about how many methods were jitted between the `Start`/`Stop` events and how many methods are still pending jitting.

adamsitnik:
@kouvel big thanks for the feedback! I've sent a fix, PTAL: https://github.com/dotnet/BenchmarkDotNet/pull/2246

KeterSCP:
Are there any plans to make these diagnosers available via attributes?

adamsitnik:
@KeterSCP There are not, but you can send a PR that adds such an attribute. Here you can find existing attributes for similar diagnosers: https://github.com/dotnet/BenchmarkDotNet/tree/master/src/BenchmarkDotNet.Diagnostics.Windows/Configs

PR #2260, opened 2023-01-31
set PublicSign to true for every OS

adamsitnik:
@AndreyAkinshin lets see if the CI passes on non-Windows OSes


AndreyAkinshin (commented):
`<PublicSign>true</PublicSign>` disables strong names; I created another PR #2263 with a proper fix.

PR #2265, opened 2023-02-10
Update How To Run: include tests

calvinhsia:
Describe how to run Benchmarks in Tests

### DIFF
--- a/docs/articles/guides/how-to-run.md
+++ b/docs/articles/guides/how-to-run.md
@@ -4,7 +4,9 @@ uid: docs.how-to-run
 
 # How to run your benchmarks
 
-There are several ways to run your benchmarks. What is important is that **BenchmarkDotNet works only with Console Apps**. It does not support any other kind of application like ASP.NET, Azure WebJobs, etc.
+There are several ways to run your benchmarks. What is important is that **BenchmarkDotNet works only with Console Apps and in Tests**. It does not support any other kind of application like ASP.NET, Azure WebJobs, etc.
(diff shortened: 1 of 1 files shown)


AndreyAkinshin:
I do not think that we should provide such a guide. Instead, we should emphasize that it is not recommended to use unit test frameworks to execute benchmarks since these frameworks can add undesired performance impact (e.g., xunit runs all tests in parallel by default, which completely distorts all the measurements). We have our own execution system with filtering capabilities, so it is possible to run a specific benchmark from the terminal and control its behavior via additional command line options.

Also, the suggestions from this PR look confusing rather than useful; see the remarks below.

> You can author tests with BenchMarks

It is not clear how to write such tests.
"BenchMarks" - strange spelling

> control their running with the Test Explorer.

What does it mean? What kind of running aspects can be controlled?
Also, an IDE is implicitly assumed, while documentation should encourage running benchmarks from the terminal to avoid the IDE perf impact.

> To see results, scroll to the bottom of the Results pane and choose Open test log.

Again, a specific IDE is assumed.
Also, NUnit/MSTest are implicitly assumed since you will not get the test output out of the box with xunit.

> You can add the [Ignore] attribute to Test methods as desired.

A specific test framework is assumed.
Also, I don't think that we should specifically highlight such features of a particular test framework in such a guide.

adamsitnik:
I wholeheartedly agree with everything that @AndreyAkinshin wrote.

On top of that I can add: https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#benchmarks-are-not-unit-tests

PR #2267, opened 2023-02-10
Bump Cake to 3.0.0

AndreyAkinshin:


adamsitnik (approved):
LGTM, the failure is unrelated, it will be fixed in 7.0.3 when https://github.com/dotnet/runtime/pull/79862 makes it to the publicly available installer.

PR #2274, opened 2023-02-21
Add Android specific GetNetCoreVersion Runtime Information capabilities.

LoopedBard3:
This adds an Android specific way to get the NetCore version of the runtime being used via Environment.Version. This is necessary because Android apps don't uncompress the downloaded apk's which causes FileNotFound errors. To double check the precision, I ran this locally. After testing the Environment.Version vs the FileVersion from the PrivateCoreLib.dll, it does appears that the Environment.Version is not as precise. The results from the test were Envirornment.Version giving 7.0.3 while the FileVersion gave 7.0.323.6910.


adamsitnik (approved):
LGTM, thank you for the fix @LoopedBard3 !

PR #2285, opened 2023-03-10
Allow custom runtimes to use Executor

tgjones:
Hello folks! This PR makes a small change to `Executor` to allow it be used for custom runtimes. The only downside I can see is that  the current `NotSupportedException` is probably a useful reminder when adding built-in runtimes, to ensure the correct behaviour is added or re-used here.

My use-case: I've written a custom runtime, and currently I need to copy/paste all of `Executor` and its supporting classes into my own codebase, which seems a bit redundant when all I really want to change are these 3 lines. Of course this particular setup for `ProcessStartInfo` is not applicable for every custom runtime - but at least it allows `Executor` to be used for those runtimes for which this is applicable.

But of course you'll have a better view of whether this change is a good idea and / or acceptable.

### DIFF
--- a/src/BenchmarkDotNet/Environments/Runtimes/CustomRuntime.cs
+++ b/src/BenchmarkDotNet/Environments/Runtimes/CustomRuntime.cs
@@ -1,0 +1,12 @@
+using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Environments
+{
+    public abstract class CustomRuntime : Runtime
+    {
+        protected CustomRuntime(RuntimeMoniker runtimeMoniker, string msBuildMoniker, string displayName)
+            : base(runtimeMoniker, msBuildMoniker, displayName)
+        {
+        }
+    }
+}
--- a/src/BenchmarkDotNet/Toolchains/Executor.cs
+++ b/src/BenchmarkDotNet/Toolchains/Executor.cs
@@ -149,7 +149,9 @@ namespace BenchmarkDotNet.Toolchains
                     start.WorkingDirectory = artifactsPaths.BinariesDirectoryPath;
                     break;
                 default:
-                    throw new NotSupportedException("Runtime = " + runtime);
+                    start.FileName = exePath;
+                    start.Arguments = args;
+                    break;
             }
             return start;
         }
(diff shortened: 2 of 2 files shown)


AndreyAkinshin:
I believe it would be better to introduce a separate base class or interface for such a purpose. E.g., like this:
```diff
Index: src/BenchmarkDotNet/Toolchains/Executor.cs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/BenchmarkDotNet/Toolchains/Executor.cs b/src/BenchmarkDotNet/Toolchains/Executor.cs
--- a/src/BenchmarkDotNet/Toolchains/Executor.cs	(revision d4de82a3f0bcd71f1f695554eef343bbfc1357ea)
+++ b/src/BenchmarkDotNet/Toolchains/Executor.cs	(date 1678458305076)
@@ -148,6 +148,10 @@
                     start.Arguments = args;
                     start.WorkingDirectory = artifactsPaths.BinariesDirectoryPath;
                     break;
+                case CustomRuntime _:
+                    start.FileName = exePath;
+                    start.Arguments = args;
+                    break;
                 default:
                     throw new NotSupportedException("Runtime = " + runtime);
             }
Index: src/BenchmarkDotNet/Environments/Runtimes/CustomRuntime.cs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/BenchmarkDotNet/Environments/Runtimes/CustomRuntime.cs b/src/BenchmarkDotNet/Environments/Runtimes/CustomRuntime.cs
new file mode 100644
--- /dev/null	(date 1678458285120)
+++ b/src/BenchmarkDotNet/Environments/Runtimes/CustomRuntime.cs	(date 1678458285120)
@@ -0,0 +1,10 @@
+using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Environments
+{
+    public abstract class CustomRuntime : Runtime
+    {
+        protected CustomRuntime(RuntimeMoniker runtimeMoniker, string msBuildMoniker, string displayName) :
+            base(runtimeMoniker, msBuildMoniker, displayName) { }
+    }
+}
\ No newline at end of file
```
@tgjones would it cover your use case?

tgjones:
Thanks! That's a better idea - and yes, it would cover my use-case. I'll modify my PR to look like that.

tgjones:
@AndreyAkinshin I've applied your suggestion.

AndreyAkinshin:
@tgjones thanks!

PR #2309, opened 2023-05-17
Overhead match workload

### LINKED ISSUE #2305: overhead and workload invocation sequences diverge
If these don't match then benchmark times will be either under or over estimated.

Context https://github.com/dotnet/runtime/issues/86033#issuecomment-1546306621

This is the "good" case where the jit isn't messing up the workload invoke codegen (which just makes the problem worse).

```C#
        public delegate System.Single OverheadDelegate();

        private void OverheadActionUnroll(System.Int64 invokeCount)
        {
            
            {
                consumer.Consume(overheadDelegate());


        public delegate  System.Numerics.Vector3 WorkloadDelegate();

        {
            
            for (System.Int64  i = 0; i < invokeCount; i++)
            {
                consumer.Consume(workloadDelegate().X);
```

```asm
;; overhead

       mov      rbp, gword ptr [rsi+38H]
       mov      rax, gword ptr [rsi+28H]
       mov      rcx, gword ptr [rax+08H]
       call     [rax+18H]BenchmarkDotNet.Autogenerated.Runnable_0+OverheadDelegate:Invoke():float:this
       vmovss   dword ptr [rbp+48H], xmm0

;; workload

       mov      rbp, gword ptr [rsi+38H]
       mov      rax, gword ptr [rsi+30H]
       mov      rcx, gword ptr [rax+08H]
       lea      rdx, [rsp+110H]
       call     [rax+18H]BenchmarkDotNet.Autogenerated.Runnable_0+WorkloadDelegate:Invoke():System.Numerics.Vector3:this
       vmovss   xmm0, dword ptr [rsp+110H]
       vmovss   dword ptr [rbp+48H], xmm0
```

May be restricted to certain return types like Vector; I haven't seen this elsewhere.

timcassell:
Matched overhead to workload return type and implementation.
Optimized `Consumer.Consume<T>(in T value)`.
Use `Consumer` more liberally in benchmarks.
Cleaned up byref returns.
Fixed pointer returns in InProcessEmitToolchain.

Fixes #2305


AndreyAkinshin:
@timcassell thanks for the PR! It will take some time for me to properly review and test it, but I will try to do it soon.

As for now, I have one high-level concern: if you need C# 11 for these changes, please do it in a separate PR. It's a significant upgrade for the source code base, and I would prefer to apply it independently from the overhead-related refactoring.

timcassell:
I just did that for raw string literal convenience. I didn't need it otherwise. Reverted.

timcassell:
I don't know if anyone actually ran into it, but I noticed the InProcessEmitToolchain tests were missing pointer returns, so I added that test and fixed it here, too (extra confidence that my changes work for both csproj gen and inprocess toolchains).

Side note - the builds are failing now, but I don't think it's related to any of my changes.
```
bash: /home/runner/work/BenchmarkDotNet/BenchmarkDotNet/.dotnet/dotnet-install.sh: No such file or directory
/home/runner/work/BenchmarkDotNet/BenchmarkDotNet/build/Build.csproj : warning NU1701: Package 'BouncyCastle 1.8.9' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' instead of the project target framework 'net7.0'. This package may not be fully compatible with your project.
```

timcassell:
I'm seeing this error in the `ReferenceTests` (`FSharpIsSupported`) in net462. Any idea how to fix it?

```
CS0234: The type or namespace name 'Unsafe' does not exist in the namespace 'System.Runtime.CompilerServices' (are you missing an assembly reference?)
```

AndreyAkinshin:
@timcassell just wanted to say that the review is still in process. I tend to procrastinate merging pull requests touching toolchains. Since we do not have 100% test coverage, changes to these critical parts demand additional verification to prevent potential not-so-obvious bugs in corner cases.
Thank you for your patience!

AndreyAkinshin:
@timcassell, as I can see if the benchmark methods return `int`, we generate the following overhead method:

```
        private  System.Int32 __Overhead() // __ is to avoid possible name conflict
        {
            System.Runtime.CompilerServices.Unsafe.SkipInit(out System.Int32 value);
return value;
        }
```

I don't think that it's a good idea to use `SkipInit` instead of `return default`. In my local experiments for Windows+Mono, such a `SkipInit`-based overhead method may actually consume more time than `[Benchmark] public int Foo() => 0`.

I would suggest restoring the original behavior (`default` for primitive types; `null` for objects).

timcassell:
> I don't think that it's a good idea to use `SkipInit` instead of `return default`. In my local experiments for Windows+Mono, such a `SkipInit`-based overhead method may actually consume more time than `[Benchmark] public int Foo() => 0`.

Interesting, I thought that would be inlined and eliminated as a noop. I guess the Mono JIT isn't that powerful.

> I would suggest restoring the original behavior (`default` for primitive types; `null` for objects).

Should I do that for any type whose size is <= IntPtr.Size? Or <= 8 (would include double and long on 32-bit runtime)? Or only primitives and references?

AndreyAkinshin:
Hmm. This is strange. I have performed the following integration experiment. I wrote the following benchmark:

```cs
[Config(typeof(Config))]
public class Benchmarks
{
    {
        public Config()
        {
            var job = Job.Default
                    .WithRuntime(MonoRuntime.Default)
                    .WithIterationCount(100)
                    .WithInvocationCount(160_000_000)
                ;
            AddJob(job);
        }
    }
    [Benchmark] public int Foo() => 0;
}
```

Also, I hacked the BenchmarkDotNet sources to force it always perform 100 overhead iterations:

```diff
diff --git a/src/BenchmarkDotNet/Engines/EngineGeneralStage.cs b/src/BenchmarkDotNet/Engines/EngineGeneralStage.cs
--- a/src/BenchmarkDotNet/Engines/EngineGeneralStage.cs	(revision 56756271f63663c406b82077209a6e630d703b1d)
+++ b/src/BenchmarkDotNet/Engines/EngineGeneralStage.cs	(date 1686227871484)
@@ -32,7 +32,7 @@
         }
 
         public IReadOnlyList<Measurement> RunOverhead(long invokeCount, int unrollFactor)
-            => RunAuto(invokeCount, IterationMode.Overhead, unrollFactor);
+            => RunSpecific(invokeCount, IterationMode.Overhead, 100, unrollFactor);
 
         public IReadOnlyList<Measurement> RunWorkload(long invokeCount, int unrollFactor, bool forceSpecific = false)
             => Run(invokeCount, IterationMode.Workload, false, unrollFactor, forceSpecific);
```

Thus, we get 100 overhead measurements and 100 workload measurements.

Here are the corresponding density plots for the master branch:
bdn-stable


And here are the plots for your PR:
bdn-pr2309


As you can see, the left mode of the overhead distribution is approximately 1 CPU cycle larger than the left mode of the workload distribution.

Raw data:
[bdn-stable.csv](https://github.com/dotnet/BenchmarkDotNet/files/11688787/bdn-stable.csv)
[bdn-pr2309.csv](https://github.com/dotnet/BenchmarkDotNet/files/11688788/bdn-pr2309.csv)

timcassell:
@AndreyAkinshin How are you generating those plots? I want to try that on my machine.

AndreyAkinshin:
First, I generated csv from summary:

```cs
var summary = BenchmarkRunner.Run<Benchmarks>();
var builder = new StringBuilder();
Thread.CurrentThread.CurrentCulture = DefaultCultureInfo.Instance;
builder.AppendLine("duration,type");
foreach (var measurement in summary.Reports.First().AllMeasurements.Where(m => m.IsOverhead() && m.IterationStage == IterationStage.Actual))
    builder.AppendLine((measurement.Nanoseconds / measurement.Operations).ToString("N4") + ",overhead");
foreach (var measurement in summary.Reports.First().AllMeasurements.Where(m => m.IsWorkload() && m.IterationStage == IterationStage.Actual))
    builder.AppendLine((measurement.Nanoseconds / measurement.Operations).ToString("N4") + ",workload");
File.WriteAllText("bdn.csv", builder.ToString());
```

(Yes, we can introduce a new exporter or reuse the existing ones, but It was the fastest way for me to obtain the first results).

Next, I used the following R script to generate plots:

```r
# Run `install.packages("ggplot2")` to install the package for the first time
library(ggplot2)

df <- read.csv("bdn.csv")
ggplot(df, aes(duration, col = type)) +
  geom_density(bw = "SJ") +
  geom_rug() +
  labs(
    x = "Duration, ns",
    y = "Density",
    col = "Method"
  )
```

timcassell:
Very interesting. When I run with those forced iteration and invocation counts, I'm seeing similar results as you. Strange that my heuristic benchmarks were having different results. Anyway, I'll do some more testing with these forced invoke counts to see what types should have default vs skipinit, and update this PR accordingly.

timcassell:
Well, it looks like `default` is almost universally faster than `Unsafe.SkipInit` when just using it to return the value, even up to a struct size of 1024 bytes.

<details><summary>Results</summary>
<p>

```
// Mono

|              Method |       Mean |     Error |    StdDev |     Median |
|-------------------- |-----------:|----------:|----------:|-----------:|
|    ReturnDefaultInt |   3.392 ns | 0.0287 ns | 0.0268 ns |   3.397 ns |
|   ReturnSkipInitInt |   3.993 ns | 0.0254 ns | 0.0225 ns |   3.988 ns |

|         DefaultLong |   3.504 ns | 0.0093 ns | 0.0082 ns |   3.507 ns |
|        SkipInitLong |   4.262 ns | 0.0849 ns | 0.1934 ns |   4.224 ns |

|       DefaultObject |   5.468 ns | 0.0371 ns | 0.0329 ns |   5.472 ns |
|      SkipInitObject |   6.319 ns | 0.0840 ns | 0.0786 ns |   6.325 ns |

|     DefaultStruct16 |  14.526 ns | 0.1066 ns | 0.0890 ns |  14.551 ns |
|    SkipInitStruct16 |  15.296 ns | 0.3015 ns | 0.2821 ns |  15.180 ns |

|     DefaultStruct32 |  15.909 ns | 0.0884 ns | 0.0827 ns |  15.973 ns |
|    SkipInitStruct32 |  17.183 ns | 0.0700 ns | 0.0584 ns |  17.192 ns |

|     DefaultStruct64 |  23.040 ns | 0.1999 ns | 0.1870 ns |  22.989 ns |
|    SkipInitStruct64 |  25.419 ns | 0.0884 ns | 0.0690 ns |  25.442 ns |

|    DefaultStruct128 | 117.386 ns | 1.0833 ns | 0.9046 ns | 117.527 ns |
|   SkipInitStruct128 | 140.688 ns | 1.7634 ns | 1.6494 ns | 140.357 ns |

|  DefaultEmptyStruct |  10.989 ns | 0.0484 ns | 0.0429 ns |  10.995 ns |
| SkipInitEmptyStruct |  10.888 ns | 0.0538 ns | 0.0449 ns |  10.890 ns |

// .Net 7

|              Method |      Mean |     Error |    StdDev |    Median |
|-------------------- |----------:|----------:|----------:|----------:|
|    ReturnDefaultInt |  2.262 ns | 0.0103 ns | 0.0086 ns |  2.263 ns |
|   ReturnSkipInitInt |  2.209 ns | 0.0139 ns | 0.0116 ns |  2.211 ns |

|         DefaultLong |  2.168 ns | 0.0414 ns | 0.0580 ns |  2.141 ns |
|        SkipInitLong |  2.186 ns | 0.0271 ns | 0.0226 ns |  2.181 ns |

|       DefaultObject |  4.535 ns | 0.0781 ns | 0.0693 ns |  4.538 ns |
|      SkipInitObject |  4.767 ns | 0.0915 ns | 0.0940 ns |  4.760 ns |

|     DefaultStruct16 |  3.080 ns | 0.0606 ns | 0.1441 ns |  3.012 ns |
|    SkipInitStruct16 |  3.312 ns | 0.0639 ns | 0.0937 ns |  3.315 ns |

|     DefaultStruct32 |  3.323 ns | 0.0202 ns | 0.0169 ns |  3.322 ns |
|    SkipInitStruct32 | 10.750 ns | 0.1241 ns | 0.1100 ns | 10.722 ns |

|     DefaultStruct64 |  4.113 ns | 0.0551 ns | 0.0489 ns |  4.107 ns |
|    SkipInitStruct64 | 13.704 ns | 0.1115 ns | 0.0989 ns | 13.688 ns |

|    DefaultStruct128 |  5.124 ns | 0.0334 ns | 0.0296 ns |  5.127 ns |
|   SkipInitStruct128 | 16.788 ns | 0.3314 ns | 0.7062 ns | 16.597 ns |

|  DefaultEmptyStruct |  2.935 ns | 0.0227 ns | 0.0201 ns |  2.930 ns |
| SkipInitEmptyStruct |  2.740 ns | 0.0278 ns | 0.0260 ns |  2.728 ns |
```

</p>
</details>

I even tried adding `SkipLocalsInitAttribute`, and it did get slightly faster, but still not quite as fast as `default`.

|                   Method |     Mean |    Error |   StdDev |
|------------------------- |---------:|---------:|---------:|
|        DefaultStruct1024 | 21.17 ns | 0.361 ns | 0.338 ns |
|       SkipInitStruct1024 | 48.10 ns | 0.922 ns | 0.906 ns |
| SkipLocalsInitStruct1024 | 33.75 ns | 0.629 ns | 0.618 ns |
|          Struct1024Field | 32.06 ns | 0.401 ns | 0.375 ns |

So, I went ahead and removed `Unsafe.SkipInit` in favor of `default` for all types.

timcassell:
@AndreyAkinshin As an experiment, I tried writing a `GetUninitializedValue` function directly in IL for this purpose, and it actually runs faster for large structs (but still slower for primitives). We could include a BDN Unsafe DLL for this written in IL.

|                     Method |      Mean |     Error |    StdDev |    Median |
|--------------------------- |----------:|----------:|----------:|----------:|
|           ReturnDefaultInt |  2.294 ns | 0.0081 ns | 0.0067 ns |  2.293 ns |
|          ReturnSkipInitInt |  2.427 ns | 0.0627 ns | 0.1800 ns |  2.347 ns |
|        GetUninitializedInt |  5.703 ns | 0.0114 ns | 0.0101 ns |  5.705 ns |
|          DefaultStruct1024 | 20.835 ns | 0.0740 ns | 0.0618 ns | 20.832 ns |
|         SkipInitStruct1024 | 46.449 ns | 0.3205 ns | 0.2677 ns | 46.384 ns |
| GetUninitializedStruct1024 |  6.853 ns | 0.0306 ns | 0.0286 ns |  6.844 ns |

```cs
[Benchmark]
public Struct1024 GetUninitializedStruct1024()
    => BenchmarkDotNet.CompilerServices.Unsafe.GetUninitializedValue<Struct1024>();
```

```cil
.method public hidebysig static !!T 
        GetUninitializedValue<T>() cil managed aggressiveinlining 
{
    .maxstack 1
    .locals (
        [0] int32 'value'
    )

    IL_0007: ldloc.0
    IL_0008: ret
} // end of method Unsafe::GetUninitializedValue
```

Thoughts?

timcassell:
@AndreyAkinshin As an experiment, I tried writing a `GetUninitializedValue` function directly in IL for this purpose, and it actually runs faster for large structs (but still slower for primitives). We could include a BDN Unsafe DLL for this written in IL.

|                     Method |      Mean |     Error |    StdDev |
|--------------------------- |----------:|----------:|----------:|
|           ReturnDefaultInt |  2.304 ns | 0.0152 ns | 0.0127 ns |
|          ReturnSkipInitInt |  2.535 ns | 0.0483 ns | 0.0452 ns |
|        GetUninitializedInt |  5.520 ns | 0.0494 ns | 0.0438 ns |
|          DefaultStruct1024 | 21.490 ns | 0.3365 ns | 0.3147 ns |
|         SkipInitStruct1024 | 48.287 ns | 0.5382 ns | 0.4495 ns |
| GetUninitializedStruct1024 |  6.830 ns | 0.0914 ns | 0.0763 ns |

```cs
[Benchmark]
public Struct1024 GetUninitializedStruct1024()
    => BenchmarkDotNet.CompilerServices.Unsafe.GetUninitializedValue<Struct1024>();
```

```cil
.method public hidebysig static !!T 
        GetUninitializedValue<T>() cil managed aggressiveinlining 
{
    .maxstack 1
    .locals (
        [0] !!T 'value'
    )

    IL_0000: ldloc.0
    IL_0001: ret
} // end of method Unsafe::GetUninitializedValue
```

Thoughts?

timcassell:
Hmm... the result of that measurement seems to depend on the runtime. In .Net 7 and Framework, I observe that `default` is always fastest. But in Mono, reading from a field is fastest.

```
// Mono
|             Method |     Mean |   Error |  StdDev |
|------------------- |---------:|--------:|--------:|
|  DefaultStruct1024 | 261.5 ns | 1.06 ns | 0.88 ns |
| SkipInitStruct1024 | 294.7 ns | 1.61 ns | 1.50 ns |
|    FieldStruct1024 | 229.6 ns | 0.45 ns | 0.40 ns |
| UnsafeAsStruct1024 | 238.2 ns | 0.50 ns | 0.44 ns |

// .Net 7
|             Method |     Mean |    Error |   StdDev |
|------------------- |---------:|---------:|---------:|
|  DefaultStruct1024 | 20.42 ns | 0.097 ns | 0.081 ns |
| SkipInitStruct1024 | 45.71 ns | 0.143 ns | 0.119 ns |
|    FieldStruct1024 | 32.54 ns | 0.073 ns | 0.065 ns |
| UnsafeAsStruct1024 | 33.00 ns | 0.099 ns | 0.083 ns |

// .Net Framework 4.7.2
|             Method |     Mean |    Error |   StdDev |
|------------------- |---------:|---------:|---------:|
|  DefaultStruct1024 | 16.64 ns | 0.196 ns | 0.174 ns |
| SkipInitStruct1024 | 73.49 ns | 0.217 ns | 0.203 ns |
|    FieldStruct1024 | 30.84 ns | 0.168 ns | 0.157 ns |
| UnsafeAsStruct1024 | 42.91 ns | 0.835 ns | 0.893 ns |
```

I also experimented with another unsafe approach, which got close to the field read in Mono, but is still slower in other runtimes.

```cs
[Benchmark]
public Struct1024 UnsafeAsStruct1024()
{
    byte b = 0;
    return Unsafe.As<byte, Struct1024>(ref b);
}
```

Can we branch the code generated depending on the runtime? We could use `default` in Core, and field read in Mono.

timcassell:
@AndreyAkinshin Latest uses field read if the struct is > 64 bytes and the runtime is old Mono, default for all other cases. Is that ok? (Tbh I'm not sure how 32-bit works, if we can even mix 32-bit host/benchmark process. I only tested 64-bit.)

AndreyAkinshin:
> Latest uses field read if the struct is > 64 bytes and the runtime is old Mono, default for all other cases.

I don't like this approach. It feels too hacky and fragile. I suspect that it would be quite challenging to maintain it and keep it relevant to all the runtimes we support. Another severe issue is the fairness of comparison between different benchmarks. Could we compare absolute nanobenchmark measurements between Mono and non-Mono? Or a benchmark with struct size of 64 and 65 bytes?

The original idea behind the overhead evaluation was to evaluate the overhead of the general BenchmarkDotNet infrastructure, like method/delegates calls and the main loop that performs invocation. The delegate call evaluation includes the instruction pointer jumps. In order to make the overhead evaluation most "honest," we try to match the signature of the overhead method to the signature of the target workload method (where such an approach is applicable). We use this signature mimicry to ensure that the measured duration of an empty method (e.g., `int Foo() => 0`). Therefore, BenchmarkDotNet always tries to provide the actual duration of the method body, excluding calling overhead.

In the case of an empty `int` method (like `int Foo() => 0`), the "zero" result seems reasonable. However, in the case of a returned struct, we don't have such an obvious baseline. As we can see, both creating a new struct instance and reading a struct instance from memory takes some time. Non-zero results in this case may be considered as expected.

Thus, I don't think we should use the actual struct type of the target workload method for the overhead method signature. This gives us two options for the overhead return type:
1. Use a primitive type. In this case, the workload method calling infrastructure shouldn't pass `.Field` to the `Consumer` to avoid additional overhead.
2. Use a fake struct type with one or two fields and pass `.Field` to `Consumer` for both overhead and workload methods.

timcassell:
So, if we were to measure the cost of creating 2 different structs

```cs
[Benchmark] public SmallStruct Small() => default;
[Benchmark] public LargeStruct Large() => default;
```

In the current version, they would both report 0. But we may want them to report different time values (since it does actually take some time).

> The original idea behind the overhead evaluation was to evaluate the overhead of the general BenchmarkDotNet infrastructure, like method/delegates calls and the main loop that performs invocation. The delegate call evaluation includes the instruction pointer jumps.

Would it be reasonable for the overhead delegate to always return `void`, and we stub in the value for the `Consumer` call in the OverheadAction?

```cs
private void OverheadActionNoUnroll(System.Int64 invokeCount)
{
    LargeStruct consumeMe = default;
    {
        overheadDelegate();
        consumer.Consume(consumeMe);
    }
}
```

> In this case, the workload method calling infrastructure shouldn't pass .Field to the Consumer to avoid additional overhead.

Currently in this PR, `.field` is only used for ref structs (since they can't be used in generics). There is no extra overhead for non-ref structs since I optimized `Consumer.Consume<T>(in T)`.

AndreyAkinshin:
> In the current version, they would both report 0. But we may want them to report different time values (since it does actually take some time).

I just checked how it works with the latest version of BenchmarkDotNet from the `master` branch:

```cs
public struct SmallStruct
{
    public long Field00 { get; }
}

public struct LargeStruct
{
    public long Field00 { get; }
    public long Field01 { get; }
    public long Field02 { get; }
    public long Field03 { get; }
    public long Field04 { get; }
    public long Field05 { get; }
    public long Field06 { get; }
    public long Field07 { get; }
    public long Field08 { get; }
    public long Field09 { get; }
    public long Field10 { get; }
    public long Field11 { get; }
    public long Field12 { get; }
    public long Field13 { get; }
    public long Field14 { get; }
    public long Field15 { get; }
    public long Field16 { get; }
    public long Field17 { get; }
    public long Field18 { get; }
    public long Field19 { get; }
    public long Field20 { get; }
    public long Field21 { get; }
    public long Field22 { get; }
    public long Field23 { get; }
    public long Field24 { get; }
    public long Field25 { get; }
    public long Field26 { get; }
    public long Field27 { get; }
    public long Field28 { get; }
    public long Field29 { get; }
    public long Field30 { get; }
    public long Field31 { get; }
    public long Field32 { get; }
    public long Field33 { get; }
    public long Field34 { get; }
    public long Field35 { get; }
    public long Field36 { get; }
    public long Field37 { get; }
    public long Field38 { get; }
    public long Field39 { get; }
    public long Field40 { get; }
    public long Field41 { get; }
    public long Field42 { get; }
    public long Field43 { get; }
    public long Field44 { get; }
    public long Field45 { get; }
    public long Field46 { get; }
    public long Field47 { get; }
    public long Field48 { get; }
    public long Field49 { get; }
}

public class Benchmarks
{
    [Benchmark] public SmallStruct Small() => default;
    [Benchmark] public LargeStruct Large() => default;
}
```

Here are the results:

```md
BenchmarkDotNet=v0.13.5.20230616-develop, OS=ubuntu 22.04
AMD Ryzen 9 7950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK=7.0.304
  [Host]     : .NET 7.0.7 (7.0.723.27404), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.7 (7.0.723.27404), X64 RyuJIT AVX2


| Method |      Mean |     Error |    StdDev |
|------- |----------:|----------:|----------:|
|  Small | 0.0000 ns | 0.0000 ns | 0.0000 ns |
|  Large | 2.8153 ns | 0.0254 ns | 0.0237 ns |

// * Warnings *
ZeroMeasurement
  Benchmarks.Small: Default -> The method duration is indistinguishable from the empty method duration
```

This looks like a reasonable result. I would like to keep this behavior.

> Would it be reasonable for the overhead delegate to always return void, and we stub in the value for the Consumer call in the OverheadAction?

I have some doubts about this idea. Let's consider the following benchmarks:

```cs
[Benchmark] public object EmptyObject() => null;
[Benchmark] public int EmptyInt() => 0;
[Benchmark] public double EmptyDouble() => 0.0;
```

With the current version of BenchmarkDotNet, all of these benchmarks have "zero" duration. I would like to keep this behavior. If void Overhead allows us to achieve this on all the runtimes, we can consider such an approach (however, I would expect some non-zero results).

> in #2111 I refactored it so that the BenchmarkType starts the clock instead of the Engine, so it would be trivial to initialize that before starting the clock if that were to be merged.

Sorry for keeping #2111 without a review for such a long time. As you can see, it's quite challenging to verify engine/toolchain-related changes: any modifications in these subsystems may lead to surprisingly unpleased side effects. I will try to make another attempt to review it and share my thoughts.

timcassell:
> I just checked how it works with the latest version of BenchmarkDotNet from the master branch:

Yeah, I mean the current version of this PR, sorry.

> I have some doubts about this idea. Let's consider the following benchmarks:
> 
> ```cs
> [Benchmark] public object EmptyObject() => null;
> [Benchmark] public int EmptyInt() => 0;
> [Benchmark] public double EmptyDouble() => 0.0;
> ```
> 
> With the current version of BenchmarkDotNet, all of these benchmarks have "zero" duration. I would like to keep this behavior. If void Overhead allows us to achieve this on all the runtimes, we can consider such an approach (however, I would expect some non-zero results).

Would that be so bad, though? After all, I would expect this
```il
IL_0000: ldc.i4.0
IL_0001: ret
```

To take some non-zero time longer than this
```il
IL_0000: ret
```

I mean yes, people may see "regressions", but it would be more accurate, wouldn't it?

> Sorry for keeping https://github.com/dotnet/BenchmarkDotNet/pull/2111 without a review for such a long time. As you can see, it's quite challenging to verify engine/toolchain-related changes: any modifications in these subsystems may lead to surprisingly unpleased side effects. I will try to make another attempt to review it and share my thoughts.

Excellent! I'll go fix the merge issues, then!

timcassell:
I experimented with that idea, and I didn't like the results I was getting. The difference between loading a local to pass to `Consume` and just passing the value that's already on the stack from the workload call is too great.

So I studied the WorkloadAction implementation to try to find another method, and I thought, why is the `Consume` call even necessary at all? So I stripped it out and replaced it with `Thread.MemoryBarrier()`. With that and the overhead method returning `void`, these are the results I got:

|      Method |  Runtime |       Mean |     Error |    StdDev |
|------------ |--------- |-----------:|----------:|----------:|
|        Void | .NET 7.0 |  0.0038 ns | 0.0089 ns | 0.0075 ns |
|         Int | .NET 7.0 |  0.0000 ns | 0.0000 ns | 0.0000 ns |
|        Long | .NET 7.0 |  0.0000 ns | 0.0000 ns | 0.0000 ns |
|      Object | .NET 7.0 |  0.0000 ns | 0.0000 ns | 0.0000 ns |
|     Pointer | .NET 7.0 |  0.0023 ns | 0.0103 ns | 0.0080 ns |
|    Struct16 | .NET 7.0 |  0.1410 ns | 0.0431 ns | 0.0819 ns |
| EmptyStruct | .NET 7.0 |  0.0003 ns | 0.0009 ns | 0.0008 ns |
|        Void |     Mono |  0.0426 ns | 0.0025 ns | 0.0019 ns |
|         Int |     Mono |  0.0397 ns | 0.0037 ns | 0.0033 ns |
|        Long |     Mono |  0.0000 ns | 0.0000 ns | 0.0000 ns |
|      Object |     Mono |  0.0387 ns | 0.0039 ns | 0.0035 ns |
|     Pointer |     Mono |  0.0000 ns | 0.0000 ns | 0.0000 ns |
|    Struct16 |     Mono | 13.8994 ns | 0.3294 ns | 0.4829 ns |
| EmptyStruct |     Mono |  5.8519 ns | 0.0284 ns | 0.0222 ns |

The default primitives and `null` are at or close to 0 (with a bit of noise in the system), which matches the current results.

Can we refactor to remove `Consumer` from the BenchmarkType altogether? I think if we call the workload method with `NoInlining` (which to my understanding was the purpose of wrapping it in a delegate) along with `Thread.MemoryBarrier`, it should prevent any dead code elimination, or out-of-order execution.

```
{
    workloadDelegate($PassArguments$); // We don't do anything with the return value, it just gets popped off the stack.
    System.Threading.Thread.MemoryBarrier();
}
```

timcassell:
@AndreyAkinshin This PR turned out to be the wrong direction, so I'm closing it in favor of #2336. I also put up a separate PR #2337 for just the `Consumer` optimization.

Issue #2312, opened 2023-05-26
CodeAnalysisTreatWarningsAsErrors=true in Directory.Build.props breaks auto-generated benchmark assembly build


viktorz:
I am experiencing failures in the build of the auto-generated benchmark assembly with the following error due treating Code Analysis warnings as errors:
```
// Found 18 benchmarks:
// <list of benchmarks>

// Build Error: Standard output:

 Standard error:
 Time Elapsed 00:00:12.85
    9 Error(s)
    0 Warning(s)

\BIN\net6.0\c5b9ac70-f596-4da8-aefa-7235c015c0d4\c5b9ac70-f596-4da8-aefa-7235c015c0d4.notcs(3721,25): error CA5394: Random is an insecure random number generator. Use cryptographically secure random number generators when randomness is required for security. [\BIN\net6.0\c5b9ac70-f596-4da8-aefa-7235c015c0d4\BenchmarkDotNet.Autogenerated.csproj]
...

MSBuild version 17.6.1+8ffc3fe3d for .NET

// BenchmarkDotNet has failed to build the auto-generated boilerplate code.
```

My solution setup is the following:
- BenchmarkDotNet version 0.13.1
- `Directory.Build.props` in same folder as the .SLN file
- `TreatWarningsAsErrors = true`, global setting in _Directory.Build.props_
- `CodeAnalysisTreatWarningsAsErrors = true`, global setting in _Directory.Build.props_
- `OutputPath = ..\BIN\`, global setting in _Directory.Build.props_
- Multi-targeting `<TargetFrameworks>net472;net6.0</TargetFrameworks>` in all .CSPROJ files

I can work-around the problem by suppressing this specific rule using `<NoWarn>CA5394</NoWarn>` in _Directory.Build.props_, but as this is the solution-wide props file, this suppresses the rule everywhere, which is not optimal. I am unable (or don't know how) to use benchmark-specific _Directory.Build.props_ that suppresses the rule only for the benchmark as I am building all .CSPROJ files in the same _\BIN\_ folder and the auto-generated project is located there.

Please advice on better way to suppress the CA rule or configure the automatic build process to ignore it without affecting the rest of the solution projects.

adamsitnik:
Hi @viktorz 

Thank you for creating a detailed bug report. We should fix it on our side, the problem is that currently I don't have the time to do it myself. The best I can do right now is to provide you some hints on what would need to be done to solve the problem.

[...]

PR #2316, opened 2023-05-30
Updated CsProj.txt, set CodeAnalysisTreatWarningsAsErrors=false

viktorz:
Auto-generated .CSPROJ file now uses <CodeAnalysisTreatWarningsAsErrors>false</CodeAnalysisTreatWarningsAsErrors> to avoid failing toolchain compilation if consumer uses global Directory.Build.props has the same flag as true. Related issue #2312


viktorz:
@dotnet-policy-service agree

adamsitnik:
Once [this](https://ci.appveyor.com/project/dotnetfoundation/benchmarkdotnet/builds/47176023) build finishes (1h from now?), you should be able to download `0.13.5.2150` from our CI feed:

https://benchmarkdotnet.org/articles/guides/nuget.html#nightly

PR #2320, opened 2023-06-05
Add support for response files

kant2002:
I make workaround here for now, but essentially issues in CommadLine library. Also place workardoud for stupid bug with quoted argumetns with space in them.

I notice issue with parsing and request for RSP on Discord where @kg and @radical discuss how to feed proper parameters in `benchmarks_ci.py`


kg:
It looks like this addresses the existing problem of `--wasmArgs \"--expose-wasm --module\"` not parsing correctly while `--wasmArgs \" --expose-wasm --module \"` does? Neat.

kant2002:
Yes, I did create workaround by injecting whitespace in BDN. For sure this should be fixed in CommandLine library, but I do not sure reaction of time of their maintainers. also coincidentally `--wasmArgs` only parameter which seems to be need whitespace, so nobody notice issue. This issue still present in CommandLine 2.9

I thought that quality of life for BDN is more important then proper fix. Will try to submit real fix a bit later.

AndreyAkinshin:
@kant2002 thanks for the PR. As I can see, this PR suggests reserving the `@` argument prefix for the response files. Do we really need a new type of notation for such a feature? Could we use the existing notation (e.g., introduce the `--reponse-file` argument)?

kant2002:
`@` is well known convention from compiler world. And also supported by System.CommandLine (not that this is matter much to me) which means eventually it would propagate across ecosystem. Also if use `--response-file` that would make it “magic” argument which somehow should be documented in Options but applied in different place (same as in this PR).

kant2002:
@adamsitnik I would like to hear your opinion on the matter. Are you have time for take a look at PR ?

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -86,6 +88,100 @@ public static (bool isSuccess, IConfig config, CommandLineOptions options) Parse
...
         }
 
+        private static IEnumerable<string> ExpandResponseFile(string[] args)
+        {
+            foreach (var arg in args)
+            {
+                if (arg.StartsWith("@"))
+                {
+                    var fileName = arg.Substring(1);
+                    if (File.Exists(fileName))
```
AndreyAkinshin, line 98: What if the given file doesn't exist? I believe we should warn the user about that rather than just silently ignore such an argument.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -86,6 +88,100 @@ public static (bool isSuccess, IConfig config, CommandLineOptions options) Parse
...
+                    {
+                        var lines = File.ReadAllLines(fileName);
```
AndreyAkinshin, line 100: Could you please add try..catch with proper exception handling? If there are some issues with the given file (e.g., we don't have permission to read the file content), it would be nice to print an error message rather than just throwing an exception.

AndreyAkinshin (changes requested):

AndreyAkinshin:
@kant2002 I think I'm OK with introducing such a `@`-syntax. You are right, such an approach is a common convention for many tools. Could you please address the above comments from the review and add some docs covering this feature?

AndreyAkinshin (approved):

AndreyAkinshin:
@kant2002 thanks!

PR #2325, opened 2023-06-11
Fix building doc specifying .net 4.6.2 as requirement

JobaDiniz:
As I'm following the [building](https://benchmarkdotnet.org/articles/contributing/building.html) documentation to locally build the project, I find out that .net `4.6.2` is required, not `4.6.1`.

This aims to fix that documentation.

### DIFF
--- a/docs/articles/contributing/building.md
+++ b/docs/articles/contributing/building.md
@@ -4,7 +4,7 @@ There are two recommended options to build BenchmarkDotNet from source:
 
 ## Visual Studio
 
-- [Visual Studio](https://www.visualstudio.com/downloads/) (Community, Professional, Enterprise) with .NET 4.6.1 SDK and F# support.
+- [Visual Studio](https://www.visualstudio.com/downloads/) (Community, Professional, Enterprise) with .NET 4.6.2 SDK and F# support.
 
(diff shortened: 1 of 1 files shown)


JobaDiniz:
@dotnet-policy-service agree

AndreyAkinshin:
@JobaDiniz thanks for the PR!
While the 4.6.2 requirement looks reasonable, I'm curious about what kind of problem you experience with 4.6.1?
Also, since we are updating this page, it makes sense to update other versions as well to make the list of requirements consistent and up-to-date (it's been a while since the last page update). Would you mind doing this?

JobaDiniz:
The solution was not compiling if I only had 4.6.1 installed. I had to install 4.6.2.
>Also, since we are updating this page, it makes sense to update other versions as well to make the list of requirements consistent and up-to-date.

I don't know what other requirements have changed. Any insights to find that out?

AndreyAkinshin:
@JobaDiniz I'm going to merge your PR and reevaluate the rest of the guide on my own. It's high time to check all the requirements and make them up-to-date. Thanks for the contribution!

timcassell:
I'm guessing this was broken due to #2333, which should be fixed in #2359.

PR #2330, opened 2023-06-15
Use Verify library for approval tests

alinasmirnova:
[ApprovalTests.Net](https://github.com/approvals/ApprovalTests.Net) is now obsolete and is no loger maintened. It is recommended to use [Verify](https://github.com/VerifyTests/Verify).


tests/BenchmarkDotNet.Tests/Portability/Cpu/CpuInfoFormatterTests.cs
```diff
@@ -28,7 +24,9 @@ public void FormatTest()
                 captions.AppendLine(CpuInfoFormatter.Format(mockCpuInfo));
             }
 
-            Approvals.Verify(captions.ToString());
+            var settings = new VerifySettings();
+            settings.UseDirectory("ApprovedFiles");
```
AndreyAkinshin, line 28: We use the same name for the directory with verified files across all the tests. It would be nice to avoid duplicating of the "ApprovedFiles" string in multiple places.

tests/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj
```diff
@@ -42,4 +42,10 @@
   <ItemGroup>
     <EmbeddedResource Include="Portability\Cpu\TestFiles\*.txt" />
   </ItemGroup>
+  <ItemGroup>
```
AndreyAkinshin, line 45: It seems that these lines were committed by mistake: I don't see any reasons to exclude the declared files from the project structure.

AndreyAkinshin (changes requested):
Thanks for the PR! It's great to have an up-to-date library for test output verification! However, I have a few minor concerns.
1. The first concern is about the command-line experience. Currently, if I introduce a change in the Verify inputs of several tests, and run the tests from the command line (e.g., using `build.ps1 --target FastTests` on Windows), Verify detects that I have locally installed Rider, next it launches Rider and spawn multiple diff windows for all the failures. This is a confusing experience for a command line tool. I would prefer disabling external diff tools outside IDEs.
2. The second concern is about naming. In this PR, the dependency on ApprovalTests was completely removed. However, the names of the relevant classes and the folders with the verified files still contain "Approved"/"Approval". I would suggest renaming them since such naming feels confusing without the actual dependency on the approval tests.

alinasmirnova:
Thanks for you feedback, I'll push some fixes today.

alinasmirnova:
@AndreyAkinshin, I've pushed all the related fixes. Could you please check the console experience again? Should be better.

AndreyAkinshin (approved):

AndreyAkinshin:
@alinasmirnova thanks!

PR #2335, opened 2023-06-21
`AggressiveOptimization` in `InProcess` toolchains

### LINKED ISSUE #1934: Ensure WorkloadActionUnroll and similar are optimized if possible

timcassell:
Follow-up to #1935, fixes #1934 for in-process toolchains.

### DIFF
--- a/src/BenchmarkDotNet/Helpers/Reflection.Emit/MethodBuilderExtensions.cs
+++ b/src/BenchmarkDotNet/Helpers/Reflection.Emit/MethodBuilderExtensions.cs
@@ -37,5 +37,14 @@ namespace BenchmarkDotNet.Helpers.Reflection.Emit
 
             return methodBuilder;
         }
+
+        public static MethodBuilder SetAggressiveOptimizationImplementationFlag(this MethodBuilder methodBuilder)
+        {
+            // AggressiveOptimization is not available in netstandard2.0, so just use the value casted to enum.
+            methodBuilder.SetImplementationFlags(
+                methodBuilder.GetMethodImplementationFlags() | (MethodImplAttributes) 512);
+
+            return methodBuilder;
+        }
     }
 }
--- a/src/BenchmarkDotNet/Helpers/Reflection.Emit/MethodBuilderExtensions.cs
+++ b/src/BenchmarkDotNet/Helpers/Reflection.Emit/MethodBuilderExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using BenchmarkDotNet.Portability;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -40,9 +41,8 @@ namespace BenchmarkDotNet.Helpers.Reflection.Emit
 
         public static MethodBuilder SetAggressiveOptimizationImplementationFlag(this MethodBuilder methodBuilder)
         {
-            // AggressiveOptimization is not available in netstandard2.0, so just use the value casted to enum.
             methodBuilder.SetImplementationFlags(
-                methodBuilder.GetMethodImplementationFlags() | (MethodImplAttributes) 512);
+                methodBuilder.GetMethodImplementationFlags() | CodeGenHelper.AggressiveOptimizationOptionForEmit);
 
             return methodBuilder;
         }
--- a/src/BenchmarkDotNet/Portability/CodeGen.cs
+++ b/src/BenchmarkDotNet/Portability/CodeGen.cs
@@ -1,0 +1,12 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace BenchmarkDotNet.Portability
+{
+    public static class CodeGenHelper
+    {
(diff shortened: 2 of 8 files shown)


src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/BenchmarkActionFactory_Implementations.cs
```diff
@@ -14,6 +15,9 @@ namespace BenchmarkDotNet.Toolchains.InProcess.NoEmit
     /// <summary>Helper class that creates <see cref="BenchmarkAction"/> instances. </summary>
     public static partial class BenchmarkActionFactory
     {
+        // Not available in netstandard2.0, so just use the value casted to enum.
+        public const MethodImplOptions AggressiveOptimization = (MethodImplOptions) 512;
```
AndreyAkinshin, line 19: Could we move this constant to a better namespace? `BenchmarkDotNet.Toolchains.InProcess.NoEmitBenchmarkActionFactory` looks like a strange place for such an important constant.

timcassell: Do you have a suggestion? I wasn't sure of a best place to put it.

src/BenchmarkDotNet/Helpers/Reflection.Emit/MethodBuilderExtensions.cs
```diff
@@ -37,5 +37,14 @@ public static MethodBuilder SetNoOptimizationImplementationFlag(this MethodBuild
 
             return methodBuilder;
         }
+
+        public static MethodBuilder SetAggressiveOptimizationImplementationFlag(this MethodBuilder methodBuilder)
+        {
+            // AggressiveOptimization is not available in netstandard2.0, so just use the value casted to enum.
+            methodBuilder.SetImplementationFlags(
+                methodBuilder.GetMethodImplementationFlags() | (MethodImplAttributes) 512);
```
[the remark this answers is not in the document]
timcassell: That's a different enum (`MethodImplOptions` vs `MethodImplAttributes`). I suppose I could make that a constant for better readability anyway.

src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/BenchmarkActionFactory_Implementations.cs
AndreyAkinshin: What about `BenchmarkDotNet.Portability`?

src/BenchmarkDotNet/Helpers/Reflection.Emit/MethodBuilderExtensions.cs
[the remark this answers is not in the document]
AndreyAkinshin: Yes, please introduce a single constant and put both instances of `MethodImplOptions.AggressiveOptimization` and `MethodImplAttributes.AggressiveOptimization` at the same place.

AndreyAkinshin (approved):

src/BenchmarkDotNet/Portability/CodeGen.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace BenchmarkDotNet.Portability
+{
+    public static class CodeGenHelper
+    {
+        // AggressiveOptimization is not available in netstandard2.0, so just use the value casted to enum.
+        public const MethodImplOptions AggressiveOptimizationOption = (MethodImplOptions) 512;
```
adamsitnik, line 9: is this value ignored by the older runtimes?

cc @AndyAyersMS who added the annotations with ifdefs

AndyAyersMS: Yes, I believe it will be ignored by older .NET Core and all .NET Framework versions.

PR #2334, opened 2023-06-20
Call benchmark method directly

timcassell:
Fixes #1133

This wraps the workload call with a `NoInlining | NoOptimization` method instead of a delegate.

<details><summary>Mac Intel x64 results</summary>
<p>

```
BenchmarkDotNet=v0.13.5.20230619-develop, OS=macOS Monterey 12.3 (21E230) [Darwin 21.4.0]
Intel Core i9-9880H CPU 2.30GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.100-preview.5.23303.2
  [Host]     : .NET 8.0.0 (8.0.23.28008), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.0 (8.0.23.28008), X64 RyuJIT AVX2
```

Master
|         Method |      Mean |     Error |    StdDev |
|--------------- |----------:|----------:|----------:|
|   OneIncrement | 0.0013 ns | 0.0037 ns | 0.0048 ns |
|   TwoIncrement | 0.0034 ns | 0.0043 ns | 0.0064 ns |
| ThreeIncrement | 0.0000 ns | 0.0000 ns | 0.0000 ns |
|  FourIncrement | 0.1652 ns | 0.0053 ns | 0.0047 ns |
|  FiveIncrement | 0.3950 ns | 0.0047 ns | 0.0039 ns |
|   SixIncrement | 0.6129 ns | 0.0038 ns | 0.0035 ns |

This PR
|         Method |      Mean |     Error |    StdDev |
|--------------- |----------:|----------:|----------:|
|   OneIncrement | 0.2852 ns | 0.0063 ns | 0.0056 ns |
|   TwoIncrement | 0.3822 ns | 0.0477 ns | 0.0742 ns |
| ThreeIncrement | 0.4122 ns | 0.0071 ns | 0.0063 ns |
|  FourIncrement | 0.4845 ns | 0.0080 ns | 0.0071 ns |
|  FiveIncrement | 0.5980 ns | 0.0082 ns | 0.0068 ns |
|   SixIncrement | 1.1828 ns | 0.0078 ns | 0.0069 ns |

</p>
</details>

<details><summary>Windows AMD x64 results</summary>
<p>

```
BenchmarkDotNet=v0.13.5.20230619-develop, OS=Windows 10 (10.0.19045.3086/22H2/2022Update)
AMD Phenom(tm) II X6 1055T Processor, 1 CPU, 6 logical and 6 physical cores
[...]

Issue #1133, opened 2019-04-19
BenchmarkDotNet (arguably) slightly overcorrects for overhead


Zhentar:
The test overhead deduction causes BDN to underreport benchmark execution times (as likely interpreted by users). The magnitude will vary depending upon hardware & the nature of test code, but should generally be on the order of 0.5ns-1.0ns.

I've noticed hints of this for a while, but only recently came to recognize what was occurring well enough to design a test that could clearly and consistently reproduce it. 

[Test Code](https://gist.github.com/Zhentar/deb1afb9bc30ed2a82583a348729dc9f)

```
BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17763.437 (1809/October2018Update/Redstone5)
Intel Core i5-6600K CPU 3.50GHz (Skylake), 1 CPU, 3 logical and 3 physical cores
.NET Core SDK=3.0.100-preview5-011351
  [Host]     : .NET Core 3.0.0-preview5-27617-04 (CoreCLR 4.6.27617.71, CoreFX 4.700.19.21614), 64bit RyuJIT
  DefaultJob : .NET Core 3.0.0-preview5-27617-04 (CoreCLR 4.6.27617.71, CoreFX 4.700.19.21614), 64bit RyuJIT
```

|         Method |      Mean |     Error |    StdDev |
|--------------- |----------:|----------:|----------:|
|   OneIncrement | 0.0000 ns | 0.0000 ns | 0.0000 ns |
|   TwoIncrement | 0.1049 ns | 0.0203 ns | 0.0180 ns |
| ThreeIncrement | 0.0771 ns | 0.0222 ns | 0.0208 ns |
|  FourIncrement | 0.3379 ns | 0.0161 ns | 0.0143 ns |
|  FiveIncrement | 0.6445 ns | 0.0265 ns | 0.0248 ns |
|   SixIncrement | 0.9659 ns | 0.0133 ns | 0.0111 ns |

One, two, and three increments are all basically the same. But past that, there's a linear 0.32ns (or roughly 1 CPU cycle) increase in execution time for each additional increment.  The first three are "free" - because they are able to run alongside the benchmark overhead instructions in the CPU pipeline, adding no effective latency to the the test harness. The execution time doesn't increase until all of that capacity has been filled, and the test flips from test harness bound to test code bound. 

[...]

PR #2336, opened 2023-06-21
Fair Return Types

### LINKED ISSUE #2305: overhead and workload invocation sequences diverge

timcassell:
Follow-up to #2334. Fixes #2305.

All return types are treated equally. No more special consume logic for `ByRef`, or `struct` return types. Return values are popped off the stack instead of passed to `Consumer` or a `NoInlining` method. This is able to be done while avoiding dead code elimination by taking advantage of the changes made in #2334.

Also fixed pointer returns in `InProcessEmitToolchain`.


timcassell:
@AndreyAkinshin I moved this to v0.14.0 milestone because I think both #2334 and this should go out in the same version so that we are only affecting measurement accuracy in a single version, rather than multiple versions. If we can't get them both in v0.14.0, let's push them both out.

tests/BenchmarkDotNet.IntegrationTests.ManualRunning/ExpectedBenchmarkResultsTests.cs
```diff
@@ -121,64 +123,102 @@ private void AssertZeroResults(Type benchmarkType, IConfig config)
...
+                // We use the default threshold here rather than using cpu resolution,
+                // because modern cpus can execute multiple instructions per clock cycle,
+                // resulting in measurements greater than 0 but less than 1 clock cycle.
+                // (example: Intel Core i9-9880H CPU 2.30GHz reports 0.2852 ns for `_field++;`)
+                bool isZero = ZeroMeasurementHelper.CheckZeroMeasurementTwoSamples(workloadMeasurements, overheadMeasurements);
                 Assert.False(isZero, $"Actual time was 0.");
 
-                isZero = ZeroMeasurementHelper.CheckZeroMeasurementTwoSamples(overheadMeasurements, workloadMeasurements, threshold);
-                Assert.True(isZero, "Overhead took more time than workload.");
+                // We don't need to check if overhead was greater. It's guaranteed less if the actual time was not zero.
```
AndreyAkinshin, line 221: 1. The current implementation of `CheckZeroMeasurementTwoSamples` is a bad one. I wrote it in 2018 when I had poor knowledge of statistics. I have done [my homework](https://aakinshin.net/tags/statistics/) since then and now I know a much better way to implement it. The problem with the current implementation is that it is not reliable and the used statistical test is underpowered: it can produce too many false-negative results (there is an actual difference between workload and overhead, but the test is always green). However, a better implementation requires a new version of Perfolizer, which means a huge upgrade in terms of used APIs. While this is a huge change, it makes total sense to include it in v0.14.0.
2. Speaking of the default threshold, I do not like the concept of "default" value in general. If we have some knowledge of the domain area, we should use it to increase accuracy. However, it is not clear how to properly use it and what we should do in the case of unknown CPU frequency. Therefore, switching to a reasonable default that is adaptive to the samples is not a bad idea. However, I do not like the current default of 1%. It is just a random value without justification. We should reconsider the default.
3. Finally, the test procedure should be adjusted. Outliers should not be removed (a proper test will handle any deviations), and the number of iterations should be increased (the default logic does not produce enough iterations for reasonable comparison). All the code I wrote here is methodically wrong and should be fixed.

Next action: I upgrade Perfolizer to the latest version so that we get a more reliable statistical engine. As a pleasant bonus, I will finally be sure of my checks of other PRs with low-level changes (now I cannot properly validate them, so I procrastinate about the merge).

timcassell:
Unfortunately, I no longer have an Intel CPU to test this, but I think the new results from this PR should address the issues you had in #2334 @AndreyAkinshin (results look good on my AMD 7800X3D CPU in .Net 8).

timcassell:
@AndyAyersMS Adam mentioned you are the expert I should tag to get your opinion on things like this. Do you have any thoughts or concerns about this approach?

timcassell:
> I am not that familiar with how robust the weaving process is... maybe it would be good (if possible) to validate it on say the dotnet/performance suite of 5000ish benchmarks.

I can give it a try.

> I guess it also means there is no simple faithful source representation for the benchmark assembly, which might trip up some users that keep the sources around and use those for investigations?

Could you elaborate? The weaver just adds `NoInlining` to `[Benchmark]` annotated methods. Yeah it's not an exact 1:1 match of source code, but C# doesn't provide a way for us to do it ourselves, even with source generators. It's a small change that I'm not sure how it would impact users doing those types of deep investigations. I would expect anyone doing that to be a power user anyway.

> What happens if you just compile and run those sources?

What do you mean? If the source project has a reference to `BenchmarkDotNet.Annotations` and it's built with dotnet, the assembly weaver will kick in whether they actually run the benchmarks or not. If they manually instantiate and use a benchmark class, it will just not inline the methods. All other behavior will be unaffected.

> Do you have examples showing the impact on reported results?

With the benchmarks from #1133, virtually no impact on my Ryzen 9800X3D, slightly more accurate on Apple M3 (a CPU architecture which BDN seems to have overhead measurement issues for nano benchmarks). It seems to mostly impact Intel CPUs and older (pre-Ryzen) AMD CPUs (see results I obtained previously in #2334. Unfortunately I no longer have those old CPUs to test again).

As for returning a value (like the issue #2305), I'm not sure if my 9800X3D is the best representative CPU compared to what dotnet/performance runs on, but I got these results:

```cs
[Benchmark] public Vector3 Vector3() => default;
```

master:

| Method  | Mean      | Error     | StdDev    |
|-------- |----------:|----------:|----------:|
| Vector3 | 0.0146 ns | 0.0026 ns | 0.0022 ns |

PR:

| Method  | Mean      | Error     | StdDev    | Median    |
|-------- |----------:|----------:|----------:|----------:|
| Vector3 | 0.0000 ns | 0.0001 ns | 0.0001 ns | 0.0000 ns |

> 
> > * Overhead return type is always `void` for a fair baseline for all benchmark return types.
> 
> Can you explain this a bit more? The overhead is now some "fixed" workload measurement regardless of benchmark method signature?

It's a fix for #2305. Not quite, it still passes all the same arguments, only the return type is fixed to `void`. My previous attempt in #2309 was a failure. Essentially, the overhead disregards the cost of the return type, so the actual result will include the cost of the return type. Practically, you could measure the cost of returning a large struct, which is currently impossible.

timcassell:
@AndyAyersMS Ignoring net462 (https://github.com/dotnet/performance/issues/4869), I tried build with these changes, and I got this error.

```
Mono.Cecil.ResolutionException: Failed to resolve System.IO.FileOptions
```

I can't find anything about `System.IO.FileOptions`, except an [enum](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileoptions), which I think is weird because it's not an assembly. Any ideas?

AndyAyersMS:
> @AndyAyersMS Ignoring net462 ([dotnet/performance#4869](https://github.com/dotnet/performance/issues/4869)), I tried build with these changes, and I got this error.
> 
> ```
> Mono.Cecil.ResolutionException: Failed to resolve System.IO.FileOptions
> ```
> 
> I can't find anything about `System.IO.FileOptions`, except an [enum](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileoptions), which I think is weird because it's not an assembly. Any ideas?

@LoopedBard3 any ideas? Android support?

LoopedBard3:
I don't have any ideas. This may be a useful reference https://github.com/mono/mono/issues/10653, but I don't have time to dig into it at the moment. Regardless, the performance repo should not block this merging as we manually update the BDN version we use: https://github.com/dotnet/performance/blob/main/eng/Versions.props#L14.

timcassell:
> the performance repo should not block this merging as we manually update the BDN version we use

Well, it's a good litmus test as to whether this approach is reliable or not, and it seems it's not. Also, you may not need to update immediately, but you will when .Net 11 time comes, and I wouldn't want to block that.

src/BenchmarkDotNet.Weaver/src/WeaveAssemblyTask.cs
```diff
@@ -0,0 +1,122 @@
...
+namespace BenchmarkDotNet.Weaver;
+
+internal class CustomAssemblyResolver : DefaultAssemblyResolver
+{
+    public override AssemblyDefinition Resolve(AssemblyNameReference name, ReaderParameters parameters)
+        // NetStandard causes StackOverflow. https://github.com/jbevain/cecil/issues/573
+        // Mscorlib fails to resolve in Visual Studio. https://github.com/jbevain/cecil/issues/966
+        // We don't care about any types from runtime assemblies anyway, so just skip resolving them.
+        => name.Name is "netstandard" or "mscorlib" or "System.Runtime" or "System.Private.CoreLib"
+            ? null
```
timcassell: My suspicion is that skipping resolving runtime assemblies is the cause of the error. I did this because of the [issue in Visual Studio](https://github.com/jbevain/cecil/issues/966), I wonder if there is a more robust solution. 🤔

timcassell: Confirmed. Resolving runtime assemblies fixes the error on command line, but also errors in VS. 😭

timcassell:
Well, I couldn't find any solution for `Mono.Cecil`. I also tried `dnlib`, but it also had issues with VS. Then I found `AsmResolver`, and it works!

[Edit] Well, almost... looks like it's not quite working on MacOS.

AndyAyersMS:
Is the weaving process compatible with benchmarking NAOT?

timcassell:
> Is the weaving process compatible with benchmarking NAOT?

It should be, yes. The weaver runs after CoreCompile target, so it should be before the native compiler starts. We have an integration test for it, and it's passing on my local machine.

timcassell:
Okay, I have gotten it to pass all tests, and tested it locally against dotnet/performance. The only caveat that should be mentioned is that AsmResolver doesn't support Arm64 Platform (for now). https://github.com/Washi1337/AsmResolver/issues/640#issuecomment-2948405936
Arm64 Platform can still be benchmarked, as long as the user's project is compiled for AnyCpu (we don't make any assembly modifications to the benchmark wrapper project).

AndyAyersMS:
@dotnet/jit-contrib any thoughts on this?

timcassell:
Interestingly, on Apple M3, an empty method is measured at 0.9ns while a method that increments a field is measured at 0.66ns, which results in the overhead measurement looking "wrong". I don't know if this is specific to M chips or Arm in general, but it doesn't happen on xArch. There does not seem to be a good way for BDN to account for this μarch quirk.

PR #1879, opened 2021-12-30
adjust defaultBenchmarkDotNetVersion on build

askazakov:
Some automatization to reconcile BenchmarkDotNet version with default version in template

If it's good enough I will fix templates for VB and F# also. If not just close PR

PR #2338, opened 2023-06-21
Improve templates

### LINKED ISSUE #1658: Pass arguments to runner in dotnet new templates
Templates for `dotnet new` were added for #1028 but don't use any passed arguments for the runner. With the global tool deprecated, this world provide a convenient way of customizing frameworks, reports, etc. from a console.

So instead of:

```csharp
{
    var summary = BenchmarkRunner.Run<$(BenchmarkName)>();
}
```

The templates could use:

```csharp
{
}
```

### LINKED ISSUE #1881: broken template for Rider
after
```
dotnet new -i BenchmarkDotNet.Templates::0.13.1
```
JetBrains Rider create broken project because `TargetFramework` contains `v0.0`

I've created [issue](https://youtrack.jetbrains.com/issue/RIDER-72787) for Rider.

But also we can fix template – for example rename `framework` parameter to `targetFramework`.
Or set defaultValue for `framework` to non-empty string

### LINKED ISSUE #2149: BenchmarkDotNet.Templates not showing in VS2022(17.3.6) new project dialog
Similar question (https://developercommunity.visualstudio.com/t/net-cli-templates-in-visual-studio-missing/1508502), might be helpful to solve this problem, but I can't do anything about it.

YegorStepanov:
The `framework` parameter must be of type `choice` (see the discussion with the VS developers below).

Currently:
VS doesn't display the templates at all.
Rider displays it with disabled dropdown but crashes when creating project.
image

The `choice` type means that we have to write all the supported frameworks by hand.
I've added all version frameworks, that have not expired end-of-life:

Also we can no longer use platform-specific TFM:
```
dotnet new benchmark -f net7.0-windows
```

But it's ok because `console` use a choice too, so we cannot create platform-specific TFM:
```
dotnet new console -f net7.0-windows
// displays an error
```

More details with screenshots: https://github.com/sayedihashimi/template-sample/issues/73
<details>
<summary>IDE screenshots</summary>

VS:
image
image

Rider:
image
</details>

fix #1881, fix #2149, fix #1658

the PR #1817 can be closed too?

### DIFF
--- a/docs/articles/guides/dotnet-new-templates.md
+++ b/docs/articles/guides/dotnet-new-templates.md
@@ -44,10 +44,10 @@ dotnet new benchmark -lang VB
 The template projects has five additional options - all of them are optional.
 
 By default a class library project targeting `netstandard2.0` is created.
-You can specify `-f` or `--frameworks` to change targeting to one or more frameworks:
+You can specify `-f` or `--framework` to change the target framework:
 
 ```log
-dotnet new benchmark -f netstandard2.0;net472
+dotnet new benchmark -f net472
 ```
 
 The option `--console-app` creates a console app project targeting `netcoreapp3.0` with an entry point:
@@ -57,7 +57,6 @@ dotnet new benchmark --console-app
 ```
 
 This lets you run the benchmarks from console (`dotnet run`) or from your favorite IDE.
-**Note:** option `-f` or `--frameworks` will be ignored when `--console-app` is set.
 
 The option `-b` or `--benchmarkName` sets the name of the benchmark class:
 
--- a/docs/articles/guides/dotnet-new-templates.md
+++ b/docs/articles/guides/dotnet-new-templates.md
@@ -43,20 +43,20 @@ dotnet new benchmark -lang VB
 
 The template projects has five additional options - all of them are optional.
 
-By default a class library project targeting `netstandard2.0` is created.
-You can specify `-f` or `--framework` to change the target framework:
+By default a console app project targeting `net6.0` is created.
+This lets you run the benchmarks from console (`dotnet run`) or from your favorite IDE.  
+
+The option `-f` or `--framework` changes the target framework:
 
 ```log
 dotnet new benchmark -f net472
 ```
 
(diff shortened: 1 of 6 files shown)


YegorStepanov:
What is the reason it creates the library and not the console application by default? It is much more common to create the console application, and it is not very clear for beginners how to run tests (they need to re-read the documentation).

templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/.template.config/template.json
```diff
@@ -98,7 +144,7 @@
       "type": "parameter",
       "datatype": "string",
       "description": "Version of BenchmarkDotNet that will be referenced.",
-      "defaultValue": "0.12.1",
```
YegorStepanov, line 101: With each new release, you need to change this manually.

Currently, if you add BDN.Templates, a project will be created with BDN v0.12.1.

docs/articles/guides/dotnet-new-templates.md
```diff
@@ -5,7 +5,7 @@ name: BenchmarkDotNet templates
 
 # BenchmarkDotNet templates
 
-BenchmarkDotNet provides project templates to setup your benchmarks easily 
+BenchmarkDotNet provides project templates to setup your benchmarks easily.
```
YegorStepanov, line 8: There is only one space, I think there should be a new paragraph here (a second space is needed)

YegorStepanov:
image

AndreyAkinshin (approved):

AndreyAkinshin:
@YegorStepanov works like a charm! Thanks for updating the templates!

templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/.template.config/template.json
askazakov: I try to handle this issue – https://github.com/dotnet/BenchmarkDotNet/pull/1879
 @AndreyAkinshin could you take a look please ?

PR #2339, opened 2023-06-21
Add the args parameter to RunAll and RunAllJoined

YegorStepanov:
Only these methods do not contain the `args` parameter.

What order of parameters should be used? (I did not break backward compatibility)
```
BenchmarkSwitcher
    RunAll(IConfig config, string[] args)  NEW
    RunAllJoined(IConfig config, string[] args) NEW
    Run(string[] args, IConfig config)
BenchmarkRunner
    Run<T>(IConfig config, string[] args)
    Run(Type, IConfig config, string[] args)
    Run(Type[], IConfig config, string[] args)
    Run(Assembly, IConfig config, string[] args)
```

I upgrade the library version because we need `settings.SkipDefault = true` (since version 2.7).

### DIFF
--- a/src/BenchmarkDotNet/BenchmarkDotNet.csproj
+++ b/src/BenchmarkDotNet/BenchmarkDotNet.csproj
@@ -17,7 +17,7 @@
     <EmbeddedResource Include="Environments\microarchitectures.txt" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


AndreyAkinshin (approved):

AndreyAkinshin:
@YegorStepanov thanks!

PR #2345, opened 2023-06-27
[wasm] Use `test-main.mjs` for net8 runtimes .

radical:
TODO: the older `--wasmMainJS` parameter should probably be re-introduced so we don't have to do this for every change in filename.


radical:
Can't merge this till https://github.com/dotnet/performance/issues/3103 is fixed.

PR #2352, opened 2023-07-04
Fix netcoreapp3.0 and older builds

timcassell:
Fixes #2333

PR #2288, opened 2023-03-13
update Microsoft.CodeAnalysis.CSharp, fixes #2266

dennishermann:

Issue #2266, opened 2023-02-10
Incompatible `Microsoft.CodeAnalysis.CSharp` versioning with Razor


matthid:
Because of https://github.com/dotnet/BenchmarkDotNet/issues/2261 we upgraded to latest but our benchmarks broke with:

```
// Build Error: Exception! 
Message: Method not found: 'Void Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions..ctor(Microsoft.CodeAnalysis.OutputKind, Boolean, System.String, System.String, System.String, System.Collections.Generic.IEnumerable`1<System.String>, Micro
soft.CodeAnalysis.OptimizationLevel, Boolean, Boolean, System.String, System.String, System.Collections.Immutable.ImmutableArray`1<Byte>, System.Nullable`1<Boolean>, Microsoft.CodeAnalysis.Platform, Microsoft.CodeAnalysis.ReportDiagnostic, Int32
, System.Collections.Generic.IEnumerable`1<System.Collections.Generic.KeyValuePair`2<System.String,Microsoft.CodeAnalysis.ReportDiagnostic>>, Boolean, Boolean, Microsoft.CodeAnalysis.XmlReferenceResolver, Microsoft.CodeAnalysis.SourceReferenceRe
solver, Microsoft.CodeAnalysis.MetadataReferenceResolver, Microsoft.CodeAnalysis.AssemblyIdentityComparer, Microsoft.CodeAnalysis.StrongNameProvider, Boolean, Microsoft.CodeAnalysis.MetadataImportOptions, Microsoft.CodeAnalysis.CSharp.NullableCo
ntextOptions)'.,
Stack trace:
   at BenchmarkDotNet.Toolchains.Roslyn.Builder.Build(GenerateResult generateResult, BuildPartition buildPartition, ILogger logger, CancellationToken cancellationToken)
   at BenchmarkDotNet.Toolchains.Roslyn.Builder.Build(GenerateResult generateResult, BuildPartition buildPartition, ILogger logger)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Build(BuildPartition buildPartition, String rootArtifactsFolderPath, ILogger buildLogger)

// BenchmarkDotNet has failed to build the auto-generated boilerplate code.
// It can be found in C:\Dev\Q-Bisect\Src\Qplix.Core.CoreBasicQueryPerformance\bin\Release\net48
[...]

PR #2359, opened 2023-07-09
Fix netcoreapp3.0 builds

timcassell:
I got the wrong version of a nuget package in #2352.

### DIFF
--- a/src/BenchmarkDotNet/BenchmarkDotNet.csproj
+++ b/src/BenchmarkDotNet/BenchmarkDotNet.csproj
@@ -24,6 +24,8 @@
     <PackageReference Include="Perfolizer" Version="0.2.1" />
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.0.2" PrivateAssets="contentfiles;analyzers" />
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="3.1.6" />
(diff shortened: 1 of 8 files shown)


AndreyAkinshin:
Why do we use different versions of `Microsoft.CodeAnalysis.CSharp` with different target frameworks? (`4.2.0`->`4.1.0` with `netstandard2.0` and `4.5.0` with `net6.0`)

timcassell:
> Why do we use different versions of `Microsoft.CodeAnalysis.CSharp` with different target frameworks? (`4.2.0`->`4.1.0` with `netstandard2.0` and `4.5.0` with `net6.0`)

Are you suggesting we just use the older version for all targets?

AndreyAkinshin:
Previously, we used `3.0.0` but it caused [compatibility issues](https://github.com/dotnet/BenchmarkDotNet/issues/2266) in some cases. Therefore, it was updated to [4.5.0](https://github.com/dotnet/BenchmarkDotNet/pull/2288/files) (the latest version at the moment of update).
I believe that there is no perfect version of `Microsoft.CodeAnalysis.CSharp` since it would cause problems for those who use other versions of this package that are not compatible with our target version. However, I would prefer that we use the same version across all the target frameworks.

AndreyAkinshin (approved):

AndreyAkinshin:
@timcassell thanks!

PR #2360, opened 2023-07-09
Fix `MosCpuInfoProvider`

timcassell:


src/BenchmarkDotNet/Portability/Cpu/MosCpuInfoProvider.cs
```diff
@@ -50,8 +50,8 @@ private static CpuInfo Load()
                 physicalCoreCount > 0 ? (int?) physicalCoreCount : null,
                 logicalCoreCount > 0 ? (int?) logicalCoreCount : null,
                 nominalClockSpeed > 0 && logicalCoreCount > 0 ? Frequency.FromMHz(nominalClockSpeed) : (Frequency?)null,
-                maxClockSpeed > 0 && logicalCoreCount > 0 ? Frequency.FromMHz(maxClockSpeed) : 0,
-                minClockSpeed > 0 && logicalCoreCount > 0 ? Frequency.FromMHz(minClockSpeed) : (Frequency?)null);
+                minClockSpeed > 0 && logicalCoreCount > 0 ? Frequency.FromMHz(minClockSpeed) : (Frequency?)null,
+                maxClockSpeed > 0 && logicalCoreCount > 0 ? Frequency.FromMHz(maxClockSpeed) : 0);
```
AndreyAkinshin, line 54: Should we also use `(Frequency?)null` for the non-defined value of `maxFrequency` instead of `0 Hz`?

timcassell: I don't know the purpose of that. I just swapped the min and max here.

AndreyAkinshin (approved):

PR #2370, opened 2023-07-14
Rebuild .Net Framework projects

### LINKED ISSUE #1773: Ability to recompile for each target runtime
Today in order to compare runtimes, I need to build against a single runtime (the lowest common denominator) with -f and then I can execute against multiple runtimes with --runtimes.  That works in the majority use case, however sometimes you want to answer the question "how does this code compare on different runtimes?", and that includes compile-time impact due to new overloads being introduced, which necessitates building against the appropriate reference assemblies for each version.

timcassell:
Resolves #1773


timcassell:
Looks like CI is unhappy trying to build the classic .Net Framework test project. Can you help me fix it @AndreyAkinshin?

AndreyAkinshin:
Here is the problem:
```
D:\a\BenchmarkDotNet\BenchmarkDotNet\tests\ClassicNetFrameworkTests\BenchmarkDotNet.IntegrationTests.ClassicNetFramework.csproj(153,5): error : This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is ..\..\packages\xunit.core.2.5.0\build\xunit.core.props.
```
We use .NET SDK for all the build-related operations that don't support classic .NET applications. If we want to fix it, we should do the following:
* Use full VS MSBuild on Windows (that adds an additional requirement for the development setup)
* Exclude this project on Linux/macOS or create a crazy build toolset that involves Mono MSBuild.

I would like to avoid this. This looks like an overcomplication of our build script with low visible benefits. Do we really need the ClassicNetFramework project?

timcassell:
> Do we really need the ClassicNetFramework project?

I just wanted to be sure that it still works properly. It worked on my Windows machine when I tested it, so I suppose I can just remove it from the sln.

timcassell:
I moved the tests to their own sln for manual execution (and I also moved the existing manual tests to that sln).

Now, the only thing we don't support is recompiling for classic Mono with different target frameworks, but I don't think it's worth adding that support, as we would need new runtime monikers (`Mono461` etc), but what do you think @AndreyAkinshin?

BenchmarkDotNet.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2027
+# Visual Studio Version 17
```
adamsitnik, line 3: We avoid updating the VS version as we don't want to force the contributors to use latest VS, however they need to install .NET 7 SDK anyway. Is this version the oldest that supports .NET 7?

src/BenchmarkDotNet/Properties/AssemblyInfo.cs
```diff
@@ -15,10 +15,14 @@
 [assembly: InternalsVisibleTo("BenchmarkDotNet.Diagnostics.Windows,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
 [assembly: InternalsVisibleTo("BenchmarkDotNet.Diagnostics.dotTrace,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
 [assembly: InternalsVisibleTo("BenchmarkDotNet.IntegrationTests.ManualRunning,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
+[assembly: InternalsVisibleTo("BenchmarkDotNet.IntegrationTests.ClassicNetFramework,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
+[assembly: InternalsVisibleTo("BenchmarkDotNet.IntegrationTests.CsProjMultipleFrameworks,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
```
adamsitnik, line 19: is this test going to be executed by the CI? https://github.com/dotnet/BenchmarkDotNet/blob/6fa7a91444b01029cc2ed81c90e237ae3e36c595/build/BenchmarkDotNet.Build/Runners/UnitTestRunner.cs#L21-L27

testsmanual/BenchmarkDotNet.IntegrationTests.ClassicNetFramework/BenchmarkDotNet.IntegrationTests.ClassicNetFramework.csproj
```diff
@@ -0,0 +1,161 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
adamsitnik, line 1: using the old project files will be expensive to maintain

testsmanual/BenchmarkDotNet.IntegrationTests.ClassicNetFramework/ClassicNetFrameworkTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            "// ### Called: Benchmark", // MainTarget1
+            "// ### Called: IterationCleanup (3)", // MainTarget1
+            "// ### Called: IterationSetup (4)", // MainTarget2
+            "// ### Called: Benchmark", // MainTarget2
+            "// ### Called: IterationCleanup (4)", // MainTarget2
+            "// ### Called: IterationSetup (5)", // MainTarget3
+            "// ### Called: Benchmark", // MainTarget3
+            "// ### Called: IterationCleanup (5)", // MainTarget3
+
+            "// ### Called: GlobalCleanup"
```
adamsitnik, line 42: the tests that verify the order of printed output are flaky. If all we care about is ensuring that .NET Framework works, we should just verify the runtime in the benchmark and throw if it's the wrong one

testsmanual/BenchmarkDotNet.IntegrationTests.ClassicNetFramework/app.config
```diff
@@ -0,0 +1,27 @@
...
+        <assemblyIdentity name="System.Memory" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.1.2" newVersion="4.0.1.2" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Reflection.Metadata" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-7.0.0.2" newVersion="7.0.0.2" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
```
adamsitnik, line 27: another thing that is going to be expensive to maintain

testsmanual/BenchmarkDotNet.IntegrationTests.CsProjNetFramework/MultipleFrameworksTest.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public MultipleFrameworksTest(ITestOutputHelper output) : base(output) { }
+
+        private static string[] GetActualLogLines(Summary summary)
+            => GetSingleStandardOutput(summary).Where(line => line.StartsWith(Prefix)).ToArray();
+
+        [Theory]
+        [InlineData(RuntimeMoniker.Net461)]
+        [InlineData(RuntimeMoniker.Net48)]
+        [InlineData(RuntimeMoniker.NetCoreApp20)]
```
adamsitnik, line 44: our CI has no .NET Core 2.0 SDK installed, it's not going to work.

adamsitnik (changes requested):
First of all, thank you for your contribution @timcassell 

This PR brings a LOT of complexity, just to support one very unique scenario that gets less and less important every year (I doubt anyone will care about comparing .NET vs .NET Framework in two years from now). I don't believe it's worth it.

I recommend that we rather simply prefer `CsProjClassicNetToolchain` over `RoslynToolchain` when users are running the benchmarks that compare multiple runtimes. It should solve this problem and require less changes.

BenchmarkDotNet.sln
timcassell: That was an automatic change by VS when I moved the project out of the sln. But yes, that also happens to be the oldest VS that supports net7.

src/BenchmarkDotNet/Properties/AssemblyInfo.cs
timcassell: No, these are all manual tests.

testsmanual/BenchmarkDotNet.IntegrationTests.CsProjNetFramework/MultipleFrameworksTest.cs
timcassell: No problem, the tests are manual only.

testsmanual/BenchmarkDotNet.IntegrationTests.ClassicNetFramework/BenchmarkDotNet.IntegrationTests.ClassicNetFramework.csproj
timcassell: Agreed. Should I just remove that test then? I guess it's not so important nowadays.

timcassell:
> I recommend that we rather simply prefer `CsProjClassicNetToolchain` over `RoslynToolchain` when users are running the benchmarks that compare multiple runtimes.

I think we could actually just always default to `CsProjClassicNetToolchain`, since in my testing, it always succeeded, even with the classic csproj. Users could then just manually specify `RoslynToolchain` if they really need it.

tests/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks/MultipleFrameworksTest.cs
```diff
@@ -0,0 +1,115 @@
...
+                expectedBenchmarkCalled, // MainTarget2
+                "// ### Called: IterationCleanup (4)", // MainTarget2
+                "// ### Called: IterationSetup (5)", // MainTarget3
+                expectedBenchmarkCalled, // MainTarget3
+                "// ### Called: IterationCleanup (5)", // MainTarget3
+
+                "// ### Called: GlobalCleanup"
+            };
+
+            SmartAssert.Equal(expectedLogLines, actualLogLines);
```
adamsitnik, line 91: the tests that verify the printed output are often flaky, we should avoid them when possible.

In this case we could pass the expected enum as string via env var and verify the compiled value in the benchmark process:

```cs
private const string TfmEnvVarName = "TfmEnvVarName";

[Theory]
[InlineData(RuntimeMoniker.Net461)]
[InlineData(RuntimeMoniker.Net48)]
[InlineData(RuntimeMoniker.NetCoreApp20)]
[InlineData(RuntimeMoniker.Net70)]
public void EachFrameworkIsRebuilt(RuntimeMoniker runtime)
{
    var config = ManualConfig.CreateEmpty().AddJob(Job.Dry.WithRuntime(runtime.GetRuntime()).WithEnvironmentVariable(TfmEnvVarName, runtime.ToString()));
    CanExecute<ValuePerTfm>(config);
}

public class ValuePerTfm
{
    private RuntimeMoniker moniker =
#if NET461
        RuntimeMoniker.Net461;
#elif NET48
        RuntimeMoniker.Net48;
#elif NETCOREAPP2_0
        RuntimeMoniker.NetCoreApp20;
#elif NET7_0
        RuntimeMoniker.Net70;
#else
        RuntimeMoniker.NotRecognized;
#endif

    [Benchmark]
    public void ThrowWhenWrong()
    {
        if (Environment.GetEnvironmentVariable(TfmEnvVarName) != moniker.ToString())
        {
            throw new InvalidOperationException($"Has not been recompiled, the value was {Environment.GetEnvironmentVariable(TfmEnvVarName)}");
        }
    }
}
```

I know it looks hacky, but it should be stable and get us what we want.

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
```diff
@@ -34,12 +35,17 @@ internal static IToolchain GetToolchain(this Runtime runtime, Descriptor descrip
             switch (runtime)
             {
                 case ClrRuntime clrRuntime:
-                    if (RuntimeInformation.IsNetCore || preferMsBuildToolchains)
-                        return clrRuntime.RuntimeMoniker != RuntimeMoniker.NotRecognized
-                            ? GetToolchain(clrRuntime.RuntimeMoniker)
-                            : CsProjClassicNetToolchain.From(clrRuntime.MsBuildMoniker);
-
-                    return RoslynToolchain.Instance;
+                    if (!preferMsBuildToolchains && RuntimeInformation.IsFullFramework &&
```
adamsitnik, line 38: the `RoslynToolchain` should also be used for classic Mono:

```suggestion
                    if (!preferMsBuildToolchains && (RuntimeInformation.IsFullFramework || RuntimeInformation.IsOldMono) &&
```

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
```diff
@@ -34,12 +35,17 @@ internal static IToolchain GetToolchain(this Runtime runtime, Descriptor descrip
...
+                        // If dotnet SDK is not installed, we use RoslynToolchain.
```
adamsitnik, line 39: thank you for covering that edge case! 👍

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
```diff
@@ -34,12 +35,17 @@ internal static IToolchain GetToolchain(this Runtime runtime, Descriptor descrip
...
+                        (!HostEnvironmentInfo.GetCurrent().IsDotNetCliInstalled()
+                        // Integration tests take too much time, because each benchmark run rebuilds the test suite and BenchmarkDotNet itself.
+                        // To reduce the total duration of the CI workflows, we just use RoslynToolchain.
+                        || XUnitHelper.IsIntegrationTest.Value))
+                        return RoslynToolchain.Instance;
```
adamsitnik, line 44: nit: this `if` became quite huge, we should use braces here

adamsitnik (commented):
@timcassell thank you for addressing my feedback, it looks much better now.

PTAL at my comments, we are almost there!

tests/BenchmarkDotNet.IntegrationTests.ManualRunning/xunit.runner.json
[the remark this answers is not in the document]
timcassell: That's a weird thing that happened with git when I moved the file then moved it back. It's not actually empty, I'm not sure why GitHub is showing it that way. I'll try to figure out how to revert it completely.

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
timcassell: Is that valid? I thought Mono would have to use a MonoRuntime.

adamsitnik: @timcassell you are right, it's not. Great catch! 👍

tests/BenchmarkDotNet.IntegrationTests.ManualRunning/xunit.runner.json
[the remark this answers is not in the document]
adamsitnik: if it's not empty then I am fine with that

PR #2381, opened 2023-07-26
Support `--cli` argument for `CsProjClassicNetToolchain`

timcassell:
For consistency with the validation error message when dotnet SDK is not installed in the default location.

### DIFF
--- a/src/BenchmarkDotNet/Diagnosers/PerfCollectProfiler.cs
+++ b/src/BenchmarkDotNet/Diagnosers/PerfCollectProfiler.cs
@@ -196,6 +196,7 @@ namespace BenchmarkDotNet.Diagnosers
         {
             string cliPath = parameters.BenchmarkCase.GetToolchain() switch
             {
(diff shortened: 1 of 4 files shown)


src/BenchmarkDotNet/Diagnosers/PerfCollectProfiler.cs
```diff
@@ -196,6 +196,7 @@ private void EnsureSymbolsForNativeRuntime(DiagnoserActionParameters parameters)
         {
             string cliPath = parameters.BenchmarkCase.GetToolchain() switch
             {
+                CsProjClassicNetToolchain classic => classic.CustomDotNetCliPath,
```
adamsitnik, line 199: The `perfcollect` profiler plugin can only be used on Linux, and `CsProjClassicNetToolchain` is needed only on Windows, so this change is redundant. 
```suggestion
```

adamsitnik (approved):
LGTM, thank you @timcassell !

PR #2386, opened 2023-07-31
Ability to specify  benchmark description in outputs

### LINKED ISSUE #1447: Ability to configure benchmark class name in outputs

Nepp3r:
Hello,
I think I've done this recomendations: 
https://github.com/dotnet/BenchmarkDotNet/pull/1651#issuecomment-1645092070

If you have some other recomendations I am open to them.

Closes #1447
Supersedes #1651


src/BenchmarkDotNet.Annotations/Attributes/BenchmarkDescriptionAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BenchmarkDotNet.Attributes
+{
+    public class BenchmarkDescriptionAttribute : Attribute
```
timcassell, line 7: Add attribute usage (class and method).

src/BenchmarkDotNet/Running/BenchmarkConverter.cs
```diff
@@ -119,6 +119,8 @@ private static ImmutableConfig GetFullMethodConfig(MethodInfo method, ImmutableC
             Tuple<MethodInfo, TargetedAttribute>[] iterationCleanupMethods,
             IConfig config)
         {
+            var abc = type.GetCustomAttributes(typeof(BenchmarkDescriptionAttribute), false);
```
timcassell, line 122: Not sure what `abc` is for, just remove.

src/BenchmarkDotNet/Running/BenchmarkConverter.cs
```diff
@@ -127,8 +129,10 @@ private static ImmutableConfig GetFullMethodConfig(MethodInfo method, ImmutableC
                                                    GetTargetedMatchingMethod(methodInfo, iterationSetupMethods),
                                                    GetTargetedMatchingMethod(methodInfo, iterationCleanupMethods),
                                                    methodInfo.ResolveAttribute<BenchmarkAttribute>(),
+                                                   classDescriptionAttribute,
+                                                   methodInfo.ResolveAttribute<BenchmarkDescriptionAttribute>(),
                                                    targetMethods,
-                                                   config));
+                                                   config)) ;
```
timcassell, line 135: Revert that extra space.

src/BenchmarkDotNet/Running/BenchmarkConverter.cs
```diff
@@ -155,18 +159,25 @@ private static MethodInfo GetTargetedMatchingMethod(MethodInfo benchmarkMethod,
...
             BenchmarkAttribute attr,
+            BenchmarkDescriptionAttribute? classDescription,
+            BenchmarkDescriptionAttribute? methodDescription,
             MethodInfo[] targetMethods,
             IConfig config)
         {
             var categoryDiscoverer = config.CategoryDiscoverer ?? DefaultCategoryDiscoverer.Instance;
+            string description = attr?.Description;
+                if (description is null)
+                    description = methodDescription?.Description;
```
timcassell, line 170: `description ??= methodDescription?.Description;`

src/BenchmarkDotNet/Running/BenchmarkConverter.cs
```diff
@@ -155,18 +159,25 @@ private static MethodInfo GetTargetedMatchingMethod(MethodInfo benchmarkMethod,
...
+                if (description is null)
+                    description = classDescription?.Description;
```
timcassell, line 172: I'm not sure if this makes sense. We want to override the `Type` description, not the method description.

timcassell (changes requested):
For the approval tests, please compare 2 types, 1 with `BenchmarkDescription` and 1 without. The table should look something like this:

```
                               Type | Method |  Job |     Mean |   Error |  StdDev | Rank | LogicalGroup | Baseline |
----------------------------------- |------- |----- |---------:|--------:|--------:|-----:|------------- |--------- |
JobBaseline_NoRenameJob_MethodsJobs |   Base | Job1 | 102.0 ns | 6.09 ns | 1.58 ns |    1 |            * |       No |
JobBaseline_NoRenameJob_MethodsJobs |    Foo | Job1 | 202.0 ns | 6.09 ns | 1.58 ns |    2 |            * |       No |
JobBaseline_NoRenameJob_MethodsJobs |    Bar | Job1 | 302.0 ns | 6.09 ns | 1.58 ns |    3 |            * |       No |
                  MyRenamedTestCase |   Base | Job1 | 102.0 ns | 6.09 ns | 1.58 ns |    4 |            * |       No |
                  MyRenamedTestCase |    Foo | Job1 | 202.0 ns | 6.09 ns | 1.58 ns |    5 |            * |       No |
                  MyRenamedTestCase |    Bar | Job1 | 302.0 ns | 6.09 ns | 1.58 ns |    6 |            * |       No |
```

tests/BenchmarkDotNet.Tests/Running/BenchmarkConverterTests.cs
```diff
@@ -278,5 +298,38 @@ public class PrivateIterationCleanup
...
+
+            [Benchmark(Description = "from Benchmark")]
+            public void ClassBenchmarkAttributeOverride() { }
+
+            [Benchmark]
+            [BenchmarkDescription("OverrideFromAttribute")]
+            public void ClassBenchmarkDescriptionAttributeOverride() { }
+
+            [Benchmark(Description = "Who are the winner?")]
+            [BenchmarkDescription("OverrideFromAttribute")]
```
timcassell, line 331: We may want a validator to prevent using both descriptions.

src/BenchmarkDotNet/Running/BenchmarkConverter.cs
```diff
@@ -127,8 +129,10 @@ private static ImmutableConfig GetFullMethodConfig(MethodInfo method, ImmutableC
                                                    GetTargetedMatchingMethod(methodInfo, iterationSetupMethods),
                                                    GetTargetedMatchingMethod(methodInfo, iterationCleanupMethods),
                                                    methodInfo.ResolveAttribute<BenchmarkAttribute>(),
+                                                   classDescriptionAttribute,
```
timcassell, line 132: You can inline the `type.ResolveAttribute` call here.

src/BenchmarkDotNet/Running/BenchmarkConverter.cs
```diff
@@ -155,18 +158,25 @@ private static MethodInfo GetTargetedMatchingMethod(MethodInfo benchmarkMethod,
...
             BenchmarkAttribute attr,
+            BenchmarkDescriptionAttribute? classDescription,
+            BenchmarkDescriptionAttribute? methodDescription,
             MethodInfo[] targetMethods,
             IConfig config)
         {
             var categoryDiscoverer = config.CategoryDiscoverer ?? DefaultCategoryDiscoverer.Instance;
+            string description = attr?.Description;
+                if (description is null)
+                    description ??= methodDescription?.Description;
```
timcassell, line 169: You don't need the `is null` check and `??=`. `??=` is fine by itself.

Nepp3r:
> For the approval tests, please compare 2 types, 1 with `BenchmarkDescription` and 1 without. The table should look something like this:
> 
> ```
>                                Type | Method |  Job |     Mean |   Error |  StdDev | Rank | LogicalGroup | Baseline |
> ----------------------------------- |------- |----- |---------:|--------:|--------:|-----:|------------- |--------- |
> JobBaseline_NoRenameJob_MethodsJobs |   Base | Job1 | 102.0 ns | 6.09 ns | 1.58 ns |    1 |            * |       No |
> JobBaseline_NoRenameJob_MethodsJobs |    Foo | Job1 | 202.0 ns | 6.09 ns | 1.58 ns |    2 |            * |       No |
> JobBaseline_NoRenameJob_MethodsJobs |    Bar | Job1 | 302.0 ns | 6.09 ns | 1.58 ns |    3 |            * |       No |
>                   MyRenamedTestCase |   Base | Job1 | 102.0 ns | 6.09 ns | 1.58 ns |    4 |            * |       No |
>                   MyRenamedTestCase |    Foo | Job1 | 202.0 ns | 6.09 ns | 1.58 ns |    5 |            * |       No |
>                   MyRenamedTestCase |    Bar | Job1 | 302.0 ns | 6.09 ns | 1.58 ns |    6 |            * |       No |
> ```

If I understand clearly, I need to add Type column to the table if class has BenchmarkDescription attribute with description, right?

Nepp3r:
@microsoft-github-policy-service agree

Nepp3r:
> It will be added automatically if more than 1 type 

I tried to change MockReport.CreateSummary, so method could get an array of types, but it would require too many changes, so I refused from it. How exactly do I need to send more than one type to the summary?

timcassell:
> > It will be added automatically if more than 1 type
> 
> I tried to change MockReport.CreateSummary, so method could get an array of types, but it would require too many changes, so I refused from it. How exactly do I need to send more than one type to the summary?

I think such changes are necessary. Please go ahead and add that functionality.

Nepp3r:
> > > It will be added automatically if more than 1 type
> > 
> > 
> > I tried to change MockReport.CreateSummary, so method could get an array of types, but it would require too many changes, so I refused from it. How exactly do I need to send more than one type to the summary?
> 
> I think such changes are necessary. Please go ahead and add that functionality.

Ok, but I tried to create new benchmark, like with instrument. I created 2 classes, both with benchmarked methods, but results were divided into 2 files, like it is 2 separate runs for benchmarking. Also I didn't find any such verified files in the _tests/BenchmarkDotNet.Tests/Exporters/VerifiedFiles_. Could you please give me the example of the test, which gave you that one result higher or explain, how can I put results of two methods from different class to one result table?

timcassell:
@Nepp3r Try this
```cs
BenchmarkSwitcher.FromTypes(typeof(Demo), typeof(Demo2)).RunAllJoined(args: args);
```

timcassell:
I actually think we should just use `System.ComponentModel.DescriptionAttribute` instead of creating a new `BenchmarkDescriptionAttribute`. We also could use it for `[Params]` fields/properties.

PR #2387, opened 2023-08-01
Adding tests for export of csv and csv Measurements

Nepp3r:
Hello,
I've noticed, that in BenchmarkDotNet.Tests.Exporters.VerifiedFiles.CommonExporterVerifyTests was todo to add CsvExporter and CsvMeasurementsExporter, so I added them and mocked verified.txt files.


Nepp3r:
@dotnet-policy-service agree

tests/BenchmarkDotNet.Tests/Exporters/CommonExporterVerifyTests.cs
```diff
@@ -43,15 +48,15 @@ public CommonExporterVerifyTests()
...
+            Thread.CurrentThread.CurrentCulture = cultureInfo;
+            EnvironmentResolver.Default.Register(EnvironmentMode.AffinityCharacteristic, () => new IntPtr(0xFF));
             var logger = new AccumulationLogger();
-
             var exporters = GetExporters();
             foreach (var exporter in exporters)
             {
-                PrintTitle(logger, exporter);
-                exporter.ExportToLog(MockFactory.CreateSummary(config.WithCultureInfo(cultureInfo)), logger);
+                PrintTitle(logger, exporter);
```
kant2002, line 57: This is not required changes, better revert them.

tests/BenchmarkDotNet.Tests/Exporters/CommonExporterVerifyTests.cs
```diff
@@ -43,15 +48,15 @@ public CommonExporterVerifyTests()
         public Task Exporters(string cultureInfoName)
         {
             var cultureInfo = CultureInfos[cultureInfoName];
-            Thread.CurrentThread.CurrentCulture = cultureInfo;
-
+            Thread.CurrentThread.CurrentCulture = cultureInfo;
+            EnvironmentResolver.Default.Register(EnvironmentMode.AffinityCharacteristic, () => new IntPtr(0xFF));
```
kant2002, line 52: Unnecessary empty line

timcassell:
> was todo to add CsvExporter and CsvMeasurementsExporter

Can you remove that todo since you did it?

timcassell (approved):
LGTM, but will wait for Andrey to take a look before merging.

AndreyAkinshin (changes requested):
I don't like this approach, specifically hacking the global instance of `EnvironmentResolver`.

The original idea behind `IResolver` was to provide an opportunity to easily customize the resolve logic for all the basic characteristics like affinity. Unfortunately, the resolver/characteristics system turned out to be overcomplicated, so it's not so easy to use it. Therefore, BenchmarkDotNet uses static instances of various resolvers here and there (like in `BenchmarkDotNet.Columns.JobCharacteristicColumn.GetValue`). A proper approach here is to somehow pass a resolver to all the places where it is needed so that it can be easily overridden in the tests.

It's always good to increase our test coverage, but I'd rather avoid introducing additional internal test-only methods for changing statistic instances. It makes our code base more fragile and increases the technical dept. A better solution is to invest efforts in refactoring the existing design so that we can write tests without using internal methods.

PR #2403, opened 2023-08-16
Add OnSummary method to Diagnoser

lihaMSFT:
This PR adds an `OnSummary` method to BenchmarkDotNet.Diagnoser. The purpose is to get BDN Summary object into IDiagnoser so we can process and save the summary.

To ensure backwards compatibility, it is defined as an extension interface to the original `IDiagnoser` interface and called only when a diagnoser object implements the new extension interface.


lihaMSFT:
@dotnet-policy-service agree company="Microsoft"

timcassell:
What's an example use-case for this? Why can't you just use IExporter?

lihaMSFT:
> What's an example use-case for this? Why can't you just use IExporter?

Ah I didn't notice this. IExporter works for my scenario. Thank you.

Closing this PR...

PR #2406, opened 2023-08-17
Refactor out base TextLogger from StreamLogger

### LINKED ISSUE #2400: Should StreamLogger be TextWriterLogger?
StreamLogger is for some reason (other than name) tied to only being useable with `StreamWriter`, which is just a `TextWriter`, as far as I can tell there is no reason for the implementation to tie into it being a `StreamWriter` and one could easily refactor to `TextWriter` or extracting a base class `TextWriterLogger` with all code, but which allows using a `TextWriter` instead. This would allow exporting to in-memory `StringWriter` instead of having to go through a `MemoryStream`.

https://github.com/dotnet/BenchmarkDotNet/blob/d058c7b6ff2434e96f177f976a0ce06e1a1643b7/src/BenchmarkDotNet/Loggers/StreamLogger.cs#L7-L28

nietras:
Fixes #2400

### DIFF
--- a/src/BenchmarkDotNet/Loggers/TextLogger.cs
+++ b/src/BenchmarkDotNet/Loggers/TextLogger.cs
@@ -9,8 +9,6 @@ namespace BenchmarkDotNet.Loggers
 
         public TextLogger(TextWriter writer) => this.writer = writer;
 
-        public void Dispose() => writer.Dispose();
-
         public virtual string Id => nameof(TextLogger);
         public int Priority => 0;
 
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Loggers/TextLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class TextLogger : ILogger, IDisposable
+    {
+        private readonly TextWriter writer;
+
+        public TextLogger(TextWriter writer) => this.writer = writer;
+
+        public void Dispose() => writer.Dispose();
+
+        public string Id => nameof(TextLogger);
```
timcassell, line 14: Please make `Id` virtual and keep `nameof(StreamLogger)`.

nietras: my bad, should be fixed now

timcassell (approved):

timcassell:
Thanks @nietras!

PR #2407, opened 2023-08-17
Make MarkdownExporter ctor and Dialect protected

### LINKED ISSUE #2399: Custom MarkdownExporter
I'm trying to avoid the GitHub MarkdownExporter highlighting certain rows using Bold. Looking at source code there is an option:

https://github.com/dotnet/BenchmarkDotNet/blob/d058c7b6ff2434e96f177f976a0ce06e1a1643b7/src/BenchmarkDotNet/Exporters/MarkdownExporter.cs#L81

However, there is no way to set this or override. Additionally, one cannot simple create a sub-class for customization as the ctor is private.

https://github.com/dotnet/BenchmarkDotNet/blob/d058c7b6ff2434e96f177f976a0ce06e1a1643b7/src/BenchmarkDotNet/Exporters/MarkdownExporter.cs#L89

Could this not be opened up? So one can reuse these parts?

Are there other ways of avoiding the bolding, and how can I in code get the file names of exported reports?

nietras:
Fixes #2399

### DIFF
--- a/src/BenchmarkDotNet/Exporters/MarkdownExporter.cs
+++ b/src/BenchmarkDotNet/Exporters/MarkdownExporter.cs
@@ -22,7 +22,7 @@ namespace BenchmarkDotNet.Exporters
         protected override string FileExtension => "md";
         protected override string FileNameSuffix => $"-{Dialect.ToLower()}";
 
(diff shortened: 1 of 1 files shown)


adamsitnik (approved):
LGTM, thank you @nietras !

Issue #1995, opened 2022-04-17
Left Justification option for summary table columns


erichiller:
It would be useful if there was an option to align/justify the Method column (leftmost column when there's only one namespace and It would be useful if there was an option to align/justify the Method column (leftmost column when there's only one namespace and class). My use case is that I typically name my methods in order of increasing specificity (opposite of DNS).

eg. this:
```
Method                            | Metric 1
----------------------------------|-------------
MajorMinorSpecificBaseline        | 1
MajorMinorSpecificDoItOneWay      | 2
SomeVeryDifferentMinorSpecific    | 3
MajorMinorSpecificDoItAnotherWay  | 4
```

Rather than the current:
```
Method                            | Metric 1
----------------------------------|-------------
       MajorMinorSpecificBaseline | 1
     MajorMinorSpecificDoItOneWay | 2
   SomeVeryDifferentMinorSpecific | 3
 MajorMinorSpecificDoItAnotherWay | 4
```

With left justification, the bases are aligned which makes it easier for me to quickly scan the column and see what methods are similar. Whereas with right justification there is no alignment between names at all.

My thought was adding something like `DefaultTextColumnJustification` to [`SummaryStyle`] with a type of `TextJustification`.

From a brief glance, I think the following would need to change:
- Currently [`SummaryTableColumn`] has a [`Justify`] property, but it is [set by whether the input `IColumn` is text or numeric], as there's no justify property on [`IColumn`]
- [`SummaryTable`] could pass the new `DefaultTextColumnJustification` from  [`SummaryStyle`] to [`SummaryTableColumn`] [here](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Reports/SummaryTable.cs#L89)

[`Justify`]: https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Reports/SummaryTable.cs#L101
[`SummaryTableColumn`]: https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Reports/SummaryTable.cs#L93

[set by whether the input `IColumn` is text or numeric]: https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Reports/SummaryTable.cs#L114

[`IColumn`]: https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Columns/IColumn.cs

[...]

PR #2410, opened 2023-08-18
feat: add text justification style

Vahdanian:
Hi, thanks for reviewing my pull request. This pull request intends to fix [https://github.com/dotnet/BenchmarkDotNet/issues/1995](this issue)

Here are some key points worth mentioning:
1- After initial investigation, I found that Justify property in "SummaryTableColumn" class does not have impact on aligning the contents of summary table. It only effects the header separator of table. (See the first attachment please)

2- I managed to add the "DefaultTextJustification" property to "SummaryStyle" and use the style to set "Justify" property inside "SummaryTableColumn". I changed the "BuildStandardText" and "BuildBoldText" methods to use "Justify" property.

3- Since I changed how the Justify Property was being initialized and the previous value was only used inside "GetJustificationIndicator" method, I changed the method to read from "IsNumeric" value of passed column. Exactly like the previous value of "Justify" property

4- I modified SummaryTableTests class and added a new method to MockFactory class to test the new style. CreateSummary method inside MockFactory did not used the specified style of config object and directly called the constructor of Summary class where style was initialized with default instance. So I added a new parameter for style.

5- Finally I modified methods "SummaryStyle" class and added a new method in "HashCode" struct to support the newly added style "DefaultTextJustification".

6- You can see the final result in second attachment.

Justify not working on columns
after changes

### DIFF
--- a/src/BenchmarkDotNet/Reports/SummaryStyle.cs
+++ b/src/BenchmarkDotNet/Reports/SummaryStyle.cs
@@ -26,11 +26,13 @@ namespace BenchmarkDotNet.Reports
         public CultureInfo CultureInfo { get; }
 
         public RatioStyle RatioStyle { get; }
-        public TextJustification DefaultTextJustification { get; }
+
+        public TextJustification TextColumnJustification { get; }
+        public TextJustification NumericColumnJustification { get; }
 
         public SummaryStyle(CultureInfo? cultureInfo, bool printUnitsInHeader, SizeUnit sizeUnit, TimeUnit timeUnit, bool printUnitsInContent = true,
             bool printZeroValuesInContent = false, int maxParameterColumnWidth = DefaultMaxParameterColumnWidth, RatioStyle ratioStyle = RatioStyle.Value,
-            TextJustification defaultTextJustification = TextJustification.Left)
+            TextJustification textColumnJustification = TextJustification.Left, TextJustification numericColumnJustification = TextJustification.Left)
         {
             if (maxParameterColumnWidth < DefaultMaxParameterColumnWidth)
                 throw new ArgumentOutOfRangeException(nameof(maxParameterColumnWidth), $"{DefaultMaxParameterColumnWidth} is the minimum.");
@@ -40,36 +42,37 @@ namespace BenchmarkDotNet.Reports
             PrintUnitsInContent = printUnitsInContent;
             SizeUnit = sizeUnit;
             TimeUnit = timeUnit;
-            PrintZeroValuesInContent = printZeroValuesInContent;
             MaxParameterColumnWidth = maxParameterColumnWidth;
             RatioStyle = ratioStyle;
             CodeSizeUnit = SizeUnit.B;
-            DefaultTextJustification = defaultTextJustification;
+            PrintZeroValuesInContent = printZeroValuesInContent;
+            TextColumnJustification = textColumnJustification;
+            NumericColumnJustification = numericColumnJustification;
         }
 
         public SummaryStyle WithTimeUnit(TimeUnit timeUnit)
             => new SummaryStyle(CultureInfo, PrintUnitsInHeader, SizeUnit, timeUnit, PrintUnitsInContent, PrintZeroValuesInContent, MaxParameterColumnWidth,
-                RatioStyle, DefaultTextJustification);
+                RatioStyle, TextColumnJustification, NumericColumnJustification);
 
         public SummaryStyle WithSizeUnit(SizeUnit sizeUnit)
             => new SummaryStyle(CultureInfo, PrintUnitsInHeader, sizeUnit, TimeUnit, PrintUnitsInContent, PrintZeroValuesInContent, MaxParameterColumnWidth,
-                RatioStyle, DefaultTextJustification);
+                RatioStyle, TextColumnJustification, NumericColumnJustification);
 
         public SummaryStyle WithZeroMetricValuesInContent()
             => new SummaryStyle(CultureInfo, PrintUnitsInHeader, SizeUnit, TimeUnit, PrintUnitsInContent, printZeroValuesInContent: true,
-                MaxParameterColumnWidth, RatioStyle, DefaultTextJustification);
+                MaxParameterColumnWidth, RatioStyle, TextColumnJustification, NumericColumnJustification);
 
         public SummaryStyle WithMaxParameterColumnWidth(int maxParameterColumnWidth)
             => new SummaryStyle(CultureInfo, PrintUnitsInHeader, SizeUnit, TimeUnit, PrintUnitsInContent, PrintZeroValuesInContent, maxParameterColumnWidth,
-                RatioStyle, DefaultTextJustification);
+                RatioStyle, TextColumnJustification, NumericColumnJustification);
 
         public SummaryStyle WithCultureInfo(CultureInfo cultureInfo)
             => new SummaryStyle(cultureInfo, PrintUnitsInHeader, SizeUnit, TimeUnit, PrintUnitsInContent, PrintZeroValuesInContent, MaxParameterColumnWidth,
-                RatioStyle, DefaultTextJustification);
+                RatioStyle, TextColumnJustification, NumericColumnJustification);
 
         public SummaryStyle WithRatioStyle(RatioStyle ratioStyle)
             => new SummaryStyle(CultureInfo, PrintUnitsInHeader, SizeUnit, TimeUnit, PrintUnitsInContent, PrintZeroValuesInContent, MaxParameterColumnWidth,
-                ratioStyle, DefaultTextJustification);
+                ratioStyle, TextColumnJustification, NumericColumnJustification);
 
         public bool Equals(SummaryStyle other)
         {
@@ -86,7 +89,8 @@ namespace BenchmarkDotNet.Reports
                    && Equals(TimeUnit, other.TimeUnit)
                    && MaxParameterColumnWidth == other.MaxParameterColumnWidth
                    && RatioStyle == other.RatioStyle
-                   && DefaultTextJustification == other.DefaultTextJustification;
+                   && TextColumnJustification == other.TextColumnJustification
+                   && NumericColumnJustification == other.NumericColumnJustification;
         }
 
         public override bool Equals(object obj) => obj is SummaryStyle summary && Equals(summary);
@@ -101,7 +105,8 @@ namespace BenchmarkDotNet.Reports
                 TimeUnit,
                 MaxParameterColumnWidth,
                 RatioStyle,
-                DefaultTextJustification);
+                TextColumnJustification,
+                NumericColumnJustification);
 
         public static bool operator ==(SummaryStyle left, SummaryStyle right) => Equals(left, right);
 
--- a/src/BenchmarkDotNet/Reports/SummaryStyle.cs
+++ b/src/BenchmarkDotNet/Reports/SummaryStyle.cs
@@ -32,7 +32,7 @@ namespace BenchmarkDotNet.Reports
 
         public SummaryStyle(CultureInfo? cultureInfo, bool printUnitsInHeader, SizeUnit sizeUnit, TimeUnit timeUnit, bool printUnitsInContent = true,
             bool printZeroValuesInContent = false, int maxParameterColumnWidth = DefaultMaxParameterColumnWidth, RatioStyle ratioStyle = RatioStyle.Value,
-            TextJustification textColumnJustification = TextJustification.Left, TextJustification numericColumnJustification = TextJustification.Left)
+            TextJustification textColumnJustification = TextJustification.Left, TextJustification numericColumnJustification = TextJustification.Right)
         {
             if (maxParameterColumnWidth < DefaultMaxParameterColumnWidth)
                 throw new ArgumentOutOfRangeException(nameof(maxParameterColumnWidth), $"{DefaultMaxParameterColumnWidth} is the minimum.");
--- a/src/BenchmarkDotNet/Reports/SummaryStyle.cs
+++ b/src/BenchmarkDotNet/Reports/SummaryStyle.cs
@@ -3,6 +3,7 @@ using System.Globalization;
 using BenchmarkDotNet.Columns;
 using BenchmarkDotNet.Helpers;
 using Perfolizer.Horology;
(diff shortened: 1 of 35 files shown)


timcassell:
It looks strange to me that there is no space on the left when it's left-justified, but there is a space on the right when it's right-justified.

tests/BenchmarkDotNet.Tests/Reports/SummaryTableTests.cs
```diff
@@ -46,19 +46,21 @@ public void PlatformTest()
         }
 
         [Fact]
-        public void NumericColumnIsRightJustified()
```
timcassell, line 49: Please add new tests rather than overwriting the old tests.

src/BenchmarkDotNet/Reports/SummaryStyle.cs
```diff
@@ -23,9 +26,11 @@ public class SummaryStyle : IEquatable<SummaryStyle>
         public CultureInfo CultureInfo { get; }
 
         public RatioStyle RatioStyle { get; }
+        public TextJustification DefaultTextJustification { get; }
```
timcassell, line 29:
```suggestion
        public TextJustification Justify { get; }
```

tests/BenchmarkDotNet.Tests/Reports/SummaryTableTests.cs
Vahdanian: For these two tests, The previous intention was completely outdated and not really working. the justify value was set for columns but it was only used to justify header separator indicator. Also please note that since numeric columns are not justified right anymore (unless user sets newly added justify style to right) and it had caused the tests to fail. I updated the tests for new design.
What do you suggest?

Vahdanian:
@timcassell
Sorry to bother but I needed to ask. What do you suggest for the tests that use a predefined .txt files in "VerifiedFiles" directory to verify the output? I started to notice some of these tests are failing because of new justify style (You can find one [here](https://github.com/dotnet/BenchmarkDotNet/blob/master/tests/BenchmarkDotNet.Tests/Attributes/ParamsAllValuesVerifyTests.cs)). Shall I update the txt files to expect columns to be justified to left by default? Or you have something else in mind? please let me know.

tests/BenchmarkDotNet.Tests/Reports/SummaryTableTests.cs
timcassell: I suggest fixing the old tests. You may need to add a separate option for text justification and numeric justification. Header justification may or may not need its own option also, but I'm not strongly opinionated on that.

timcassell:
> Shall I update the txt files to expect columns to be justified to left by default?

That sound reasonable to me. Especially since the old test expects it (but seems broken currently).

Vahdanian:
@timcassell
I separated styling for numeric and text columns and fixed broken tests, Can you review the changes?

src/BenchmarkDotNet/Reports/SummaryStyle.cs
```diff
@@ -24,8 +27,12 @@ public class SummaryStyle : IEquatable<SummaryStyle>
 
         public RatioStyle RatioStyle { get; }
 
+        public TextJustification TextColumnJustification { get; }
+        public TextJustification NumericColumnJustification { get; }
+
         public SummaryStyle(CultureInfo? cultureInfo, bool printUnitsInHeader, SizeUnit sizeUnit, TimeUnit timeUnit, bool printUnitsInContent = true,
-            bool printZeroValuesInContent = false, int maxParameterColumnWidth = DefaultMaxParameterColumnWidth, RatioStyle ratioStyle = RatioStyle.Value)
+            bool printZeroValuesInContent = false, int maxParameterColumnWidth = DefaultMaxParameterColumnWidth, RatioStyle ratioStyle = RatioStyle.Value,
+            TextJustification textColumnJustification = TextJustification.Left, TextJustification numericColumnJustification = TextJustification.Left)
```
timcassell, line 35:
```suggestion
            TextJustification textColumnJustification = TextJustification.Left, TextJustification numericColumnJustification = TextJustification.Right)
```

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithNotAllowedFlagsEnumError.verified.txt
```diff
@@ -7,9 +7,9 @@ Frequency: 2531248 Hz, Resolution: 395.0620 ns, Timer: TSC
   DefaultJob : extra output line
 
 
-    Method | ParamProperty |     Mean |   Error |  StdDev |
----------- |-------------- |---------:|--------:|--------:|
- Benchmark |             0 | 102.0 ns | 6.09 ns | 1.58 ns |
+Method    | ParamProperty | Mean     | Error   | StdDev  | 
+--------- |-------------- |---------:|--------:|--------:|
+Benchmark | 0             | 102.0 ns | 6.09 ns | 1.58 ns |
```
timcassell, line 12: Can those extra spaces on the end be removed?

tests/BenchmarkDotNet.Tests/Reports/SummaryTableTests.cs
```diff
@@ -59,6 +61,7 @@ public void NumericColumnIsRightJustified()
         public void TextColumnIsLeftJustified()
         {
             var config = ManualConfig.Create(DefaultConfig.Instance).AddColumn(new ParamColumn("Param"));
+            config.SummaryStyle = MockFactory.CreateSummaryStyle(textColumnJustification: SummaryTable.SummaryTableColumn.TextJustification.Left);
```
timcassell, line 64: Please revert these tests for the default values, and add new tests with override values.

timcassell (changes requested):

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithNotAllowedFlagsEnumError.verified.txt
Vahdanian: > Can those extra spaces on the end be removed?

I should add some logic to TrimEnd() the "TableHeaderSeparator" and TableColumnSeparator" when we want to print them for last column. I wasn't sure that's what we want but if its okay with you I can add it.

Vahdanian:
@timcassell 
Hi, extra spaces are removed in 3d commit and numeric columns are now justified to right by default

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
```diff
@@ -7,9 +7,9 @@ Frequency: 2531248 Hz, Resolution: 395.0620 ns, Timer: TSC
   DefaultJob : extra output line
 
 
-    Method | ParamProperty |     Mean |   Error |  StdDev |
----------- |-------------- |---------:|--------:|--------:|
- Benchmark |         False | 102.0 ns | 6.09 ns | 1.58 ns | ^
- Benchmark |          True | 202.0 ns | 6.09 ns | 1.58 ns | ^
+Method    | ParamProperty |     Mean |   Error |  StdDev |
```
timcassell, line 10: Can you make all headers uniform justification? All left or all right, independent of the column's justification. It may make sense to use the `TextJustification` property (remove "Column" from the name).

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
```diff
@@ -7,9 +7,9 @@ Frequency: 2531248 Hz, Resolution: 395.0620 ns, Timer: TSC
...
+--------- |-------------- |---------:|--------:|--------:|
+Benchmark | False         | 102.0 ns | 6.09 ns | 1.58 ns | ^
+Benchmark | True          | 202.0 ns | 6.09 ns | 1.58 ns | ^
```
timcassell, line 13: Can you re-add the space in front of the first column?

timcassell (changes requested):

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
Vahdanian: I'm a little confused. Correct me if I'm wrong, you want to have 2 styles: "TextJustification" and "NumericColumnJustification". "TextJustification" is used for styling header texts as well as text columns. "NumericColumnJustification" is used for styling numeric columns.

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
Vahdanian: The removed space was due to adding spaces to start of columns which was the behaviour for all the columns. You would see more spaces if texts of two rows had different length. Now we add spaces to justify based on the given style. For a text column with default styling value (left justification) we only add spaces to the end of column

Vahdanian:
@timcassell 
Do you have a clue why some of the tests failed? It says some of the memory diagnoser tests are failing but I checked them on my machine and they were fine.

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
timcassell: I would like the headers to have uniform justification. The way I see it, there are 3 options for this.

1. Always left-justify headers (or right-justify), regardless of settings.
2. Use the same settings as text cells (think of the settings for each type of cell, rather than column).
3. Add a separate setting for header justification.

@AndreyAkinshin Do you have an opinion on this?

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
timcassell: I understand, but I would like to avoid other changes to the table besides justification.

timcassell:
> @timcassell Do you have a clue why some of the tests failed? It says some of the memory diagnoser tests are failing but I checked them on my machine and they were fine.

It's unrelated.

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
Vahdanian: > I understand, but I would like to avoid other changes to the table besides justification.

Ok, This is fixed in my latest commit.

tests/BenchmarkDotNet.Tests/Attributes/VerifiedFiles/ParamsAllValuesVerifyTests.BenchmarkShouldProduceSummary_WithAllValuesOfBool.verified.txt
Vahdanian: > I would like the headers to have uniform justification. The way I see it, there are 3 options for this.
> 
> 1. Always left-justify headers (or right-justify), regardless of settings.
> 2. Use the same settings as text cells (think of the settings for each type of cell, rather than column).
> 3. Add a separate setting for header justification.
> 
> @AndreyAkinshin Do you have an opinion on this?

I implemented the first option. Headers are now always justified to left.

tests/BenchmarkDotNet.Tests/Exporters/VerifiedFiles/CommonExporterVerifyTests.Exporters_ru-RU.verified.txt
```diff
@@ -429,9 +429,9 @@ Job=LongRun  IterationCount=100  LaunchCount=3
 WarmupCount=15  
 
 {noformat}
-||Method ||    Mean ||Error ||  StdDev ||     P67 ||
-|    Foo | 1,000 ns |    NA | 0,000 ns | 1,000 ns |
-|    Bar | 1,000 ns |    NA | 0,000 ns | 1,000 ns |
+| Method ||Mean     ||Error ||StdDev   ||P67      ||
```
timcassell, line 432: Looks like a `|` got removed on accident. I think that's necessary for this format.

Vahdanian: > Looks like a `|` got removed on accident. I think that's necessary for this format.

Yep, you are right! You can check it again now, it is fixed.

timcassell (approved):
Thanks @Vahdanian!

Vahdanian:
> Thanks @Vahdanian!

You're very welcome! I'm glad I could contribute to the project 😄

PR #2412, opened 2023-08-20
Add HardwareIntrinsics AVX-512 info

nietras:
WIP untested just to ask whether this is looks like path. Feel free to edit code directly if need be or do PR to PR.


timcassell:
We will need to update the sdk to net8 before the CI can run the tests, which I think will have to wait until net8 is officially released.
Also, we need someone with AVX512 hardware to confirm this works.

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
```diff
@@ -52,6 +54,12 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
                 {
                     case Platform.X86:
                     case Platform.X64:
+                        if (IsX86Avx512FSupported) yield return "AVX-512F";
+                        if (IsX86Avx512BWSupported) yield return "AVX-512BW";
+                        if (IsX86Avx512CDSupported) yield return "AVX-512CD";
+                        if (IsX86Avx512DQSupported) yield return "AVX-512DQ";
+                        if (IsX86Avx512VbmiSupported) yield return "AVX-512VBMI";
+
```
tannergooding, line 62: The other main path is an `if/else if` pattern.

This new code will yield/return several AVX-512F ISAs and not in their reverse hierarchical order.

Is that intentional?

--------

I'd imagine we want to cover `Avx512F` in the main if/else chain and the others to be on top to be "inline" with how the other works.

Notably certain combinations also have well-defined names.

That is, `x86-64-v4` is the formal definition that includes `AVX512F, AVX512BW, AVX512CD, AVX512DQ, AVX512VL`

`x86-64-v3` is `AVX, AVX2, BMI1, BMI2, F16C, FMA, LZCNT, MOVBE, OSXSAVE`

`x86-64-v2` is `CMPXCHG16B, POPCNT, SSE3, SSE4.1, SSE4.2, SSSE3`

`x86-64-v1` (baseline) is `CMOV, CX8, x87FPU, FXSR, MMX, OSFXSR, SCE, SSE, SSE2`

tannergooding:
> Also, we need someone with AVX512 hardware to confirm this works.

Can confirm that it works.

> We will need to update the sdk to net8 before the CI can run the tests, which I think will have to wait until net8 is officially released.

Notably, I think it would be beneficial if this support was added earlier. The runtime repo (and various other repos) are all using BDN to validate performance of preview versions of .NET.

BDN itself adding in some preview support for the latest version of .NET, including ISAs like this, makes the entire .NET preview experience that much better. It's also notably not a lot of code and can itself be considered "preview" until the point that version of .NET does ship as stable.

timcassell:
BDN already supports .Net 8 preview. This would be for our internal CI. I'll defer to @adamsitnik in regards to that.

src/BenchmarkDotNet/Portability/Cpu/HardwareIntrinsics.cs
nietras: I probably argue against using `x86-x64-v?` since it not immediately obvious what's what.

Instead perhaps we could list it as `AVX-512(X/Y/Z)` e.g. `AVX-512(F/BW/CD/DQ/VL)`. This makes set obvious and immediately visible.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <AssemblyTitle>BenchmarkDotNet</AssemblyTitle>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0;net8.0</TargetFrameworks>
```
adamsitnik, line 5: We just merged the update to .NET 8 SDK in #2462, could you please sync your branch?

nietras: Merged but this doesn't change this file?

adamsitnik: > Merged but this doesn't change this file?

Correct.

tannergooding (approved):

nietras:
👍

adamsitnik (approved):
@nietras thank you for your contribution!

PR #2413, opened 2023-08-21
Fix linux crash from disassembler

### LINKED ISSUE #2404: Using `DisassemblyDiagnoser` in GitHub Actions
I attempted to add `DisassemblyDiagnoser` to a benchmark class and run it in GH Actions default runners and I haven't been successful with Linux and macOS.

## Linux

```
Failed to set up high priority. Make sure you have the right permissions. Message: Permission denied`
```

After running it with sudo:
```
Process completed with exit code 139
```

## macOS

(similar to https://github.com/dotnet/BenchmarkDotNet/issues/1499):

```
// * Diagnostic Output - DisassemblyDiagnoser *
No benchmarks were disassembled
```

Interestingly, when ran with `sudo` an `asm-md` file was created, but an empty one.

----

Link to the actual benchmark runs:
https://github.com/lynx-chess/Lynx/actions/runs/5882072009
https://github.com/lynx-chess/Lynx/actions/runs/5882730576 (with sudo)

Branch where I tried to add the `DisassemblyDiagnoser`: https://github.com/lynx-chess/Lynx/tree/dotnet/90667

timcassell:
https://github.com/dotnet/runtime/pull/90794

Fixes #2404

### DIFF
--- a/src/BenchmarkDotNet/Disassemblers/ClrMdV2Disassembler.cs
+++ b/src/BenchmarkDotNet/Disassemblers/ClrMdV2Disassembler.cs
@@ -15,9 +15,9 @@ namespace BenchmarkDotNet.Disassemblers
     internal abstract class ClrMdV2Disassembler
     {
         // Translating an address to a method can cause AV and a process crash (https://github.com/dotnet/BenchmarkDotNet/issues/2070).
-        // It was fixed in https://github.com/dotnet/runtime/pull/79846,
-        // and most likely will be backported to 7.0.2 very soon (https://github.com/dotnet/runtime/pull/79862).
-        protected static readonly bool IsVulnerableToAvInDac = !RuntimeInformation.IsWindows() && Environment.Version < new Version(7, 0, 2);
+        // It was partially fixed in https://github.com/dotnet/runtime/pull/79846,
+        // fully fixed in https://github.com/dotnet/runtime/pull/90797.
+        protected static readonly bool IsVulnerableToAvInDac = !RuntimeInformation.IsWindows() && Environment.Version < new Version(8, 0);
 
         internal DisassemblyResult AttachAndDisassemble(Settings settings)
         {
--- a/src/BenchmarkDotNet/Disassemblers/ClrMdV2Disassembler.cs
+++ b/src/BenchmarkDotNet/Disassemblers/ClrMdV2Disassembler.cs
@@ -8,12 +8,36 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using BenchmarkDotNet.Portability;
+using BenchmarkDotNet.Helpers;
 
 namespace BenchmarkDotNet.Disassemblers
 {
     // This Disassembler uses ClrMd v2x. Please keep it in sync with ClrMdV1Disassembler (if possible).
     internal abstract class ClrMdV2Disassembler
     {
+        private static readonly ulong MinValidAddress = GetMinValidAddress();
+
+        private static ulong GetMinValidAddress()
+        {
+            // https://github.com/dotnet/BenchmarkDotNet/pull/2413#issuecomment-1688100117
+            if (RuntimeInformation.IsWindows())
+                return 65536;
+            if (RuntimeInformation.IsMacOS())
+                return RuntimeInformation.GetCurrentPlatform() switch
+                {
+                    Environments.Platform.X86 or Environments.Platform.X64 => 4096,
+                    Environments.Platform.Arm64 => 0x100000000,
+                    _ => throw new NotSupportedException($"{RuntimeInformation.GetCurrentPlatform()} is not supported")
+                };
+            if (RuntimeInformation.IsLinux())
+            {
+                string? minAddrResult = ProcessHelper.RunAndReadOutput("/bin/bash", "sudo cat /proc/sys/vm/mmap_min_addr");
+                ulong.TryParse(minAddrResult, out ulong minAddress);
+                return minAddress;
+            }
+            throw new NotSupportedException($"{System.Runtime.InteropServices.RuntimeInformation.OSDescription} is not supported");
+        }
+
         internal DisassemblyResult AttachAndDisassemble(Settings settings)
         {
             using (var dataTarget = DataTarget.AttachToProcess(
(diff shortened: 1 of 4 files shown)


adamsitnik (commented):
@timcassell thank you for providing a quick fix!

After reading https://github.com/dotnet/runtime/pull/90794#issuecomment-1684035240 I wonder whether we should simply unconditionally not try to map `0` and `-1`. Thoughts?

timcassell:
> After reading https://github.com/dotnet/runtime/pull/90794#issuecomment-1684035240 I wonder whether we should simply unconditionally not try to map 0 and -1. Thoughts?

I was also thinking about that, but I'm not familiar enough with the disassembler code. If you want to open a PR for that, we can close this in favor of that.

timcassell:
@adamsitnik It looks like the IntelDisassembler already checks for it `if (address > ushort.MaxValue)` but the Arm64Disassembler does not.

timcassell:
I changed it to always check for `0` or `-1`, though I'm not sure if it needs to only check that in non-windows.

timcassell:
@janvorli Is `-1` address invalid in every environment? Also are addresses less than `ushort.MaxValue` invalid in every environment?

@leculver It might make sense to explicitly check for `-1` address in ClrMd also.

janvorli:
> Is `-1` address invalid in every environment? 

Yes, -1 is treated in a special way in DAC. See e.g.:
https://github.com/dotnet/runtime/blob/4a5695b6ea0e461e6527de057420872369345282/src/coreclr/debug/daccess/dacfn.cpp#L384-L388

>Also are addresses less than `ushort.MaxValue` invalid in every environment?

This is true for Windows (actually, less or equal to that value are invalid). For Unix, this can differ based on the memory page size. On Linux, it is 4096 in most cases and 65536 for the case when the Linux kernel is configured for 64kB page size (some arm64 systems use that). But it can be different. For example, in WSL2 on Windows, it is 65536. The `/proc/sys/vm/mmap_min_addr` can be used to get the actual value on Linux.
On macOS, it is also different. On x64 it is be 4096 and on arm64 it is 4GB (0x100000000).

timcassell:
@janvorli How can I use that to get the value on Linux? I tried `/proc/sys/vm/mmap_min_addr` on a GitHub runner to test, and it gave back `Permission denied`.

leculver:
> @janvorli How can I use that to get the value on Linux? I tried `/proc/sys/vm/mmap_min_addr` on a GitHub runner to test, and it gave back `Permission denied`.

You are trying to run it, if that's the command you gave.

`cat /proc/sys/vm/mmap_min_addr`
or
`sudo cat /proc/sys/vm/mmap_min_addr`
if it's protected

src/BenchmarkDotNet/Disassemblers/ClrMdV2Disassembler.cs
```diff
@@ -8,16 +8,35 @@
...
+            if (RuntimeInformation.IsMacOS())
+                return RuntimeInformation.GetCurrentPlatform() switch
+                {
+                    Environments.Platform.X86 or Environments.Platform.X64 => 4096,
+                    Environments.Platform.Arm64 => 0x100000000,
+                    _ => throw new NotSupportedException($"{RuntimeInformation.GetCurrentPlatform()} is not supported")
+                };
+            if (RuntimeInformation.IsLinux())
+            {
+                string? minAddrResult = ProcessHelper.RunAndReadOutput("/bin/bash", "sudo cat /proc/sys/vm/mmap_min_addr");
```
leculver, line 34: Why not File.ReadAllText("/proc/sys/vm/mmap_min_addr")?  (be sure to catch exceptions)

Either way, I don't think that'll work on all flavors of linux.  You might need a coarser check,

timcassell: Or maybe it's not protected after all. I tried without sudo and it worked.

janvorli: This check would not work for dump debugging in case you are debugging it on a different distro from the one the dump was taken on and they differed in the null zone size. I am not sure if it is worth trying to be precise here. Runtime itself uses the target page size on Unix and 64kB on Windows to determine whether an invalid access is a null reference or an access violation, so I guess the same thing would be sufficient here.

janvorli: Ah, right, I've forgotten that this is for benchmarkdotnet usage only. So you are right, this is for live processing only. 
You can get the page size using `Environment.SystemPageSize`

timcassell: @janvorli Thanks. Do you know if the value for Android is the same as Linux, and iOS the same as MacOS?

janvorli: I don't know, but I don't think we support disassembling on any of these. AFAIK the disassembler supports coreclr and coreclr doesn't support them.

timcassell:
@adamsitnik This should be good to go, if you could take a look.

adamsitnik (commented):
The changes look OK, but we have lost a possibility to map certain addresses.

Every time we introduce changes to the disassembler I run a sample benchmark with `--disasmFilter *` to get the disassembly of everything that we can on Win x64, Lin x64 and arm64.

```cmd
dotnet run -c Release -f net7.0 --filter BenchmarkDotNet.Samples.IntroDisassembly.SumLocal --disasmFilter '*'
```

I've run this command on master branch, created a copy of the exported asm file and then run it against your branch. When eyeballing the diff I've found:

image

The smallest repro is:

```cmd
dotnet run -c Release -f net7.0 --filter BenchmarkDotNet.Samples.IntroDisassembly.SumLocal --disasmFilter "BenchmarkDotNet.Characteristics.CharacteristicSetPresenter..ctor()"
````

@timcassell PTAL

timcassell:
@adamsitnik I spent a few hours trying to figure out why that was happening. I finally found that if I delete my `BenchmarkDotNet.Samples\bin` directory before running, it works as expected. If I run it twice in a row without deleting it, I get the missing ctor method. I see the same behavior on master, so it's unrelated to my changes.

adamsitnik (approved):
LGTM, thank you @timcassell 

> I spent a few hours trying to figure out why that was happening. I finally found that if I delete my BenchmarkDotNet.Samples\bin directory before running, it works as expected. If I run it twice in a row without deleting it, I get the missing ctor method. I see the same behavior on master, so it's unrelated to my changes.

@janvorli it would be great if we could debug it one day

janvorli:
I have experimented with this a bit - just changing the runtime version in the test command line to 8.0. I cannot reproduce the behavior that @timcassell has seen, it fails to discover the method name for the constructor call all the time, even if I deleted the bin/obj folders before. 
However, I've re-read some of my notes from the time I have implemented decoding the target from the stubs it calls and I've found that a case that doesn't work is when the target is called via a ready to run delayload helper. So, I've tried to disable ready to run by setting `DOTNET_ReadyToRun=0` and the `System.Object..ctor` got decoded correctly. And then I've set it to 1 - and it was not decoded. So, I guess that is the culprit.
My notes say that real target extraction for calls via the ready to run delayload helpers would not be reasonably feasible from the disassembler code.

PR #2418, opened 2023-08-27
Enable nullability for BenchmarkDotNet.Annotations

alinasmirnova:


AndreyAkinshin (changes requested):
Thanks for the PR! Changes in BenchmarkDotNet.Annotations look good. However, we should avoid custom changes in CodeAnnotations.cs since it prevents us from updating this file to the actual version. (Which is also a good thing to do. In Rider, the latest version can be obtained in Settings->Editor->Inspection Settings->Code Annotations->Copy C# implementation to clipboard).

alinasmirnova:
So, you suggest updating this file to the last version and disabling nullability for it, right? Sounds like a great idea, I'll change it.

AndreyAkinshin (approved):
Thanks, looks good to me.

Issue #2054, opened 2022-07-28
Custom logging/visualization during the benchmark run


caaavik-msft:
Currently, if we want to do some custom visualization or export of benchmark results, we have to wait until all the benchmarks have been completed to be able to interpret the results by using a custom Exporter. As an example, the only way to see the results of benchmarks as they run is by what is passed through to the logger by BDN ([link to relevant source](https://github.com/dotnet/BenchmarkDotNet/blob/8c963ba0c6f29d676a49b0188dc195bb428ba60b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L177-L179)). In addition to this, the benchmark runner in BDN does lots of logging at various stages and there is no way to customize these logs or to use them meaningfully outside the context of a console/terminal.

For large projects which have lots of benchmarks that take a long time to run, it would be good to be able to hook into the various stages of the benchmark running so that we don't have to wait until the whole run is completed to be able to get custom exported information.

For this issue, I wanted to suggest a new feature to hook into the benchmark running to be able to subscribe to "events". An example of an event might be "RunStarted" or "RunCompleted" and those events would have read-only data attached to them containing the additional data about the event. I think as well, that all the logging code that is done today by BDN could be moved out to use this same abstraction, that way we can replace or extend it with our own custom logging. Not only would this be useful for customising the log output, but it could also provide the ability to build tools on top of BDN that display live execution results graphically rather than textually.

Diagnosers do somewhat implement this functionality with the `Handle(HostSignal, DiagnoserActionParameters)` method, but they are only run within the context of a single benchmark execution. If we want to subscribe to events such as when a benchmark case has completed running and the `BenchmarkReport` is available such as the example linked above, then it would be difficult to see how `IDiagnoser` could be extended to accommodate such an event.

I am willing to implement this but wanted to get some confirmation that this is a desirable feature for the project. If given the go-ahead, I can follow up with a proposal for how it could be implemented. Alternatively, if this feature is a bit too ambitious with wanting to have events for everything that needs to be logged, I think I still would like to look into something that lets me handle `BenchmarkReport` objects as they are created.

adamsitnik:
> I am willing to implement this but wanted to get some confirmation that this is a desirable feature for the project. If given the go-ahead, I can follow up with a proposal for how it could be implemented.

@caaavik-msft I confirm that this is a desirable feature. Please go ahead and send the proposal.

caaavik-msft:
[...]

PR #2420, opened 2023-08-28
Add event processor functionality

caaavik-msft:
This addresses #2054 which is a feature request to have hooks which can enable other tools that are invoking BDN to display progress rather than having to parse live console output or to wait for all the benchmarks to complete to have the exporters run. 

A full proposal and explanation of the events/hooks in this PR are covered in this comment: https://github.com/dotnet/BenchmarkDotNet/issues/2054#issuecomment-1202229339

There are potentially many more hooks that could be added like events for when a benchmark run enters a different stage, but they would likely need to be added as custom events sent through the Broker and supported by all the different executors/toolchains, which I didn't want to add at this stage. I'm open as well to modifying or adding extra events before merging this in to see what people think would be useful.

I've also included a `EventHandlerBase` class which is a base class that implements all the hooks as virtual methods, so that implementing classes only need to override the hooks they care about.


cincuranet:
Any reason why `CompositeEventHandler` does not derive from `EventHandlerBase`?

caaavik-msft:
> Any reason why `CompositeEventHandler` does not derive from `EventHandlerBase`?

`EventHandlerBase` is just an implementation of the interface with virtual methods so that custom event handlers don't have to implement every hook and can just implement the hooks they care about. The `CompositeEventHandler` composes multiple event handlers into one, so by nature we want to ensure that every hook is implemented and handled, if we used the EventHandlerBase we may forget to update `CompositeEventHandler` if we add a new hook.

src/BenchmarkDotNet/EventHandlers/EventHandlerBase.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.Results;
+using BenchmarkDotNet.Validators;
+
+namespace BenchmarkDotNet.EventHandlers
+{
+    public class EventHandlerBase : IEventHandler
```
adamsitnik, line 10: I know that all existing BDN config switches expose interfaces, but in this case I wonder whether we should just expose an `abstract` class with all the methods being `virtual`. Benefits:
- to sign up for a single event, the users would need to override only one method
- if we wanted to introduce a new event in the future, we would not break the existing users, as their code would still compile.

What do you think?

src/BenchmarkDotNet/EventHandlers/IEventHandler.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.Results;
+using BenchmarkDotNet.Validators;
+
+namespace BenchmarkDotNet.EventHandlers
+{
+    public interface IEventHandler
```
adamsitnik, line 10: `EventHandler` has a very specific meaning in C#: https://learn.microsoft.com/en-us/dotnet/api/system.eventhandler?view=net-7.0

Adding an `I` in front of it may suggest that it's an interface that `EventHandler` implements.

I believe that we should find a better name. I don't have any good suggestions, but perhaps @bartonjs could throw us a hand here?

src/BenchmarkDotNet/EventHandlers/IEventHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        void HandleStartValidationStage();
+        void HandleUnsupportedBenchmark(BenchmarkCase benchmarkCase);
```
adamsitnik, line 13: To me the **Handle**UnsupportedBenchmark name suggests that you can actually do something with the unsuported benchmark (like stop benchmarking or continue). Since it's not the case here, I think we should use the `On$Abc` rather than `Handle$Abc` naming convention.

src/BenchmarkDotNet/EventHandlers/IEventHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+        void HandleValidationError(ValidationError validationError);
+        void HandleStartBuildStage();
+        void HandleBuildFailed(BenchmarkCase benchmarkCase, BuildResult buildResult);
```
adamsitnik, line 16: we have events for build start and failure, what about successful completion?

src/BenchmarkDotNet/EventHandlers/IEventHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+        void HandleStartRunStage();
+        void HandleRunBenchmarksInType(Type type, IReadOnlyList<BenchmarkCase> benchmarks);
+        void HandleCompletedBenchmarksInType(Type type, Summary summary);
```
adamsitnik, line 19: the naming is inconsistent: Start&Stop vs Run&Completed. We should choose one schema and stick to it.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -54,13 +55,24 @@ internal static Summary[] Run(BenchmarkRunInfo[] benchmarkRunInfos)
...
                 var (supportedBenchmarks, validationErrors) = GetSupportedBenchmarks(benchmarkRunInfos, resolver);
 
                 validationErrors.AddRange(Validate(supportedBenchmarks));
 
+                foreach (var validationError in validationErrors)
+                {
+                    var validationErrorEventHandler = validationError.BenchmarkCase != null
+                        ? validationError.BenchmarkCase.Config.GetCompositeEventHandler()
+                        : eventHandler;
+                    validationErrorEventHandler.HandleValidationError(validationError);
```
adamsitnik, line 73: I am not sure whether we should be allowing to have the event handlers only for selected benchmarks.

```suggestion
                    eventHandler.HandleValidationError(validationError);
```

@caaavik-msft what do you think? do you see any real life usage examples of that?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -179,6 +209,8 @@ internal static Summary[] Run(BenchmarkRunInfo[] benchmarkRunInfos)
                 logger.WriteLineInfo($"//   {benchmark.DisplayInfo}");
             logger.WriteLine();
 
+            var eventHandler = config.GetCompositeEventHandler();
```
adamsitnik, line 212: could you please pass an instance of composite handler from the calling method here (we already do that for logger)?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -334,15 +369,26 @@ private static ImmutableArray<ValidationError> Validate(params BenchmarkRunInfo[
...
-                        .FailOnError
-                        .Validate(new ValidationParameters(joinedCases, null))
-                    );
+                var compatibilityValidationErrors = ConfigCompatibilityValidator
+                    .FailOnError
+                    .Validate(new ValidationParameters(joinedCases, null));
+
+                foreach (var error in compatibilityValidationErrors)
+                {
+                    validationErrors.Add(error);
```
adamsitnik, line 378: you have achieved exactly the same effect, but with more code. Did you intend to raise some events here?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -334,15 +369,26 @@ private static ImmutableArray<ValidationError> Validate(params BenchmarkRunInfo[
...
+                }
             }
 
             foreach (var benchmarkRunInfo in benchmarks)
-                validationErrors.AddRange(benchmarkRunInfo.Config.GetCompositeValidator().Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config)));
+            {
+                var config = benchmarkRunInfo.Config;
+                var validator = config.GetCompositeValidator();
+                var eventHandler = config.GetCompositeEventHandler();
```
adamsitnik, line 386: this variable is not used, did you mean to use it?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -550,15 +596,21 @@ private static (BenchmarkRunInfo[], List<ValidationError>) GetSupportedBenchmark
...
+            var runInfos = benchmarkRunInfos.Select(info =>
+                {
+                    var eventHandler = info.Config.GetCompositeEventHandler();
+                    return new BenchmarkRunInfo(
+                        info.BenchmarksCases.Where(benchmark =>
+                        {
+                            var errors = benchmark.GetToolchain().Validate(benchmark, resolver).ToArray();
+                            validationErrors.AddRange(errors);
+                            if (errors.Any())
+                                eventHandler.HandleUnsupportedBenchmark(benchmark);
```
adamsitnik, line 608: `OnValidaitonError` seems to be a better name (not every validation error is a hard stop)

tests/BenchmarkDotNet.IntegrationTests/EventHandlerTests.cs
```diff
@@ -0,0 +1,298 @@
...
+            Assert.Equal(typeof(ClassA).GetMethod(nameof(ClassA.Method1)), (events[2].Args[0] as BenchmarkCase).Descriptor.WorkloadMethod);
+            Assert.Equal(nameof(IEventHandler.HandleBuildFailed), events[3].EventType);
+            Assert.Equal(typeof(ClassA).GetMethod(nameof(ClassA.Method2)), (events[3].Args[0] as BenchmarkCase).Descriptor.WorkloadMethod);
+            Assert.Equal(nameof(IEventHandler.HandleStartRunStage), events[4].EventType);
+        }
+
+        private List<LoggingEventHandler.EventData> RunBenchmarksAndRecordEvents(Type[] types, IValidator? validator = null, IToolchain? toolchain = null)
+        {
+            var eventHandler = new LoggingEventHandler();
+            var job = new Job(Job.Dry);
```
adamsitnik, line 160: thank you for using `Dry` job here 👍

tests/BenchmarkDotNet.IntegrationTests/EventHandlerTests.cs
```diff
@@ -0,0 +1,298 @@
...
+                Events.Add(new EventData(nameof(HandleUnsupportedBenchmark), new object[] { benchmarkCase }));
+            }
+
+            public void HandleValidationError(ValidationError validationError)
+            {
+                Events.Add(new EventData(nameof(HandleValidationError), new object[] { validationError }));
+            }
+        }
+    }
+}
```
adamsitnik, line 298: big thanks for adding plenty of tests! 🥇

adamsitnik (commented):
Overall it looks good, but let's iterate on simplifying it a bit and unifying the naming.

Thank you for your contribution @caaavik-msft !

src/BenchmarkDotNet/EventHandlers/EventHandlerBase.cs
bartonjs: I was about to say something very similar.  The fact that this is "one type to handle all kinds of events" means an interface is not suitable.

src/BenchmarkDotNet/EventHandlers/IEventHandler.cs
bartonjs: Event...Processor?  Event...Broker?

timcassell: EventListener?

caaavik-msft: what about `BenchmarkEventHandler`? I'm going to use this for now but if we want to go with something else then am happy to change it again

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
caaavik-msft: The only case I can think of would be for the BenchmarkDotNet.Samples project where we might want to demonstrate a custom event handler but not apply it to every benchmark that is being run. But I suppose even for that case it is probably acceptable since usually only one type is run at a time when using the samples.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
caaavik-msft: I have removed `OnUnsupportedBenchmark` since they surface in `OnValidationError` anyways

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -69,16 +76,34 @@ internal static Summary[] Run(BenchmarkRunInfo[] benchmarkRunInfos)
...
                 var buildPartitions = BenchmarkPartitioner.CreateForBuild(supportedBenchmarks, resolver);
                 var buildResults = BuildInParallel(compositeLogger, rootArtifactsFolderPath, buildPartitions, in globalChronometer);
+
+                foreach (var buildPartition in buildPartitions)
+                {
+                    var buildResult = buildResults[buildPartition];
+                    if (!buildResult.IsBuildSuccess)
+                    {
+                        foreach (var benchmark in buildPartition.Benchmarks)
+                            benchmarkEventHandler.OnBuildFailed(benchmark.BenchmarkCase, buildResult);
```
timcassell, line 98: Why not just do `OnBuildComplete` so the listener can check for build success?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -69,16 +76,34 @@ internal static Summary[] Run(BenchmarkRunInfo[] benchmarkRunInfos)
...
 
+                benchmarkEventHandler.OnEndValidationStage();
+
                 int totalBenchmarkCount = supportedBenchmarks.Sum(benchmarkInfo => benchmarkInfo.BenchmarksCases.Length);
                 int benchmarksToRunCount = totalBenchmarkCount - (idToResume + 1); // ids are indexed from 0
                 compositeLogger.WriteLineHeader("// ***** BenchmarkRunner: Start   *****");
                 compositeLogger.WriteLineHeader($"// ***** Found {totalBenchmarkCount} benchmark(s) in total *****");
                 var globalChronometer = Chronometer.Start();
 
+                benchmarkEventHandler.OnStartBuildStage();
```
timcassell, line 87: It might be handy to pass in how many partitions are being built.

src/BenchmarkDotNet/EventHandlers/CompositeBenchmarkEventHandler.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.Results;
+using BenchmarkDotNet.Validators;
+
+namespace BenchmarkDotNet.EventHandlers
+{
+    public class CompositeBenchmarkEventHandler : BenchmarkEventHandlerBase
```
adamsitnik, line 10: this type does not need to be `public` 

```suggestion
    internal sealed class CompositeBenchmarkEventHandler : BenchmarkEventHandlerBase
```

src/BenchmarkDotNet/EventHandlers/BenchmarkEventHandlerBase.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.Results;
+using BenchmarkDotNet.Validators;
+
+namespace BenchmarkDotNet.EventHandlers
+{
+    public abstract class BenchmarkEventHandlerBase
```
adamsitnik, line 10: this is very subjective, but `EventProcessor` suggested by @bartonjs sounds best to me.

```suggestion
    public abstract class EventProcessor
```

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -640,6 +673,16 @@ void AddLogger(ILogger logger)
             return new CompositeLogger(loggers.Values.ToImmutableHashSet());
         }
 
+        private static BenchmarkEventHandlerBase CreateCompositeEventHandler(BenchmarkRunInfo[] benchmarkRunInfos)
+        {
+            var eventHandlers = new HashSet<BenchmarkEventHandlerBase>();
```
adamsitnik, line 678: this logic can be moved the the ctor of the composite event handler/processor

adamsitnik (commented):
We are almost there, thank you for addressing the feedback @caaavik-msft !

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -390,8 +380,19 @@ private static ImmutableArray<ValidationError> Validate(params BenchmarkRunInfo[
...
+                .ToDictionary(
+                    buildPartition => buildPartition,
+                    buildPartition =>
+                    {
+                        var buildResult = Build(buildPartition, rootArtifactsFolderPath, buildLogger);
+
+                        // If the generation was successful, but the build was not, we will try building sequentially
+                        // so don't send the OnBuildComplete event yet.
+                        if (buildPartitions.Length <= 1 || !buildResult.IsGenerateSuccess || buildResult.IsBuildSuccess)
+                            eventProcessor.OnBuildComplete(buildPartition, buildResult);
```
timcassell, line 392: It looks like this is being called concurrently, which necessitates synchronization. Perhaps it would be better to take the parallel query and force it back to sequential to invoke on the same thread?

```cs
buildResults = buildPartitions
    .AsParallel()
    .Select(buildPartition => (buildPartition, buildResult: Build(buildPartition, rootArtifactsFolderPath, buildLogger)))
    .AsSequential()
    .ToDictionary(...)
```

caaavik-msft: Nice suggestion, I did not know that `AsSequential` existed. Adding that in now.

timcassell (approved):

caaavik-msft:
@adamsitnik are we good to merge this in now?

src/BenchmarkDotNet/EventProcessors/CompositeEventProcessor.cs
```diff
@@ -0,0 +1,102 @@
...
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.Results;
+using BenchmarkDotNet.Validators;
+
+namespace BenchmarkDotNet.EventProcessors
+{
+    internal sealed class CompositeEventProcessor : EventProcessorBase
+    {
+        private readonly IReadOnlyCollection<EventProcessorBase> eventProcessors;
```
adamsitnik, line 13: We should prefer concrete types over abstractions for private fields (https://github.com/dotnet/roslyn-analyzers/pull/6370)

```suggestion
        private readonly HashSet<EventProcessorBase> eventProcessors;
```

src/BenchmarkDotNet/EventProcessors/CompositeEventProcessor.cs
```diff
@@ -0,0 +1,102 @@
...
+            foreach (var info in benchmarkRunInfos)
+                eventProcessors.AddRange(info.Config.GetEventProcessors());
+
+            this.eventProcessors = eventProcessors;
+        }
+
+        public CompositeEventProcessor(IReadOnlyCollection<EventProcessorBase> eventProcessors)
+        {
+            this.eventProcessors = eventProcessors;
+        }
```
adamsitnik, line 28: it seems that this ctor is not used anywhere

```suggestion
```

src/BenchmarkDotNet/EventProcessors/EventProcessorBase.cs
```diff
@@ -5,15 +5,15 @@
...
+    public abstract class EventProcessorBase
     {
         public virtual void OnStartValidationStage() { }
         public virtual void OnValidationError(ValidationError validationError) { }
         public virtual void OnEndValidationStage() { }
-        public virtual void OnStartBuildStage() { }
-        public virtual void OnBuildFailed(BenchmarkCase benchmarkCase, BuildResult buildResult) { }
+        public virtual void OnStartBuildStage(IReadOnlyList<BuildPartition> partitions) { }
+        public virtual void OnBuildComplete(BuildPartition benchmarkCase, BuildResult buildResult) { }
         public virtual void OnEndBuildStage() { }
```
adamsitnik, line 17: what is the difference between `OnBuildComplete` and `OnEndBuildStage`? The first is for one partition and the latter is for all of them?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -390,8 +380,19 @@ private static ImmutableArray<ValidationError> Validate(params BenchmarkRunInfo[
...
                 .AsParallel()
-                .Select(buildPartition => (buildPartition, buildResult: Build(buildPartition, rootArtifactsFolderPath, buildLogger)))
-                .ToDictionary(result => result.buildPartition, result => result.buildResult);
+                .ToDictionary(
+                    buildPartition => buildPartition,
+                    buildPartition =>
+                    {
+                        var buildResult = Build(buildPartition, rootArtifactsFolderPath, buildLogger);
+
+                        // If the generation was successful, but the build was not, we will try building sequentially
```
adamsitnik: thank you for adding the comment 👍

src/BenchmarkDotNet/EventProcessors/EventProcessorBase.cs
```diff
@@ -5,15 +5,15 @@
 using BenchmarkDotNet.Toolchains.Results;
 using BenchmarkDotNet.Validators;
 
-namespace BenchmarkDotNet.EventHandlers
+namespace BenchmarkDotNet.EventProcessors
 {
-    public abstract class BenchmarkEventHandlerBase
+    public abstract class EventProcessorBase
```
adamsitnik, line 10: in the past we used the `Base` suffix for abstract types, but from what I've learned from API Reviews it's an anti-pattern

```suggestion
    public abstract class EventProcessor
```

adamsitnik (commented):
LGTM, but please rename the type before I hit the merge button.

Thank you @caaavik-msft !

src/BenchmarkDotNet/EventProcessors/EventProcessorBase.cs
caaavik-msft: Yes, that is the difference between them.

src/BenchmarkDotNet/Configs/ConfigExtensions.cs
```diff
@@ -132,5 +134,4 @@ private static ManualConfig With(this IConfig config, Action<ManualConfig> addAc
             addAction(manualConfig);
             return manualConfig;
         }
-    }
 }
```
adamsitnik, line 137: this should fix the build

```suggestion
    }
}
```

src/BenchmarkDotNet/Configs/ConfigExtensions.cs
```diff
@@ -110,6 +111,7 @@ public static class ConfigExtensions
         [Obsolete("This method will soon be removed, please start using .AddLogicalGroupRules() instead.")]
         [EditorBrowsable(EditorBrowsableState.Never)] public static IConfig With(this IConfig config, params BenchmarkLogicalGroupRule[] rules) => config.AddLogicalGroupRules(rules);
         [PublicAPI] public static ManualConfig AddLogicalGroupRules(this IConfig config, params BenchmarkLogicalGroupRule[] rules) => config.With(c => c.AddLogicalGroupRules(rules));
+        [PublicAPI] public static ManualConfig AddEventHandler(this IConfig config, params IEventHandler[] eventHandlers) => config.With(c => AddEventHandler(c, eventHandlers));
```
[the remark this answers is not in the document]
mawosoft: @adamsitnik , @caaavik-msft

> nit: it's unlikely that the users will need more than one instance

While this may be true, this makes it the only one of the `Add...` methods that doesn't accept a params array. Personally, I would prefer the API consistency.

src/BenchmarkDotNet/Configs/ManualConfig.cs
```diff
@@ -221,6 +224,12 @@ public ManualConfig AddLogicalGroupRules(params BenchmarkLogicalGroupRule[] rule
             return this;
         }
 
+        public ManualConfig AddEventProcessor(EventProcessor eventProcessor)
```
mawosoft, line 227: @adamsitnik, @caaavik-msft

> nit: it's unlikely that the users will need more than one instance

While this may be true, this makes it the only one of the `Add...` methods that doesn't accept a params array. Personally, I would prefer the API consistency.

mawosoft:
Also, `ManualConfig.Add(IConfig config)` and, by extension, `ManualConfig.Union(IConfig globalConfig, IConfig localConfig)` and `ManualConfig.Create(IConfig config)` don't copy the event processor(s) from the source.

timcassell:
@caaavik-msft Can you address @mawosoft's comment in another PR (merging configs)?

PR #2421, opened 2023-09-01
More nullability warnings fixes

alinasmirnova:


AndreyAkinshin:
@alinasmirnova thanks!

PR #2423, opened 2023-09-02
Fix builds when --keepFiles is specified

### LINKED ISSUE #1067: Conditions in Directory.Build.props fail
While trying out the most recent changes which allow benchmarking with dotnet core 3.0 desktop project, which btw, work GREAT, I noticed that when running BenchmarkDotNet I get errors for Tasks which should not run!

Those tasks have conditions which are **normally** not met but for the benchmark project they did.
I **deleted** the tasks, the benchmark ran without issues.

The cause was this ItemGroup:
```
  <ItemGroup Condition="!$(MSBuildProjectName.Contains('Test')) And !$(MSBuildProjectName.Contains('Demo'))">
    <PackageReference Include="PdbGit" Version="3.0.41">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
    </PackageReference>
  </ItemGroup>
```
While thinking about this, I noticed that the conditions are on the MSBuildProjectName. I believe that the temporary generated projects get a GUID as name, which causes my condition to return "true" (does not contain Test).

The workaround for me, is to delete the ItemGroup manually, which is okay for now.

I wanted to discuss a possible solution, while I believe there is no perfect solution I was hoping for a compromise.

I would suggest that we change the project names to have the original name in it, extended with the GUID. The main question is how, and this should be documented.

Suggestions:

- MSBuildProjectName-GUID
- GUID-MSBuildProjectName

Maybe you have some other ideas?

I will be more than happy to supply an example repository if this helps, but I could imagine that the case is quite clear? I can also take a look at writing a PR, I just wanted to get into a small discussion before doing so, not that things move into the wrong direction.

### LINKED ISSUE #2411: Jobs can mess up builds with --keepFiles
Jobs with provided Ids pass along those Ids when modifications are made.
CsProj is always generated with the same filename `BenchmarkDotNet.Autogenerated.csproj`.
When `--keepFiles` is specified, the generated directory uses the job Id instead of a guid.

Combine those factors, and when you try something like this, the builds will be incorrect.

```cs
DefaultConfig.Instance
    .AddJob(Job.ShortRun.WithRuntime(RuntimeMoniker.Net70.GetRuntime())) // Id is ShortRun
    .AddJob(Job.ShortRun.WithRuntime(RuntimeMoniker.NativeAot70.GetRuntime())) // Id is ShortRun
    .KeepBenchmarkFiles(true);
```

Orthogonally, I think the API is bad. `Id` is generally expected to be a unique identifier (or at least semi-unique), which is clearly not the case here. Ideally, the `Id` would be generated always internally, and we would only expose setting a `Name` rather than `Id`. But of course that would be a hefty breaking change.

timcassell:
Fixes #2411
Fixes #1067

### DIFF
--- a/src/BenchmarkDotNet/Running/BuildPartition.cs
+++ b/src/BenchmarkDotNet/Running/BuildPartition.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Reflection;
+using System.Threading;
 using BenchmarkDotNet.Characteristics;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Environments;
@@ -18,15 +19,20 @@ namespace BenchmarkDotNet.Running
 {
     public class BuildPartition
     {
+        // We use an auto-increment global counter instead of Guid to guarantee uniqueness per benchmark run (Guid has a small chance to collide),
+        // assuming there are fewer than 4 billion build partitions (a safe assumption).
+        private static int s_partitionCounter;
+
         public BuildPartition(BenchmarkBuildInfo[] benchmarks, IResolver resolver)
         {
             Resolver = resolver;
             RepresentativeBenchmarkCase = benchmarks[0].BenchmarkCase;
             Benchmarks = benchmarks;
-            var keepBenchmarkFiles = benchmarks[0].Config.Options.IsSet(ConfigOptions.KeepBenchmarkFiles);
-            var guid = Guid.NewGuid().ToString();
-            ProgramName = keepBenchmarkFiles ? RepresentativeBenchmarkCase.Job.FolderInfo : guid;
-            ProgramDirectory = keepBenchmarkFiles ? Path.Combine(RepresentativeBenchmarkCase.Job.FolderInfo, guid) : guid;
+            // Combine the benchmark's assembly name, folder info, and build partition id.
+            string benchmarkAssemblyName = RepresentativeBenchmarkCase.Descriptor.Type.Assembly.GetName().Name;
+            string folderInfo = RepresentativeBenchmarkCase.Job.FolderInfo;
+            int id = Interlocked.Increment(ref s_partitionCounter);
+            ProgramName = $"{benchmarkAssemblyName}-{folderInfo}-{id}";
             LogBuildOutput = benchmarks[0].Config.Options.IsSet(ConfigOptions.LogBuildOutput);
             GenerateMSBuildBinLog = benchmarks[0].Config.Options.IsSet(ConfigOptions.GenerateMSBuildBinLog);
         }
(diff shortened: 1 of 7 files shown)


src/BenchmarkDotNet/Running/BuildPartition.cs
```diff
@@ -18,12 +19,20 @@ namespace BenchmarkDotNet.Running
 {
     public class BuildPartition
     {
+        // We use an auto-increment global counter instead of Guid to guarantee uniqueness per benchmark run (Guid has a small chance to collide),
```
adamsitnik, line 22: Is the reason behind the two bug reports a Guid collision?

timcassell: No, the Guid collision is a side issue. The issues are the program name itself (and thus the output folder based on that program name).

tests/BenchmarkDotNet.IntegrationTests/InProcessEmitTest.cs
```diff
@@ -68,10 +69,11 @@ private void DiffEmit(Summary summary)
...
                 return;
 
-            var caseName = summary.BenchmarksCases.First().Job.ToString();
+            var benchmarkCase = summary.BenchmarksCases.First();
+            var caseName = $"{benchmarkCase.Descriptor.Type.Assembly.GetName().Name}-{benchmarkCase.Job.FolderInfo}";
             NaiveRunnableEmitDiff.RunDiff(
-                $@"{caseName}.exe",
-                $@"{caseName}Emitted.dll",
+                $@"{caseName}-{BuildPartition.s_partitionCounter}.exe",
+                $@"{caseName}-{BuildPartition.s_partitionCounter - 1}Emitted.dll",
```
adamsitnik, line 76: Could you please add a comment explaining why `-1` is needed here?

src/BenchmarkDotNet/Running/BuildPartition.cs
```diff
@@ -18,12 +19,20 @@ namespace BenchmarkDotNet.Running
...
+        internal static int s_partitionCounter;
+
         public BuildPartition(BenchmarkBuildInfo[] benchmarks, IResolver resolver)
         {
             Resolver = resolver;
             RepresentativeBenchmarkCase = benchmarks[0].BenchmarkCase;
             Benchmarks = benchmarks;
-            ProgramName = benchmarks[0].Config.Options.IsSet(ConfigOptions.KeepBenchmarkFiles) ? RepresentativeBenchmarkCase.Job.FolderInfo : Guid.NewGuid().ToString();
+            // Combine the benchmark's assembly name, folder info, and build partition id.
+            string benchmarkAssemblyName = RepresentativeBenchmarkCase.Descriptor.Type.Assembly.GetName().Name;
```
adamsitnik, line 32: IIRC Assembly Name should not contain any characters that are invalid for folder names. 👍

adamsitnik (approved):
LGTM, thank you @timcassell !

PR #2433, opened 2023-09-21
Fix build errors with latest sdk

timcassell:


adamsitnik (commented):
Which package that we use has known critical vulnerability? Do we have to use it?

timcassell:
It's dotnet core 2.0, the oldest core runtime we support.

adamsitnik (approved):
LGTM, thank you @timcassell !

PR #2434, opened 2023-09-22
Fix Event Processors not being copied in ManualConfig.Add

caaavik-msft:
Also add params support for `AddEventProcessor` to be consistent with other addable things.

This addresses additional feedback from @mawosoft on the previous PR: https://github.com/dotnet/BenchmarkDotNet/pull/2420#issuecomment-1729756976


adamsitnik (approved):
LGTM, thank you @caaavik-msft !

PR #2435, opened 2023-09-23
Treat warnings not as errors in manual test project

timcassell:
Followup to #2433


adamsitnik:
Perhaps we should just ignore all warnings for this project?

timcassell:
> Perhaps we should just ignore all warnings for this project?

Good idea. More are likely to appear as it ages.

adamsitnik (approved):
LGTM!

PR #2438, opened 2023-09-26
Add a VSTest Adapter

caaavik-msft:
This is a feature that builds upon my recent work on adding support for event processors. This is a fully featured and working VSTest Adapter for running BenchmarkDotNet benchmarks. Below you can see two examples of what the experience looks like running the BDN sample benchmarks in Visual Studio's Test Explorer and VS Code's C# Dev Kit Testing UI. Apologies if the images are difficult to read, I would recommend opening them in another tab to view them.

### Visual Studio's Test Explorer  
image

### Visual Studio Code's Testing UI with C# Dev Kit
image

### Why VSTest?
While most people associate VSTest with being a framework for running unit tests, integration tests, and other tests which have a boolean outcome of "Passed" or "Failed", there is nothing inherent to VSTest that restricts it to these cases and it turns out it is flexible enough to support performance tests as well. While VSTest does not have any functionality built in today for storing numerical results and doing anything meaningful with them, its benefits are that it is a standardised test discovery and execution protocol to build on top of with good IDE support in the .NET ecosystem, and it may be preferable over using the CLI to invoke benchmarks.

### Seeing the benchmark results
One feature of VSTest is that you can add arbitrary messages onto your test results, so what I did was just put the benchmark results in those messages, and so you can see in the Visual Studio screenshot above that the results are easy to view from the Test Explorer window. Unfortunately, C# Dev Kit's testing functionality does not display these messages as of the time of writing. However as an alternative, the full BDN console output is forwarded to VSTest as well and you can view it in the Tests Output Window on Visual Studio, or the Test Explorer Output Window for VS Code as seen on the bottom-right of the VS Code screenshot above.

### Quirks around .NET Framework support
I'd also like to point out that for .NET Framework it was quite difficult to get this working with proper assembly loading. The approach I ended up with was invoking BenchmarkDotNet in a new AppDomain that is configured to have the same base directory as the BenchmarkDotNet dll/exe. Since there is communication across AppDomains, there is quite a bit of code in this PR spent towards "remoting" and passing test cases and test results across this boundary. Note this is only talking about the BDN program itself being compiled for .NET Framework, this doesn't apply for example if you are compiling the BDN program with .NET 6 and  having benchmarks that target .NET Framework.

### How to add VSTest support to an existing project.
Only three lines need to be added to the project file of the BDN project, and you can see these changes done in the `BenchmarkDotNet.Samples` and `BenchmarkDotNet.Samples.FSharp` projects in this PR. The first change is setting the property `GenerateProgramFile` to `false` which will instruct the test SDK to not generate a program file. The other two changes would be to add references to the `BenchmarkDotNet.TestAdapter` and `Microsoft.NET.Test.Sdk` libraries. Should this PR be accepted, I imagine there would be some work needed to get this published as a new NuGet package, so I would need some assistance with that so that people can start using this.

### Future work
There are three things that I did not complete in this PR which I would like to do in a future PR after discussing it some more in a separate issue:
- Adding CancellationToken support to BDN to cancel the current benchmark run at any time
  - My hack to get this working in this PR was to throw an exception inside the event processor on the `OnStartRunBenchmark` event, but I think in the long run having proper CancellationToken support would be preferable
- Adding a way to specify a default global config so that the VSTest adapter can use it
  - Right now, it just loads everything with the default config with no way to modify this.
- An entrypoint-less BDN project structure, much like how unit test projects work today where the entry point is generated at build time. This would be combined with the above feature of specifying a default global config. The default entry point that is generated could just be a call to `BenchmarkSwitcher`. 

### More reading about VSTest
If you are unfamiliar with VSTest, and wanted to understand it more so that you can review this PR better, the docs can be found in the [`docs` folder in the `microsoft/vstest` repo](https://github.com/microsoft/vstest/tree/main/docs). In particular I would recommend reading the [Test Platform Architecture](https://github.com/microsoft/vstest/blob/main/docs/RFCs/0001-Test-Platform-Architecture.md) and the [Adapter Extensibility](https://github.com/microsoft/vstest/blob/main/docs/RFCs/0004-Adapter-Extensibility.md) pages. I also learnt a lot by reading the [MSTest GitHub repo](https://github.com/microsoft/testfx) as well.

### Other comments
- As for the package name, I chose `BenchmarkDotNet.TestAdapter` because VSTest will by default load and packages that end in `.TestAdapter`, this same approach is used almost universally for other testing libraries too. If we want to change it, as long as it ends in `.TestAdapter` it should be ok.
- The TestAdapter project is not too big, so I didn't add too many abstractions, but as a result the code is a bit messy in some places. I think there are definitely places that could be structured better so if you have ideas I am happy to implement them to make this easier to maintain.
- This PR was my first time interacting with `AppDomains` and creating them, as well as how to pass things across `AppDomain` boundaries. I think it's likely that some of my code around that space is not following best practices or I have not implemented it in a way that handles all the edge cases. I can confirm though that the sample benchmarks in this repo all run perfectly fine on .NET framework for me, but that was the extent of my testing.


timcassell:
Very cool! I'm curious why in your screenshots, it seems the Visual Studio details are very minimal (not even including the summary table) compared to VS Code.

As for creating a new nuget package, that'd fall on @AndreyAkinshin.

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// <summary>
+        /// This will create the given type in a child AppDomain when used in .NET Framework.
+        /// If not in the .NET Framework, it will use the current
+        /// </summary>
+        /// <param name="type">The type to create.</param>
+        /// <param name="source">The dll or exe of the benchmark project.</param>
+        /// <returns>The created object.</returns>
+        private static object CreateIsolatedType(Type type, string source)
+        {
+#if NETFRAMEWORK
```
timcassell, line 183: I think it'd make more sense to do a runtime lookup than multi-targeting. `if (RuntimeInformation.IsFullFramework)`

AndreyAkinshin:
@caaavik-msft thanks for the contribution! The idea of using VSTest Adapter to run benchmarks looks interesting. However, I would need some time to properly review it and validate that the adapter doesn't produce a significant impact on the obtained measurements.

src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// Converts a BDN BenchmarkCase to a VSTest TestCase.
+        /// </summary>
+        /// <param name="benchmarkCase">The BenchmarkCase to convert.</param>
+        /// <param name="source">The dll or exe of the benchmark project.</param>
+        /// <param name="includeJobInName">Whether or not the display name should include the job name.</param>
+        /// <returns>The VSTest TestCase.</returns>
+        internal static TestCase ToVSTestCase(this BenchmarkCase benchmarkCase, string source, bool includeJobInName=false)
+        {
+            var benchmarkMethod = benchmarkCase.Descriptor.WorkloadMethod;
+            var fullClassName = benchmarkCase.Descriptor.Type.FullName;
```
adamsitnik, line 29: I am not sure whether this particular line to responsible for that, but for the generic types we display the number of generic type arguments in the benchmark name:

image

You can use following method to get a nice output for all kinds of types (it caused us a lot of trouble in the past): https://github.com/dotnet/BenchmarkDotNet/blob/d49417314f6f361779e007042182f0a7ce13b45e/src/BenchmarkDotNet/Extensions/ReflectionExtensions.cs#L39

src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+                DisplayName = FullNameProvider.GetBenchmarkName(benchmarkCase),
+                Id = GetTestCaseId(benchmarkCase),
+            };
+
+            if (includeJobInName)
+            {
+                vsTestCase.DisplayName += $" [{benchmarkCase.GetUnrandomizedJobDisplayInfo()}]";
+            }
+
+            var benchmarkAttribute = benchmarkMethod.ResolveAttribute<BenchmarkAttribute>();
```
adamsitnik, line 53: will it work for attributes that derive from `[Benchmark]` attribute?

src/BenchmarkDotNet.TestAdapter/BenchmarkExecutor.cs
```diff
@@ -0,0 +1,86 @@
...
+                    b.Config.AddEventProcessor(eventProcessor).AddLogger(logger).CreateImmutableConfig()))
+                .ToArray();
+
+            // Run all the benchmarks, and ensure that any tests that don't have a result yet are sent.
+            BenchmarkRunner.Run(benchmarks);
+            eventProcessor.SendUnsentTestResults();
+        }
+
+        /// <summary>
+        /// Stop the benchmarks when next able.
```
adamsitnik, line 79: It's possible to stop the run? WOW that is impressive! 🥇

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
```diff
@@ -0,0 +1,208 @@
...
+            var appBase = Path.GetDirectoryName(source);
+            var setup = new AppDomainSetup { ApplicationBase = appBase };
+            var domainName = $"Isolated Domain for {type.Name}";
+            var appDomain = AppDomain.CreateDomain(domainName, null, setup);
```
adamsitnik, line 187: Is it really necessary? What do we benefit out of creating a dedicated app domain?

adamsitnik (commented):
Big thanks for your contribution @caaavik-msft ! 

It works very smoothly! I added some comments, most of them are just questions.

Here are some extra questions unrelated to the code:

1. Currently all tests and benchmarks are hard to distinguish. Is there any way we could categorize them? For example by using `Traits` and always assigning all benchmarks to "perf" trait so users could group them: test vs benchmarks ([sample](http://dontcodetired.com/blog/post/Grouping-and-Filtering-Tests-in-Visual-Studio-Test-Explorer))

image

2. The "Test Detail Summary" does not contain the table that we all love. Why is that?

image

Is it because the tooling does not support displaying tables? Does it support HTML (we have an exporter for that)?

If it's impossible to support it, could we swap the order and print histogram first? So the results would be "somehow" visualized for those who have little space on the screen?

image

3. Would it be possible to show the benchmarks only for Release configuration? So the users don't try to run them in Debug just to get an error?

Thank you!

timcassell:
> 1. Currently all tests and benchmarks are hard to distinguish. Is there any way we could categorize them? For example by using `Traits` and always assigning all benchmarks to "perf" trait so users could group them: test vs benchmarks ([sample](http://dontcodetired.com/blog/post/Grouping-and-Filtering-Tests-in-Visual-Studio-Test-Explorer))

👍 

> 2. The "Test Detail Summary" does not contain the table that we all love. Why is that?
> 
> Is it because the tooling does not support displaying tables? Does it support HTML (we have an exporter for that)?

OP shows an image with the table. It looks to me like selecting an individual benchmark result displays only the histogram for that benchmark, but you can select the group which displays the summary table of all the benchmarks in that group. It makes sense to me.

> 3. Would it be possible to show the benchmarks only for Release configuration? So the users don't try to run them in Debug just to get an error?

I think debugging could be useful, maybe there's a way to force a Debug config, and display a warning about it? Or maybe we can add a `[DebugBuild]` attribute that does the same thing as `DebugBuildConfig` (and similar for `DebugInProcess`), and only show those benchmarks with that attribute.

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
caaavik-msft: I'm unable to do this as `AppDomainSetup` and related methods are not supported in `netstandard2.0`. They are supported in `net6.0` though, but I'm not sure how to get around this without the multi-targeting.

samples/BenchmarkDotNet.Samples/BenchmarkDotNet.Samples.csproj
```diff
@@ -11,6 +11,7 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
     <DebugSymbols>true</DebugSymbols>
     <NoWarn>$(NoWarn);CA1018;CA5351;CA1825</NoWarn>
+    <GenerateProgramFile>false</GenerateProgramFile>
```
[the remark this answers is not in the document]
caaavik-msft: Yes, it is mandatory. Without this, when the Test SDK is referenced, it will automatically copy into the project a default entry point and the compilation will fail.

Having said this, my other PR solves this exact issue and provides a way for this property to be set just by referencing the NuGet Package, so when this PR is merged in we no longer need this property to be set in the csproj: https://github.com/dotnet/BenchmarkDotNet/pull/2445

samples/BenchmarkDotNet.Samples/BenchmarkDotNet.Samples.csproj
```diff
@@ -19,10 +20,12 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="System.Drawing.Common" Version="4.7.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
```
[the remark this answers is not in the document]
caaavik-msft: Unfortunately this is required as this library provides and runs the vstest executables which load the test adapter for everything to work. It doesn't work as a transitive dependency. I spent a long time searching and wasn't able to find any way to avoid this. Perhaps if we want to avoid this we could try speaking to the VSTest team to see if they have any suggestions.

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
timcassell: Is the `AppDomainSetup` required? Can you just use the `CreateDomain(string)` version that's part of netstandard2.0?

src/BenchmarkDotNet/Properties/AssemblyInfo.cs
```diff
@@ -16,11 +16,13 @@
 [assembly: InternalsVisibleTo("BenchmarkDotNet.Diagnostics.dotTrace,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
 [assembly: InternalsVisibleTo("BenchmarkDotNet.IntegrationTests.ManualRunning,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
 [assembly: InternalsVisibleTo("BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
+[assembly: InternalsVisibleTo("BenchmarkDotNet.TestAdapter,PublicKey=" + BenchmarkDotNetInfo.PublicKey)]
```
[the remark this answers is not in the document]
caaavik-msft: We would need these:
- `ReflectionExtensions` for `GetCorrectCSharpTypeName`, `ResolveAttribute`, and `GetRunnableBenchmarks`
- `GenericBenchmarksBuilder` for `GetRunnableBenchmarks`
- `AssemblyExtensions` for `IsDebug`

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
caaavik-msft: Unfortunately it's needed for .NET Framework so I can set the `ApplicationBase` for the new domain so that it will load all the assemblies from the BDN directory rather than from the VSTest directory. There is no supported cross-platform way to achieve this without cross-targeting.

See: https://github.com/dotnet/runtime/issues/28026#issuecomment-444562466

An alternative though would be that we don't support the VSTest Adapter for .NET Framework projects

src/BenchmarkDotNet.TestAdapter/Remoting/SerializationHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace BenchmarkDotNet.TestAdapter.Remoting
+{
+    /// <summary>
+    /// A set of helper methods for serializing and deserializing the VSTest TestCases and TestReports.
+    /// </summary>
+    internal static class SerializationHelpers
+    {
+        public static string Serialize<T>(T data)
+        {
+            return JsonDataSerializer.Instance.Serialize(data, version: 7);
```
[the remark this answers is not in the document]
caaavik-msft: I have added a comment explaining the version number and what might require it to be updated

src/BenchmarkDotNet.TestAdapter/BenchmarkExecutor.cs
caaavik-msft: Yep! But it does seem a bit hacky abusing the EventHandler to throw an exception to cause this rather than proper cancellation support. Perhaps this would be good to add proper support for in the future.

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
caaavik-msft: Without this, it is impossible to get .NET Framework to resolve the correct assemblies as otherwise it will try load them from the directory that the vstest console executable lives in rather than the BenchmarkDotNet folder. .NET Core is smarter around assembly resolution and does not need this. The same approach of using an AppDomain to load the assemblies for .NET Framework is also used by MSTest and XUnit libraries, so I fear that it may be the only option.

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
timcassell: Another alternative would be to use reflection to use that API. But I'm not strong-minded about it. It's probably fine to multi-target if all the testing libraries do the same.

caaavik-msft:
I've added some code to support VSTest traits. All benchmarks will have a BenchmarkDotNet trait, and all BenchmarkCategory attributes are surfaced as traits too:
image

Tim is correct in that the reason that we don't get the standard BDN output table is because each test case in the VSTest Explorer window is for each individual set of arguments. The Test Explorer does not expose any way for us to provide information for a group of tests so that we can display a comparison for them easily. To see all the tests compared, it is still possible to do so by viewing the "Tests" output window.

Also I have made the change for now to make it so that when Debug mode is selected, that the Test Explorer does not show any benchmarks. I think maybe there might be a nicer way to allow benchmarks in debug mode. Instead of preventing debug assemblies from being searched, another option would be to allow them if the `InProcess` toolchain is used which the user might be able to set with a assembly-level `IConfigSource` which checks if the current assembly is debug and loads the `InProcess` toolchain when that occurs.

src/BenchmarkDotNet/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
adamsitnik: That is more methods than I expected. Let's keep `InternalsVisibleTo` for now

adamsitnik (commented):
Thank you for addressing my feedback. The last thing I need is a doc with a few screen shots and most important info (build in Release, how to see the table etc). You could put it here: https://github.com/dotnet/BenchmarkDotNet/tree/master/docs/articles/features

Thank you for your amazing contribution @caaavik-msft !

src/BenchmarkDotNet.TestAdapter/VSTestEventProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Threading;
+
+namespace BenchmarkDotNet.TestAdapter
+{
+    /// <summary>
+    /// An event processor which will pass on benchmark execution information to VSTest.
+    /// </summary>
+    internal class VSTestEventProcessor : EventProcessor
```
AndreyAkinshin, line 21: I would suggest using `Vs` instead of `VS` in all symbol names. Motivation:
* Miscorosft tend to use `Vs` in their own projects, e.g., see https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
* Names `VSTestEventProcessor` cause warnings for some linters. `Vs` is a more safe way to avoid warnings with different tools.

src/BenchmarkDotNet.TestAdapter/Remoting/MessageLoggerWrapper.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+using System;
+
+namespace BenchmarkDotNet.TestAdapter
```
AndreyAkinshin, line 4: It seems that `.Remoting` is missing the namespace declaration.

src/BenchmarkDotNet.TestAdapter/Remoting/SerializationHelpers.cs
```diff
@@ -7,14 +7,18 @@ namespace BenchmarkDotNet.TestAdapter.Remoting
     /// </summary>
     internal static class SerializationHelpers
     {
+        // Version number of the VSTest protocol that the adapter supports. Only needs to be updated when
```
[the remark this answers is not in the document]
AndreyAkinshin: Could we improve this comment further and provide a link to additional info, where we can learn about different versions of the vstest protocol?

src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+            var benchmarkFullName = $"{fullClassName}.{benchmarkMethodName}";
+
+            // Display name has arguments as well.
+            var displayMethodName = FullNameProvider.GetMethodName(benchmarkCase);
+            if (includeJobInName)
+                displayMethodName += $" [{benchmarkCase.GetUnrandomizedJobDisplayInfo()}]";
+
+            var displayName = $"{fullClassName}.{displayMethodName}";
+
+            var vsTestCase = new TestCase(benchmarkFullName, VSTestAdapter.ExecutorUri, assemblyPath)
```
AndreyAkinshin, line 38: Currently, several test cases may have equal `fullyQualifiedName`. It will be better to have a unique FQN for each test case.

src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+    internal static class BenchmarkCaseExtensions
+    {
+        /// <summary>
+        /// Converts a BDN BenchmarkCase to a VSTest TestCase.
+        /// </summary>
+        /// <param name="benchmarkCase">The BenchmarkCase to convert.</param>
+        /// <param name="assemblyPath">The dll or exe of the benchmark project.</param>
+        /// <param name="includeJobInName">Whether or not the display name should include the job name.</param>
+        /// <returns>The VSTest TestCase.</returns>
+        internal static TestCase ToVSTestCase(this BenchmarkCase benchmarkCase, string assemblyPath, bool includeJobInName=false)
```
AndreyAkinshin, line 24: `includeJobInName=false`: missing spaces around `=`

src/BenchmarkDotNet.TestAdapter/VSTestAdapter.cs
```diff
@@ -0,0 +1,206 @@
...
+        private static void ValidateSourceIsAssemblyOrThrow(string source)
+        {
+            if (string.IsNullOrEmpty(source))
+                throw new ArgumentException($"'{nameof(source)}' cannot be null or whitespace.", nameof(source));
+
+            if (!Path.HasExtension(source))
+                throw new NotSupportedException($"Missing extension on source '{source}', must have the extension '.dll' or '.exe'.");
+
+            var extension = Path.GetExtension(source);
+            if (!string.Equals(extension, ".dll") && !string.Equals(extension, ".exe"))
```
AndreyAkinshin, line 202: It will be safer to ignore casing in this comparison (extensions `.DLL` and `.EXE` are not common, but they do exist).

AndreyAkinshin (changes requested):
Thanks again for a great pull request! (And sorry for such a long review.)

In general, looks good to me. I left some minor comments. Once they are resolved, I'm ready to merge the PR.

src/BenchmarkDotNet.TestAdapter/BenchmarkDotNet.TestAdapter.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\build\common.props" />
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
```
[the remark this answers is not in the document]
caaavik-msft: Missed this review comment from earlier, multi-targeting is needed due to the AppDomain dependency that is not supported with netstandard2.0

src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs
caaavik-msft: `fullyQualfiedName` is a legacy property now and is not intended to be used as a unique identifier for a test, rather the `Id` property is the only thing that needs to uniquely identify a case.

Despite it being legacy, it still exists and is required for the purposes of extracting the project/namespace/class/testgroup hierarchy view that is shown in the VS Test Explorer: https://github.com/microsoft/vstest/blob/main/docs/RFCs/0033-Hierarchy-TestCase-Property.md#overview

caaavik-msft:
Have made two more changes to the functionality:
- The histogram now displays above the other statistics:
image
- When the assembly is compiled with `Debug`, it will allow you to run benchmarks that are using an `InProcess` toolchain. I've tested out as well that if I put a breakpoint inside a BDN method and run the test using Debug in Test Explorer, that it hits the breakpoint, so should make it easier to debug your benchmarks this way too.

caaavik-msft:
@adamsitnik I was writing up the feature doc and I realised that I think there are two things that need to be done before people can really start using this feature and for it to make sense to be documented:
1. The BenchmarkDotNet.TestAdapter project needs to be published to NuGet
2. This PR needs to be merged (or discussed if we even want it): https://github.com/dotnet/BenchmarkDotNet/pull/2445

At the very least, the PR from the 2nd bullet point makes it so that `GenerateProgramFile` gets set to false by any projects that references it as a PackageReference, so we don't need people to include this line in their project file to support it. It would also be neat to include that `GenerateBDNEntryPoint` from this PR is available, and you could use it in collaboration with an assembly-level `IConfigSource` to allow specifying a default configuration that can be used when running through VSTest. 

Another thing we could do if we want to is extend the PR above further by making it so that referencing `BenchmarkDotNet` includes the `TestAdapter` project as a dependency so that everyone gets the VSTest adapter without having to import any extra dependencies. I'm not sure exactly how that would look, but I think it can be achieved with a custom nuspec?

adamsitnik (commented):
> The BenchmarkDotNet.TestAdapter project needs to be published to NuGet

We will publish the docs to benchmarkdotnet.org after we ship new packages to nuget.org. So when writing the docs you can assume that this package exists.

> This PR needs to be merged (or discussed if we even want it)

I've provided my feedback, sorry for the delay.

> referencing `BenchmarkDotNet` includes the `TestAdapter` project as a dependency

Currently I am not convinced that it should be the default. For years we have been telling the users to close all the apps (including VS) when running the benchmarks. Just to minimize the chance of other apps consuming resources during benchmark run and affecting the stability of the benchmarks.
For now we can ship it as a standalone package with good docs and wait for user feedback.


FWIW For me the most important aspect of the doc is what needs to be done to get it working and how to find the results.

caaavik-msft:
I have included in this PR the functionality from https://github.com/dotnet/BenchmarkDotNet/pull/2445. From the discussion there, I realised that this was better to include as part of the Test Adapter itself.

Just to reply to some comments left on that PR here since they are applicable here too:

> So far, we’ve supported only Console Apps. Will this change make it possible to support libraries?
> If so, what command will be needed to run the benchmarks? dotnet test?
> If so, what will be the behavior for libraries that target .NET standard rather than a specific runtime (example: netstandard2.0)?

Unfortunately, for the VSTest host to be launched, the `Microsoft.NET.Test.Sdk` package needs to be referenced, and that package being referenced will force any project that references it to be a Console Application even if it does not have the `OutputType` set to `exe`. This is why it is necessary to always have an entry point, and so this functionality means that a BDN project can look like a library in the same way that unit tests projects look like libraries since they don't have a defined `Main`. When using the auto-generated entry point, an executable will still be produced that just runs the benchmarks using the `BenchmarkSwitcher`

> Edge case: will it work with the "old" project files?

By old do you mean before the SDK-style project structure? All this functionality does is inject a custom property through a nuget package reference, so I think it should but I'm not sure how to create an older project file to test it on.

> BTW I am not convinced that this is the right direction. Currently all BDN users are used to the fact that only Console Apps are supported (providing an entry point is a must-have). With this change, they won’t need to provide the entry point, but they will still have to do something to get it working (setting GenerateBDNEntryPoint to true). From my perspective it’s less intuitive (something new) and it does not solve the problem of passing the config.

I think now that I have moved this under the TestAdapter project it should alleviate most of the concerns? This way it is not seen as the recommended way to go, but just an option if you are intending to use the VSTest adapter and want an experience similar to that of other unit testing libraries.

I'll get the docs finished ASAP as well now.

src/BenchmarkDotNet.TestAdapter/Package/BenchmarkDotNet.TestAdapter.props
```diff
@@ -0,0 +1,12 @@
+﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Disables generation of MSTest entrypoint -->
+    <GenerateProgramFile>false</GenerateProgramFile>
+    <EntryPointSourceDirectory>$(MSBuildThisFileDirectory)..\entrypoints\</EntryPointSourceDirectory>
+  </PropertyGroup>
+  <ItemGroup Condition="'$(GenerateBDNEntryPoint)' == 'true'">
```
timcassell, line 7: Do we need the condition still? I would expect the entry point to always be generated if I'm using the test adapter.

samples/BenchmarkDotNet.Samples/BenchmarkDotNet.Samples.csproj
timcassell, line 14: This is no longer needed since you added it in the test adapter props.

timcassell:
> * When the assembly is compiled with `Debug`, it will allow you to run benchmarks that are using an `InProcess` toolchain. I've tested out as well that if I put a breakpoint inside a BDN method and run the test using Debug in Test Explorer, that it hits the breakpoint, so should make it easier to debug your benchmarks this way too.

(For future) it would be nice to also be able to debug build benchmarks (the default configuration). I wonder if there's a way to automatically attach a debugger to the benchmark process when it's launched if the host process has a debugger attached.

src/BenchmarkDotNet.TestAdapter/Package/BenchmarkDotNet.TestAdapter.props
caaavik-msft: Yes, this is still needed for projects that already have an entry point or want to use their own because of the customisations that it has. If we don't have this condition, then the compilation will fail because there are two entry points detected.

samples/BenchmarkDotNet.Samples/BenchmarkDotNet.Samples.csproj
caaavik-msft: The ability to copy across the `.props` file so that this property is not needed only happens via `PackageReference`, not `ProjectReference`. I have added a comment now to explain this.

src/BenchmarkDotNet.TestAdapter/Package/BenchmarkDotNet.TestAdapter.props
timcassell: >  for projects that already have an entry point

I don't think that's a concern. Using the test adapter is changing how the benchmarks are ran and so users should remove their own entry point. The same way you don't expect to write an entry point for unit test libraries.

> or want to use their own because of the customisations that it has

Ok, that's a valid point. But I still think it's not a concern, as advanced usages like that can force the entry point used, even if 2 exist. See https://github.com/dotnet/BenchmarkDotNet/blob/b3b2d913a9de86147167f3fe65d770969c9541a5/src/BenchmarkDotNet/Templates/CsProj.txt#L29

caaavik-msft: Fair points, what if I just invert the default condition instead so it is auto-generated by default, with a way to disable it? Then it would behave the same as MSTest which provides a way to conditionally disable the entry point generation with `GenerateProgramFile`. Also I did not realise that was what `StartupObject` was for, that's neat.

caaavik-msft:
> > * When the assembly is compiled with `Debug`, it will allow you to run benchmarks that are using an `InProcess` toolchain. I've tested out as well that if I put a breakpoint inside a BDN method and run the test using Debug in Test Explorer, that it hits the breakpoint, so should make it easier to debug your benchmarks this way too.
> 
> (For future) it would be nice to also be able to debug build benchmarks (the default configuration). I wonder if there's a way to automatically attach a debugger to the benchmark process when it's launched if the host process has a debugger attached.

I know this doesn't solve the general problem, but I could maybe modify the auto-generated entry point so that if it's compiled with `Debug` that it will launch it with an `InProcess` toolchain by default. That way, most projects that are using the VSTest adapter will get this functionality for free.

src/BenchmarkDotNet.TestAdapter/Package/BenchmarkDotNet.TestAdapter.props
timcassell: I think that's a fine idea to swap the default. I wonder if it can be disabled from the csproj like that, though? If the props file imports the entry point file before the csproj disables the property, it will essentially be ignored. How does MSTest do it?

timcassell:
> I know this doesn't solve the general problem, but I could maybe modify the auto-generated entry point so that if it's compiled with `Debug` that it will launch it with an `InProcess` toolchain by default. That way, most projects that are using the VSTest adapter will get this functionality for free.

Let's not do that. It will break any benchmarks that expect to be built. Let's just leave this as a future improvement.

caaavik-msft:
One thing I forgot to mention as well, is that it seems that there are two ways to define what being in "debug" mode means. I originally did the check using `AssemblyExtensions.IsDebug` which apparently looks at `IsJITTrackingEnabled` on the `Debuggable` attribute, but I discovered that for F# projects this was returning true always even when compiled in Release mode so they stopped working with the adapter. I changed that line of code now to check `IsJITOptimizerDisabled` instead to test whether debug mode is on.

src/BenchmarkDotNet.TestAdapter/Package/BenchmarkDotNet.TestAdapter.props
caaavik-msft: From my testing, it seems that the csproj runs before the props file is imported, which makes sense because the csproj is the one that loads the package reference. So after inverting this condition, it all works as described.

timcassell: Can we hijack `GenerateProgramFile`? Something like this?

```xml
<PropertyGroup>
    <EntryPointSourceDirectory>$(MSBuildThisFileDirectory)..\entrypoints\</EntryPointSourceDirectory>
</PropertyGroup>
<ItemGroup Condition="'$(GenerateProgramFile)' != 'false'">
    <Compile Include="$(EntryPointSourceDirectory)EntryPoint.cs" Condition="'$(Language)' == 'C#'" Visible="false"/>
    <CompileAfter Include="$(EntryPointSourceDirectory)EntryPoint.fs" Condition="'$(Language)' == 'F#'" Visible="false"/>
    <Compile Include="$(EntryPointSourceDirectory)EntryPoint.vb" Condition="'$(Language)' == 'VB'" Visible="false"/>
</ItemGroup>
<PropertyGroup>
    <!-- Disables generation of MSTest entrypoint -->
    <GenerateProgramFile>false</GenerateProgramFile>
</PropertyGroup>
```

caaavik-msft: I managed to get it working, turns out that PropertyGroup has higher precedence than ItemGroup when outside of a Target, but when inside a Target they will be executed in order. I've moved this into a target now and it works perfectly.

caaavik-msft:
I've added the docs page now: https://github.com/dotnet/BenchmarkDotNet/blob/c432f887d274e9aa7f902ed75353a0d99615c804/docs/articles/features/vstest.md

I'm not sure how it looks when on the website though, but the markdown rendering looks good.

timcassell:
@caaavik-msft Maybe Visual Studio works differently than Rider, but I can't see the table in the group summary.

image

caaavik-msft:
I've tried a bunch of things but have not been able to find a way to do have anything display in the output section for a test group. It doesn't seem like there is any way to display information there,

timcassell (approved):

docs/articles/features/vstest.md
```diff
@@ -0,0 +1,75 @@
...
+After doing this, you can set your build configuration to `Release`, run a build, and you should be able to see the benchmarks in your IDE's VSTest integration.
+
+## Setting a default configuration
+
+Previously, it was common for the default configuration to be defined inside the entry point. Since the entry point is not used when running benchmarks through VSTest, the default configuration must be specified using a custom `IConfigSource` attribute instead that is set on the assembly. 
+
+First, create a custom attribute that implements `IConfigSource` like below, making sure that it has `Assembly` as one of the attribute targets:
+
+```csharp
+class MyDefaultConfigSourceAttribute : Attribute, IConfigSource
```
adamsitnik, line 52: We already have a type for that:

https://github.com/dotnet/BenchmarkDotNet/blob/b3b2d913a9de86147167f3fe65d770969c9541a5/src/BenchmarkDotNet/Attributes/ConfigAttribute.cs#L6-L11

Both solutions require to create new dedicated types. But IMO creating a type that just derives from `ManualConfig` is easier. Example:

https://github.com/dotnet/BenchmarkDotNet/blob/b3b2d913a9de86147167f3fe65d770969c9541a5/tests/BenchmarkDotNet.IntegrationTests.EnabledOptimizations/OptimizationsEnabledInCsproj.cs#L8-L18

adamsitnik (commented):
Last comment from my side, I promise!

Thank you for providing great docs @caaavik-msft !

docs/articles/features/vstest.md
caaavik-msft: Thanks! just tested it out and that all works fine with the VSTest adapter

adamsitnik (approved):
LGTM, thank you for your contribution @caaavik-msft !

CameronAavik:
Thanks @AndreyAkinshin for getting published and fixing the issues with the packing!

AndreyAkinshin:
@CameronAavik thanks again for the awesome PR!

PR #2444, opened 2023-10-13
Changed console color for Help from DarkGreen to Green

danielmertens:
I had a look at the other colors but with only 16 colors available on the `ConsoleColor` enum, there isn't much I can do.

For now I changed the color for Help from DarkGreen to Green, which makes it a lot more readable.


AndreyAkinshin:
While `ConsoleColor.Green` may look better than `ConsoleColor.DarkGreen` in some dark-themed terminals, it may also look less readable in some light-themed terminals. That's how the suggested change looks in the default 'Light Background' color preset of [iTerm2](https://iterm2.com/):
image

PR #2458, opened 2023-11-01
Adds Metrics Columns to Benchmark Report Output

### LINKED ISSUE #2233: Q: Include hardware counters in XML output
Hi. I've lost a couple of hours now trying to coax BenchmarkDotNet to output hardware counters in a XML or JSON file. Default job configuration outputs them in the console report and CSV file, but not matter what I do (even following CSV export example), I run into the following troubles:

- The HW counters are missing in the XML or JSON file, but present in the CSV output
- Configuring CSV exporter to write machine-readable output (as below) does not work; i get units in values columns

HW counters DO get shown in the console summary.

Any suggestions?

Sample code:
```
// In main: var summaries = BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(config: new CustomConfig());

internal class CustomConfig : ManualConfig
{
    public CustomConfig() {
        AddJob(Job.Default);
        AddLogger(BenchmarkDotNet.Loggers.ConsoleLogger.Default);
        AddColumn(BenchmarkDotNet.Columns.)

        AddExporter(new CsvExporter(
            CsvSeparator.CurrentCulture,
            new SummaryStyle(
                cultureInfo: System.Globalization.CultureInfo.InvariantCulture,
                printUnitsInHeader: true,
                printUnitsInContent: false,
                timeUnit: Perfolizer.Horology.TimeUnit.Millisecond,
                sizeUnit: BenchmarkDotNet.Columns.SizeUnit.KB)));
        AddHardwareCounters(HardwareCounter.TotalIssues, HardwareCounter.CacheMisses, HardwareCounter.BranchInstructions, HardwareCounter.BranchMispredictions);
    }
}
```

nazulg:
Fix #2233

 Adds Metrics Columns to Benchmark Report for XmlExporter, and adds columns for output based on Hardware Counters when using ManualConfig. 

Included verified text files.


src/BenchmarkDotNet/Configs/ManualConfig.cs
```diff
@@ -196,6 +196,10 @@ public ManualConfig AddJob(params Job[] newJobs)
         public ManualConfig AddHardwareCounters(params HardwareCounter[] newHardwareCounters)
         {
             hardwareCounters.AddRange(newHardwareCounters);
+            columnProviders.AddRange(
+                newHardwareCounters
+                    .Select(h => new MetricColumn(new HardwareCounterDescriptor(h)))
+                    .Select(c => c.ToProvider()));
```
timcassell, line 202: This can be done with a single Select.

timcassell: Also, this looks kind of strange to me. The column providers by default are added from `DefaultColumnProviders`. I think the hardware counters just need to be added to the metrics when the report is generated so it will be picked up automatically by the default `MetricsColumnProvider`.

nazulg: Hi @timcassell,

I checked and the DefaultColumnProviders add the Metrics columns, but they are not being used on the ManualConfig.cs unless the user calls ManualConfig.CreateMinimal(). So the following code in ManuaklConfig.cs may work for when they create an empty ManualConfig and add Hardware Counters? 


```cs       
        public ManualConfig AddHardwareCounters(params HardwareCounter[] newHardwareCounters)
        {
            hardwareCounters.AddRange(newHardwareCounters);
            AddColumnProvider(DefaultColumnProviders.Metrics);
            return this;
        }

timcassell: So the exporter already works when using default config? Or am I misunderstanding?

nazulg: Yes, it works. Only the XmlExporter was missing the Metrics data, that one needs the change to the file SummaryDto.cs.

The Exporter in a ManualConfig also works, but by calling AddHardwareCounters in a ManualConfig does not add the Counters Columns by default (to be printed and exported), so the columns will need to be added manually. The code above tries to solve this by adding the metric columns by default when Hardware Counters are added to a ManualConfig.

nazulg:
@dotnet-policy-service agree

timcassell (approved):
LGTM. @nazulg If you think it's ready, you can mark this as ready to review, then I can merge.

nazulg:
Hi @timcassell, it is ready, I marked it ready for review, thank you!

timcassell:
Thanks @nazulg

PR #2461, opened 2023-11-06
Adds a custom exporter option on command line args

### LINKED ISSUE #1967: Allow custom exporter options to be defined
My usecase would be extending a benchmark project to include the option of an exporter that'd write the summary to a sql-database.

Thanks in advance

nazulg:
Fix #1967

This code adds the ability to include a custom exporter from command line arguments.

The user can use the option --customExporter "{assembly-qualified name of exporter}" and it will load the exporter into the config. 

Please let me know your thoughts on this approach.

Thank you.

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -49,12 +49,9 @@ namespace BenchmarkDotNet.ConsoleArguments
             { "verylong", Job.VeryLongRun }
         };
 
-        public static void RegisterCustomExporter(string commandLineName, IExporter exporter) => CustomExporters.Add(commandLineName, exporter);
-        private static readonly IDictionary<string, IExporter> CustomExporters = new Dictionary<string, IExporter>();
-
         [SuppressMessage("ReSharper", "StringLiteralTypo")]
         [SuppressMessage("ReSharper", "CoVariantArrayConversion")]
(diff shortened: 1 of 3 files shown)


timcassell:
It doesn't make sense to have the `RegisterCustomExporter` function if you're just going to create it from the type and ignore the registered name. I like being able to use it from the type, but it should also work from the registered name.

Also, the command line option should support multiple values (like `--runtimes`). And I would prefer `--exporters`

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -339,6 +356,7 @@ private static IConfig CreateConfig(CommandLineOptions options, IConfig globalCo
                 config.AddJob(baseJob);
 
             config.AddExporter(options.Exporters.SelectMany(exporter => AvailableExporters[exporter]).ToArray());
+            config.AddExporter(CustomExporters.Select(c => c.Value).ToArray());
```
timcassell, line 359: The registered exporters should only be added if the user specified it in the command line.

nazulg:
I updated the PR. In this change,  a custom exporter can be added to the existing exporters collection. 

If the exporter name is not found in the default list, then it tries to create an exporter from the argument as it will be considered to be a type name using the assembly name.

timcassell:
LGTM, unsure if we do want to support registering custom names (to be able to use simplified names rather than full assembly-qualified types). What do you think @adamsitnik?

timcassell:
#3212

PR #2462, opened 2023-11-08
update SDK to .NET 8

adamsitnik:
Currently it's just a draft. It would be nice to have for #2412, where we would not need to use reflection to figure out whether AVX512 is supported.

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -20,12 +20,10 @@ jobs:
       - uses: actions/checkout@v3
       - name: Run task 'build'
         shell: cmd
-        run: |
-          ./build.cmd build
+        run: ./build.cmd build
       - name: Run task 'in-tests-core'
         shell: cmd
-        run: |
-          ./build.cmd in-tests-core -e
+        run: ./build.cmd in-tests-core -e
       - name: Upload test results
         uses: actions/upload-artifact@v3
         if: always()
@@ -42,12 +40,10 @@ jobs:
       - uses: actions/checkout@v3
       - name: Run task 'build'
         shell: cmd
-        run: |
-          ./build.cmd build
+        run: ./build.cmd build
       - name: Run task 'in-tests-full'
         shell: cmd
-        run: |
-          ./build.cmd in-tests-full -e
+        run: ./build.cmd in-tests-full -e
       - name: Upload test results
         uses: actions/upload-artifact@v3
         if: always()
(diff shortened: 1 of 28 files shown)


AndreyAkinshin:
@adamsitnik I tried to fix issues in this branch, but we have a problem with the Windows GitHub actions workflows. For some reason, it uses the system .NET SDK (`C:\Program Files\dotnet\sdk\7.0.403\`) instead of ours. I have already explicitly passed our SDK to PATH and BenchmarkDotNet.Build, but it doesn't work. At the moment, I'm out of ideas on how to fix it. =(

adamsitnik:
@AndreyAkinshin thanks for your help!

I would expect https://github.com/dotnet/BenchmarkDotNet/commit/8a02ec28d55529f9be0ea66d843049738b2be8fa to fix the problem, but it did not?!?

AndreyAkinshin:
@adamsitnik I have found a bug with setting the PATH variable in `build.ps1`. Now the bug is fixed and `build.cmd` uses our .NET SDK for all kinds of operations on all OSes. Let's wait for the test run results.

.github/workflows/run-tests.yaml
```diff
@@ -18,6 +18,9 @@ jobs:
         run: Set-MpPreference -DisableRealtimeMonitoring $true
         shell: powershell
       - uses: actions/checkout@v3
+      - name: Run task 'sdk-check'
```
adamsitnik, line 21: @AndreyAkinshin  I assume we want to remove the SDK checks because they were temporary?

tests/BenchmarkDotNet.IntegrationTests/BuildTimeoutTests.cs
```diff
@@ -31,8 +31,8 @@ public void WhenBuildTakesMoreTimeThanTheTimeoutTheBuildIsCancelled()
                 .AddJob(Job.Dry
                     .WithRuntime(NativeAotRuntime.Net70)
```
adamsitnik, line 32: I just found a bug in my part.

```suggestion
                    .WithRuntime(NativeAotRuntime.Net80)
```

tests/BenchmarkDotNet.IntegrationTests/MemoryDiagnoserTests.cs
```diff
@@ -84,7 +84,7 @@ public void MemoryDiagnoserSupportsNativeAOT()
         [FactEnvSpecific("We don't want to test MonoVM twice (for .NET Framework 4.6.2 and .NET 7.0)", EnvRequirement.DotNetCoreOnly)]
```
adamsitnik, line 84:
```suggestion
        [FactEnvSpecific("We don't want to test MonoVM twice (for .NET Framework 4.6.2 and .NET 8.0)", EnvRequirement.DotNetCoreOnly)]
```

adamsitnik (commented):
Overall it LGTM, I found only some minor nits.

adamsitnik:
@AndreyAkinshin before I've applied my suggestions all CI legs were green! 🥇

AndreyAkinshin:
@adamsitnik I have removed all the temporary stuff and cleaned up the history. If all tests are green, feel free to merge the PR, there are no more upcoming updates from my side.

AndreyAkinshin (approved):

AndreyAkinshin:
@adamsitnik everything is green!

PR #2467, opened 2023-11-16
Create empty Directory.Build.props to ignore other prop files

### LINKED ISSUE #2371: Generated benchmarks ignore --framework flag when running out of process
hello
I am hitting an error where when I run our benchmarks out of process with target net7.0 (like this: dotnet run -c release --framework net7.0 -r win-x64),
Bechnmarkdotnet generates a csproj named "BenchmarkDotNet.Autogenerated" that contains the following configuration:
<TargetFramework>net7.0</TargetFramework>
but because this project is part of a solution where Directory.Build.Props is used with configuration:
<TargetFrameworks>netstandard2.0;net7.0</TargetFrameworks>
the configuration TargetFrameworks takes precedence over TargetFramework and i get error:
error NU1201: Project <project name> is not compatible with netstandard2.0.

is there a way to configure benchmarkdotnet to put TargetFrameworks in the generated project instead of TargetFramework?
or any other way to get around this problem
thank you for your help

adamsitnik:
It should not affect the WASM benchmarks (cc @radical), as they have their own generator:

https://github.com/dotnet/BenchmarkDotNet/blob/630622b6df3192f766ffa03ff07b5086e70cb264/src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs#L52

https://github.com/dotnet/BenchmarkDotNet/blob/630622b6df3192f766ffa03ff07b5086e70cb264/src/BenchmarkDotNet/Templates/WasmCsProj.txt#L3-L5

which links a dedicated props file: https://github.com/dotnet/performance/blob/main/src/benchmarks/micro/MicroBenchmarks.Wasm.props

But I am not sure about dotnet/performance in general, as IIRC the props file there alters some settings like build output path. This is going to require a detailed testing (@cincuranet  @LoopedBard3 @DrewScoggins @caaavik-msft @sblom)

fixes #2371

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
@@ -87,6 +87,22 @@ namespace BenchmarkDotNet.Toolchains.CsProj
                 .ToString();
 
             File.WriteAllText(artifactsPaths.ProjectFilePath, content);
+
+            // We create an empty Directory.Build.props file next to the project file to ignore other prop files
+            // https://github.com/dotnet/BenchmarkDotNet/issues/2371#issuecomment-1814709479
+            string propsFilePath = Path.Combine(Path.GetDirectoryName(artifactsPaths.ProjectFilePath), "Directory.Build.props");
+            if (!File.Exists(propsFilePath))
+            {
+                File.WriteAllText(propsFilePath, @"
+<Project>
+  <!-- This is an empty Directory.Build.props file to prevent projects which reside
+       under this directory to use any of the repository local settings. -->
+  <PropertyGroup>
+    <ImportDirectoryPackagesProps>false</ImportDirectoryPackagesProps>
+    <ImportDirectoryBuildTargets>false</ImportDirectoryBuildTargets>
+  </PropertyGroup>
+</Project>");
+            }
         }
 
         /// <summary>


timcassell:
Perhaps we could adjust the `csproj` to make it work, instead of adding a `Directory.Build.props`?

Old:
```xml
<Project Sdk="$SDKNAME$">
  <PropertyGroup>
    <ImportDirectoryBuildProps>false</ImportDirectoryBuildProps>
    <ImportDirectoryBuildTargets>false</ImportDirectoryBuildTargets>
  </PropertyGroup>
</Project>
```

New:
```xml
<Project>
  <PropertyGroup>
    <ImportDirectoryBuildProps>false</ImportDirectoryBuildProps>
    <ImportDirectoryBuildTargets>false</ImportDirectoryBuildTargets>
  </PropertyGroup>
  
  <Sdk Name="$SDKNAME$">
</Project>
```

(I forgot where I read that that could work.)

adamsitnik:
> I'm not entirely sure it will work

The solution that I am proposing here was recommended in https://github.com/dotnet/BenchmarkDotNet/issues/2371#issuecomment-1814709479 by @ViktorHofer who is the MSBuild expert on the .NET Team. It just works.

ViktorHofer:
Yeah, you could defer the SDK import but that would make the project file look very different. The Directory.Build.props solution is my preferred approach.

timcassell:
> But I am not sure about dotnet/performance in general, as IIRC the props file there alters some settings like build output path. 

That was addressed in #2393. The cli arguments override whatever is in the props, even if we don't ignore it. I'm not sure if other properties are consequential, but this looks good to me otherwise. Should we expose an API for whether or not to ignore props files?

timcassell:
> It should not affect the WASM benchmarks (cc @radical), as they have their own generator:
> ...
> which links a dedicated props file: https://github.com/dotnet/performance/blob/main/src/benchmarks/micro/MicroBenchmarks.Wasm.props

The directory props are imported before the project, so there it would be importing 2 props files, no? Wouldn't we want to ignore the directory props in the case of wasm also?

Also, the NativeAot generator needs this as well.

timcassell:
It seems this is no longer needed since the build system passes paths to the cli.

PR #2480, opened 2023-12-07
feat: Create dependabot.yml for automatic updates for nuget packages and github actions

wmundev:
This enables github dependabot so it is easier to keep packages up to date

at the moment, only minor and patch updates are enabled for nuget and for github actions, major versions will be updated

See examples of dependabot how the PRs will be created like from my other repository where i use dependabot
- https://github.com/wmundev/weather/pull/93
- https://github.com/wmundev/weather/pull/91

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,35 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the default location of `.github/workflows`. (You don't need to specify `/.github/workflows` for `directory`. You can use `directory: "/"`.)
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 2
+  - package-ecosystem: "nuget"
+    directory: "/src"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 2
+    groups:
+      src_projects_minor_patch_updates:
+        update-types:
+        - "minor"
+        - "patch"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
+  - package-ecosystem: "nuget"
+    directory: "/tests"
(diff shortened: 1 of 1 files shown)


wmundev:
@dotnet-policy-service agree

AndreyAkinshin:
Thanks for the PR. However, I don't think that we need such a workflow. In BenchmarkDotNet, we are quite conservative about NuGet package updates. Here are some potential problems:
* If BenchmarkDotNet itself references a package, we try to keep the lowest version possible. If a user project shares some package references with BenchmarkDotNet (e.g., `Iced`), but the BenchmarkDotNet reference version is higher, it prevents the user from benchmarking.
* In tests, we intestinally use `Newtonsoft.Json` v13.0.1 instead of the latest v13.0.3 to make it possible to test the benchmark installation of various NuGet packages (see `BenchmarkDotNet.IntegrationTests.NuGetReferenceTests`)
* We use `Verify.Xunit` v20.8.2 instead of the latest v22.8.0 because v20.8.2 is the last `Verify.Xunit` version that supports net461
* We use `Perfolizer` v0.2.1 instead of the latest v0.3.5 because the upgrade would require a refactoring (that's on me, but some preparation work on both sides is needed; the task is "work in progress")

At the moment, I'm the primary person who maintains the package version updates. I perform the package upgrades only when I have enough time to properly verify that everything works fine with the new versions. And I don't need automatic reminders.

Therefore, I don't see a need for such a workflow.

wmundev:
@AndreyAkinshin no worries, thanks for considering my Pr!

if it helps, you can also disable the dependabot nuget updates and only enable the github actions dependabot updates which helps to keep only the github actions up to date which hopefully reduces the risks 

have a nice day ahead!

AndreyAkinshin:
Well, again, I believe that we don't have problems with NuGet version maintenance that should be addressed using automatic workflows. I feel like such a workflow will bring its own problems but will not help us actually do the work.

PR #2485, opened 2023-12-15
[Bug Fix] High Performance Plan Fix

### LINKED ISSUE #2319: [Bug] Pointless power plan change
You should not change power plant to High Performance if it is already Ultimate Performance.

MattFromRVA:
Fixes #2319

### DIFF
--- a/src/BenchmarkDotNet/Running/PowerManagementApplier.cs
+++ b/src/BenchmarkDotNet/Running/PowerManagementApplier.cs
@@ -72,6 +72,14 @@ namespace BenchmarkDotNet.Running
                     isInitialized = true;
                 }
 
+                Guid ultimatePerformanceGuid = PowerPlansDict[PowerPlan.UltimatePerformance];
+                Guid highPerformanceGuid = PowerPlansDict[PowerPlan.HighPerformance];
+                if (userCurrentPowerPlan == ultimatePerformanceGuid && guid == highPerformanceGuid)
+                {
+                    logger.WriteLineInfo($"Current power plan is already Ultimate Performance. Not changing to High Performance.");
+                    return;
+                }
+
                 if (PowerManagementHelper.Set(guid))
                 {
                     powerPlanChanged = true;


Alluseri:
Been 1.5 years since this PR was proposed. And yet...

timcassell:
I don't think the change is robust enough. If a user explicitly configures high performance power plan, this will ignore it. It should only be ignored if it's not configured.

MattFromRVA:
Sounds good @timcassell. I will take a look at this when I have some free time.

MattFromRVA:
@timcassell Ok I have pushed my changes.

My understanding of the original intent in #2319 was to never downgrade to High Performance if the current plan is Ultimate. However, your recent comment makes sense that the code should always use the explicit power plan.

So, I’ve updated the code to only remain on Ultimate if the High Performance plan is not configured—otherwise, the requested plan will be set as expected.

Please let me know if this matches what you’re looking for.

timcassell:
I can't tell what you're testing here, but it doesn't look like it's testing what a user would actually do. Users set the powerplan via `Job.WithPowerPlan(...)`.

The benchmark runner retrieves the value and applies it here:

https://github.com/dotnet/BenchmarkDotNet/blob/daa233906cc5aba9aa94d4bc1b3dc449e42806e9/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L235-L236

To check if the user explicitly set the value, you can do `job.HasValue(EnvironmentMode.PowerPlanModeCharacteristic)`.

MattFromRVA:
@timcassell Thanks for the additional insight. Looks like I cannot properly test this change as I just realized I cannot change my laptop's settings to Ultimate. I pushed my latest commit if someone else wants to test/make changes.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -232,8 +232,24 @@ private static Summary Run(BenchmarkRunInfo benchmarkRunInfo,
...
+
+                    // If the plan is not explicitly set and would downgrade Ultimate, skip changing
+                    if (!userExplicitlySetPlan && requestedPlan == PowerManagementApplier.Map(PowerPlan.HighPerformance))
+                    {
+                        var ultimateGuid = PowerManagementApplier.Map(PowerPlan.UltimatePerformance);
+                        var currentPlan = PowerManagementHelper.CurrentPlan;
+                        if (currentPlan.HasValue && currentPlan.Value == ultimateGuid)
+                        {
+                            logger.WriteLineInfo("Already on Ultimate Performance; not switching to High Performance since no plan explicitly set in Job.");
+                            continue;
```
timcassell, line 248: This doesn't work, you're skipping the entire benchmark here. You should rather invert the branches, or move it to another function.

PR #2490, opened 2023-12-28
Update the default framework on the templates to net8.0

julesrx:
This PR updates the default framework on the templates to net8.0 which is a LTS release (was net6.0).
Useful to bench newer C# features, like UTF-8 string literals.


julesrx:
@dotnet-policy-service agree

AndreyAkinshin:
@julesrx thanks!

PR #2488, opened 2023-12-28
Update clrmd to 3.1

timcassell:
Fixes #2383 

I left the x86/x64 disassemblers on v1, those can be updated separately.

Also, as discussed in #2383, this breaks netcoreapp3.0 and older, so we should probably release this in 0.14.0.

PR #2505, opened 2024-01-14
Remove support for old core runtimes

timcassell:
Follow-up to #2488.
Explicitly removed support for netcoreapp3.0 and older build targets, instead of letting the build system fail.


AndreyAkinshin (approved):
No concerns from my side. If @adamsitnik has no objections, we can merge it.

timcassell:
> No concerns from my side. If @adamsitnik has no objections, we can merge it.

Cool, but we should wait for #2488 to be merged first.

timcassell:
Merging since #2488 was merged.
@AndreyAkinshin Probably should make the next version v0.15.0 since #2488 was very breaking for old runtimes.

Issue #2506, opened 2024-01-14
Include Environment Variables in JSON output in easier-to-parse form


edyoung:
If I run a set of benchmarks with

`dotnet run -c:Release -- -f *Log* --runtimes net8.0 --envVars x:1 y:2`

And export a JSON report, the JSON contains the environment variables embedded in the DisplayInfo:

```json
"DisplayInfo":"LogBenchmark.TrackTrace: Job-OTYZFP(EnvironmentVariables=x=1,y=2, Runtime=.NET 8.0, Toolchain=net8.0, IterationCount=3, LaunchCount=1, WarmupCount=3)"
```

But I would like to be able to extract the environment variable values more easily from the report for further processing. What if there was an additional property bag in the report? 

Something like:

```json
 "EnvironmentVars": {
    "x": "1",
    "y:" "2"
  }
```

adamsitnik:
I am supportive of this idea. Please go ahead and send a PR.

edyoung:
The linked PR is complete afaik but needs a review.

PR #2507, opened 2024-01-15
Environmentvars in report

edyoung:
Addresses #2506 

Adds a property `EnvironmentVariables` to the Benchmark JSON output. The value is an object with keys and values for the environment variables, like this:

```json
"EnvironmentVariables":{
    "foo":"bar"
    }
```
If no variables are specified the output includes an empty object, eg

```json
"EnvironmentVariables": {}
```

Should other exporters like XmlExporter be updated for symmetry?

The production code change just in JsonExporterBase. The other changes are for the unit tests.


src/BenchmarkDotNet/Exporters/Json/JsonExporterBase.cs
```diff
@@ -74,6 +74,7 @@ public override void ExportToLog(Summary summary, ILogger logger)
                 { "Method", report.BenchmarkCase.Descriptor.WorkloadMethod.Name },
                 { "MethodTitle", report.BenchmarkCase.Descriptor.WorkloadMethodDisplayInfo },
                 { "Parameters", report.BenchmarkCase.Parameters.PrintInfo },
+                { "EnvironmentVariables", report.BenchmarkCase.Job.Environment.EnvironmentVariables ?? new List<Jobs.EnvironmentVariable>() },
```
timcassell, line 77: Maybe `.ToDictionary` would make it look better?

edyoung: Thanks, I think that is an improvement. .ToDictionary() can throw if the list contains duplicate keys, but BenchmarkDotNet already ensures that the list doesn't contain duplicates elsewhere, so I don't believe we need to account for that here.

kant2002:
@adamsitnik or @AndreyAkinshin what do you think about this PR?

AndreyAkinshin:
@edyoung thank you very much for the PR.

Right now I'm working on a new serialization format that would naturally handle all possible metadata values (including environment variables). I'm also going to provide a set of API for easy manipulation and analysis of the data. I will post updates once I obtain a ready-to-use prototype.

The current json format is going to be deprecated in favor of the new one. I appreciate the effort you put into the PR, but I'm going to close since I feel that it doesn't make sense to extend the existing json format.

PR #2511, opened 2024-01-24
Sdk Validation Fix

### LINKED ISSUE #2199: Deadlock when SDK is not installed
```C#
[DryJob(RuntimeMoniker.Net50)] //this can be changed for example to RuntimeMoniker.NetCoreApp20
public class Benchmarks
{
    [Benchmark] public void Method() { }
}
```

I waited 20 min (10min is a timeout), nothing has changed.

```
// Benchmark: BenchmarkClass1.Method: Default-.NET 5.0(Runtime=.NET 5.0)
// *** Execute ***
// Launch: 1 / 1
// Execute: dotnet 8ba28a4a-8517-449f-95a6-54f11551ed2b.dll --anonymousPipes 7480 7564 --benchmarkName BenchmarkDotNet.Samples.BenchmarkClass1.Method --job "Default-.NET
 5.0" --benchmarkId 0 in C:\src\github\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\bin\Release\net6.0\8ba28a4a-8517-449f-95a6-54f11551ed2b\bin\Release\net5.0
```

This happens even when a target benchmark framework is not selected:
```xaml
result is the same:
<TargetFramework>net6.0</TargetFramework>
<TargetFrameworks>net6.0;net5.0</TargetFrameworks>
```

MattFromRVA:
Fixes #2199

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -363,9 +363,12 @@ namespace BenchmarkDotNet.Running
                         .Validate(new ValidationParameters(joinedCases, null))
                     );
             }
-
+            var sdkValidator = SdkValidator.Instance;
             foreach (var benchmarkRunInfo in benchmarks)
-                validationErrors.AddRange(benchmarkRunInfo.Config.GetCompositeValidator().Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config)));
+            {
+                var sdkValidationErrors = sdkValidator.Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config));
+                validationErrors.AddRange(sdkValidationErrors);
+            }
 
             return validationErrors.ToImmutableArray();
         }
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -363,16 +363,8 @@ namespace BenchmarkDotNet.Running
                         .Validate(new ValidationParameters(joinedCases, null))
                     );
             }
-            var sdkValidator = SdkValidator.Instance;
             foreach (var benchmarkRunInfo in benchmarks)
-            {
-                var validators = benchmarkRunInfo.Config.GetValidators().ToList();
-                validators.Add(sdkValidator);
-                var compositeValidator = new CompositeValidator(validators.ToImmutableHashSet());
-
-                var currentValidationErrors = compositeValidator.Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config));
-                validationErrors.AddRange(currentValidationErrors);
-            }
+                validationErrors.AddRange(benchmarkRunInfo.Config.GetCompositeValidator().Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config)));
 
             return validationErrors.ToImmutableArray();
         }
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -366,8 +366,12 @@ namespace BenchmarkDotNet.Running
             var sdkValidator = SdkValidator.Instance;
             foreach (var benchmarkRunInfo in benchmarks)
             {
-                var sdkValidationErrors = sdkValidator.Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config));
-                validationErrors.AddRange(sdkValidationErrors);
+                var validators = benchmarkRunInfo.Config.GetValidators().ToList();
+                validators.Add(sdkValidator);
+                var compositeValidator = new CompositeValidator(validators.ToImmutableHashSet());
+
+                var currentValidationErrors = compositeValidator.Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config));
+                validationErrors.AddRange(currentValidationErrors);
             }
 
             return validationErrors.ToImmutableArray();
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -363,6 +363,7 @@ namespace BenchmarkDotNet.Running
                         .Validate(new ValidationParameters(joinedCases, null))
                     );
             }
+
             foreach (var benchmarkRunInfo in benchmarks)
                 validationErrors.AddRange(benchmarkRunInfo.Config.GetCompositeValidator().Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config)));
 
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs
@@ -26,13 +26,15 @@ namespace BenchmarkDotNet.Toolchains.CsProj
         [PublicAPI] public static readonly IToolchain NetCoreApp80 = From(NetCoreAppSettings.NetCoreApp80);
         [PublicAPI] public static readonly IToolchain NetCoreApp90 = From(NetCoreAppSettings.NetCoreApp90);
 
-        internal CsProjCoreToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath, ISdkProvider sdkProvider)
+        private readonly ISdkProvider sdkProvider;
+
+        public ISdkProvider SdkProvider => sdkProvider;
+
+        internal CsProjCoreToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, ISdkProvider sdkProvider)
             : base(name, generator, builder, executor, sdkProvider)
         {
-            CustomDotNetCliPath = customDotNetCliPath;
+            this.sdkProvider = sdkProvider;
         }
-
-        internal string CustomDotNetCliPath { get; }
 
         [PublicAPI]
         public static IToolchain From(NetCoreAppSettings settings)
@@ -40,8 +42,7 @@ namespace BenchmarkDotNet.Toolchains.CsProj
                 new CsProjGenerator(settings.TargetFrameworkMoniker, settings.CustomDotNetCliPath, settings.PackagesPath, settings.RuntimeFrameworkVersion),
                 new DotNetCliBuilder(settings.TargetFrameworkMoniker, settings.CustomDotNetCliPath),
                 new DotNetCliExecutor(settings.CustomDotNetCliPath),
-                settings.CustomDotNetCliPath,
-                new DotNetSdkProvider());
+                new DotNetSdkProvider() { CustomDotNetCliPath = settings.CustomDotNetCliPath });
 
         public override IEnumerable<ValidationError> Validate(BenchmarkCase benchmarkCase, IResolver resolver)
         {
@@ -50,7 +51,7 @@ namespace BenchmarkDotNet.Toolchains.CsProj
                 yield return validationError;
             }
 
-            if (IsCliPathInvalid(CustomDotNetCliPath, benchmarkCase, out var invalidCliError))
+            if (IsCliPathInvalid(sdkProvider.CustomDotNetCliPath, benchmarkCase, out var invalidCliError))
             {
                 yield return invalidCliError;
             }
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs
@@ -28,7 +28,7 @@ namespace BenchmarkDotNet.Toolchains.CsProj
 
         private readonly ISdkProvider sdkProvider;
 
-        public ISdkProvider SdkProvider => sdkProvider;
+        internal ISdkProvider SdkProvider => sdkProvider;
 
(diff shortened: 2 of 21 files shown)


src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -363,9 +363,16 @@ private static ImmutableArray<ValidationError> Validate(params BenchmarkRunInfo[
...
             foreach (var benchmarkRunInfo in benchmarks)
-                validationErrors.AddRange(benchmarkRunInfo.Config.GetCompositeValidator().Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config)));
+            {
+                var validators = benchmarkRunInfo.Config.GetValidators().ToList();
+                validators.Add(sdkValidator);
+                var compositeValidator = new CompositeValidator(validators.ToImmutableHashSet());
+
+                var currentValidationErrors = compositeValidator.Validate(new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config));
+                validationErrors.AddRange(currentValidationErrors);
+            }
```
timcassell, line 375: This looks like a very strange place to add the validator. I think you should instead add the validation to `CsProjCoreToolchain.Validate`. (And other toolchains that need it. Maybe it would be useful to make them inherit a common base class with that validation.)

MattFromRVA:
@timcassell I am not sure why the linux and macos checks are failing. According to the result report all tests have passed. Am I missing something? https://github.com/dotnet/BenchmarkDotNet/pull/2511/checks?check_run_id=21320355764

timcassell:
> @timcassell I am not sure why the linux and macos checks are failing. According to the result report all tests have passed. Am I missing something? https://github.com/dotnet/BenchmarkDotNet/pull/2511/checks?check_run_id=21320355764

I ran into the same issue in some of my PRs. I opened an issue for it. #2520. You'll have to search the logs or run it locally to find the culprit.

MattFromRVA:
> > @timcassell I am not sure why the linux and macos checks are failing. According to the result report all tests have passed. Am I missing something? https://github.com/dotnet/BenchmarkDotNet/pull/2511/checks?check_run_id=21320355764
> 
> I ran into the same issue in some of my PRs. I opened an issue for it. #2520. You'll have to search the logs or run it locally to find the culprit.

Ok sounds good. I will do some digging. @timcassell

src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs
```diff
@@ -26,8 +26,8 @@ public class CsProjCoreToolchain : Toolchain, IEquatable<CsProjCoreToolchain>
         [PublicAPI] public static readonly IToolchain NetCoreApp80 = From(NetCoreAppSettings.NetCoreApp80);
         [PublicAPI] public static readonly IToolchain NetCoreApp90 = From(NetCoreAppSettings.NetCoreApp90);
 
-        internal CsProjCoreToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath)
-            : base(name, generator, builder, executor)
+        internal CsProjCoreToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath, ISdkProvider sdkProvider)
```
timcassell, line 29: Why don't we merge the `customDotNetCliPath` into the `ISdkProvider`? They seem very related, and the validation likely would change depending on the custom path.

MattFromRVA: > Why don't we merge the `customDotNetCliPath` into the `ISdkProvider`? They seem very related, and the validation likely would change depending on the custom path.

Will do. Should I do the same thing with `customDotNetCliPath` in `NativeAotToolchain` as well?

timcassell: Yes, please. All the toolchains that use it.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjCoreToolchain.cs
```diff
@@ -26,21 +26,23 @@ public class CsProjCoreToolchain : Toolchain, IEquatable<CsProjCoreToolchain>
         [PublicAPI] public static readonly IToolchain NetCoreApp80 = From(NetCoreAppSettings.NetCoreApp80);
         [PublicAPI] public static readonly IToolchain NetCoreApp90 = From(NetCoreAppSettings.NetCoreApp90);
 
-        internal CsProjCoreToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath)
-            : base(name, generator, builder, executor)
+        private readonly ISdkProvider sdkProvider;
+
+        public ISdkProvider SdkProvider => sdkProvider;
```
timcassell, line 31:
```suggestion
        internal ISdkProvider SdkProvider => sdkProvider;
```

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -15,7 +15,7 @@ public class MonoAotToolchain : Toolchain
         public static readonly IToolchain Instance = new MonoAotToolchain();
 
         [PublicAPI]
-        public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(), new Executor())
+        public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(), new Executor(), null)
```
timcassell, line 18: If not every toolchain needs an `ISdkProvider`, why is it part of the base toolchain?

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolchain.cs
```diff
@@ -13,8 +13,8 @@ public class WasmToolchain : Toolchain
     {
         private string CustomDotNetCliPath { get; }
 
-        private WasmToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath)
-            : base(name, generator, builder, executor)
+        private WasmToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, string customDotNetCliPath, ISdkProvider sdkProvider)
```
timcassell, line 16:
```suggestion
        private WasmToolchain(string name, IGenerator generator, IBuilder builder, IExecutor executor, ISdkProvider sdkProvider)
```

src/BenchmarkDotNet/Toolchains/NativeAot/NativeAotToolchain.cs
```diff
@@ -37,26 +38,30 @@ public class NativeAotToolchain : Toolchain
             .TargetFrameworkMoniker("net9.0")
             .ToToolchain();
 
+#pragma warning disable CS0649
+        private readonly ISdkProvider sdkProvider;
+#pragma warning restore CS0649
+
+        public ISdkProvider SdkProvider => sdkProvider;
```
timcassell, line 45:
```suggestion
        internal ISdkProvider SdkProvider => sdkProvider;
```

src/BenchmarkDotNet/Validators/ISdkProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace BenchmarkDotNet.Validators
+{
+    public interface ISdkProvider
+    {
+        IEnumerable<string> GetInstalledSdks();
+        string CustomDotNetCliPath { get; set; }
```
timcassell, line 8:
```suggestion
        string CustomDotNetCliPath { get; }
```

src/BenchmarkDotNet/Validators/SdkValidator.cs
```diff
@@ -0,0 +1,142 @@
...
+                {
+                    yield return new ValidationError(
+                        TreatsWarningsAsErrors,
+                        $"The required SDK for {runtimeMoniker} is not installed",
+                        benchmark);
+                }
+            }
+        }
+
+        private bool IsSdkInstalled(RuntimeMoniker runtimeMoniker)
```
timcassell, line 41: This should probably be updated to check the `CustomDotNetCliPath`.

tests/BenchmarkDotNet.Tests/Validators/FakeSdkProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+            this.installedSdks = installedSdks;
+            this.CustomDotNetCliPath = customDotNetCliPath;
+        }
+
+        public IEnumerable<string> GetInstalledSdks()
+        {
+            if (!string.IsNullOrEmpty(CustomDotNetCliPath))
+            {
+                return installedSdks;
+            }
```
timcassell, line 22: I don't understand this check. Both paths return the same thing.

timcassell (changes requested):

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
MattFromRVA: I forgot about this one. I am not familiar with Mono and just put `null` for the time being while I did more research. From what I found it is possible that Mono can use the .NET SDK since .NET 5. 

I will work on its implementation.

src/BenchmarkDotNet/Validators/SdkValidator.cs
MattFromRVA: Great catch. The path info is passed into `DotNetSdkProvider` but I forgot to update my method there. I will revise it.

tests/BenchmarkDotNet.Tests/Validators/FakeSdkProvider.cs
MattFromRVA: Oops. Me neither. 😄

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
timcassell: Please do that in a separate PR if you're going to do it. But it's the same thing for in process toolchains and Roslyn toolchain.

MattFromRVA: ah ok. I just committed some more changes but I will update my code.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolchain.cs
```diff
@@ -11,12 +11,12 @@ namespace BenchmarkDotNet.Toolchains.MonoWasm
     [PublicAPI]
     public class WasmToolchain : Toolchain
     {
-        private string CustomDotNetCliPath { get; }
+        private readonly string CustomDotNetCliPath;
```
timcassell, line 14: I don't think this field is needed.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmToolchain.cs
```diff
@@ -51,7 +51,6 @@ public static IToolchain From(NetCoreAppSettings netCoreAppSettings)
                         // aot builds can be very slow
                         logOutput: netCoreAppSettings.AOTCompilerMode == MonoAotLLVM.MonoAotCompilerMode.wasm),
                     new WasmExecutor(),
-                    netCoreAppSettings.CustomDotNetCliPath,
                     new DotNetSdkProvider());
```
timcassell, line 54: Need to pass the custom path to the provider.

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
```diff
@@ -15,7 +15,7 @@ public class MonoAotToolchain : Toolchain
         public static readonly IToolchain Instance = new MonoAotToolchain();
 
         [PublicAPI]
-        public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(), new Executor(), null)
+        public MonoAotToolchain() : base("MonoAot", new Generator(), new MonoAotBuilder(), new Executor(), new DotNetSdkProvider())
```
timcassell, line 18: It appears that MonoAotToolchain uses a variant of Roslyn. So it doesn't need the ISdkProvider (unless Roslyn toolchain needs to validate the sdk also, but it seems different than csproj toolchains).

src/BenchmarkDotNet/Validators/DotNetSdkProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+{
+    public class DotNetSdkProvider : ISdkProvider
+    {
+        private string _customDotNetCliPath;
+
+        public string CustomDotNetCliPath
+        {
+            get => _customDotNetCliPath;
+            set => _customDotNetCliPath = value;
+        }
```
timcassell, line 17: Just make this a read only property and pass the string into the constructor (can make it an optional argument).

src/BenchmarkDotNet/Validators/ISdkProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace BenchmarkDotNet.Validators
+{
+    public interface ISdkProvider
```
timcassell, line 5: `ISdkProvider` name implies that a different sdk besides dotnet could be used. Does that make sense? Does `RoslynToolchain` use an sdk that isn't dotnet?

timcassell: Also, I think this interface should be moved to `BenchmarkDotNet.Toolchains` namespace.

src/BenchmarkDotNet/Toolchains/Mono/MonoAotToolchain.cs
MattFromRVA: ah ok. Good info. I will definitely need a different approach in implementing this validation to just certain toolchains then.

MattFromRVA:
@timcassell I am closing this PR and opening a new one. After doing some more reworking I found that the scope of this .Net validation is much smaller than I was originally assuming with Toolchains. 

The new PR has a simpler approach instead of using inheritance which was messing up other Toolchains. 

I appreciate all the feedback so far.

PR #2508, opened 2024-01-23
Gather assembly references

timcassell:
Fixes #1197
Fixes #2344
Also fixes legacy `packages.config` issues with `CsProjClassicNetToolchain` (no open issue, but was mentioned in https://github.com/dotnet/BenchmarkDotNet/issues/2405#issuecomment-1683212942).
And this is a pre-requisite to adding support for #2289.

PR #2515, opened 2024-02-04
Use `CsProjClassicNetToolchain` as default toolchain in .Net Framework

timcassell:
Don't merge before #2508.


timcassell:
No longer needed since #2682 correctly handles which toolchain to use by default.

PR #2525, opened 2024-02-22
Update toolchain docs

MattFromRVA:
Updated non-default dotnet cli path example.

I was receiving compiler warnings for `Add` and `With` when using the provided doc example. I have updated them to the newer `AddJob` and `WithToolchain` methods.


adamsitnik (approved):
LGTM, thank you @MattFromRVA !

PR #2531, opened 2024-02-26
Fix wasm toolchain

timcassell:
I think this should fix https://github.com/dotnet/performance/issues/3984, but I couldn't get wasm to work locally, so I couldn't test it. PTAL @LoopedBard3

PR #2532, opened 2024-02-27
Add Integration Test for WASM

caaavik-msft:
I'm making this PR a draft since I was not able to get it to a condition where it is good to merge in, however I think it will be a useful start for someone who is more familiar with WASM to complete it. There also would need to be some changes to the CI machines to install the wasm-tools workload and the v8 engine so that wasm can run.

This PR also includes the changes from https://github.com/dotnet/BenchmarkDotNet/pull/2531 but I think that change should go in separately and can be removed from this PR. I only include it as I needed it to test out that everything is working. I can confirm though that the issue that caused WASM runs to fail was reproducible by this integration test.

The test doesn't pass yet on my system though, but I believe it is failing because of the stuff in test-main.js which I copied from the dotnet/runtime repository and not because of any issues with BDN itself. I wasn't able to figure out how to create a minimal test-main.js to use for this integration test and just used the one from the dotnet/runtime repository, but I think maybe it relies on some v8 features or specific versions that I wasn't able to figure out. Here is the error message I get when it tries to run the js:

    TypeError: dotnet.withVirtualWorkingDirectory(...).withEnvironmentVariables(...).withDiagnosticTracing(...).withExitOnUnhandledError(...).withExitCodeLogging(...).withElementOnExit(...).withInteropCleanupOnExit(...).withAssertAfterExit(...).withDumpThreadsOnNonZeroExit is not a function


tests/BenchmarkDotNet.IntegrationTests/WasmTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        public class WasmBenchmark
+        {
+            [Benchmark]
+            public void Check()
+            {
+                // See if there are any other checks to run to make sure the benchmark is running in a WASM environment.
+
+                if (RuntimeInformation.GetCurrentRuntime().RuntimeMoniker != RuntimeMoniker.WasmNet70)
```
adamsitnik, line 46: I would expect it to be 8.0?

```suggestion
                if (RuntimeInformation.GetCurrentRuntime().RuntimeMoniker != RuntimeMoniker.WasmNet80)
```

adamsitnik (commented):
@caaavik-msft Thank you for working on this! It would be great to have the coverage and finally stop randomly breaking WASM support.

> There also would need to be some changes to the CI machines to install the wasm-tools workload and the v8 engine so that wasm can run.

We are using GH actions, so this should be doable. Here you can take a look to see how we are using npm to install some other js stuff:

https://github.com/dotnet/BenchmarkDotNet/blob/0a41e16424cb256caacfdaf490782be83f5b57f4/.github/workflows/run-tests.yaml#L113-L118

Our WASM doc is outdated, but it can be helpful too: https://benchmarkdotnet.org/articles/configs/toolchains.html#wasm

caaavik-msft:
@radical @lewing Are one of you able to assist with writing a minimal test-main.js to be used in a WASM BDN Integration Test that is simple enough we don't have to worry about using specific versions of v8. I took the one from the runtime repo but I think it might depend on the version defined in the ChromeVersion.props file. Once we get this change in we can prevent regressions like https://github.com/dotnet/performance/issues/3984 from occurring again.

radical:
cc @pavelsavara @maraf

tests/BenchmarkDotNet.IntegrationTests/WasmTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            var dotnetVersion = "net8.0";
+            var logger = new OutputLogger(Output);
+            var netCoreAppSettings = new NetCoreAppSettings(dotnetVersion, null, "Wasm");
+            var mainJsPath = Path.Combine(AppContext.BaseDirectory, "AppBundle", "test-main.js");
+
+            var config = ManualConfig.CreateEmpty()
+                .AddLogger(logger)
+                .AddJob(Job.Dry
+                    .WithArguments([new MsBuildArgument($"/p:WasmMainJSPath={mainJsPath}")])
+                    .WithRuntime(new WasmRuntime(dotnetVersion, moniker: RuntimeMoniker.WasmNet70, javaScriptEngineArguments: "--expose_wasm --module"))
```
timcassell, line 32:
```suggestion
                    .WithRuntime(new WasmRuntime(dotnetVersion, moniker: RuntimeMoniker.WasmNet80, javaScriptEngineArguments: "--expose_wasm --module"))
```

timcassell:
#2531 was merged, so you can rebase and remove those changes.

pavelsavara:
> Are one of you able to assist with writing a minimal `test-main.js`

`test-main.js` is not great template and it also says so in the header.

I don't know what kind of setup this perf test requires, but here is [simple V8 sample](https://github.com/dotnet/runtime/blob/main/src/mono/sample/wasm/console-v8/main.mjs)

Also this is [public API ](https://github.com/dotnet/runtime/blob/8f33d62892130fab9ca1fd8bf6c31ba2c16bae25/src/mono/browser/runtime/dotnet.d.ts#L56-L131) you want to use to configure it.

caaavik-msft:
@pavelsavara it seems that script is sufficient to get things working without any more customisation, thanks! 

@adamsitnik @AndreyAkinshin CI is fully working now setting up everything needed for the wasm integration test to run. However, I am not sure if maybe that stuff better belongs inside the build scripts rather than in the github actions yaml. Particularly the part where we install the wasm-tools workload. Let me know what you think

adamsitnik (approved):
It looks great to me @caaavik-msft, big thanks for preventing future WASM breaks!

adamsitnik:
> However, I am not sure if maybe that stuff better belongs inside the build scripts rather than in the github actions yaml. Particularly the part where we install the wasm-tools workload. Let me know what you think

IMO if it turns out to be a problem we can improve it later.

@AndreyAkinshin thoughts?

caaavik-msft:
Thanks to @ilonatommy who sent me the changes needed to add this functionality to the BenchmarkDotNet.Build project, I think those concerns should be addressed now and we should be ok to merge this

ilonatommy (approved):

AndreyAkinshin:
@caaavik-msft thank you very much for WasmTests support! Great job with extending `build.cmd`! I like this approach: it's consistent with the other build workflows and provides a simple way to configure the local environment. It was pretty easy to run it locally.

P.S. I have added [additional docs](https://github.com/dotnet/BenchmarkDotNet/commit/63626bb357b802c7f724744f2f29ce2921a351dc) about local setup.

PR #2538, opened 2024-03-05
Add support for wasm and Mono AOT LLVM monikers to SDK Validator

caaavik-msft:
These monikers were missed in the recent SDK Validation feature.


timcassell (approved):
Thanks!

Issue #2536, opened 2024-02-29
Add integration tests for `MonoAOTLLVMToolchain`


timcassell:
To make sure it works and fails properly (#2533).

PR #2539, opened 2024-03-08
Fix Mono AOT LLVM

caaavik-msft:
Similar to what broke with the WASM runs, it seems that Mono AOT LLVM also will always append the `publish` directory onto the end of the `OutDir` and `OutputPath` that we pass in, so we need to remove it from the end of those paths and then append it the executable path and working directory.

~~We are doing some more testing on this first before it can be ready to merge however, but this seems like the likely fix.~~
Testing is done, this PR is ready for a merge.

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -571,19 +571,19 @@ namespace BenchmarkDotNet.ConsoleArguments
                     return MakeWasmJob(baseJob, options, "net9.0", runtimeMoniker);
 
                 case RuntimeMoniker.MonoAOTLLVM:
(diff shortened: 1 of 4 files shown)


caaavik-msft:
I'm fixing a 2nd bug in this PR, if you specified a moniker `monoaotllvmnet90` it was still using the non .NET specific MonoAOTLLVM moniker in the runtime and so the `DotNetSdkValidator` was failing.

LoopedBard3 (approved):
LGTM with successful local tests when using monoaotllvmnet90 runtime.

timcassell:
Do we need to account for `MonoAOTLLVM` and `Wasm` monikers in the sdk validator? And should we deprecate those?

caaavik-msft:
I have created another issue for that: https://github.com/dotnet/BenchmarkDotNet/issues/2540

timcassell:
It would also be nice to get integration tests for this (#2536).

timcassell (approved):

timcassell:
Thanks for the fix!

Issue #2540, opened 2024-03-08
Unable to use `wasm` or `monoaotllvm` as runtime


caaavik-msft:
With the recent change from https://github.com/dotnet/BenchmarkDotNet/pull/2523, we are unable to use `wasm` or `monoaotllvm` in the `--runtimes` argument like we could previously because the DotNetSdkValidator does not support it. Instead we have to use one which has the .NET version in it such as `wasmnet9.0` or `monoaotllvmnet9.0`. I am not sure what the correct fix for this should be, but the dotnet/performance repo uses these two values for our WASM and Mono AOT LLVM runs, so this is blocking our ability to merge in the latest BDN.

LoopedBard3:
Put up a PR that should fix the issue for both wasm and monoaotllvm: https://github.com/dotnet/BenchmarkDotNet/pull/2541

PR #2541, opened 2024-03-08
Add Wasm and MonoAOTLLVM to DotNetSdkVersionValidator

LoopedBard3:
Add Wasm and MonoAOTLLVM plain moniker support to DotNetSdkVersionValidator. I used the same check mentioned in the issue for determining the sdkversion, with an additional substring to remove the 'net' portion of the MsBuildMoniker (eg. net9.0).
Tested locally for MonoAOTLLVM successfully.

Fix: #2540


caaavik-msft (approved):

timcassell:
I think we should use the version of the runtime rather than the MsBuildMoniker. `CoreRuntime.TryGetVersion` That way we can be sure we're not getting something like `net8.0-windows` and we can make sure it's using a supported version (3.1 should be bumped up to 5.0/6.0). Now that I mention it, the config parser doesn't enforce the minimum version there either, so it would probably just explode if the host version is too old (do we care, though?).

LoopedBard3:
Updated and tested locally for MonoAOTLLVM successfully.

timcassell (approved):

PR #2549, opened 2024-03-30
DotMemoryDiagnoser implementation

martinothamar:
Implements `DotMemoryDiagnoser` to match `DotTraceDiagnoser`

* The code is pretty similar between dotMemory and dotTrace impl, because the underlying APIs/CLIs are similar. Could do something to share code here
  * Is there any reason we can't use the [PID config](https://github.com/JetBrains/profiler-self-api/blob/80b61830f41b42fb036adcf2af1083c6fab21173/JetBrains.Profiler.SelfApi/src/DotTrace.cs#L421) on the APIs for the Jetbrains library directly? I could refactor to use that instead, both for dotTrace and dotMemory
* Could refactor out common code and either publish a `Common` package or just use `<Compile Include="..." />`

Sample output of `IntroDotMemoryDiagnoser`

image


martinothamar:
There was a segfault happening for the integration test that I don't quite understand. I can no longer reproduce it after the last commit. I tried to debug some dumps locally but was unable to find relevant exceptions, so my only guess atm is that the crash originates from unmanaged code. Since changing the init logic seemed to affect things, and the crash always happening during `InProcessEmitToolchain`,  I'm guessing it has something to do with attaching/detaching multiple times in the same process or something

In any case, I would like to test out using only the JetBrains API directly

martinothamar:
The last commit simplifies the implementation of the tool by using only the Jetbrains API and more clearly manages the lifetime of it

AndreyAkinshin (approved):

AndreyAkinshin:
@martinothamar thank you very much for this PR!

PR #2558, opened 2024-04-12
Simplify dotTrace diagnoser, dedupe common code

martinothamar:
Similar approach to [#2549](https://github.com/dotnet/BenchmarkDotNet/pull/2549#issuecomment-2028807304) for simplifying `DotTraceDiagnoser`.

Also deduped some code by creating a shared project (using `<Compile Include="..." />` to avoid another package).
Pulled this code into `JetBrains` folder so that the projects are ordered next to eachother in file explorers.
Made sure to update build script and references to include the new folder.

Other options:
* Just not dedupe
* Keep the common code in one of the projects and just include from there
* ?


timcassell:
Could you use shared project for this?

image

martinothamar:
TIL, did not know that existed :smile: 

I wanted to test that out now but I'm not sure if it's possible to create that project type for me since I'm working on Linux atm. .NET SDK, Rider and VSCode C# DevKit all seem to be missing support for creating this project type. 

Reading up on this issue: https://github.com/dotnet/sdk/issues/2511
There seems to be some complaints about the UX of it, but that it is fully usable. Some people suggest `Microsoft.Build.NoTargets`, while the "official" recommendation seems to be using a folder + globbing includes? I thought having a plain folder to contain the files would make intellisense/tooling/IDE experience worse but testing that out now its completely fine and everything works as expected. So personally I think the plain folder approach seems to be nice enough and lightweight. I pushed a commit now  to do that

I could try out shared projects, but I would have to hand-code it unless someone on a Windows PC could scaffold it for me. What do you think?

timcassell:
Ah, I didn't realize that was legacy project type. I think just using the folder approach is good.

BenchmarkDotNet.sln
```diff
@@ -47,13 +47,15 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "templates", "templates", "{
...
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BenchmarkDotNet.Diagnostics.dotTrace", "src\BenchmarkDotNet.Diagnostics.dotTrace\BenchmarkDotNet.Diagnostics.dotTrace.csproj", "{C5BDA61F-3A56-4B59-901D-0A17E78F4076}"
-EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks", "tests\BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks\BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks.csproj", "{AACA2C63-A85B-47AB-99FC-72C3FF408B14}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BenchmarkDotNet.TestAdapter", "src\BenchmarkDotNet.TestAdapter\BenchmarkDotNet.TestAdapter.csproj", "{4C9C89B8-7C4E-4ECF-B3C9-324C8772EDAC}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BenchmarkDotNet.Diagnostics.dotMemory", "src\BenchmarkDotNet.Diagnostics.dotMemory\BenchmarkDotNet.Diagnostics.dotMemory.csproj", "{2E2283A3-6DA6-4482-8518-99D6D9F689AB}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "JetBrains", "JetBrains", "{AD5887B3-2F6C-48B6-B338-A4E286AFFBBD}"
```
timcassell, line 54: Looks like the project was not removed from the sln when you deleted it.

src/JetBrains/Shared/Helper.cs
```diff
@@ -0,0 +1,87 @@
+using System;
+using System.Reflection;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Portability;
+
+namespace BenchmarkDotNet.JetBrains.Shared
```
timcassell, line 6: nit

```suggestion
namespace BenchmarkDotNet.JetBrains
```

src/JetBrains/Shared/Progress.cs
```diff
@@ -2,9 +2,10 @@
 using System.Diagnostics;
 using BenchmarkDotNet.Loggers;
 
-namespace BenchmarkDotNet.Diagnostics.dotTrace
+namespace BenchmarkDotNet.JetBrains.Shared
 {
-    public class Progress : IProgress<double>
+
+    internal sealed class Progress : IProgress<double>
```
timcassell, line 8: This was public before. Is it ok to remove it from public API? @AndreyAkinshin

BenchmarkDotNet.sln
martinothamar: I think this is just the solution folder, so that the solution explorer view matches what's in the filesystem 
image

src/JetBrains/Shared/Progress.cs
martinothamar: True, I forgot to mention that. I figured since it's not used by any configuration or anything it makes sense to not expose it? Users could have referenced it for some reason, but not because of any interaction with BenchmarkDotNet AFAICT.

If we keep it public I probably have to do something about the namespace, so that both packages don't expose the same type in the same namespace, unless that's not a big deal since we don't expect anyone to import `BenchmarkDotNet.JetBrains` regardless of which package is used?

src/JetBrains/BenchmarkDotNet.Diagnostics.dotMemory/BenchmarkDotNet.Diagnostics.dotMemory.csproj
```diff
@@ -10,7 +10,11 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\BenchmarkDotNet\BenchmarkDotNet.csproj" />
+        <ProjectReference Include="..\..\BenchmarkDotNet\BenchmarkDotNet.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <Compile Include="..\Shared\*.cs" />
```
timcassell, line 17: Recursively includes items in sub-folders (just in case for future).

```suggestion
        <Compile Include="..\Shared\**/*.cs" />
```

src/JetBrains/BenchmarkDotNet.Diagnostics.dotTrace/BenchmarkDotNet.Diagnostics.dotTrace.csproj
```diff
@@ -10,7 +10,11 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\BenchmarkDotNet\BenchmarkDotNet.csproj" />
+        <ProjectReference Include="..\..\BenchmarkDotNet\BenchmarkDotNet.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <Compile Include="..\Shared\*.cs" />
```
timcassell, line 17:
```suggestion
        <Compile Include="..\Shared\**/*.cs" />
```

AndreyAkinshin:
@martinothamar thanks for the PR! Indeed, it makes sense to unify dotTrace/dotMemory diagnosers and eliminate code duplication. However, the presented approach inherits the disadvantages of the poor design of the original diagnosers. Both were born as prototypes. Now, it's time to refactor them. I have come up with a better solution for this problem, see #2627.

Issue #2548, opened 2024-03-28
Cannot build NativeAOT 8.0 on Arch Linux


Jisu-Woniu:
I wrote a simple benchmark, with NativeAOT enabled:

```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Running;

BenchmarkRunner.Run<MyBench>();


[SimpleJob(RuntimeMoniker.NativeAot80)]
public class MyBench
{
    [Benchmark]
    public void Bench() { }
}
```

But when I run `dotnet run -c Release`, it fails with:

```plaintext
...
// Build Error: Standard output: 
  
 Standard error: 
 MSBuild version 17.8.5+b5265ef37 for .NET
$HOME/.nuget/packages/microsoft.net.illink.tasks/8.0.3/build/Microsoft.NET.ILLink.targets(210,5): warning : Property 'TrimmerDefaultAction' is deprecated in .NET 7 and higher and will be ignored. Use TrimMode instead. [<PROJECT_DIR>/bin/Release/net8.0/136c2e4d-c592-4318-baf9-67ab64a593fc/BenchmarkDotNet.Autogenerated.csproj]
  Generating native code
EXEC : error : Unrecognized instruction set avx-512f [<PROJECT_DIR>/bin/Release/net8.0/136c2e4d-c592-4318-baf9-67ab64a593fc/BenchmarkDotNet.Autogenerated.csproj]
  System.CommandLine.CommandLineException: Unrecognized instruction set avx-512f
     at System.CommandLine.Helpers.ConfigureInstructionSetSupport(String, Int32, Boolean, TargetArchitecture, TargetOS, String, String, Logger, Boolean) + 0x9d5
     at ILCompiler.Program.Run() + 0x59b
     at ILCompiler.ILCompilerRootCommand.<>c__DisplayClass227_0.<.ctor>b__0(ParseResult result) + 0x310
[...]

PR #2566, opened 2024-04-22
Fix names of instructions sets for PublishAot scenarios

MichalStrehovsky:
Ran into this on a machine that supports AVX-512.

I would have expected this path not to be reached (in .NET 8 we added `native` and BDN should be using that when targeting .NET 8), but that's orthogonal.

Cc @adamsitnik


timcassell:
Looks like this fixes #2548.

timcassell (approved):

adamsitnik (approved):
Thanks a lot for the fix @MichalStrehovsky !

PR #2579, opened 2024-06-06
Support required properties

aelij:


adamsitnik:
@aelij some of the tests are failing, PTAL.

aelij:
@adamsitnik Sorry the PR slipped my mind. All tests pass now, I took a different approach.

src/BenchmarkDotNet/Code/CodeGenerator.cs
```diff
@@ -186,7 +187,15 @@ private static DeclarationsProvider GetDeclarationsProvider(Descriptor descripto
             return new NonVoidDeclarationsProvider(descriptor);
         }
 
+        private static string GetParamsInitializer(BenchmarkCase benchmarkCase)
+            => string.Join(
+                ", ",
+                benchmarkCase.Parameters.Items
+                    .Where(parameter => !parameter.IsArgument && !parameter.IsStatic)
+                    .Select(parameter => $"{parameter.Name} = default"));
```
timcassell, line 195: I think it would be better to assign the actual value, that way `init` will also work. You will also need to change the `GetParamsContent` implementation to skip instance values. A rename to `GetInstanceParams` and `GetStaticParams` would be in order as well.

timcassell (changes requested):

src/BenchmarkDotNet/Code/CodeGenerator.cs
aelij: That's what I tried initially (see previous commit) but the problem is that dynamic parameters won't work since they require the instance.
Also `required` support is essential not to get nullable warnings, `init` is a less useful IMO.

timcassell: I can't see your initial commit. Are you referring to ParamsSource?

aelij: I forced-pushed, it's in the PR history:
https://github.com/dotnet/BenchmarkDotNet/pull/2579/commits/469f034adf1722c4069109ac96efc125fc53333a
Yes, ParamsSource

aelij: Sorry it's this one - https://github.com/dotnet/BenchmarkDotNet/commit/32495a953cb473a053bce18a944c738986b05818

aelij: I even named the methods as you suggested :)

timcassell (approved):
Since the "SmartParameter" supports instance methods, it looks like this is the only way to do this. Thanks @aelij!

src/BenchmarkDotNet/Code/CodeGenerator.cs
timcassell: I agree, `init` doesn't seem particularly useful for benchmarks. It just would've been nice to support it "for free" if it was possible.

timcassell:
@aelij I just realized after merging that you didn't add tests for InProcess toolchains. Would you mind doing a follow-up to add those?

aelij:
Sure

timcassell:
InProcessTest and InProcessEmitTest

PR #2594, opened 2024-07-22
fix: Modify logging behavior when explicitly adding NullLogger with custom config

### LINKED ISSUE #190: Feature request: quiet mode

filzrev:
This PR intended to fix problem no console log is outputted when following conditions met.
- Using custom config that adding `NullLogger` explicitly
- Using custom config as `Global Config` (passing config as `BenchmarkSwitcher::Run` parameters)..

Currently BenchmarkDotNet use `ConsoleLogger` to print errors and list of benchmarks if logger is not enabled.
But when explicitly set `NullLogger` it's not works and no console logs are outputted.

**Background**

I'm trying to write custom progress logger that resolve #190 issue.
And it'll be implemented by using custom `EventProcessor` that is added from v0.13.9.

So It need to suppress default `ConsoleLogger` by adding `NullLogger`. 
Because when no logger specified following error is recorded.
> No loggers defined, you will not see any progress!

It works almost as intended.
But when passing custom config to `BenchmarkSwitcher::Run` argument. console log is not outputted.

### DIFF
--- a/src/BenchmarkDotNet/Configs/ConfigExtensions.cs
+++ b/src/BenchmarkDotNet/Configs/ConfigExtensions.cs
@@ -122,7 +122,7 @@ namespace BenchmarkDotNet.Configs
         internal static ILogger GetNonNullCompositeLogger(this IConfig config)
         {
             // if user did not provide any loggers, we use the ConsoleLogger to somehow show the errors to the user
-            if (config == null || !config.GetLoggers().Any())
+            if (config == null || !config.GetLoggers().Any() || config.GetLoggers().All(x => x is NullLogger))
                 return new CompositeLogger(ImmutableHashSet.Create(ConsoleLogger.Default));
 
             return new CompositeLogger(config.GetLoggers().ToImmutableHashSet());


AndreyAkinshin:
@filzrev Sorry, but I do not understand the original problem. If we pass a custom config with NullLogger to BenchmarkRunner or BenchmarkSwitcher, it is expected that we will not see the output. What's wrong with this behavior?

filzrev:
> If we pass a custom config with NullLogger to BenchmarkRunner or BenchmarkSwitcher, it is expected that we will not see the output. What's wrong with this behavior?

The logger that is returned from `GetNonNullCompositeLogger` is used for following purpose.
- Print config parse errors
- Print `--list` outputs 

https://github.com/dotnet/BenchmarkDotNet/blob/ca5dfdf106affb0922f5b8002a67272a27d3fd03/src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs#L93-L123

And this logger is not used for running actual benchmarks.

I'm output console logs by using `ConsoleLogger.Default` (via custom `EventProcessor`).
So default console logs should be suppressed by setting `NullLogger`.

But when setting `NullLogger` to global config.
Above type of logs are also suppressed.

filzrev:
Close this PR,
The reason is that the following workaround is available.

1. Add custom logger that delegate logging to `ConsoleLogger.Default`
2. Add custom EventProcessor. and when `OnStartValidationStage` is called. set custom logger to **quiet** mode

PR #2604, opened 2024-08-05
feat(BenchmarkDotNet.TestAdapter): Throw error when you are compiling benchmark project with in not Release mode and with Optimize set to false

workgroupengineering:
Throw error when you are compiling benchmark project with in not `Release` mode and with `Optimize` set to `false`


src/BenchmarkDotNet.TestAdapter/build/BenchmarkDotNet.TestAdapter.targets
```diff
@@ -0,0 +1,24 @@
...
+
+    <!--Begin workaround -->
+    <MakeDir Directories="$(IntermediateOutputPath)BDNTA"/>
+    <WriteLinesToFile File="$(IntermediateOutputPath)BDNTA\build.stamp"
+        Lines="$([System.Guid]::NewGuid().ToString('d'))"
+        Overwrite="true" />
+    <!--End workaround -->
+    
+    <Error Code="BDNTAE00001"
+           Text="You are compiling $(MSBuildProjectName) with in not Release mode and with Optimize set to false."
```
Saibamen, line 19:
```suggestion
           Text="You are compiling $(MSBuildProjectName) with in not Release mode and/or with Optimize set to false."
```

AndreyAkinshin:
We definitely don't want to fail compilation in Debug. Compilation should always be performed.

At the moment, TestAdapter just doesn't discover benchmarks for the non-Release configuration (except for in-process benchmarks). If we want to enhance the experience for in-process benchmarks or support discovery in the non-Release mode, we should reuse the existing validation checks and print the validation message in the test output.

PR #2613, opened 2024-08-12
Fix known high severity vulnerabilities

adamsitnik:
I've tried to build and run our samples today and got following errors:

```log
PS D:\projects\BenchmarkDotNet\samples\BenchmarkDotNet.Samples> dotnet run -c Release -f net8.0 --filter *Counters* --list flat
C:\Program Files\dotnet\sdk\9.0.100-preview.7.24402.8\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(187,5): warning NETSDK1215: Targeting .NET Standard prior to 2.0 is no longer recommended. See https://aka.ms/dotnet/dotnet-standard-guidance for more details.
D:\projects\BenchmarkDotNet\src\BenchmarkDotNet.Annotations\BenchmarkDotNet.Annotations.csproj : error NU1903: Warning As Error: Package 'System.Text.RegularExpressions' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-cmhx-cq75-c4mj
D:\projects\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj : error NU1903: Warning As Error: Package 'System.Net.Http' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-7jgj-8wvc-jh57
D:\projects\BenchmarkDotNet\samples\BenchmarkDotNet.Samples\BenchmarkDotNet.Samples.csproj : error NU1903: Warning As Error: Package 'System.Text.RegularExpressions' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-cmhx-cq75-c4mj
D:\projects\BenchmarkDotNet\src\BenchmarkDotNet.Diagnostics.dotMemory\BenchmarkDotNet.Diagnostics.dotMemory.csproj : error NU1903: Warning As Error: Package 'System.Net.Http' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-7jgj-8wvc-jh57
D:\projects\BenchmarkDotNet\src\BenchmarkDotNet.Diagnostics.dotTrace\BenchmarkDotNet.Diagnostics.dotTrace.csproj : error NU1903: Warning As Error: Package 'System.Net.Http' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-7jgj-8wvc-jh57
D:\projects\BenchmarkDotNet\src\BenchmarkDotNet.Diagnostics.dotMemory\BenchmarkDotNet.Diagnostics.dotMemory.csproj : error NU1903: Warning As Error: Package 'System.Text.RegularExpressions' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-cmhx-cq75-c4mj
D:\projects\BenchmarkDotNet\src\BenchmarkDotNet.Diagnostics.dotTrace\BenchmarkDotNet.Diagnostics.dotTrace.csproj : error NU1903: Warning As Error: Package 'System.Text.RegularExpressions' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-cmhx-cq75-c4mj
```

My first thought was that the fix will be straightforward: just update `System.Text.RegularExpressions` and `System.Net.Http` to most recent versions. I've quickly realized that these are transitive dependencies.

In case of `BenchmarkDotNet.Annotations`, this project targets `netstandard1.0` (it's just a project with attributes, almost no logic at all and we wanted to target lowest tfm possible). This gives us a dependency to https://www.nuget.org/packages/NETStandard.Library/1.6.1. Can we just update it? No, because it has not been updated since 2018. The following warning suggests that it's on purpose:

> warning NETSDK1215: Targeting .NET Standard prior to 2.0 is no longer recommended.

I could just remove the `netstandard1.0` TFM from `BenchmarkDotNet.Annotations`, but `BenchmarkDotNet.Diagnostics.dotTrace` and `BenchmarkDotNet.Diagnostics.dotMemory` both depend on https://www.nuget.org/packages/JetBrains.Profiler.SelfApi/, which depends on https://www.nuget.org/packages/JetBrains.HabitatDetector/ which depends on https://www.nuget.org/packages/JetBrains.FormatRipper/ which has the same dependency:

image

As a quick workaround I've decided to just add a dependency to these two packages (`System.Text.RegularExpressions` and `System.Net.Http`) to `BenchmarkDotNet.Annotations`, which all BDN packages depend on.

The alternatives I've considered:
- Open an issue and send a PR to https://www.nuget.org/packages/JetBrains.FormatRipper/ to add these explicit dependencies to `System.Text.RegularExpressions` and `System.Net.Http`.
- Open an issue and send a PR to https://www.nuget.org/packages/JetBrains.Profiler.SelfApi/ to change the supported monikers: from `net46` to `net462` (this would pick up the `netstandard2.0` dependency of `JetBrains.HabitatDetector` and solve the problem). But this would be a breaking change (cc @AndreyAkinshin).

cc @ericstj

### DIFF
--- a/samples/BenchmarkDotNet.Samples.FSharp/BenchmarkDotNet.Samples.FSharp.fsproj
+++ b/samples/BenchmarkDotNet.Samples.FSharp/BenchmarkDotNet.Samples.FSharp.fsproj
@@ -24,6 +24,5 @@
     <PackageReference Update="FSharp.Core" Version="4.6.0" />
     <PackageReference Update="System.ValueTuple" Version="4.5.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
-    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
   </ItemGroup>
 </Project>
--- a/samples/BenchmarkDotNet.Samples.FSharp/BenchmarkDotNet.Samples.FSharp.fsproj
+++ b/samples/BenchmarkDotNet.Samples.FSharp/BenchmarkDotNet.Samples.FSharp.fsproj
@@ -13,7 +13,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\BenchmarkDotNet\BenchmarkDotNet.csproj" />
-    <ProjectReference Include="..\..\src\BenchmarkDotNet.TestAdapter\BenchmarkDotNet.TestAdapter.csproj"/>
+    <ProjectReference Include="..\..\src\BenchmarkDotNet.TestAdapter\BenchmarkDotNet.TestAdapter.csproj" />
   </ItemGroup>
   
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
@@ -24,5 +24,6 @@
     <PackageReference Update="FSharp.Core" Version="4.6.0" />
     <PackageReference Update="System.ValueTuple" Version="4.5.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
   </ItemGroup>
 </Project>
--- a/src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
+++ b/src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <AssemblyTitle>Basic BenchmarkDotNet attributes that can be used to annotate your benchmarks</AssemblyTitle>
-    <TargetFrameworks>netstandard1.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <NoWarn>$(NoWarn);1701;1702;1705;1591;3005;NU1702;CA1825</NoWarn>
     <AssemblyName>BenchmarkDotNet.Annotations</AssemblyName>
     <PackageId>BenchmarkDotNet.Annotations</PackageId>
@@ -13,9 +13,5 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="..\BenchmarkDotNet\Helpers\CodeAnnotations.cs" Link="Attributes\CodeAnnotations.cs" />
-  </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
-    <PackageReference Include="System.Net.Http" Version="4.3.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
   </ItemGroup>
 </Project>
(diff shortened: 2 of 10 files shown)


adamsitnik:
>  Open an issue and send a PR to https://www.nuget.org/packages/JetBrains.Profiler.SelfApi/ to change the supported monikers: from `net46` to `net462` (this would pick up the `netstandard2.0` dependency of `JetBrains.HabitatDetector` and solve the problem).

I gave it a try and it has not solved the problem. Considering https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0 the right solution would be to remove `netstandard1.x` support in all the mentioned packages?

Numpsy:
I don't know if it'd help for this situation, but for the

> In case of BenchmarkDotNet.Annotations

case, if you wanted to keep the old .NET Standard 1.x version, I think some of the Serilog libraries used to set
```
<DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```
for  .NET Standard 1.x targets to remove the automatic dependency on ```NETStandard.Library``` and then directly reference the minimum bits that they actually use (and then the references to Http and RegularExpression might go away altogether, rather than needing to be updated)

samples/BenchmarkDotNet.Samples.FSharp/BenchmarkDotNet.Samples.FSharp.fsproj
```diff
@@ -24,5 +24,6 @@
     <PackageReference Update="FSharp.Core" Version="4.6.0" />
     <PackageReference Update="System.ValueTuple" Version="4.5.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
```
ericstj, line 27: You shouldn't do this.  You don't use it.  What brought it in?  Could you update something higher up the stack to avoid bringing it completely?

src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
```diff
@@ -14,4 +14,8 @@
   <ItemGroup>
     <Compile Include="..\BenchmarkDotNet\Helpers\CodeAnnotations.cs" Link="Attributes\CodeAnnotations.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
```
ericstj, line 18: This should also be conditioned on ``$(TargetFramework)' == 'netstandard2.0'`

ericstj (commented):
If you can, get all your dependencies to, at the very least, add new target frameworks that are newer than `netstandard1.x`.  When targeting `netstandard2.0` or newer no framework packages are referenced at all.  Ideally they'd remove the `netstandard1.x` targets as well - but that's not necessary to expose fewer dependencies to consumers.

If you can't get your dependencies to do that, then updating to the latest version of NETStandard.Library may solve the issue without needing to add unused dependencies since the latest version of this package will avoid bringing in the old packages on frameowrks where they aren't needed.

timcassell:
Is there any value in retaining netstandard1.0 target? BenchmarkDotNet itself has a minimum of netstandard2.0, so I don't see what we gain by it.

AndreyAkinshin:
I vote for dropping `netstandard1.0`. This target brings a lot of problems (and I expect more problems in the future), while there is not much value in supporting obsolete runtimes.

Speaking of `JetBrains.FormatRipper`, the `netstandard2.0` target [was added](https://github.com/JetBrains/format-ripper/commit/3c9bec1dce8e6f74ed1696496a72666293721624). Hopefully, it will solve the transitive dependency problems in dotTrace/dotMemory packages (once the new versions of the JetBraind dependencies are available).

timcassell (approved):

PR #2616, opened 2024-08-15
Clean up unsupported native AOT flags

MichalStrehovsky:
* `llcOptimizationPreference` was renamed to `OptimizationPreference` and became supported in .NET 8 (or maybe .NET 7, it doesn't matter)
* `IlcGenerateStackTraceData` is now `StackTraceSupport` and supported. (Also got renamed in .NET 8 or maybe in 7 already.)

I'm leaving the old names of properties for backcompat. We could also delete them. I don't know how much BDN cares about .NET versions that are out of support.

The last one, `IlcGenerateCompleteTypeMetadata` is unsupported, we don't test it, it's a mode for troubleshooting trimming issues for people who ignore trimming warnings. I don't think BDN should be setting this. You'll only find bugs (https://github.com/dotnet/runtime/issues/106439). So deleting that. Not deleting the API because it looks to be a public API.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
+++ b/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
@@ -134,9 +134,10 @@ $@"<?xml version=""1.0"" encoding=""utf-8""?>
     <RunAnalyzers>false</RunAnalyzers>
     <PublishAot Condition="" '$(TargetFramework)' != 'net6.0' "">true</PublishAot>
     <IlcOptimizationPreference>{ilcOptimizationPreference}</IlcOptimizationPreference>
+    <OptimizationPreference>{ilcOptimizationPreference}</OptimizationPreference>
     {GetTrimmingSettings()}
-    <IlcGenerateCompleteTypeMetadata>{ilcGenerateCompleteTypeMetadata}</IlcGenerateCompleteTypeMetadata>
     <IlcGenerateStackTraceData>{ilcGenerateStackTraceData}</IlcGenerateStackTraceData>
+    <StackTraceSupport>{ilcGenerateStackTraceData}</StackTraceSupport>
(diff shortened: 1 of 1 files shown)


src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -134,9 +134,10 @@ private string GenerateProjectForNuGetBuild(BuildPartition buildPartition, Artif
     <RunAnalyzers>false</RunAnalyzers>
     <PublishAot Condition="" '$(TargetFramework)' != 'net6.0' "">true</PublishAot>
```
am11, line 135: nit:
```suggestion
    <PublishAot Condition=""$([MSBuild]::VersionGreaterThan('$(TargetFrameworkVersion)', '6.0'))"">true</PublishAot>
```

am11:
@MichalStrehovsky, there is another warning due to line 167:

> /Users/am11/.nuget/packages/microsoft.net.illink.tasks/9.0.0-rc.1.24410.5/build/Microsoft.NET.ILLink.targets(220,5): warning : Property 'TrimmerDefaultAction' is deprecated in .NET 7 and higher and will be ignored. Use TrimMode instead.

This patch should fix it:

```diff
-                : "<TrimMode>link</TrimMode><TrimmerDefaultAction>link</TrimmerDefaultAction>";
+                : @"<TrimMode>link</TrimMode>
    <TrimmerDefaultAction Condition=""$([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)', '7.0'))"">link</TrimmerDefaultAction>";
```

MichalStrehovsky:
> This patch should fix it:

We'd also need to condition TrimMode - valid values of TrimMode after .NET 7 are `partial` and `full` (with the latter as the default). I'd honestly just delete the "less than or equal to .NET 6" support for native AOT, those were all various experimental unsupported products the perf of which is not very interesting anymore.

These are however all orthogonal to what this PR is doing.

adamsitnik (approved):
LGTM, thank you @MichalStrehovsky and @am11 !

PR #2622, opened 2024-08-22
Fix deadlock in GetDotNetSdkVersion

Korexio:
If the `dotnet` command writes to standard error this can cause a deadlock in `GetDotNetSdkVersion`.
While `dotnet --version` usually does not write to standard error, this can happen if the dotnet executable is instrumented or if a custom dotnet command is used.
The reason for this is that standard error is redirected but not read by GetDotNetSdkVersion, similar to #1629.
The fix is to simply not redirect standard error as it is not read anyway.


timcassell (approved):

adamsitnik:
@Korexio thank you for providing a very simple fix with a detailed explanation!

Issue #2577, opened 2024-05-27
Processor model is not detected on Azure arm64


EgorBo:
When I run any benchmark on Azure ARM64, it reports
```
BenchmarkDotNet v0.13.12, Ubuntu 22.04.4 LTS (Jammy Jellyfish)
Unknown processor
```
E.g. https://github.com/dotnet/runtime/pull/102705#issuecomment-2132450520

`cat /proc/cpuinfo` (which BDN uses) outputs:
```
$ cat /proc/cpuinfo
processor       : 0
BogoMIPS        : 50.00
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp
CPU implementer : 0x41
CPU architecture: 8
CPU variant     : 0x3
CPU part        : 0xd0c
CPU revision    : 1

processor       : 1
BogoMIPS        : 50.00
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp
CPU implementer : 0x41
CPU architecture: 8
CPU variant     : 0x3
CPU part        : 0xd0c
CPU revision    : 1
...
```
and `lscpu` is:
```
$ lscpu
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 16
  On-line CPU(s) list:  0-15
Vendor ID:              ARM
  Model name:           Neoverse-N1
    Model:              1
    Thread(s) per core: 1
    Core(s) per socket: 16
    Socket(s):          1
    Stepping:           r3p1
    BogoMIPS:           50.00
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp
```

So at least `Neoverse-N1, 16 cores` would be better than `Unknown processor`.

From my understanding, on Azure we have
1) Neoverse-N1 (Ampere Altra)
2) Neoverse-N2 ([Cobalt 100](https://techcommunity.microsoft.com/t5/azure-compute-blog/announcing-the-preview-of-new-azure-vms-based-on-the-azure/ba-p/4146353)) (e.g. [this is how](https://github.com/llvm/llvm-project/pull/77793/files) it's recognized in LLVM)

I'll try to implement it myself if nobody grabs it 🙂

[...]

Issue #2358, opened 2023-07-08
System.MissingMethodException: Method not found: 'System.String Perfolizer.Horology.TimeInterval.ToString(System.Globalization.CultureInfo, System.String, Perfolizer.Common.UnitPresentation)'


jwosty:
BenchmarkDotNet was working in my project until recently. Now, my benchmarks are failing with:

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.MissingMethodException: Method not found: 'System.String Perfolizer.Horology.TimeInterval.ToString(System.Globalization.CultureInfo, System.String, Perfolizer.Common.UnitPresentation)'.
   at BenchmarkDotNet.Reports.Measurement.ToString()
   at BenchmarkDotNet.Engines.Engine.RunIteration(IterationData data)
   at BenchmarkDotNet.Engines.EngineFactory.Jit(Engine engine, Int32 jitIndex, Int32 invokeCount, Int32 unrollFactor)
   at BenchmarkDotNet.Engines.EngineFactory.CreateReadyToRun(EngineParameters engineParameters)
   at BenchmarkDotNet.Autogenerated.Runnable_20.Run(IHost host, String benchmarkName) in /Users/jwostenberg/Code/FSharp.Logf/benchmark/bin/Release/net7.0/07fdf749-2a40-4a4d-a6a9-58c9cdbbb84a/07fdf749-2a40-4a4d-a6a9-58c9cdbbb84a.notcs:line 3618
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
[...]

PR #2626, opened 2024-08-27
Bump Perfolizer + Initial Phd adoption

AndreyAkinshin:
This commit is a first step of the grand API refactoring.

* Cpu Detection Upgrade
  * All CPU-related detection logic moved from RuntimeInformation to CpuDetector
  * BrandString formatting moved to Perfolizer (see CpuBrandHelper)
  * New serializable class for Cpu information: PhdCpu from Perfolizer
* Os Detection Upgrade
  * All OS-related detection logic moved from RuntimeInformation to OsDetector
  * BrandString formatting moved to Perfolizer (see OsBrandHelper)
  * New serializable class for Os information: PhdOs from Perfolizer
* Initial adoption of Phd (Performance History Data)
  * Phd from Perfolizer is our future way to keep and process gathered data and measurements
  * Only the initial implementation is available, it still misses a lot of features
  * Examples of Phd json and corresponding new Summary Tables are in BenchmarkDotNet/tests/BenchmarkDotNet.Tests/Phd/VerifiedFiles/Phd.PhdTableTest*
  * Phd can be tried via [PhdExporter]
  * Documentation is coming once we have a more complete implementation


src/BenchmarkDotNet/BenchmarkDotNet.csproj
```diff
@@ -13,14 +13,13 @@
...
     <EmbeddedResource Include="Templates\*" Exclude="bin\**;obj\**;**\*.xproj;packages\**;@(EmbeddedResource)" />
-    <EmbeddedResource Include="Environments\microarchitectures.txt" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.9.1" />
     <PackageReference Include="Gee.External.Capstone" Version="2.3.0" />
     <PackageReference Include="Iced" Version="1.17.0" />
     <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="2.2.332302" />
-    <PackageReference Include="Perfolizer" Version="[0.3.17]" />
+    <PackageReference Include="Perfolizer" Version="[0.4.0]" />
```
adamsitnik, line 22: Just for my education: why is the version `[0.4.0]` rather than just `0.4.0`?

src/BenchmarkDotNet/Detectors/Cpu/ICpuDetector.cs
```diff
@@ -0,0 +1,12 @@
+using Perfolizer.Phd.Dto;
+
+namespace BenchmarkDotNet.Detectors.Cpu;
+
+/// <summary>
+/// Loads the <see cref="PhdCpu"/> for the current hardware
+/// </summary>
+public interface ICpuDetector
```
adamsitnik, line 8: does this interface need to be public? can it be internal?

src/BenchmarkDotNet/Detectors/CpuDetector.cs
```diff
@@ -0,0 +1,29 @@
...
+using BenchmarkDotNet.Detectors.Cpu;
+using BenchmarkDotNet.Detectors.Cpu.Linux;
+using BenchmarkDotNet.Detectors.Cpu.macOS;
+using BenchmarkDotNet.Detectors.Cpu.Windows;
+using BenchmarkDotNet.Extensions;
+using Perfolizer.Phd.Dto;
+
+namespace BenchmarkDotNet.Detectors;
+
+public class CpuDetector(params ICpuDetector[] detectors) : ICpuDetector
```
adamsitnik, line 12: similar to `ICpuDetector`, does it need to be public?

src/BenchmarkDotNet/Phd/BdnJob.cs
```diff
@@ -0,0 +1,8 @@
+using Perfolizer.Phd.Dto;
+
+namespace BenchmarkDotNet.Phd;
+
+public class BdnJob : PhdJob
```
adamsitnik, line 5: what is the point of having this empty and public class? Also, it's another thing called Job..

adamsitnik (commented):
I've taken a quick look and left some questions.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
timcassell: #2358

src/BenchmarkDotNet/Detectors/Cpu/ICpuDetector.cs
AndreyAkinshin: Eventually, I want to come up with an API design, in which all such features can be easily redefined by the user. Imagine a request like #2577. Right now, if there is a bug in CPU detection and we want to introduce a fix, the verification feedback loop is too long: we should merge the potential fix in master, deploy the nightly version, next ask the user to upgrade their BDN to the latest nightly and check how it works. It would be much more convenient if the user had the ability to play with a potential fix in their local project without a need for updated nightly packages. Also, such a way allows the user to fix the problem on the spot without waiting for the next official BenchmarkDotNet release.

However, I don't have a strict opinion on the exact API. We can change it whatever we want.

src/BenchmarkDotNet/Detectors/CpuDetector.cs
AndreyAkinshin: From time to time, I check if the given version of BenchmarkDotNet is capable of detecting the CPU (or other environment properties) correctly in the given environment. Right now, I use `HostEnvironmentInfo.GetInformation()`, which is not very convenient. Also, I want to make features like CPU detection reusable. We have many cool detection features in the library, but when I need them somewhere else, I have to copy-paste the code and next maintain such a pseudo-fork. It would be much better if I could just reference `BenchmarkDotNet` and call the relevant method. Again, we can discuss how the API should look, but I definitely want to have a public function for CPU detection.

src/BenchmarkDotNet/Phd/BdnJob.cs
AndreyAkinshin: > what is the point of having this empty and public class?

This is a work-in-progress solution, which is not fully ready yet. I've merged it to master mostly because of the Cpu/Os refactoring (it was a long-lasting branch and I was tired of solving merge conflicts). I guess it's a good idea to move it away from master to a feature branch until it's fully ready. **Will do it.**

> Also, it's another thing called Job.

The current design of jobs and configs has several problems. Here are two main problems I want to solve:
1. The current design with `CharacteristicObject` is overcomplicated, it's challenging to work with it. I want to come up with a simpler design. However, if I start refactoring jobs right now, we will get a lot of breaking changes. Instead, I want to build a new solution side-by-side with the existing one preserving the existing solution. When everything is ready, we can switch to the new API.
2. 
3. Current jobs are not deserializable. This means that once we export the benchmark results to files (with job descriptions inside), we can't load them back. But we need this information if we want to work with the results later. Manual recreation of the original jobs (and other config entities) based on the current xml/json exported files is quite challenging and, in some places, just impossible. In the new design, I want to have a set of extremely simple DTO entities, which can be easily serialized/deserialized at any time. My current attempt focuses on the `Phd*` classes (which provide some common ground with other tools) and the `Bdn*` classes (which extend `Phd*` with BenchmarkDotNet-specific properties).

src/BenchmarkDotNet/Detectors/CpuDetector.cs
adamsitnik: FWIW I am using `--info` for that:

https://github.com/dotnet/BenchmarkDotNet/blob/ca5dfdf106affb0922f5b8002a67272a27d3fd03/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs#L155-L156

> We have many cool detection features in the library

I see. In such a case we should consider moving such reusable helpers to a standalone package, so the users don't need to reference BenchmarkDotNet.dll itself as it's quite big.

src/BenchmarkDotNet/Phd/BdnJob.cs
adamsitnik: > it was a long-lasting branch and I was tired of solving merge conflicts

I understand. 

> The current design with CharacteristicObject is overcomplicated, it's challenging to work with it.

I agree. In case you have some idea you can express it as an issue with some pseudocode, I would be happy to provide feedback (I am slowly getting better at API design, this year I even got some APIs approved by the .NET API review with the first attempt).

adamsitnik: BTW I am not sure if we should be using the `Phd` word, as it will most likely be mapped to `PhD` by most of our users ;p

AndreyAkinshin: @adamsitnik @timcassell It's been a while, but I have finally resolved the remaining issues related to this refactoring:

-  All `Phd*` DTO classes have been renamed to `*Info` classes (e.g., `CpuInfo`, `OsInfo`, etc.).
-  The remaining usages of `Phd` that refer to the new format or approach have been renamed to `Perfonar`, a new codename that avoids false associations.
-  All new classes related to this refactoring are marked as internal, so we have made minimal changes to the public API surface.

PR #2641, opened 2024-09-21
Add ConfigOption for disabling the .NET SDK Validator

caaavik-msft:
We are in the process of upgrading the .NET runtime repository to target the .NET 10 TFM, and so when this happens we have some benchmarks which are configured to target .NET 9, but the .NET 10 SDK will be installed. We wanted to add a config option that lets us disable the SDK validator so that we can do this without getting a validation error.

### DIFF
--- a/src/BenchmarkDotNet/Configs/ConfigOptions.cs
+++ b/src/BenchmarkDotNet/Configs/ConfigOptions.cs
@@ -48,7 +48,11 @@ namespace BenchmarkDotNet.Configs
         /// <summary>
         /// Continue the execution if the last run was stopped.
         /// </summary>
-        Resume = 1 << 9
+        Resume = 1 << 9,
+        /// <summary>
+        /// Determines if the .NET SDK Version validator should be entirely turned off
+        /// </summary>
+        DisableDotnetSdkVersionValidator = 1 << 10,
     }
 
     internal static class ConfigOptionsExtensions
--- a/src/BenchmarkDotNet/Validators/DotNetSdkVersionValidator.cs
+++ b/src/BenchmarkDotNet/Validators/DotNetSdkVersionValidator.cs
@@ -1,4 +1,5 @@
-﻿using BenchmarkDotNet.Environments;
+﻿using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Running;
 using System;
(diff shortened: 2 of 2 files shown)


timcassell:
What's the actual error you're getting? I wonder if the validator can be fixed rather than disabled.

caaavik-msft:
In our specific case, we are passing in wasmnet90 but the SDK being installed is .NET 10 SDK. So it was invalid because there is no .NET 9 SDK installed, which is what we want at this point. Longer term we probably want to find a way to handle wasm that doesn't require us to have to specify the .NET version in the runtime.

timcassell:
> In our specific case, we are passing in wasmnet90 but the SDK being installed is .NET 10 SDK. So it was invalid because there is no .NET 9 SDK installed, which is what we want at this point.
> 
> The error is "The required .NET Core SDK version 9.0 for runtime moniker WasmNet90 is not installed."

Is running wasm for net9.0 supported with the net10 sdk?

> Longer term we probably want to find a way to handle wasm that doesn't require us to have to specify the .NET version in the runtime.

Using `wasm` moniker doesn't already work? It looks like the config parser should use the current runtime version for it. And the validator uses the same logic.

https://github.com/dotnet/BenchmarkDotNet/blob/3a2d115ace6dc7a68a3c00a001243e098e3e4229/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs#L550-L551

timcassell:
Oh I see, you aren't using `wasm` moniker because of the `mainJs` branching.

https://github.com/dotnet/BenchmarkDotNet/blob/3a2d115ace6dc7a68a3c00a001243e098e3e4229/src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs#L58

Is that actually still relevant? Can that branch be removed and we just always use `test-main.js`?

lewing:
> > In our specific case, we are passing in wasmnet90 but the SDK being installed is .NET 10 SDK. So it was invalid because there is no .NET 9 SDK installed, which is what we want at this point.
> > The error is "The required .NET Core SDK version 9.0 for runtime moniker WasmNet90 is not installed."
> 
> Is running wasm for net9.0 supported with the net10 sdk?

The .NET SDK can target previous versions, so it is fine to target 9 in a 10.0.xxx SDK, it is a required part of the support matrix.  During the part of the year when the runtime and sdk are being updated to support a new tfm (now) that feature can be critical to resolve what would otherwise be circular dependencies between the runtime and sdk.

timcassell:
> The .NET SDK can target previous versions, so it is fine to target 9 in a 10.0.xxx SDK, it is a required part of the support matrix. During the part of the year when the runtime and sdk are being updated to support a new tfm (now) that feature can be critical to resolve what would otherwise be circular dependencies between the runtime and sdk.

So the Core SDK is backwards compatible just like the Framework SDK. In that case, we do need to update the validator to account for that backwards compatibility.

We can also update the `WasmExecutor` to fix the `wasm` moniker for the `test-main.js` path.

Either or both of those things should make this PR unnecessary, I think.

caaavik-msft:
Will close this PR in favour of an alternative solution

PR #2642, opened 2024-09-22
Add .NET 10 support

am11:
This strongly-typed enums based implementation is not scaling very well; but for now, lets add 10.0 and soon after lets see if we can make it less verbose to ease up future updates a bit: https://github.com/dotnet/runtime/pull/106599#issuecomment-2365265969

cc @adamsitnik, @lewing


src/BenchmarkDotNet.Annotations/Jobs/RuntimeMoniker.cs
```diff
@@ -115,6 +115,11 @@ public enum RuntimeMoniker
         /// </summary>
         Net90,
 
+        /// <summary>
+        /// .NET 10.0
+        /// </summary>
+        Net100,
```
timcassell, line 121: Should we start adding `_` here?
```suggestion
        Net10_0,
```

am11: I think we should really consider moving away from these strong type names. This is too much ceremony to issue a nice message. If the underlying SDK doesn't support the TFM, it issues a very reasonable, comprehensive and googleable error message. For the other use-case, consider string parameter `[SupportedFramework("netX.Y")]` rather than these unscalable abstractions which we can't get rid of even after the TFM is rendered unsupported by rest of the ecosystem.

am11: Added underscore. However, the way SDK (and other parts of ecosystem) is handling these validation is they rely on common implementation in the SDK. If the underlying SDK doesn't support the TFM, it will report the error. Then there is the new concept of ["torn SDK"](https://github.com/dotnet/sdk/blob/main/documentation/general/torn-sdk.md) in .NET 9, which makes the tooling download the required SDK. For NativeAOT, LLVMAOT, Mono runtime etc. there are flags like `--aot` available on command line (e.g. `dotnet new console --aot`).

BDN has gone in other direction with these dedicated monikers, `nativeaot9.0` etc., which IMHO is an unnecessary maintenance headache, one we can easily avoid by using stringly arguments.

am11: Added underscore. However, the way SDK (and other parts of ecosystem) is handling these validations is they rely on common implementation in the SDK. If the underlying SDK doesn't support the TFM, it will report the error. Then there is the new concept of ["torn SDK"](https://github.com/dotnet/sdk/blob/main/documentation/general/torn-sdk.md) in .NET 9, which makes the tooling download the required SDK. For NativeAOT, LLVMAOT, Mono runtime etc. there are flags like `--aot` available on command line (e.g. `dotnet new console --aot`).

BDN has gone in other direction with these dedicated monikers, `nativeaot9.0` etc., which IMHO is an unnecessary maintenance headache, one we can easily avoid by using stringly arguments.

caaavik-msft:
I've tested this PR against a local build from https://github.com/dotnet/runtime/pull/106599 and it fails because it seems that the .NET 10 SDK doesn't actually allow to target .NET 10, rather it only allows targetting .NET 9. To get things to work we do need to make changes to the SDK validator to allow backwards compatibility. However, this change will still be useful to have in once we do have .NET 10 targeting supported

am11:
Yup, it's a chicken-egg situation atm, and we will soon be on alpha1 in runtime repo (based on how it was done last year, around this time). I think we can pass --corerun option in local builds that points to built corerun under `artifacts/bin` or `artifacts/tests/coreclr/<platform>/Tests/Core_Root` to get the ball rolling?

src/BenchmarkDotNet.Annotations/Jobs/RuntimeMoniker.cs
timcassell: I don't disagree. Feel free to open an issue for suggestions on improvements.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
timcassell, line 1: GitHub won't let me add a comment on the actual line... I think the `TryParse` methods needs to be updated to account for the underscore.

```cs
internal static bool TryParse(string runtime, out RuntimeMoniker runtimeMoniker)
{
    int index = runtime.IndexOf('-');
    if (index >= 0)
    {
        runtime = runtime.Substring(0, index);
    }

    // Monikers older than Net 10 don't use any version delimiter, newer monikers use _ delimiter.
    if (Enum.TryParse(runtime.Replace(".", string.Empty), ignoreCase: true, out runtimeMoniker))
    {
        return true;
    }
    return Enum.TryParse(runtime.Replace('.', '_'), ignoreCase: true, out runtimeMoniker);
}
```

I don't have a 10.0 sdk version, so I didn't test.

tests/BenchmarkDotNet.Tests/ConfigParserTests.cs
```diff
@@ -388,6 +388,7 @@ public void NetFrameworkMonikerParsedCorrectly(string tfm)
         [InlineData("net70")]
         [InlineData("net80")]
         [InlineData("net90")]
+        [InlineData("net10_0")]
```
timcassell, line 391:
```suggestion
        [InlineData("net7.0")]
        [InlineData("net8.0")]
        [InlineData("net9.0")]
        [InlineData("net10.0")]
```

am11:
IMHO, the best direction would be to introduce something akin to https://learn.microsoft.com/dotnet/api/system.runtime.versioning.supportedosplatformattribute, e.g. `[SupportedFramework("netX.0")]`, which will scale indefinitely. As mentioned, this hardcoded enum approach is not scaling which was the very reason why `[SupportedPlatform("")]` (relatively new API since net5.0) was made string-based instead of enum-based (which would have been a nightmare to maintain over years to come).

Perhaps this (net10.0 transition) is a good time to consider such a change and deprecate the old enum (before completely removing it after a few releases). Also, the psuedo-TFMs like nativeaot9.0 etc. are also a bit unnatural compared to what the rest of the ecosystem is using: `net9.0`+`--aot` (e.g. `dotnet new console -f net9.0 --aot`) or simply PublishAot=true.

timcassell:
I don't understand how `[SupportedFramework("netX.0")]` would be used. I think you should open an issue with more details rather than discuss it here. And/or create a WIP PR showcasing a new system.

LoopedBard3:
I think now may be a good time to revisit this PR. Is there anything specifically blocking it or just another pass to clean up the suggested changes? We are hitting an issue for our performance wasm testing due to wasmnet100/wasmnet10_0. FYI @caaavik-msft

LoopedBard3:
@timcassell is there anything else that needs to be done before merging this? It seems maybe the TryParse update, but wanted to check if there was anything else.

timcassell:
I think that's the only thing.

timcassell:
@LoopedBard3 Regarding wasm specifically, you could also send a PR to fix the wasm logic so you can use the simple `Wasm` moniker and not need to continually update the moniker every year. https://github.com/dotnet/BenchmarkDotNet/pull/2641#issuecomment-2365334620

LoopedBard3:
I have a PR against the base branch for this PR here: https://github.com/am11/BenchmarkDotNet/pull/1. If @am11 accepts it or makes the changes, we should merge this, otherwise I will make a new PR tomorrow and link it back to this one.

src/BenchmarkDotNet.Diagnostics.dotMemory/DotMemoryDiagnoser.cs
```diff
@@ -81,27 +81,32 @@ internal override bool IsSupported(RuntimeMoniker runtimeMoniker)
...
             case RuntimeMoniker.MonoAOTLLVMNet70:
             case RuntimeMoniker.MonoAOTLLVMNet80:
             case RuntimeMoniker.MonoAOTLLVMNet90:
+            case RuntimeMoniker.MonoAOTLLVMNet10_0:
             case RuntimeMoniker.Mono60:
             case RuntimeMoniker.Mono70:
             case RuntimeMoniker.Mono80:
+            case RuntimeMoniker.Mono10_0:
             case RuntimeMoniker.Mono90:
 #pragma warning disable CS0618 // Type or member is obsolete
```
adamsitnik, line 111: this is just a pedantic nit

```suggestion
            case RuntimeMoniker.Mono90:
            case RuntimeMoniker.Mono10_0:
#pragma warning disable CS0618 // Type or member is obsolete
```

src/BenchmarkDotNet/Toolchains/NativeAot/NativeAotToolchain.cs
```diff
@@ -40,6 +40,14 @@ public class NativeAotToolchain : Toolchain
             .TargetFrameworkMoniker("net9.0")
             .ToToolchain();
 
+        /// <summary>
+        /// compiled as net10.0, targets latest NativeAOT build from the .NET 9 feed: https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet10/nuget/v3/index.json
```
adamsitnik, line 44: minor copy-paste error

```suggestion
        /// compiled as net10.0, targets latest NativeAOT build from the .NET 10 feed: https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet10/nuget/v3/index.json
```

templates/templates/BenchmarkDotNet.BenchmarkProjectTemplate.CSharp/.template.config/template.json
```diff
@@ -34,6 +34,14 @@
       "description": "The target framework for the project.",
       "datatype": "choice",
       "choices": [
+        {
+          "choice": "net10.0",
+          "description": ".NET 10"
```
adamsitnik, line 39: .NET 10 is not even in Preview 1 yet so we don't have to update the templates right now. But on the other hand if we don't do it now, we may simply forget about it. So let's merge it in the current form

adamsitnik (approved):
LGTM. Since .NET Team needs it, I am going to merge it in the current form and cherry pick changes from https://github.com/am11/BenchmarkDotNet/pull/1/files.

Thank you @am11 and others involved!

PR #2645, opened 2024-09-23
Change .NET SDK Validator to account for backwards compatibility

caaavik-msft:
This is an alternative approach that was discussed in #2641 to account for backwards compatibility in the SDK (e.g. that you can target previous versions from a later version of the SDK). I have verified locally on my machine that with this BDN installed, that we can run the wasm benchmarks with `--runtimes wasmnet90` passed in while having a .NET 10 SDK installed via the changes in https://github.com/dotnet/runtime/pull/106599 so merging this and using it in the dotnet/performance repository should get everything unblocked.

### DIFF
--- a/src/BenchmarkDotNet/Validators/DotNetSdkVersionValidator.cs
+++ b/src/BenchmarkDotNet/Validators/DotNetSdkVersionValidator.cs
@@ -21,26 +21,20 @@ namespace BenchmarkDotNet.Validators
             {
                 yield return cliPathError;
             }
-            else if (TryGetSdkVersion(benchmark, out string requiredSdkVersion))
+            else if (TryGetSdkVersion(benchmark, out Version requiredSdkVersion))
             {
                 var installedSdks = GetInstalledDotNetSdks(customDotNetCliPath);
-                if (!installedSdks.Any(sdk => sdk.StartsWith(requiredSdkVersion)))
+                if (!installedSdks.Any(sdk => sdk >= requiredSdkVersion))
                 {
-                    yield return new ValidationError(true, $"The required .NET Core SDK version {requiredSdkVersion} for runtime moniker {benchmark.Job.Environment.Runtime.RuntimeMoniker} is not installed.", benchmark);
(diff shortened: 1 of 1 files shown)


caaavik-msft:
Pinging @lewing @am11 @timcassell as they were involved in the previous discussion

am11:
LGTM, whatever unblocks https://github.com/dotnet/performance/pull/4470 at this point and get the .NET 10 builds rolling. 😅

lewing (approved):

src/BenchmarkDotNet/Validators/DotNetSdkVersionValidator.cs
```diff
@@ -113,17 +107,38 @@ private static IEnumerable<string> GetInstalledDotNetSdks(string? customDotNetCl
...
+                            if (secondDot < 0)
+                                continue;
+
+                            if (Version.TryParse(line.Substring(0, secondDot), out var version))
+                            {
+                                versions.Add(version);
+                            }
+                        }
+
+                        return versions;
```
am11, line 131: Do need to account for framework versions too? i.e.
```suggestion
                        return lines.Select(line => Version.Parse(line.Split(' ')[0]));
```

timcassell (approved):

src/BenchmarkDotNet/Validators/DotNetSdkVersionValidator.cs
timcassell: Framework sdk version are retrieved in a separate method.

am11: There are also few others: netcoreapp3.1 and netstandrd2.1 etc. (not sure if tests are exercising them, but we do have tests for those TFMs).

timcassell: This code is comparing major and minor version, so netcoreapp3.1 should work fine.
BDN doesn't support netstandard builds (it must be executable).

PR #2646, opened 2024-09-24
Get full parsable version part in .NET SDK Validator

caaavik-msft:
Unfortunately there was a bug introduced in https://github.com/dotnet/BenchmarkDotNet/pull/2645 that I missed which this PR will fix. Essentially if you use `--runtimes monoaotllvm` or `--runtimes wasm`, it would actually return `new Version(9, 0, 0)` because it defers to `CoreRuntime.TryGetVersion()` to get the version. The code that parsed the output of `dotnet --list-sdks` was only looking at the major and minor versions, and so the comparison of `new Version(9, 0, 0) <= new Version(9, 0)` was returning false. With this change we will read the full version string including the feature band/patch version.

In the dotnet/runtime performance tests we don't use `--runtimes wasm` but we do use `--runtimes monoaotllvm` to run the Mono AOT benchmarks, so this will fix that one. I'm doing some more testing locally but wanted to get the PR out now so it can get some reviews and get it merged quickly.

### DIFF
--- a/src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs
+++ b/src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs
@@ -194,7 +194,7 @@ namespace BenchmarkDotNet.Environments
         }
 
         // Version.TryParse does not handle thing like 3.0.0-WORD
-        private static string GetParsableVersionPart(string fullVersionName) => new string(fullVersionName.TakeWhile(c => char.IsDigit(c) || c == '.').ToArray());
+        internal static string GetParsableVersionPart(string fullVersionName) => new string(fullVersionName.TakeWhile(c => char.IsDigit(c) || c == '.').ToArray());
 
         private static CoreRuntime GetPlatformSpecific(CoreRuntime fallback)
         {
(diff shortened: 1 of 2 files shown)


caaavik-msft:
@timcassell @am11 @lewing

timcassell:
Out of curiosity, how did you run into the situation of `(9, 0, 0)` vs `(9, 0)`? From the code in the previous PR, it looks like you were only parsing and comparing major and minor.

caaavik-msft:
> Out of curiosity, how did you run into the situation of `(9, 0, 0)` vs `(9, 0)`? From the code in the previous PR, it looks like you were only parsing and comparing major and minor.

[This line](https://github.com/dotnet/BenchmarkDotNet/blob/5fe0c78539fe5664477649e6e3adc7cb4e97df5f/src/BenchmarkDotNet/Validators/DotNetSdkVersionValidator.cs#L242) was returning `(9, 0, 0)` and the `GetInstalledDotNetSdks` was returning `(9, 0)`.

caaavik-msft:
Was able to reproduce the version error on my local machine and confirmed that this PR makes it all work correctly. Also just confirmed that `--wasmnet90` still works with .NET 10 SDK installed and that is still working. So we should be all good to merge if it looks good.

timcassell (approved):

lewing (approved):

lewing:
merge away :)

PR #2658, opened 2024-10-23
Auto-hide feature added for columns via Exception and Threading diagnoser configuration

### LINKED ISSUE #2629: Feature Request: Add ability to automatically hide metric columns if value is not set
When using `ThreadingDiagnoser`/`ExceptionDiagnoser`.
Following metric columns are displayed **always** .

- `Exceptions`
- `Completed Work Items`
- `Lock Contentions`

Is it able to **automatically** hide these columns if all metric values are `0` ?.

It seems to be able to change behaviors by modifying following code.
- [MetricColumn::IsAvailable](https://github.com/dotnet/BenchmarkDotNet/blob/ca5dfdf106affb0922f5b8002a67272a27d3fd03/src/BenchmarkDotNet/Columns/MetricColumn.cs#L29-L31)
- [IMetricDescriptor::GetIsAvailable](https://github.com/dotnet/BenchmarkDotNet/blob/ca5dfdf106affb0922f5b8002a67272a27d3fd03/src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs#L65)

**Excample benchmark code**

```csharp
 [ShortRunJob]
 [ThreadingDiagnoser]
 // [HideColumns(Column.CompletedWorkItems, Column.LockContentions, Column.Exceptions)]
 public class Benchmarks
 {
     [Benchmark]
     public void Benchmark()
     {
         Thread.Sleep(1);
     }
 }
```

**Console Output**

```markdown
| Method    | Mean     | Error    | StdDev   | Exceptions | Completed Work Items | Lock Contentions |
|---------- |---------:|---------:|---------:|-----------:|---------------------:|-----------------:|
| Benchmark | 15.57 ms | 0.364 ms | 0.020 ms |          - |                    - |                - |
```

Unreal11:
Closes #2629

### DIFF
--- a/samples/BenchmarkDotNet.Samples/IntroExceptionDiagnoser.cs
+++ b/samples/BenchmarkDotNet.Samples/IntroExceptionDiagnoser.cs
@@ -6,20 +6,10 @@ namespace BenchmarkDotNet.Samples
     [ExceptionDiagnoser]
     public class IntroExceptionDiagnoser
     {
-        [Benchmark]
-        public void ThrowExceptionRandomly()
+        [Benchmark] public void NoThrow() { }
+        [Benchmark] public void Throw()
         {
-            try
-            {
-                if (new Random().Next(0, 5) > 1)
-                {
-                    throw new Exception();
-                }
-            }
-            catch
-            {
-                // ignored
-            }
+            try { throw new Exception(); } catch { }
         }
     }
 }
--- a/src/BenchmarkDotNet/Diagnosers/ExceptionDiagnoser.cs
+++ b/src/BenchmarkDotNet/Diagnosers/ExceptionDiagnoser.cs
@@ -14,7 +14,11 @@ namespace BenchmarkDotNet.Diagnosers
 {
     public class ExceptionDiagnoser : IDiagnoser
     {
-        public static readonly ExceptionDiagnoser Default = new ExceptionDiagnoser();
+        public static readonly ExceptionDiagnoser Default = new ExceptionDiagnoser(new ExceptionDiagnoserConfig(true));
+
+        public ExceptionDiagnoser(ExceptionDiagnoserConfig config) => Config = config;
+
+        public ExceptionDiagnoserConfig Config { get; }
 
         private ExceptionDiagnoser() { }
 
@@ -32,7 +36,10 @@ namespace BenchmarkDotNet.Diagnosers
 
         public IEnumerable<Metric> ProcessResults(DiagnoserResults results)
         {
-            yield return new Metric(ExceptionsFrequencyMetricDescriptor.Instance, results.ExceptionFrequency);
+            if (!Config.DisplayExceptionsIfZeroValue && results.ExceptionFrequency == 0)
+            {
+                yield return new Metric(ExceptionsFrequencyMetricDescriptor.Instance, results.ExceptionFrequency);
+            }
         }
 
         public IEnumerable<ValidationError> Validate(ValidationParameters validationParameters) => Enumerable.Empty<ValidationError>();
--- a/src/BenchmarkDotNet/Diagnosers/ExceptionDiagnoser.cs
+++ b/src/BenchmarkDotNet/Diagnosers/ExceptionDiagnoser.cs
@@ -20,8 +20,6 @@ namespace BenchmarkDotNet.Diagnosers
 
         public ExceptionDiagnoserConfig Config { get; }
 
-        private ExceptionDiagnoser() { }
-
         public IEnumerable<string> Ids => new[] { nameof(ExceptionDiagnoser) };
 
         public IEnumerable<IExporter> Exporters => Array.Empty<IExporter>();
@@ -36,7 +34,7 @@ namespace BenchmarkDotNet.Diagnosers
 
         public IEnumerable<Metric> ProcessResults(DiagnoserResults results)
         {
-            if (!Config.DisplayExceptionsIfZeroValue && results.ExceptionFrequency == 0)
+            if (Config.DisplayExceptionsIfZeroValue || results.ExceptionFrequency != 0)
             {
                 yield return new Metric(ExceptionsFrequencyMetricDescriptor.Instance, results.ExceptionFrequency);
             }
--- a/src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs
+++ b/src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs
@@ -15,7 +15,11 @@ namespace BenchmarkDotNet.Diagnosers
 {
     public class ThreadingDiagnoser : IDiagnoser
     {
-        public static readonly ThreadingDiagnoser Default = new ThreadingDiagnoser();
+        public static readonly ThreadingDiagnoser Default = new ThreadingDiagnoser(new ThreadingDiagnoserConfig(true, true));
+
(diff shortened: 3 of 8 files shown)


Unreal11:
dotnet-policy-service agree

Unreal11:
@dotnet-policy-service agree

src/BenchmarkDotNet/Diagnosers/ExceptionDiagnoser.cs
```diff
@@ -14,7 +14,11 @@ namespace BenchmarkDotNet.Diagnosers
...
     public class ExceptionDiagnoser : IDiagnoser
     {
-        public static readonly ExceptionDiagnoser Default = new ExceptionDiagnoser();
+        public static readonly ExceptionDiagnoser Default = new ExceptionDiagnoser(new ExceptionDiagnoserConfig(true));
+
+        public ExceptionDiagnoser(ExceptionDiagnoserConfig config) => Config = config;
+
+        public ExceptionDiagnoserConfig Config { get; }
 
         private ExceptionDiagnoser() { }
```
timcassell, line 23:
```suggestion
```

src/BenchmarkDotNet/Diagnosers/ExceptionDiagnoser.cs
```diff
@@ -32,7 +36,10 @@ public void Handle(HostSignal signal, DiagnoserActionParameters parameters) { }
 
         public IEnumerable<Metric> ProcessResults(DiagnoserResults results)
         {
-            yield return new Metric(ExceptionsFrequencyMetricDescriptor.Instance, results.ExceptionFrequency);
+            if (!Config.DisplayExceptionsIfZeroValue && results.ExceptionFrequency == 0)
```
timcassell, line 39:
```suggestion
            if (Config.DisplayExceptionsIfZeroValue || results.ExceptionFrequency != 0)
```

src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs
```diff
@@ -15,7 +15,11 @@ namespace BenchmarkDotNet.Diagnosers
...
     public class ThreadingDiagnoser : IDiagnoser
     {
-        public static readonly ThreadingDiagnoser Default = new ThreadingDiagnoser();
+        public static readonly ThreadingDiagnoser Default = new ThreadingDiagnoser(new ThreadingDiagnoserConfig(true, true));
+
+        public ThreadingDiagnoser(ThreadingDiagnoserConfig config) => Config = config;
+
+        public ThreadingDiagnoserConfig Config { get; }
 
         private ThreadingDiagnoser() { }
```
timcassell, line 24:
```suggestion
```

src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs
```diff
@@ -33,8 +37,15 @@ public void Handle(HostSignal signal, DiagnoserActionParameters parameters) { }
 
         public IEnumerable<Metric> ProcessResults(DiagnoserResults results)
         {
-            yield return new Metric(CompletedWorkItemCountMetricDescriptor.Instance, results.ThreadingStats.CompletedWorkItemCount / (double)results.ThreadingStats.TotalOperations);
-            yield return new Metric(LockContentionCountMetricDescriptor.Instance, results.ThreadingStats.LockContentionCount / (double)results.ThreadingStats.TotalOperations);
+            if (!Config.DisplayWorkItemsColumnIfZeroValue && results.ThreadingStats.CompletedWorkItemCount == 0)
```
timcassell, line 40:
```suggestion
            if (Config.DisplayWorkItemsColumnIfZeroValue || results.ThreadingStats.CompletedWorkItemCount != 0)
```

src/BenchmarkDotNet/Diagnosers/ThreadingDiagnoser.cs
```diff
@@ -33,8 +37,15 @@ public void Handle(HostSignal signal, DiagnoserActionParameters parameters) { }
...
+            {
+                yield return new Metric(CompletedWorkItemCountMetricDescriptor.Instance, results.ThreadingStats.CompletedWorkItemCount / (double)results.ThreadingStats.TotalOperations);
+            }
+
+            if (!Config.DisplayLockContentionsColumnIfZeroValue && results.ThreadingStats.LockContentionCount == 0)
```
timcassell, line 45:
```suggestion
            if (Config.DisplayLockContentionsColumnIfZeroValue || results.ThreadingStats.LockContentionCount != 0)
```

timcassell:
What if one benchmark has zero value and another has non-zero value?

```cs
[Benchmark] public void Void() { }
[Benchmark] public void Throws() throw new Exception();
```

May need new acceptance tests to validate the table output.

Unreal11:
> What if one benchmark has zero value and another has non-zero value?
> 
> ```cs
> [Benchmark] public void NoThrow() { }
> [Benchmark] public void Throw()
> {
>     try { throw new Exception(); } catch { }
> }
> ```
> 
> May need new acceptance tests to validate the table output.

What behavior do we expect in this case? Something like that? Removed other colums for better readability
| Method   | Exceptions |
|--------     | -----------  |
| NoThrow |                - |
| Throw      |       1.0000 |

timcassell:
> What behavior do we expect in this case? Something like that? Removed other colums for better readability
> 
> Method	Exceptions
> NoThrow	-
> Throw	1.0000

Yes, that looks like what I would expect. I think with the current code it will display `?` instead of `-` if you configure it to not display zero.

samples/BenchmarkDotNet.Samples/IntroExceptionDiagnoser.cs
timcassell, line 1: I don't think the sample needs to change.

Unreal11: That's draft, sorry) Just tested new output

Unreal11:
> > What behavior do we expect in this case? Something like that? Removed other colums for better readability
> > Method	Exceptions
> > NoThrow	-
> > Throw	1.0000
> 
> Yes, that looks like what I would expect. I think with the current code it will display `?` instead of `-` if you configure it to not display zero.

At the moment it displays 'NA' instead of '-' if its configured to not display zero. Is it acceptable? And is there any chance you can specify examples of tests for validation table output?

case displayExceptions = true(default behavior):

| Method  | Mean          | Error     | StdDev    | Median        | Exceptions |
|-------- |--------------:|----------:|----------:|--------------:|-----------:|
| NoThrow |     0.0002 ns | 0.0004 ns | 0.0003 ns |     0.0000 ns |          - |
| Throw   | 2,260.6319 ns | 5.4942 ns | 4.8705 ns | 2,262.6339 ns |     1.0000 |

case displayExceptions = false:
| Method  | Mean          | Error     | StdDev    | Median        | Exceptions |
|-------- |--------------:|----------:|----------:|--------------:|-----------:|
| NoThrow |     0.0008 ns | 0.0014 ns | 0.0012 ns |     0.0002 ns |         NA |
| Throw   | 2,282.4955 ns | 7.6377 ns | 7.1443 ns | 2,281.0249 ns |     1.0000 |

timcassell:
Superseded by #2673

PR #2661, opened 2024-10-28
Prevent premature end of the Benchmark process at Ctrl-C, fixes #2483

### LINKED ISSUE #2483: Pressing Ctrl-C does not reset power plan back
Also mentioned here: https://github.com/dotnet/BenchmarkDotNet/issues/2319#issuecomment-1577583183

1. Start benchmark
2. Wait for power plan to change to high performance
3. Press Ctrl-C

Benchmark exits, but the power plan remains set to high performance. Using v0.13.10.

leonvandermeer:
Set ConsoleCancelEventArgs.Cancel to true so that Benchmark process continues and PowerPlan is reverted at end of aborted Benchmark.

Note that at Ctrl-C (or Ctrl-Break) press, current benchmark fails (as all its processes are killed). The [main run loop](../blob/f8e0a5c23883fa6ea4a975cd08a0071c4648472c/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L161) continues running remaining benchmarks.

As it would (imho) require some complicated code to set `stop` to true at Ctrl-C press, I decided to document current behavior.

(If you do like to have all benchmarks aborted at Ctrl-C press, let me know and I'll implement such behavior)

### DIFF
--- a/src/BenchmarkDotNet/Helpers/DisposeAtProcessTermination.cs
+++ b/src/BenchmarkDotNet/Helpers/DisposeAtProcessTermination.cs
@@ -1,0 +1,37 @@
+﻿using System;
+
+namespace BenchmarkDotNet.Helpers
+{
+    /// <summary>
+    /// Ensures that Dispose is called at termination of the Process.
+    /// </summary>
+    public class DisposeAtProcessTermination : IDisposable
+    {
+        public DisposeAtProcessTermination()
+        {
+            Console.CancelKeyPress += OnCancelKeyPress;
+            AppDomain.CurrentDomain.ProcessExit += OnProcessExit;
+            // It does not make sense to include a finalizer. As we are subscribed to static events,
+            // it will never be called.
+        }
+
+        /// <summary>
+        /// Called when the user presses Ctrl-C or Ctrl-Break.
+        /// </summary>
+        private void OnCancelKeyPress(object? sender, ConsoleCancelEventArgs e)
+        {
+            if (!e.Cancel) { Dispose(); }
+        }
+
+        /// <summary>
+        /// Called when the user clicks on the X in the upper right corner to close the Benchmark's Window.
+        /// </summary>
+        private void OnProcessExit(object? sender, EventArgs e) => Dispose();
+
+        public virtual void Dispose()
+        {
+            Console.CancelKeyPress -= OnCancelKeyPress;
+            AppDomain.CurrentDomain.ProcessExit -= OnProcessExit;
+        }
+    }
+}
--- a/src/BenchmarkDotNet/Helpers/DisposeAtProcessTermination.cs
+++ b/src/BenchmarkDotNet/Helpers/DisposeAtProcessTermination.cs
@@ -1,18 +1,38 @@
 ﻿using System;
+using System.Runtime.InteropServices;
 
 namespace BenchmarkDotNet.Helpers
 {
     /// <summary>
-    /// Ensures that Dispose is called at termination of the Process.
+    /// Ensures that explicit Dispose is called at termination of the Process.
     /// </summary>
-    public class DisposeAtProcessTermination : IDisposable
+    /// <remarks>
+    /// <para>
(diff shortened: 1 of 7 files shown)


timcassell:
I don't think it's desirable for ctrl-c to abort only a single benchmark. I use it frequently to abort a benchmark run altogether.

leonvandermeer:
Tim, I first tried to implement a kind of 'ConsoleControl' that ensures the process is not terminated (by setting [ConsoleCancelEventArgs.Cancel](https://source.dot.net/System.Console/R/db8b1199a72b1825.html) to `true` from `Console.CancelKeyPress` event handler. And then bail out from the [benchmark runner loop](https://github.com/dotnet/BenchmarkDotNet/blob/346bbab62a508fbce8179965ba05452e7a361367/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L206). More or less similar to the CancellationToken that's passed to `ExecuteAsync` in a .NET Generic Host worker app.
However, that would not work during compilation stage. It also did not work for the inprocess toolchain.

While browsing BDN's source code for what to do, I discovered that some other classes already reverted system state at process termination. Then I choose to follow that pattern for both the power settings and the console title.

timcassell:
I like your solution. One thing that I wonder is whether `DisposeAtProcessTermination` should inherit from `CriticalFinalizerObject` and implement the full disposable pattern.

src/BenchmarkDotNet/Helpers/DisposeAtProcessTermination.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+
+namespace BenchmarkDotNet.Helpers
+{
+    /// <summary>
+    /// Ensures that Dispose is called at termination of the Process.
+    /// </summary>
+    public class DisposeAtProcessTermination : IDisposable
```
timcassell, line 8:
```suggestion
    internal abstract class DisposeAtProcessTermination : IDisposable
```

leonvandermeer: I hope you are OK with the class being public:

..\src\BenchmarkDotNet.Diagnostics.Windows\EtwDiagnoser.cs(19,27,19,39): error CS0060: Inconsistent accessibility: base class 'DisposeAtProcessTermination' is less accessible than class 'EtwDiagnoser\<TStats>'

timcassell: Oh I missed that those other classes are public. In that case, it's fine.

leonvandermeer:
> I like your solution. One thing that I wonder is whether `DisposeAtProcessTermination` should inherit from `CriticalFinalizerObject` and implement the full disposable pattern.

I added some xml documentation to explain the design and reason not to implement full disposable pattern. Please read [^1].

Constrained Execution Regions and related classes as `CriticalFinalizerObject` (a good explanation is written [here](https://learn.microsoft.com/en-us/dotnet/framework/performance/constrained-execution-regions#critical-finalizer-object)) are intended for high availability processes like SQL Server that must stay alive. The clr guarantees that your finalizer is called completely and in some more situations [^2].

A finalizer is only needed when a class maintains native resources (e.g. a handle). `DisposeAtProcessTermination` does not and derived classes should not do either directly.
Derived classes shall delegate maintaining and closing of native resources to a more granular level, preferably using SafeHandle derived classes (and delegate explicit disposal to their Idisposable fields).
Note that some system state is pure managed code and does not keep any native resource alive (e.g. set and revert of Console.Title)

[^1]: Note that when BDN process is killed via e.g. task manager, not any finalizer will be called (and not any system state is reverted). But that's a more rare occasion, whereas Ctrl-C to abort is a feature.
[^2]: This is only supported on .Net Framework, not in .Net Core.

timcassell:
> A finalizer is only needed when a class maintains native resources

While not technically native resources, I would argue that system state falls in that category.

timcassell (approved):
Thanks @leonvandermeer!

PR #2665, opened 2024-11-15
Native AOT projects should also copy SettingsWeWantToCopy

keegan-caruso:
See: https://github.com/dotnet/BenchmarkDotNet/issues/2428#issuecomment-2477793449

I'm happy to add tests here, but I didn't see any good examples of similar tests currently present. I ran some of the integration tests and checked the outputs and those matched expectations.

relates to #2428


timcassell (approved):
Thanks @keegan-caruso!

PR #2670, opened 2024-11-24
Implement WakeLock so that the system does not enter sleep

leonvandermeer:
Implement WakeLock so that the system does not enter sleep while benchmarks are running.

As a happy user of BDN, I was annoyed by the fact that long running benchmarks were not completed when I returned to my system. All because my system enters sleep, after not being touched for a while.

The test code needs to enumerate power requests. I decided not to pinvoke [undocumented api](https://github.com/diversenok/Powercfg) as maintenance is difficult if it changes in future windows version. It gave me a chance to learn more about parsers in general 😂. Imo, maintaining the parser when `powercfg /requests` output would change is less difficult.

In case you prefer usage of [Superpower library](https://github.com/datalust/superpower) to parse, I'm willing to change that.

All test code is in integration tests, since `powercfg` needs elevation. I imagine developers dislike failing unit tests when they run those locally.


src/BenchmarkDotNet/Configs/WakeLockType.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace BenchmarkDotNet.Configs
+{
+    public enum WakeLockType
+    {
+        /// <summary>
+        /// Allows the system to enter sleep and/or turn off the display while benchmarks are running.
+        /// </summary>
+        No,
```
timcassell, line 8:
```suggestion
        None,
```

docs/articles/samples/IntroWakeLock.md
```diff
@@ -0,0 +1,32 @@
+---
+uid: BenchmarkDotNet.Samples.IntroWakeLock
+---
+
+## Sample: IntroWakeLock
+
+Running Benchmarks usually takes enough time such that the system enters sleep or turns of the display.
```
timcassell, line 7:
```suggestion
Running benchmarks may sometimes take enough time such that the system enters sleep or turns off the display.
```

src/BenchmarkDotNet/Configs/DefaultConfig.cs
```diff
@@ -87,6 +87,8 @@ public IEnumerable<IValidator> GetValidators()
 
         public TimeSpan BuildTimeout => TimeSpan.FromSeconds(120);
 
+        public WakeLockType WakeLock => WakeLockType.No;
```
timcassell, line 90: Should the default be system?

timcassell (commented):
I appreciate all the tests! Do any other OSes besides Windows need this treatment?

src/BenchmarkDotNet/Running/WakeLock.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace BenchmarkDotNet.Running;
+
+internal partial class WakeLock
+{
+    public static WakeLockType GetWakeLockType(BenchmarkRunInfo[] benchmarkRunInfos) =>
+        benchmarkRunInfos.Select(static i => i.Config.WakeLock).Max();
+
+    public static IDisposable Request(WakeLockType wakeLockType, string reason) =>
+        wakeLockType == WakeLockType.No || !OsDetector.IsWindows() ? null : new WakeLockSentinel(wakeLockType, reason);
```
timcassell, line 15: What is the minimum windows version for this API?

leonvandermeer: Power(Create|Set|Clear)Request are [>= Windows 7;Windows Server 2008 R2 ](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-powercreaterequest#requirements)

CloseHandle is >= Windows 2000 😉

timcassell: Can you add the windows version check like in TaskbarProgress?

src/BenchmarkDotNet/Configs/DefaultConfig.cs
timcassell: Wait, does this require administrator privileges? If so, we can't use it by default. I don't see it called out in the docs, only in the test comments.

leonvandermeer: The feature itself does **not** require admin privileges. Any process can create power requests.

`powercfg.exe /requests` does, so one must run the test code with elevation. That's why I put all test code in integration tests.

docs/articles/guides/console-args.md
```diff
@@ -243,6 +243,7 @@ dotnet run -c Release -- --filter * --runtimes net6.0 net8.0 --statisticalTest 5
 * `--platform` the Platform that should be used. If not specified, the host process platform is used (default). AnyCpu/X86/X64/Arm/Arm64/LoongArch64.
 * `--runOncePerIteration` run the benchmark exactly once per iteration.
 * `--buildTimeout` build timeout in seconds.
+* `--preventSleep` prevents the system from entering sleep or turning off the display. None/RequireSystem/RequireDisplay.
```
timcassell, line 246: I find it a little strange that the code config uses `WakeLock` while the console arg uses `--preventSleep`. Can the console arg also be `--wakelock` or `--wakeLock`?

src/BenchmarkDotNet/Configs/WakeLockType.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        /// <summary>
+        /// Forces the system to be in the working state while benchmarks are running.
+        /// </summary>
+        RequireSystem,
+
+        /// <summary>
+        /// Allows the system to enter sleep and/or turn off the display while benchmarks are running.
+        /// </summary>
+        None,
```
timcassell, line 13: I know we chose to have system be the default, but I think `None` should be the 0 enum value, and these be in their order of precedence.

src/BenchmarkDotNet/Configs/WakeLockType.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace BenchmarkDotNet.Configs
+{
+    public enum WakeLockType
+    {
+        /// <summary>
+        /// Forces the system to be in the working state while benchmarks are running.
+        /// </summary>
+        RequireSystem,
```
timcassell, line 8: Can we drop the "Require"? It reads weird to me. It's already implied by the enum name.
```suggestion
        System,
```

src/BenchmarkDotNet/Configs/WakeLockType.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        /// <summary>
+        /// Allows the system to enter sleep and/or turn off the display while benchmarks are running.
+        /// </summary>
+        None,
+
+        /// <summary>
+        /// Forces the display to be on while benchmarks are running.
+        /// </summary>
+        RequireDisplay
```
timcassell, line 18:
```suggestion
        Display
```

src/BenchmarkDotNet/Configs/WakeLockType.cs
leonvandermeer: System be default means that None must 'beat' system and display must 'beat' none, so that combining the different config values result in the 'highest' value. I was lazy and reordered the enum members. Indeed, making this ordering in ManualConfig.GetWakeLock explicit results in better readable and maintanable code.

timcassell (approved):
LGTM! I would like a second review before merging this new functionality, though. @adamsitnik or @AndreyAkinshin may have different opinions on what the default should be.

leonvandermeer:
Hi @AndreyAkinshin and @adamsitnik . Happy New Year[^1]!

Do you have time to look at this feature? I will be very happy if my contribution makes it into the product.

[^1]: Also applies to all other BDN team members.

tests/BenchmarkDotNet.IntegrationTests/StringExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+internal static class StringExtensions
+{
+    private static readonly char[] newLine = Environment.NewLine.ToCharArray();
+
+    /// <summary>
+    /// Creates an array of strings by splitting this string at each occurrence of a newLine separator.
+    /// Contrary to calling <code>value.Split('\r', '\n')</code>, this method does not return an empty
+    /// string when CR is followed by LF.
+    /// </summary>
+    public static IEnumerable<string> EnumerateLines(this string value)
```
adamsitnik, line 15: Please use `StringReader` and its [ReadLine](https://learn.microsoft.com/en-us/dotnet/api/system.io.stringreader.readline?view=net-9.0) method instead.

adamsitnik (commented):
@leonvandermeer big thanks for your high quality contribution!

Overall the code looks good, but some feedback needs to be addressed before the code gets merged. PTAL at my comments.

src/BenchmarkDotNet/Running/WakeLock.PInvoke.cs
```diff
@@ -0,0 +1,77 @@
...
+        public static SafePowerHandle PowerCreateRequest(string reason)
+        {
+            REASON_CONTEXT context = new REASON_CONTEXT()
+            {
+                Version = POWER_REQUEST_CONTEXT_VERSION,
+                Flags = POWER_REQUEST_CONTEXT_FLAGS.POWER_REQUEST_CONTEXT_SIMPLE_STRING,
+                SimpleReasonString = reason
+            };
+            SafePowerHandle safePowerHandle = PowerCreateRequest(context);
+            if (safePowerHandle.IsInvalid) { throw new Win32Exception(); }
```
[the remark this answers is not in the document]
leonvandermeer: I personally cannot see which use case would cause  `PowerCreateRequest`, `PowerSetRequest` and `PowerClearRequest` to fail except for the situations we already covered (e.g. high enough windows version). This makes the 3 methods to throw only when there's a programming error in BDN's Wakelock code. And for programming errors I always throw.

I can also understand your reasoning to not annoy users and thus will update the code.

I'm very curious if users will see the logging (and in which cases).

tests/BenchmarkDotNet.IntegrationTests/WakeLockTests.cs
```diff
@@ -0,0 +1,185 @@
...
+                ping = EventWaitHandle.OpenExisting(PingEventName),
+                pong = EventWaitHandle.OpenExisting(PongEventName);
+            ping.Set();
+            pong.WaitOne(testTimeout);
+        }
+    }
+
+    private string GetPowerRequests(string? expectedReason = null)
+    {
+        Assert.True(IsAdministrator(), "'powercfg /requests' requires administrator privileges and must be executed from an elevated command prompt.");
```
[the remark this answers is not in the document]
leonvandermeer: Good idea, thanks.

While implementing, I also wanted the new enum to work for non-windows systems, in case future BDN tests needs root / sudo on unix. Browsing for how to check this on unix, I stumbled accross `System.Environment.IsPrivilegedProcess` property [^1], added in net8.0. You'll see that appear in the rework.

With usage of the code from `System.Security.Principal` for net462, I was able to implement it for all frameworks / platforms 😃 .

[^1]: which inspired me to name the new enum member `IsPrivilegedProcess`.

src/BenchmarkDotNet/Configs/DefaultConfig.cs
```diff
@@ -87,6 +87,8 @@ public IEnumerable<IValidator> GetValidators()
 
         public TimeSpan BuildTimeout => TimeSpan.FromSeconds(120);
 
+        public WakeLockType WakeLock => WakeLockType.System;
```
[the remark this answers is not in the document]
leonvandermeer: When you adjust power and sleep settings [^1] to not go to sleep and always leave the screen on, then:
* While BDN is running, both you (via the settings) and BDN (via the API) want the system to be running and screen on.
* When your benchmarks end (after your long time), as far as the finished BDN concerned, the system may enter sleep and the screen may be switched off. However, because of your system settings, the system and screen stay on.

Conclusion: same behavior as without this feature.

[^1] See https://support.microsoft.com/en-us/windows/how-to-adjust-power-and-sleep-settings-in-windows-26f623b5-4fcc-4194-863d-b824e5ea7679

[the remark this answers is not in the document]
timcassell: In other words, it behaves like a counter? When an app requests a wake lock, the system increments a counter, and as long as the counter is > 0, it will respect it? That's what I would expect.

src/BenchmarkDotNet/Running/WakeLock.PInvoke.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        private enum POWER_REQUEST_CONTEXT_FLAGS : uint
+        {
+            POWER_REQUEST_CONTEXT_DETAILED_STRING = 2U,
+            POWER_REQUEST_CONTEXT_SIMPLE_STRING = 1U,
+        }
+
+        public enum POWER_REQUEST_TYPE
+        {
+            PowerRequestDisplayRequired = 0,
```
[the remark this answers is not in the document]
leonvandermeer: Reading microsoft's documentation on this topic, it seems that DisplayRequired implies SystemRequired. However, they leave implementation of this to the caller of the API. I know of undocumented enum members [^1] and guess that some combinations are not allowed. Thus not numbering 1,2,4,... and prevents developers from combining [^2]. But indeed a bit awkward API. Maybe there's also some history involved, as system sleep is quite complicated.

[^1]: https://github.com/diversenok/Powercfg/blob/master/Readme.md#request-types
[^2]: When calling PowerSetRequest multiple times with invalid combinations, I can imagine that PowerSetRequest returns false on the 2nd call. That would be a programming error in which case my pinvoke throws correctly Win32Exception 😉.

src/BenchmarkDotNet/Configs/DefaultConfig.cs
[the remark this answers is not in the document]
leonvandermeer: That's correct. Quoting microsoft's documentation: [Increments the count of power requests of the specified type for a power request object.](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-powersetrequest)

leonvandermeer:
FYI, I'm finished with this rework round. Please take next steps to merge the feature.

timcassell:
It looks to me like all the feedback was addressed, so I will go ahead and merge.

I actually recently ran into this issue on a MacBook, so hopefully someone can extend this to other OSes in the future.

Thanks @leonvandermeer!

src/BenchmarkDotNet/Running/WakeLock.cs
```diff
@@ -0,0 +1,69 @@
...
+                    {
+                        PInvoke.PowerClearRequest(safePowerHandle, PInvoke.POWER_REQUEST_TYPE.PowerRequestDisplayRequired);
+                    }
+                    PInvoke.PowerClearRequest(safePowerHandle, PInvoke.POWER_REQUEST_TYPE.PowerRequestSystemRequired);
+                }
+                catch (Win32Exception ex)
+                {
+                    logger.WriteLineError($"Unable to allow the system from entering sleep or turning off the display (error message: {ex.Message}).");
+                }
+                safePowerHandle.Dispose();
```
adamsitnik, line 64: the disposal should be put into a finally block

adamsitnik (commented):
Thank you for addressing my feedback and providing high quality contribution to BenchmarkDotNet @leonvandermeer 👍

PR #2681, opened 2024-12-22
Fix Windows path too long

timcassell:
Found while trying to run manual tests. Introduced in #2423


AndreyAkinshin (approved):
Nit: it would be nice to extract 36 to a constant to avoid duplication. The rest is great, thanks!

AndreyAkinshin (approved):

Issue #1927, opened 2022-02-20
Is there any issue with dotnet <Benchmark>.dll on Linux arm  or arm64


Acruval:
No problem on windows environment. dotnet <Benchmark>.dll 
I check first https://benchmarkdotnet.org/articles/guides/troubleshooting.htmlbefore write

// Validating benchmarks:
// ***** BenchmarkRunner: Start   *****
// ***** Found 2 benchmark(s) in total *****
// ***** Building 1 exe(s) in Parallel: Start   *****
Unable to find .sln or .csproj file. Will use current directory /home/********/Benchmark31 to search for project file. If you don't use .sln file on purpose it should not be a problem.
// ***** Done, took 00:00:00 (0.33 sec)   *****
// Found 2 benchmarks:
//   BenchMatchTalib.RecognizeEx: Job-ERJEFG(Platform=Arm, Runtime=.NET Core 3.1)
//   BenchMatchTalib.Recognize: Job-ERJEFG(Platform=Arm, Runtime=.NET Core 3.1)

// Generate Exception: Unable to find Benchmark in /home/*******/Benchmark31 and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj

// BenchmarkDotNet has failed to build the auto-generated boilerplate code.
// It can be found in /home/********/Benchmark31/eb6f0439-1aab-4077-a60d-1cd1d2d33baa
// Please follow the troubleshooting guide: https://benchmarkdotnet.org/articles/guides/troubleshooting.html
// ***** BenchmarkRunner: Finish  *****

adamsitnik:
Hi @Acruval 

Have you copied the `.dll` from one machine to another?

BenchmarkDotNet generates some code that takes care of common microbenchmark pitfalls, builds it (using .NET SDK) and runs every benchmark in a dedicated process. To build the benchmark it needs the source code.

If for some reason you can't provide the source code or SDK you can use the [InProcessToolchain](https://benchmarkdotnet.org/articles/samples/IntroInProcess.html) that runs the benchmarks in the same process.

Acruval:
Hi Adam Sitnik.
Thanks for your quick reply. I am aware you are working on this in your
sparse time.
If fat I was generating portable core in my windows X64 and deploying it on
ARM64.
Finally, I got thanks for your help.

  public class BenchmarkConfig : ManualConfig
    {
        public BenchmarkConfig()
        {
            var job = new Job
            {
                Environment =
                {
[...]

Issue #1785, opened 2021-08-26
does not work with vb.net


eschneider999:
does not work with vb.net

Unable to find .sln or .csproj file. Will use current directory G:\Dev\FrozenElephant\SymbioticORM\Benchmarks\bin\netcoreapp2.0 to search for project file. If you don't use .sln file on purpose it should not be a problem.

// Generate Exception: Unable to find Benchmarks in G:\Dev\FrozenElephant\SymbioticORM\Benchmarks\bin\netcoreapp2.0 and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj

Issue #1178, opened 2019-06-08
New to BenchmarkDotNet, Need help


AlenToma:
Hi, so im new to BenchmarkDotNet and i really loved it when i saw it youtube. 

So i created a test project with the name Benchmarklb to test my library, But i get a strange result and many exception that i really dont understand why those happen. 

Here is my class that im testing

```csharp 
    [ClrJob(baseline: true), CoreJob, CoreRtJob]
    //[RPlotExporter, RankColumn]
    public class FastDeepCloner_tests
    {
        [Benchmark]
        public void ProxyTest()
        {
            var pUser = DeepCloner.CreateProxyInstance<User>();
            pUser.Name = "testo";
        }



        [Benchmark]
        public void Clone()
        {
            var user = new List<User>();
            user.Add(new User() { Name = "Alen" });

            var cloned = DeepCloner.Clone(user);

        }


        [Benchmark]
        public void DynamicClone()
        {
            dynamic originalAnonnymousObject = new
            {
                prop1 = "p1"
            ,
                prop2 = 10
            };
            dynamic clonedAnonnymousObject = DeepCloner.CloneDynamic(originalAnonnymousObject);

        }

        [Benchmark]
        public void CloneCollection()
        {
            var user = new UserCollections();
            user.Add(new User() { Name = "Alen" });
            user.TestValue = "Alen";
            user.us = new User() { Name = "Toma" };
            var cloned = DeepCloner.Clone(user);

        }
    }
```

And i simple start my application with
```csharp
 static void Main(string[] args)
        {
            var summary = BenchmarkRunner.Run<FastDeepCloner_tests>(/*new MyConfig()*/);
        }
```

the logs im getting
```
// Validating benchmarks:
// ***** BenchmarkRunner: Start   *****
// ***** Found 12 benchmark(s) in total *****
// ***** Building 3 exe(s) in Parallel: Start   *****
// ***** Done, took 00:00:00 (0.55 sec)   *****
// Found 12 benchmarks:
//   FastDeepCloner_tests.ProxyTest: Clr(Runtime=Clr)
//   FastDeepCloner_tests.Clone: Clr(Runtime=Clr)
//   FastDeepCloner_tests.DynamicClone: Clr(Runtime=Clr)
//   FastDeepCloner_tests.CloneCollection: Clr(Runtime=Clr)
//   FastDeepCloner_tests.ProxyTest: Core(Runtime=Core)
[...]

Issue #1019, opened 2019-01-16
Generate Exception: Unable to find <Project> in <Folder> and its subfolders


ilyaext:
I have a project Tests.Performance.csproj DotNet Core 2.2, with BenchmarkDotNet 0.11.3
The output Tests.Performance.dll is in sub-folder: \bin\Release. The program runs fine when the solution file is in this folder. When the solution is in another folder I get the following exceptions:

Unable to find .sln file. Will use current directory D:\S6\pc\src\Tests\Tests.Performance\bin\Release to search for project file. If you don't use .sln file on purpose it should not be a problem.

// Generate Exception: Unable to find Tests.Performance in D:\SW\pc\src\Tests.Performance\bin\Release and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj

Could you please help with it?

ilyaext:
Would also be very helpfully if I could add a path to project by configuration. In this case I will not be required to have the same assembly name as the project name. It would probably solve and the problem above

adamsitnik:
When we generate the boilerplate code we need to reference the project that actually defines the benchmarks.

What we do as of today is that we walk the folders down until we find a folder with `.sln` file. From there, we search for `$dllName.csproj` including subfolders. 

> Would also be very helpfully if I could add a path to project by configuration. In this case I will not be required to have the same assembly name as the project name. 

I think that this is too rare edge case to extend our configuration with another option. The error message is clear, this is our requirement to make it work.

If you have a better idea for how we can search for the project file, please let me know.

PR #2684, opened 2025-01-01
Add support for user-supplied project file detection

Genbox:
This update adds support for "Locators" in the IConfig interface. A Locator is an extensibility point where users can define path-handling logic that differs from the built-in logic.

For example, if the user sets `AssemblyName` in their csproj, they are forced to use the InProcess job, or else BenchmarkDotnet gives an error with:
> Generate Exception: System.NotSupportedException: Unable to find <csproj> in <folder> and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj

The error is caused by the built-in logic trying to find a project file relative to the project root, which has the same name as the DLL. Unfortunately, [lots of projects](https://github.com/search?q=lang%3Axml+AssemblyName+&type=code) use AssemblyName to generate a short name, different from the project name (Like GNU utilities) or prepend the owners name (like GitHub/Nuget prefixes) to avoid assembly name collisions.

The issue has been repeatedly raised on the BenchmarkDotnet issue tracker and StackOverlow. For issues, see #1019, #498, #1178, #1785 and #1927.

### The code
ILocator is a new public interface where users can define their logic. The interface is implemented like this:

```csharp
private class CustomLocator : ILocator
{
    public LocatorType LocatorType => LocatorType.ProjectFile;

    public FileInfo Locate(DirectoryInfo rootDirectory, Type type)
    {
        return new FileInfo(Path.Combine(rootDirectory.FullName, @"Src\Benchmarks\MyBenchmarks.csproj"));
    }
}
```
The LocatorType is not strictly necessary, as the locator is only used in a single place right now. However, if there are other places where the user could benefit from the extensibility, it is necessary to have to distinguish between locators.

If that is not wanted, let me know, and I'll remove it and rename ILocator to IProjectLocator to specialize the code for its intended purpose.

Users can add their locator to a ManualConfig like so:

```csharp
// To use both ProjectLocator.Default and custom locator:
IConfig config = ManualConfig.CreateMinimumViable()
                           .AddLocator(new CustomLocator());

// To use only the custom locator:
IConfig config = ManualConfig.CreateEmpty()
                           .AddLocator(new CustomLocator());
```

### Notes
- ProjectLocator is a new class that contains the old logic to find the project file.
- DefaultConfig uses ProjectLocator.Default
- DebugConfig also uses ProjectLocator.Default
- ManualConfig defaults to no ILocator, but `CreateMinimumViable` includes ProjectLocator.Default
- I've added tests to ensure 1) duplicate locators are supported 2) that configs contains the correct locators

### DIFF
--- a/src/BenchmarkDotNet/Locators/ILocator.cs
+++ b/src/BenchmarkDotNet/Locators/ILocator.cs
@@ -1,0 +1,10 @@
+using System;
+using System.IO;
+
+namespace BenchmarkDotNet.Locators;
+
+public interface ILocator
+{
+    LocatorType LocatorType { get; }
+    FileInfo Locate(DirectoryInfo rootDirectory, Type type);
+}
--- a/src/BenchmarkDotNet/Locators/LocatorType.cs
+++ b/src/BenchmarkDotNet/Locators/LocatorType.cs
@@ -1,0 +1,6 @@
+namespace BenchmarkDotNet.Locators;
+
+public enum LocatorType
+{
+    ProjectFile
+}
--- a/src/BenchmarkDotNet/Locators/ProjectLocator.cs
+++ b/src/BenchmarkDotNet/Locators/ProjectLocator.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using BenchmarkDotNet.Exporters;
+
+namespace BenchmarkDotNet.Locators;
+
+public class ProjectLocator : ILocator
+{
+    public static ProjectLocator Default { get; } = new ProjectLocator();
+
+    public LocatorType LocatorType => LocatorType.ProjectFile;
+
+    public FileInfo Locate(DirectoryInfo rootDirectory, Type type)
+    {
+        // important assumption! project's file name === output dll name
+        string projectName = type.GetTypeInfo().Assembly.GetName().Name;
+
+        var possibleNames = new HashSet<string> { $"{projectName}.csproj", $"{projectName}.fsproj", $"{projectName}.vbproj" };
+        var projectFiles = rootDirectory
+                           .EnumerateFiles("*proj", SearchOption.AllDirectories)
+                           .Where(file => possibleNames.Contains(file.Name))
+                           .ToArray();
+
+        if (projectFiles.Length == 0)
+        {
+            throw new NotSupportedException(
+                $"Unable to find {projectName} in {rootDirectory.FullName} and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj");
+        }
+        else if (projectFiles.Length > 1)
+        {
+            throw new NotSupportedException(
+                $"Found more than one matching project file for {projectName} in {rootDirectory.FullName} and its subfolders: {string.Join(",", projectFiles.Select(pf => $"'{pf.FullName}'"))}. Benchmark project names needs to be unique.");
+        }
+
+        return projectFiles[0];
+    }
+}
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
@@ -11,6 +11,7 @@ using BenchmarkDotNet.Characteristics;
 using BenchmarkDotNet.Extensions;
 using BenchmarkDotNet.Helpers;
 using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Locators;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.DotNetCli;
@@ -71,7 +72,7 @@ namespace BenchmarkDotNet.Toolchains.CsProj
         protected override void GenerateProject(BuildPartition buildPartition, ArtifactsPaths artifactsPaths, ILogger logger)
         {
             var benchmark = buildPartition.RepresentativeBenchmarkCase;
-            var projectFile = GetProjectFilePath(benchmark.Descriptor.Type, logger);
+            var projectFile = GetProjectFilePath(benchmark, logger);
 
             var xmlDoc = new XmlDocument();
             xmlDoc.Load(projectFile.FullName);
@@ -246,8 +247,10 @@ namespace BenchmarkDotNet.Toolchains.CsProj
         /// returns a path to the project file which defines the benchmarks
         /// </summary>
         [PublicAPI]
-        protected virtual FileInfo GetProjectFilePath(Type benchmarkTarget, ILogger logger)
+        protected virtual FileInfo GetProjectFilePath(BenchmarkCase benchmark, ILogger logger)
         {
+            var benchmarkTarget = benchmark.Descriptor.Type;
+
             if (!GetSolutionRootDirectory(out var rootDirectory) && !GetProjectRootDirectory(out rootDirectory))
             {
                 logger.WriteLineError(
@@ -255,27 +258,21 @@ namespace BenchmarkDotNet.Toolchains.CsProj
                 rootDirectory = new DirectoryInfo(Directory.GetCurrentDirectory());
             }
 
-            // important assumption! project's file name === output dll name
-            string projectName = benchmarkTarget.GetTypeInfo().Assembly.GetName().Name;
-
-            var possibleNames = new HashSet<string> { $"{projectName}.csproj", $"{projectName}.fsproj", $"{projectName}.vbproj" };
-            var projectFiles = rootDirectory
-                .EnumerateFiles("*proj", SearchOption.AllDirectories)
-                .Where(file => possibleNames.Contains(file.Name))
-                .ToArray();
-
-            if (projectFiles.Length == 0)
+            List<string> notFound = new List<string>();
+            foreach (ILocator locator in benchmark.Config.GetLocators())
             {
-                throw new NotSupportedException(
-                    $"Unable to find {projectName} in {rootDirectory.FullName} and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj");
-            }
-            else if (projectFiles.Length > 1)
-            {
-                throw new NotSupportedException(
-                    $"Found more than one matching project file for {projectName} in {rootDirectory.FullName} and its subfolders: {string.Join(",", projectFiles.Select(pf => $"'{pf.FullName}'"))}. Benchmark project names needs to be unique.");
+                if (locator.LocatorType != LocatorType.ProjectFile)
+                    continue;
+
+                var path = locator.Locate(rootDirectory, benchmarkTarget);
+
+                if (path.Exists)
+                    return path;
+
+                notFound.Add(path.FullName);
             }
 
-            return projectFiles[0];
+            throw new NotSupportedException($"Unable to find project file in {rootDirectory}. Attempted location(s): " + string.Join(", ", notFound));
         }
 
         public override bool Equals(object obj) => obj is CsProjGenerator other && Equals(other);
--- a/src/BenchmarkDotNet/Configs/DebugConfig.cs
+++ b/src/BenchmarkDotNet/Configs/DebugConfig.cs
@@ -8,6 +8,7 @@ using BenchmarkDotNet.EventProcessors;
 using BenchmarkDotNet.Exporters;
 using BenchmarkDotNet.Filters;
 using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Locators;
 using BenchmarkDotNet.Loggers;
 using BenchmarkDotNet.Order;
 using BenchmarkDotNet.Reports;
@@ -58,6 +59,7 @@ namespace BenchmarkDotNet.Configs
         public IEnumerable<IValidator> GetValidators() => Array.Empty<IValidator>();
         public IEnumerable<IColumnProvider> GetColumnProviders() => DefaultColumnProviders.Instance;
         public IEnumerable<IExporter> GetExporters() => Array.Empty<IExporter>();
+        public IEnumerable<ILocator> GetLocators() => new[] { ProjectLocator.Default };
         public IEnumerable<ILogger> GetLoggers() => new[] { ConsoleLogger.Default };
         public IEnumerable<IDiagnoser> GetDiagnosers() => Array.Empty<IDiagnoser>();
         public IEnumerable<IAnalyser> GetAnalysers() => Array.Empty<IAnalyser>();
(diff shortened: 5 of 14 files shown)


src/BenchmarkDotNet/Locators/ILocator.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using System.IO;
+
+namespace BenchmarkDotNet.Locators;
+
+public interface ILocator
+{
+    LocatorType LocatorType { get; }
+    FileInfo Locate(DirectoryInfo rootDirectory, Type type);
```
timcassell, line 9: I don't think custom locators need the root directory that BDN uses. They can find the root directory again themselves if they need it. I also think it makes more sense to pass the assembly, rather than the type.

```suggestion
    FileInfo Locate(Assembly assembly);
```

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -246,36 +247,32 @@ private static string GetIndentedXmlString(XmlDocument doc)
...
+        protected virtual FileInfo GetProjectFilePath(BenchmarkCase benchmark, ILogger logger)
         {
+            var benchmarkTarget = benchmark.Descriptor.Type;
+
             if (!GetSolutionRootDirectory(out var rootDirectory) && !GetProjectRootDirectory(out rootDirectory))
             {
                 logger.WriteLineError(
                     $"Unable to find .sln or .csproj file. Will use current directory {Directory.GetCurrentDirectory()} to search for project file. If you don't use .sln file on purpose it should not be a problem.");
                 rootDirectory = new DirectoryInfo(Directory.GetCurrentDirectory());
             }
```
timcassell, line 259: This can be moved to the ProjectLocator also.

src/BenchmarkDotNet/Locators/ProjectLocator.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        var possibleNames = new HashSet<string> { $"{projectName}.csproj", $"{projectName}.fsproj", $"{projectName}.vbproj" };
+        var projectFiles = rootDirectory
+                           .EnumerateFiles("*proj", SearchOption.AllDirectories)
+                           .Where(file => possibleNames.Contains(file.Name))
+                           .ToArray();
+
+        if (projectFiles.Length == 0)
+        {
+            throw new NotSupportedException(
```
timcassell, line 29: If we're now supporting multiple locators, this should not throw. Perhaps the ILocator should use `bool TryLocate(..., out string errorMessage)` instead.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -246,36 +247,32 @@ private static string GetIndentedXmlString(XmlDocument doc)
...
-                throw new NotSupportedException(
-                    $"Found more than one matching project file for {projectName} in {rootDirectory.FullName} and its subfolders: {string.Join(",", projectFiles.Select(pf => $"'{pf.FullName}'"))}. Benchmark project names needs to be unique.");
+                if (path.Exists)
+                    return path;
+
+                notFound.Add(path.FullName);
             }
 
-            return projectFiles[0];
+            throw new NotSupportedException($"Unable to find project file in {rootDirectory}. Attempted location(s): " + string.Join(", ", notFound));
```
timcassell, line 275: I know `NotSupportedException` was used previously, but it looks weird to me. I think it should be `FileNotFoundException` instead.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -246,36 +247,32 @@ private static string GetIndentedXmlString(XmlDocument doc)
         /// returns a path to the project file which defines the benchmarks
         /// </summary>
         [PublicAPI]
-        protected virtual FileInfo GetProjectFilePath(Type benchmarkTarget, ILogger logger)
+        protected virtual FileInfo GetProjectFilePath(BenchmarkCase benchmark, ILogger logger)
```
timcassell, line 250: Noting the breaking change to a public API.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -246,36 +247,32 @@ private static string GetIndentedXmlString(XmlDocument doc)
...
             {
                 logger.WriteLineError(
                     $"Unable to find .sln or .csproj file. Will use current directory {Directory.GetCurrentDirectory()} to search for project file. If you don't use .sln file on purpose it should not be a problem.");
                 rootDirectory = new DirectoryInfo(Directory.GetCurrentDirectory());
             }
 
-            // important assumption! project's file name === output dll name
-            string projectName = benchmarkTarget.GetTypeInfo().Assembly.GetName().Name;
+            List<string> notFound = new List<string>();
+            foreach (ILocator locator in benchmark.Config.GetLocators())
```
timcassell, line 262: I think the default locator should be filtered out, and used last if none of the custom locators return a valid path.

src/BenchmarkDotNet/Locators/ILocator.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using System.IO;
+
+namespace BenchmarkDotNet.Locators;
+
+public interface ILocator
```
timcassell, line 6: We're always returning a `FileInfo`, so a better interface name is in order.

```suggestion
public interface IFileLocator
```

src/BenchmarkDotNet/Locators/LocatorType.cs
```diff
@@ -0,0 +1,6 @@
+namespace BenchmarkDotNet.Locators;
+
+public enum LocatorType
+{
+    ProjectFile
+}
```
timcassell, line 6: In line with the interface rename.

```suggestion
public enum FileLocatorType
{
    Project
}
```

timcassell (changes requested):

timcassell:
Also the `DotNetCliGenerator.GetBuildArtifactsDirectoryPath` may need to be updated to support the new locators.

src/BenchmarkDotNet/Locators/ILocator.cs
timcassell: If this is intended to be extended in the future, it probably makes sense to pass in a `LocatorArgs` class so it can be updated without breaking changes.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
Genbox: GetSolutionRootDirectory is also used in DotNetCliGenerator.cs:54 which is the only reason I didn't move them to the locator. I'll see what can be done.

src/BenchmarkDotNet/Locators/ProjectLocator.cs
Genbox: Is `TryLocate()` or `FileInfo? Locate()` preferred in BenchmarkDotNet? I'll make it into a TryLocate for now.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
Genbox: Agree

src/BenchmarkDotNet/Locators/ProjectLocator.cs
timcassell: I think it depends on the usage. Some places use `??` to get fallback values, but since this is a public API, I think `Try*` makes more sense.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
Genbox: It might make more sense to let the default locator stay in `GetProjectFilePath` and run the locators first then. What do you think?

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
Genbox: I've let some of the code stay in DotNetCliGenerator as I have no project to test with. I'll focus on CsProj for now.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
Genbox: Done

Genbox:
Update: ProjectLocator no longer exists. Instead, we simply give the user the ability to extend ManualConfig with a set of IFileLocators, which are executed before the default logic inside `CsProjGenerator.GetProjectFilePath`.

Genbox:
For future reference and to show that it works.

I have a project that [sets AssemblyName]((https://github.com/Genbox/FastData/blob/26bda9ced5f37745f2972727020ddbe4de16d7f0/Src/Directory.Build.props#L18)) like this: `<AssemblyName>Genbox.$(MSBuildProjectName)</AssemblyName>`

With this code:
image

I get this error (as expected):
image

After changing it to use the new Locator API:
image

It works as expected:
image

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -266,12 +285,12 @@ protected virtual FileInfo GetProjectFilePath(Type benchmarkTarget, ILogger logg
 
             if (projectFiles.Length == 0)
             {
-                throw new NotSupportedException(
+                throw new FileNotFoundException(
                     $"Unable to find {projectName} in {rootDirectory.FullName} and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj");
```
timcassell, line 289: Could add a mention of the new file locators to the message.

timcassell:
Can you please add an integration test?

src/BenchmarkDotNet/Locators/LocatorArgs.cs
```diff
@@ -0,0 +1,16 @@
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Running;
+
+namespace BenchmarkDotNet.Locators;
+
+public class LocatorArgs
```
timcassell, line 6:
```suggestion
public class FileLocatorArgs
```

Genbox:
> Can you please add an integration test?

The only solution I can think of is a separate project. Ideally, it would be a test inBenchmarkDotNet.IntegrationTests, but I cannot think of a way to do that. Do you have any ideas?

timcassell:
Why can't you do it in `BenchmarkDotNet.IntegrationTests`? Just do like your example above:

```cs
fileInfo = new FileInfo(Path.Combine(dir, "../../../BenchmarkDotNet.IntegrationTests.csproj"));
```

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -266,12 +288,18 @@ protected virtual FileInfo GetProjectFilePath(Type benchmarkTarget, ILogger logg
...
             if (projectFiles.Length == 0)
             {
-                throw new NotSupportedException(
-                    $"Unable to find {projectName} in {rootDirectory.FullName} and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj");
+                string message;
+
+                if (notFound.Count > 0)
+                    message = $"Unable to find {projectName} in any of the paths: {string.Join(", ", notFound)} or in {rootDirectory.FullName} and its subfolders";
+                else
+                    message = $"Unable to find {projectName} in {rootDirectory.FullName} and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj";
```
timcassell, line 296:
```suggestion
                    message = $"Unable to find {projectName} in {rootDirectory.FullName} and its subfolders. Most probably the name of output exe is different than the name of the .(c/f)sproj. You can add an IFileLocator to the config if this is on purpose.";
```

Genbox:
> Why can't you do it in `BenchmarkDotNet.IntegrationTests`? Just do like your example above:
> 
> ```cs
> fileInfo = new FileInfo(Path.Combine(dir, "../../../BenchmarkDotNet.IntegrationTests.csproj"));
> ```

Well, I was thinking I have to change the AsemblyName of BenchmarkDotNet.IntegrationTests.csproj, which will affect all the other tests. Adding a FileLocator that points to IntegrationTests.csproj could still work, even if FileLocators don't. The fallback will find the project instead, thereby hiding a bug.

timcassell:
Ah, that's a good point. Yes, you can create a new project and set it up like `BenchmarkDotNet.IntegrationTests.ConfigPerAssembly`.

Genbox:
An integration test has been implemented.

tests/BenchmarkDotNet.IntegrationTests.FileLocators/BenchmarkDotNet.IntegrationTests.FileLocators.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\build\common.props" />
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
timcassell, line 4:
```suggestion
    <TargetFrameworks>net462;net8.0</TargetFrameworks>
```

Genbox:
Please check that the csproj has the correct properties set. There were package ids and other stuff in the other projects that seemed irrelevant.

The tests also fails when running as debug. Not sure if this is intended or not. It is not something I've caused :)

tests/BenchmarkDotNet.IntegrationTests/BenchmarkDotNet.IntegrationTests.csproj
```diff
@@ -31,6 +31,12 @@
     <ProjectReference Include="..\..\src\BenchmarkDotNet.Diagnostics.dotTrace\BenchmarkDotNet.Diagnostics.dotTrace.csproj" />
     <ProjectReference Include="..\..\src\BenchmarkDotNet.Diagnostics.dotMemory\BenchmarkDotNet.Diagnostics.dotMemory.csproj" />
   </ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <!-- This test project is only relevant for .NET 5+ projects -->
```
timcassell, line 36: This is also relevant for .Net Framework projects using sdk-style projects using `CsProjClassicNetToolchain`. You can ensure that toolchain is used in the test instead of `RoslynToolchain` by passing the toolchain in the config. You can test both `CsProjClassicNetToolchain.Net462` and `CsProjCoreToolchain.NetCoreApp80`.

timcassell (changes requested):

tests/BenchmarkDotNet.IntegrationTests/BenchmarkDotNet.IntegrationTests.csproj
Genbox: I've used .NET Core for so long that I completely forgot csproj-style projects were also made available for the .NET Framework. 😆

Genbox:
> LGTM. We can update the `GetBuildArtifactsDirectoryPath` to support the new file locators separately.
> 
> Thanks @Genbox!

Likewise! Thanks for your help and contributions. Very much appreciated.

PR #2688, opened 2025-01-16
Constant stack size

### LINKED ISSUE #1120: Large Spike from WorkloadWarmup to WorkloadActual
I'm comparing the Contains method of 3 .NET HashSets using N with params ranging from 10,000 Contains calls to 100,000 calls (my Params increment by 1,000, so 91 params in total).  I then graph the results and notice large spikes in the graph.  I've run multiple times and noticed large spikes every time along with some smaller spikes.  There should never be spikes (a point that is higher and then the next point is lower) in the graph because for every N Param, I'm just doing the same work as the prior N plus a little more work.  I noticed in the .log file that the WorkloadWarmup mean times look ok, and it is only the WorkloadActual times that have this spike.

Below is the part of the log file that shows the spike.  It happens when calling Contains for the C5 HashSet for a param of 41,000, but a similar spike has occurred for the other 2 HashSets that I'm testing at slightly different N values.  Something seems to happen between the warmup and actual that causes this.

// **************************
// Benchmark: MinMaxIntRangeContains10_000to100_000.C5_Contains: RyuJitX64(Jit=RyuJit, Platform=X64) [N=41000]
// *** Execute ***
// Launch: 1 / 1
// Execute: E:\Proj\FastHashSet\HashSetBench\bin\Release\c96c584c-7da8-41d1-b399-10ff677c388d.exe --benchmarkName "HashSetBench.MinMaxIntRangeContains10_000to100_000.C5_Contains(N: 41000)" --job "RyuJitX64" --benchmarkId 95 in 
// BeforeAnythingElse

// Runtime=.NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3362.0
// GC=Concurrent Workstation
// Job: RyuJitX64(Jit=RyuJit, Platform=X64)

OverheadJitting  1: 1 op, 203913.14 ns, 203.9131 us/op
WorkloadJitting  1: 1 op, 1102405.40 ns, 1.1024 ms/op

OverheadJitting  2: 16 op, 180363.39 ns, 11.2727 us/op
WorkloadJitting  2: 16 op, 13302838.71 ns, 831.4274 us/op

WorkloadPilot    1: 16 op, 20696629.34 ns, 1.2935 ms/op
WorkloadPilot    2: 32 op, 41250574.89 ns, 1.2891 ms/op
WorkloadPilot    3: 64 op, 82138206.56 ns, 1.2834 ms/op
WorkloadPilot    4: 128 op, 164356482.28 ns, 1.2840 ms/op
WorkloadPilot    5: 256 op, 328456687.85 ns, 1.2830 ms/op
WorkloadPilot    6: 512 op, 657090691.48 ns, 1.2834 ms/op

OverheadWarmup   1: 512 op, 1662.34 ns, 3.2467 ns/op
OverheadWarmup   2: 512 op, 831.17 ns, 1.6234 ns/op
OverheadWarmup   3: 512 op, 831.17 ns, 1.6234 ns/op
OverheadWarmup   4: 512 op, 831.17 ns, 1.6234 ns/op
OverheadWarmup   5: 512 op, 831.17 ns, 1.6234 ns/op
OverheadWarmup   6: 512 op, 831.17 ns, 1.6234 ns/op

OverheadActual   1: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual   2: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual   3: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual   4: 512 op, 1108.22 ns, 2.1645 ns/op
OverheadActual   5: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual   6: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual   7: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual   8: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual   9: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual  10: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual  11: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual  12: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual  13: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual  14: 512 op, 831.17 ns, 1.6234 ns/op
OverheadActual  15: 512 op, 831.17 ns, 1.6234 ns/op

WorkloadWarmup   1: 512 op, 414978749.81 ns, 810.5054 us/op
WorkloadWarmup   2: 512 op, 414891754.26 ns, 810.3355 us/op
WorkloadWarmup   3: 512 op, 414943840.77 ns, 810.4372 us/op
WorkloadWarmup   4: 512 op, 414920291.02 ns, 810.3912 us/op
WorkloadWarmup   5: 512 op, 414833018.41 ns, 810.2207 us/op
WorkloadWarmup   6: 512 op, 415077935.82 ns, 810.6991 us/op
WorkloadWarmup   7: 512 op, 414927217.42 ns, 810.4047 us/op

// BeforeActualRun
WorkloadActual   1: 512 op, 604296028.68 ns, 1.1803 ms/op
WorkloadActual   2: 512 op, 603948323.53 ns, 1.1796 ms/op
WorkloadActual   3: 512 op, 603905933.98 ns, 1.1795 ms/op
WorkloadActual   4: 512 op, 603851908.08 ns, 1.1794 ms/op
WorkloadActual   5: 512 op, 604080202.14 ns, 1.1798 ms/op
WorkloadActual   6: 512 op, 604648720.83 ns, 1.1810 ms/op
WorkloadActual   7: 512 op, 604043353.71 ns, 1.1798 ms/op
WorkloadActual   8: 512 op, 604067734.62 ns, 1.1798 ms/op
WorkloadActual   9: 512 op, 604017310.45 ns, 1.1797 ms/op
WorkloadActual  10: 512 op, 603970488.01 ns, 1.1796 ms/op
WorkloadActual  11: 512 op, 603755492.63 ns, 1.1792 ms/op
WorkloadActual  12: 512 op, 603723354.15 ns, 1.1791 ms/op
WorkloadActual  13: 512 op, 603893743.52 ns, 1.1795 ms/op
WorkloadActual  14: 512 op, 604155284.29 ns, 1.1800 ms/op
WorkloadActual  15: 512 op, 603900669.92 ns, 1.1795 ms/op

// AfterActualRun
WorkloadResult   1: 512 op, 604295179.04 ns, 1.1803 ms/op
WorkloadResult   2: 512 op, 603947473.90 ns, 1.1796 ms/op
WorkloadResult   3: 512 op, 603905084.35 ns, 1.1795 ms/op
WorkloadResult   4: 512 op, 603851058.45 ns, 1.1794 ms/op
WorkloadResult   5: 512 op, 604079352.50 ns, 1.1798 ms/op
WorkloadResult   6: 512 op, 604042504.07 ns, 1.1798 ms/op
WorkloadResult   7: 512 op, 604066884.99 ns, 1.1798 ms/op
WorkloadResult   8: 512 op, 604016460.81 ns, 1.1797 ms/op
WorkloadResult   9: 512 op, 603969638.37 ns, 1.1796 ms/op
WorkloadResult  10: 512 op, 603754643.00 ns, 1.1792 ms/op
WorkloadResult  11: 512 op, 603722504.51 ns, 1.1791 ms/op
WorkloadResult  12: 512 op, 603892893.89 ns, 1.1795 ms/op
WorkloadResult  13: 512 op, 604154434.65 ns, 1.1800 ms/op
WorkloadResult  14: 512 op, 603899820.28 ns, 1.1795 ms/op
GC:  0 0 0 0 0

// AfterAll

Mean = 1.1796 ms, StdErr = 0.0001 ms (0.01%); N = 14, StdDev = 0.0003 ms
Min = 1.1791 ms, Q1 = 1.1795 ms, Median = 1.1796 ms, Q3 = 1.1798 ms, Max = 1.1803 ms
IQR = 0.0003 ms, LowerFence = 1.1790 ms, UpperFence = 1.1803 ms
ConfidenceInterval = [1.1793 ms; 1.1800 ms] (CI 99.9%), Margin = 0.0003 ms (0.03% of Mean)
Skewness = 0.28, Kurtosis = 2.41, MValue = 2

timcassell:
Fixes #1120.

1. Refactored engine stages to make the stack size constant for each benchmark invocation.
2. Applied `AggressiveOptimization` to engine methods to eliminate tiered-JIT as a potential variable in the engine when running iterations. (See also https://github.com/AndreyAkinshin/perfolizer/pull/19)

I tested this on Ryzen 7 9800X3D, and got the same results on both master and this PR. I could not repro the measurement spikes from the issue (the graph was smooth).

I tested on Apple M3 and got these results.

Master:

image

PR:

image

Observations to note
- The after results are much less spikey for longer.
  - I can't say why the last bit of the graph is also spikey, it could have to do with the fact that I ran on a MacBook Air and it could have throttled, but I can't prove it (the benchmark took a very long time to run, so I only ran it once). I also did not include the changes in https://github.com/AndreyAkinshin/perfolizer/pull/19 in this test, which could also be a factor.
- The after results are larger, at the upper end of the spikes from the results with master.
  - I suspect that the benchmarks are sensitive to the alignment of the stack when it is invoked. I'm not sure if BDN should take this into account by default, and if so, what it should do about it (we already have `[MemoryRandomization]` that randomizes the stack for each iteration).


src/BenchmarkDotNet/Engines/Engine.cs
```diff
@@ -148,11 +319,15 @@ public RunResults Run()
...
         }
 
+        [MethodImpl(CodeGenHelper.AggressiveOptimizationOption)]
         public Measurement RunIteration(IterationData data)
         {
             // Initialization
             long invokeCount = data.InvokeCount;
             int unrollFactor = data.UnrollFactor;
+            if (invokeCount % unrollFactor != 0)
+                throw new ArgumentOutOfRangeException($"InvokeCount({invokeCount}) should be a multiple of UnrollFactor({unrollFactor}).");
```
Corniel, line 329: I would argue that the arguments validated should be mentioned:

```suggestion
                throw new ArgumentOutOfRangeException(nameof(data), $"InvokeCount({invokeCount}) should be a multiple of UnrollFactor({unrollFactor}).");
```

timcassell: Yeah makes sense. I just copy/pasted that from the engine stage.

AndreyAkinshin (changes requested):
@timcassell I really love this approach! Maintaining a constant stack size should help achieve more stable measurements. I also like the idea of a unified method for running iterations in the engine instead of delegating it to the `*Stage.Run*()` methods.

Now, we need to develop a better design. The `IEngineStageEvaluator` seems unnecessary since we already have `IStoppingCriteria`; merging these entities would be beneficial. We can also simplify the existing solution by eliminating `EngineWarmupStage`, `EnginePilotStage`, and `EngineActualStage`, as their purpose is to provide running and stopping strategies. Since we have a sealed hierarchy of stages that should not be extended externally, we can introduce an advanced dispatcher that builds an `IStoppingCriteria`/`IEngineStageEvaluator` based on `IterationStage`, `IterationMode`, and other state properties without calling overridden `Engine*Stage` methods.

What do you think?

timcassell:
> Now, we need to develop a better design. The `IEngineStageEvaluator` seems unnecessary since we already have `IStoppingCriteria`; merging these entities would be beneficial.

I initially looked at re-using `IStoppingCriteria`, but it wouldn't quite work for this purpose due to how the pilot stage builds up the count. I implemented it this way to prevent any breaking changes, since stages and `IStoppingCriteria` are public. But if we're cool if breaking those public types (or removing them from public surface entirely), then we can certainly adjust the implementation.

> We can also simplify the existing solution by eliminating `EngineWarmupStage`, `EnginePilotStage`, and `EngineActualStage`, as their purpose is to provide running and stopping strategies. Since we have a sealed hierarchy of stages that should not be extended externally, we can introduce an advanced dispatcher that builds an `IStoppingCriteria`/`IEngineStageEvaluator` based on `IterationStage`, `IterationMode`, and other state properties without calling overridden `Engine*Stage` methods.

I'm not sure what you have in mind there. It seems to me that the `EnumerateStages()` method _is_ that dispatcher. You just want to remove the stages altogether and move all that logic to that iterator?

AndreyAkinshin:
> but it wouldn't quite work for this purpose due to how the pilot stage builds up the count

We can rework how the pilot stage exposes the chosen invocation count; changing the API is not a problem.

> But if we're cool with breaking those public types (or removing them from public surface entirely)

Initially, I made it public to enable experiments with stopping criteria without needing to build a new version of BenchmarkDotNet. However, I don't feel that anyone is actually using it. Therefore, I believe it's safe to completely rework the API and reduce the public surface.

> I'm not sure what you have in mind there.

Currently, the `Engine*Stage` classes serve primarily as wrappers for two methods (Auto/Specific) that implement the running and stopping strategy. What if we extracted each strategy into a separate class that manages the stopping behavior and includes additional features, such as exposing the chosen invocation count? `EnumerateStages` could return instances of this class instead of tuples `(IterationStage stage, IterationMode mode, IEngineStageEvaluator evaluator)`. This change would allow us to eliminate the existing `Engine*Stage` classes, resulting in a design that is more conducive to testing.

timcassell:
@AndreyAkinshin Since we're ok with breaking changes, what do you think of making the measurement indices 0-based? I noticed they are 1-based while I was working on this, and I couldn't figure out why. Maybe just an off-by-one bug that was never caught? There are no tests for the measurement index.

AndreyAkinshin:
@timcassell The measurement iteration indices are primarily intended for presentation to the user, rather than for internal logic. From a usability perspective, 1-based numbering tends to be more intuitive and user-friendly. For example, we typically refer to the first iteration as “Iteration 1” rather than “Iteration 0”—just as spreadsheet rows in Excel start from 1 to match user expectations.

Using 0-based indices in this context can easily lead to confusion. Consider a case where the output shows:
```
Iteration 0  
Iteration 1  
Iteration 2  
```
If someone refers to “the second iteration,” it’s ambiguous whether they mean `Iteration 1` (which is technically the second) or `Iteration 2` (which has the number 2). With 1-based numbering, this ambiguity is avoided.

Since the main purpose of these indices is to improve readability and clarity for the end user, we chose 1-based indexing intentionally.

timcassell:
> Initially, I made it public to enable experiments with stopping criteria without needing to build a new version of BenchmarkDotNet. However, I don't feel that anyone is actually using it. Therefore, I believe it's safe to completely rework the API and reduce the public surface.

Well, it looks like there was never any way to inject custom stopping criteria. Users would have to implement their own `IEngine` anyway.

AndreyAkinshin (approved):
It looks much better now. Thanks for implementing this!

Issue #2701, opened 2025-03-01
InProcessNoEmitToolchain + NativeAOT Support?


eliphatfs:
I have the same use cases as #2694 to benchmark the same binary on different machines.

I use this config object:

```csharp
    public class BenchmarkConfig : ManualConfig
    {
        public BenchmarkConfig()
        {
            AddJob(Job.Default.WithToolchain(InProcessNoEmitToolchain.Instance));
        }
    }
```

However, when running a published NativeAOT (.NET 9.0.2) executable, I get:
```
System.InvalidOperationException: Bug: type BenchmarkDotNet.Toolchains.InProcess.NoEmit.InProcessNoEmitRunner+Runnable not found.
   at BenchmarkDotNet.Toolchains.InProcess.NoEmit.InProcessNoEmitRunner.Run(IHost, BenchmarkCase) + 0x370
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at BenchmarkDotNet.Toolchains.Results.ExecuteResult.LogIssues(ILogger, BuildResult) + 0x1f
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.RunExecute(ILogger, BenchmarkCase, BenchmarkId, IToolchain, BuildResult, IResolver, IDiagnoser, Int32) + 0xeb
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Execute(ILogger, BenchmarkCase, BenchmarkId, IToolchain, BuildResult, IResolver) + 0x572
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.RunCore(BenchmarkCase, BenchmarkId, ILogger, IResolver, BuildResult) + 0xa3
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Run(BenchmarkRunInfo, Dictionary`2, IResolver, ILogger, EventProcessor, List`1, String, String, Int32, StartedClock&, Int32&, TaskbarProgress) + 0x7f6
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Run(BenchmarkRunInfo[]) + 0x982
   at BenchmarkDotNet.Running.BenchmarkRunner.RunWithDirtyAssemblyResolveHelper(Type, IConfig, String[]) + 0x50
   at BenchmarkDotNet.Running.BenchmarkRunner.RunWithExceptionHandling(Func`1) + 0x20
[...]

PR #2702, opened 2025-03-01
Basic support of `InProcessNoEmitRunner` for NativeAOT.

eliphatfs:
Supports benchmark methods with `void` or `Task` as return type.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/InProcessNoEmitRunner.cs
+++ b/src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/InProcessNoEmitRunner.cs
@@ -1,4 +1,5 @@
-﻿using System;
+using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using BenchmarkDotNet.Engines;
 using BenchmarkDotNet.Environments;
--- a/src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/InProcessNoEmitRunner.cs
+++ b/src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/InProcessNoEmitRunner.cs
@@ -15,7 +15,9 @@ namespace BenchmarkDotNet.Toolchains.InProcess.NoEmit
     /// </summary>
     internal class InProcessNoEmitRunner
     {
+#if NET6_0_OR_GREATER
         [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(Runnable))]
+#endif
         public static int Run(IHost host, BenchmarkCase benchmarkCase)
         {
             // the first thing to do is to let diagnosers hook in before anything happens
--- a/src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/InProcessNoEmitRunner.cs
+++ b/src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/InProcessNoEmitRunner.cs
@@ -15,6 +15,7 @@ namespace BenchmarkDotNet.Toolchains.InProcess.NoEmit
     /// </summary>
     internal class InProcessNoEmitRunner
     {
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(Runnable))]
         public static int Run(IHost host, BenchmarkCase benchmarkCase)
         {
             // the first thing to do is to let diagnosers hook in before anything happens


eliphatfs:
See #2701.

timcassell:
Does it also fix `int` or `Task<int>` returns?

src/BenchmarkDotNet/Toolchains/InProcess/NoEmit/InProcessNoEmitRunner.cs
```diff
@@ -15,6 +15,7 @@ namespace BenchmarkDotNet.Toolchains.InProcess.NoEmit
     /// </summary>
     internal class InProcessNoEmitRunner
     {
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(Runnable))]
```
timcassell, line 18:
```suggestion
#if NET6_0_OR_GREATER
        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(Runnable))]
#endif
```

eliphatfs:
It doesn't fix int or Task<int> returns.
I don't think of a general way for supporting arbitrary return types (user-defined classes/structs etc.).
For basic types (bool, byte, int, uint, long, etc.) though we may be able to enumerate them and either call them explicitly as special cases or use in a dummy helper class and declare dynamic dependency to that helper class. Would you like me to do that?

eliphatfs:
For the general return type to work we may expose a generic helper class for users to declare themselves the dependencies with. Not zero-instrumentation, but may be acceptable.

timcassell:
> For the general return type to work we may expose a generic helper class for users to declare themselves the dependencies with. Not zero-instrumentation, but may be acceptable.

That sounds like a better option. I don't want to special case a bunch of types.

timcassell:
On second thought, I don't really want to add complex APIs just for this toolchain. The ultimate solution will be #1770.

Tests are failing, please check.

timcassell (approved):
Thanks @eliphatfs

MichalStrehovsky:
If you were okay harcoding assembly name into `string inProcessRunnableTypeName = $"{typeof(InProcessNoEmitRunner).FullName}+{nameof(Runnable)}";` (i.e. add a comma and put the assembly name after the comma) and use `Type.GetType(inProcessRunnableTypeName)` instead of the `typeof(InProcessNoEmitRunner).GetTypeInfo().Assembly.GetType(inProcessRunnableTypeName)`, this would "just work", no `DynamicDependency` needed.

Type.GetType with a constant string gets fully analyzed and produces no trimming warnings.

`DynamicDependency` is always a last resort thing because it means there is code that cannot be analyzed __somewhere__ and if that code ever changes and people forget about the DynamicDependency, that's a bug...

MichalStrehovsky:
(The intrinsic APIs are doc'd at https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/intrinsic-requiresunreferencedcode-apis)

eliphatfs:
Hi Michal, thank you for the valuable insights and comments! Are there any way through these intrinsics we may also solve the missing metadata error from MakeGeneric BenchmarkAction<T> calls where T is the return type of the methods users attributed with [Benchmark]?

MichalStrehovsky:
There are certain patterns where `MakeGeneric` might be AOT safe - those are documented at https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/intrinsic-requiresdynamiccode-apis#methodinfomakegenericmethodtype-method-net-9 (note this is a different doc from above - one is for trimming, the other is for AOT).

If it involved valuetypes, this can only be safe if analysis can see all possible instantiations ahead of time.

So `typeof(SomeType).GetMethod("GenericMethod").MakeGenericMethod(typeof(int))` is safe, or `typeof(SomeType).GetMethod("GenericMethod").MakeGenericMethod(typeof(T))` is safe, but if the type is obtained from a place like "return type of a method annotated `Benchmark`", there's no way to express that.

You'd need a source generator that generates the `MakeGeneric` calls in a shape that can be analyzed (or better yet, just source-generate the C# to construct a delegate to the instantiated method if it's visible/accessible instead of reflecting at runtime).

eliphatfs:
I also agree using a source generator is the ultimate solution, just as Tim has mentioned above.

MichalStrehovsky:
> relying on internal analysis details of the compiler

These are documented behaviors (documented on learn.microsoft.com on the pages I linked to) and as such are contractual, same way as e.g. `Console.WriteLine` is (but you can always p/invoke into `printf` if you don't trust WriteLine, same way as the `DynamicDependency`, it's just more awkward to do so).

If benchmark.net means AOT support seriously, eventually it will enable the [trimming/AOT analyzers](https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/) and then the Assembly.GetType use would immediately get flagged as a warning. Rewriting it to use `Type.GetType` with a constant string would clear that warning.

PR #2703, opened 2025-03-01
Missing dotnet 9.0 support.

VAllens:
Added support for `RuntimeMoniker.NativeAot90` in the `RuntimeMonikerExtensions.cs` file, which returns `NativeAotRuntime.Net90`.


timcassell:
While you're at it, can you also add the missing `Mono90`?

VAllens:
> While you're at it, can you also add the missing `Mono90`?

Sure, I didn't realize it was missing as well.

timcassell (approved):
Thanks @VAllens!

PR #2709, opened 2025-03-20
Exporters.Plotting: Move to ScottPlot 5.0.54

bclehmann:
Hey, I noticed that this project was using ScottPlot 5.0.25, which causes issues for me trying to benchmark ScottPlot (I'm a ScottPlot maintainer). Without csproj trickery with package aliases we're not able to reference two versions of a package in one project, it'll just pick the newer one, which is incompatible in some respects with 5.0.25.

There are some breaking changes, since despite my misgivings we don't use SemVer (I'm a ScottPlot maintainer). These are mostly all fairly trivial field renames. For these I believe they are just marked obsolete, not removed yet.

I don't know if you would consider it a breaking change to update ScottPlot versions, I think it only would be if the project under test was also using ScottPlot. In that case, I think for most of these issues you can sprinkle a couple `#pragma warning disable`s around.

### DIFF
--- a/src/BenchmarkDotNet.Exporters.Plotting/BenchmarkDotNet.Exporters.Plotting.csproj
+++ b/src/BenchmarkDotNet.Exporters.Plotting/BenchmarkDotNet.Exporters.Plotting.csproj
@@ -14,6 +14,6 @@
     <ProjectReference Include="..\BenchmarkDotNet\BenchmarkDotNet.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="ScottPlot" Version="5.0.25" />
+    <PackageReference Include="ScottPlot" Version="5.0.54" />
   </ItemGroup>
 </Project>
--- a/src/BenchmarkDotNet.Exporters.Plotting/ScottPlotExporter.cs
+++ b/src/BenchmarkDotNet.Exporters.Plotting/ScottPlotExporter.cs
@@ -204,13 +204,13 @@ namespace BenchmarkDotNet.Exporters.Plotting
                 .ToDictionary(t => t.jobId, t => palette.GetColor(t.index));
 
             plt.Legend.IsVisible = true;
-            plt.Legend.Location = Alignment.UpperRight;
-            plt.Legend.Font.Size = this.FontSize;
+            plt.Legend.Alignment = Alignment.UpperRight;
+            plt.Legend.FontSize = this.FontSize;
             var legend = data.Select(d => d.JobId)
                 .Distinct()
                 .Select((label, index) => new LegendItem()
                 {
-                    Label = label,
+                    LabelText = label,
                     FillColor = legendPalette[label]
                 })
                 .ToList();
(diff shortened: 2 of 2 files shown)


bclehmann:
For what it's worth, I don't know which features you're looking to add to this exporter, but if you have a wishlist I would be happy to help implement some of it.

timcassell:
I think there shouldn't be any issues updating the version. But I'm curious why you can't benchmark? `BenchmarkDotNet.Exporters.Plotting` shouldn't be a dependency of `BenchmarkDotNet`, so you should be able to run benchmarks as long as you haven't added that package yourself.

bclehmann:
> I think there shouldn't be any issues updating the version. But I'm curious why you can't benchmark? BenchmarkDotNet.Exporters.Plotting shouldn't be a dependency of BenchmarkDotNet, so you should be able to run benchmarks as long as you haven't added that package yourself.

Oh I wanted to see the charts, there's no problem without using this exporter.

I am now seeing that the generated bar charts look fine, but the box plots don't show up?
ScottPlotBench Benchmarks DataLogger-20250319-220747-DataLogger-boxplot

I'll go look and see if that's a new issue introduced by this PR.

timcassell:
> For what it's worth, I don't know which features you're looking to add to this exporter, but if you have a wishlist I would be happy to help implement some of it.

Ideally everything from `RPlotExporter` (I don't know all of its features though), then we can deprecate and eventually remove it in favor of Scott.

bclehmann:
> I am now seeing that the generated bar charts look fine, but the box plots don't show up?

Ah, I see the issue:

```
// * Detailed results *
DataLogger.DataLoggerRender: Job-DKXHKF(IterationCount=1, IterationTime=200ms, LaunchCount=1, WarmupCount=1) [Points=100]
Runtime = .NET 8.0.13 (8.0.1325.6609), X64 RyuJIT AVX2; GC = Concurrent Workstation
Mean = 999.952 us, StdErr = 0.000 us (0.00%), N = 1, StdDev = 0.000 us
Min = 999.952 us, Q1 = 999.952 us, Median = 999.952 us, Q3 = 999.952 us, Max = 999.952 us
IQR = 0.000 us, LowerFence = 999.952 us, UpperFence = 999.952 us
ConfidenceInterval = [NaN us; NaN us] (CI 99.9%), Margin = NaN us (NaN% of Mean)
Skewness = NaN, Kurtosis = NaN, MValue = 2
[999.952 us ; 999.953 us) | @
```

Min = Q1 = Med = Q3 = Max. Not sure how that's the case, I don't think I have done anything to break the stats. Is there something special a consumer has to do to opt into more stats then just (what I presume) is the mean?

bclehmann:
> It looks like you set the iteration count to 1.

Aye, I sure did🤦It works fine on our default config, I was just testing on a faster config out of laziness:
ScottPlotBench Benchmarks DataLogger-20250319-230229-DataLogger-boxplot

timcassell (approved):
Thanks @bclehmann

PR #2711, opened 2025-03-24
Exporters.Plotting: ScottPlot histogram export

bclehmann:
It's my understanding that the existing R plot exporters includes a density plot, but I don't know what kind of smoothing that goes through. ScottPlot does support calculating this from a histogram but it doesn't support the necessary kernel density estimation. So your options are to either compute the discrete PDF (unlikely to be appropriate unless there's hundreds of bins) or assume a normal distribution (almost certainly not appropriate). And then you can potentially add smoothing on top of that (e.g. turn it into a bezier or compute a moving average).

I figure that a histogram is at least part of the way there, and I might look at adding a KDE implementation into ScottPlot, since it would unlock a few plot types that we've never supported before.

On real data it looks something like this: 
ScottPlotBench Benchmarks DataLogger-20250323-182101-DataLogger-histogramplot

On test data it looks more like this:
MockSummary-N9-JobBaseline_MethodsJobs-histogramplot

I'm noticing that the chart is fairly messy, since there are multiple histograms in the same colour (as they're in the same job). Unless I did something wrong while grouping the data, this suggests that either histograms may not be a great option to have enabled by default (if this is typical of real-world usage) or that there might need to be separate test cases which are less busy (if this isn't typical of real-world usage).

This may be unrelated, but I noticed that the R exporter shows multiple plots in one image. This is supported in ScottPlot, so if you want to go that route that is an option. Here's a couple of examples: https://scottplot.net/cookbook/5.0/MultiplotRecipes/

### DIFF
--- a/src/BenchmarkDotNet.Exporters.Plotting/ScottPlotExporter.cs
+++ b/src/BenchmarkDotNet.Exporters.Plotting/ScottPlotExporter.cs
@@ -415,9 +415,6 @@ namespace BenchmarkDotNet.Exporters.Plotting
                 }
             }
 
-            // Tell the plot to autoscale with a small padding below the boxes.
-            plt.Axes.Margins(bottom: 0.05, right: .2);
-
             plt.PlottableList.AddRange(annotations);
 
             plt.SavePng(fileName, this.Width, this.Height);
--- a/src/BenchmarkDotNet.Exporters.Plotting/ScottPlotExporter.cs
+++ b/src/BenchmarkDotNet.Exporters.Plotting/ScottPlotExporter.cs
@@ -8,6 +8,7 @@ using BenchmarkDotNet.Properties;
 using BenchmarkDotNet.Reports;
 using ScottPlot;
 using ScottPlot.Plottables;
+using ScottPlot.Statistics;
 
 namespace BenchmarkDotNet.Exporters.Plotting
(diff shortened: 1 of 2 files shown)


bclehmann:
FWIW I've began work on KDE in ScottPlot, I expect to have a PR up this evening and then we'll see if the eponymous Scott wants it in the library:

image

AndreyAkinshin (changes requested):
@bclehmann thanks for the PR! We definitely need a histogram support here, but the current overlapping view looks confusing. Here what I got from `ScottPlotExporterTests.HistogramPlots`:
MockSummary-N9-Invalid_TwoMethodBaselines-histogramplot
MockSummary-N9-MethodBaseline_MethodsParamsJobs-histogramplot
MockSummary-N9-NoBaseline_MethodsParamsJobs_GroupByParams-histogramplot

For the initial implementation, we need something that presents information in a clearer way.

bclehmann:
Those are all fair comments, I suspect the best move is to go with a density plot like the existing R code. It should be more readable. The support for that has recently been merged into ScottPlot and I suspect there'll be a release soon: https://github.com/ScottPlot/ScottPlot/pull/4869

For what it's worth I'm sorry for the slow response time, I've been moving and been busy with work.

PR #2713, opened 2025-03-31
Add test for required properties in InProcess

aelij:


aelij:
@timcassell I only modified `InProcessEmitTest` because in `InProcessTest` the benchmark class is manually instaniated

timcassell:
> @timcassell I only modified `InProcessEmitTest` because in `InProcessTest` the benchmark class is manually instaniated

You can add a new test like `InProcessBenchmarkAllCasesSupported`.

aelij:
Sorry I don't follow. Add which test in what class? And what would be the difference from the test I already added, doesn't it already use the InProcess toolchain?

timcassell:
There are 2 in process toolchains, InProcessTest tests the InProcessNoEmitToolchain, InProcessEmitTest tests the InProcessEmitToolchain. We want to test both.

tests/BenchmarkDotNet.IntegrationTests/InProcessTest.cs
```diff
@@ -258,6 +270,13 @@ public decimal InvokeOnceValueType()
                 return DecimalResult;
             }
 
+            [Benchmark]
+            public string InvokeOnceReturnParamValue()
```
timcassell, line 274: I don't think this extra benchmark method is necessary. Adding the required property should be enough.

tests/BenchmarkDotNet.IntegrationTests/InProcessTest.cs
```diff
@@ -220,6 +225,13 @@ public void InProcessBenchmarkAllCasesSupported()
         [UsedImplicitly(ImplicitUseTargetFlags.WithMembers)]
         public class BenchmarkAllCases
         {
+            [Params("a", "b")]
+#if NET8_0_OR_GREATER
+            public required string ParamProperty { get; set; }
```
timcassell, line 230: It might be better to put this in a separate type, then you can assert the value in the benchmark method.

tests/BenchmarkDotNet.IntegrationTests/InProcessEmitTest.cs
```diff
@@ -133,6 +133,13 @@ public void InProcessBenchmarkEmitsSameIL(Type benchmarkType)
         [UsedImplicitly(ImplicitUseTargetFlags.WithMembers)]
         public class BenchmarkAllCases
         {
+            [Params("a", "b")]
+#if NET8_0_OR_GREATER
+            public required string ParamProperty { get; set; }
```
timcassell, line 138: Same here, move this to a separate type and assert the value.

aelij:
If I understand correctly, there's no point adding a test to `InProcessEmitTest` because of this:

> We can't use Roslyn toolchain for .NET Core because we don't know which assemblies to reference and .NET Core does not support dynamic assembly saving

And .NET Framework doesn't support `required`.

timcassell:
> If I understand correctly, there's no point adding a test to `InProcessEmitTest` because of this:
> 
> > We can't use Roslyn toolchain for .NET Core because we don't know which assemblies to reference and .NET Core does not support dynamic assembly saving
> 
> And .NET Framework doesn't support `required`.

We can still test it on that toolchain. Just do the same test. We don't need to test the IL matches for it for now (I think latest .Net added assembly saving functionality, but it's out of scope for this).

tests/BenchmarkDotNet.IntegrationTests/InProcessTest.cs
```diff
@@ -217,6 +217,32 @@ public void InProcessBenchmarkAllCasesSupported()
...
+            }
+        }
+
+        public class ParamsTestRequiredProperty
+        {
+            [Params("a", "b")]
+            public required string ParamProperty { get; set; }
+
+            [Benchmark]
+            public void Benchmark() => Console.WriteLine($"// ### New Parameter {ParamProperty} ###");
```
timcassell, line 242: I think asserting in the benchmark method is cleaner, no need to parse the log. Something like this:

```suggestion
            const string expected = "a";
            
            [Params(expected)]
            public required string ParamProperty { get; set; }

            [Benchmark]
            public void Benchmark() => Assert.Equals(expected, ParamProperty);
```

timcassell (approved):

Issue #2712, opened 2025-03-30
BenchmarkDotNet should warn if there are no benchmarks


richardcocks:
I've got a small benchmark that I was about to open a bug report because it appeared to "crash" during Benchmark Validation.

It shows `// Validating benchmarks:` and then the process exits without any error or warning. 

I'm an idiot, and it turns out I forgot to annotate any methods at all with `[Benchmark]`.

A warning that "No benchmarks found" would help idiots like me realise their mistake without thinking that validation is crashing.

CodingwithKetan:
@timcassell, can I work on this issue?

timcassell:
Yes

AvishaiDotan:
@CodingwithKetan Have you had a chance to make any progress on this?

It looks like checking whether `supportedBenchmarks.Length == 0` could be a good way to handle this. If there are no supported benchmarks, logging an informative message and returning a summary array might be sufficient. Reference:

https://github.com/dotnet/BenchmarkDotNet/blob/21da339839f71f1b064090e2f61edc6b47596835/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L63

```c#
if (supportedBenchmarks.Length == 0)
{
    compositeLogger.WriteLineError("// No benchmarks have been found");
    return new[] { Summary.ValidationFailed(title, resultsFolderPath, logFilePath,
        ImmutableArray.Create(new ValidationError(true, "No benchmarks have been found"))) };
}
``` 

If you're not actively working on this, I’d be happy to pick it up — just let me know.

CodingwithKetan:
@AvishaiDotan, as you got solution, so you can work on this.

PR #2718, opened 2025-04-17
Log a warning when an empty benchmark is run

AvishaiDotan:
Fix https://github.com/dotnet/BenchmarkDotNet/issues/2712

I've added a warning using the composite logger to notify when an empty benchmark is executed.
Also included a test to ensure the warning is logged correctly.


src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -578,20 +574,41 @@ private static void LogTotalTime(ILogger logger, TimeSpan time, int executedBenc
...
+                    new BenchmarkRunInfo(
+                        validBenchmarks,
+                        benchmarkRunInfo.Type,
+                        benchmarkRunInfo.Config
+
+                    ));
+
+
+            }
+            return (runInfos.ToArray(), validationErrors);
```
timcassell, line 611: This doesn't cover the case empty benchmarkRunInfos.

AvishaiDotan:
```c#
            if (benchmarkRunInfos.Length == 0)
            {
                validationErrors.Add(new ValidationError(true, $"No benchmarks were found."));
                return (Array.Empty<BenchmarkRunInfo>(), validationErrors);
            }
``` 

Will do it

AvishaiDotan:
Hi @timcassell,

Apologies for the continuous test runs. I prefer working directly on the branch, and every time I push changes—even when the PR is marked as a draft—it still triggers the tests.

As a side note, it might be a good idea to update the CI/CD configuration so tests only run when the PR is not in draft mode.

timcassell:
Error is printed when I use

```cs
BenchmarkRunner.Run<Benchmark>();
```

But nothing is printed when I use

```cs
BenchmarkRunner.Run(typeof(Program).Assembly);
```

AvishaiDotan:
> Error is printed when I use
> 
> ```cs
> BenchmarkRunner.Run<Benchmark>();
> ```
> 
> But nothing is printed when I use
> 
> ```cs
> BenchmarkRunner.Run(typeof(Program).Assembly);
> ```

I've tested running the assembly without any benchmarks using the following methods, and the implementation did not fail in either case:

1. Mocking with AssemblyBuilder
2. Referencing a new project that doesn't contain any benchmarks

In both scenarios, the implementation worked as expected.

Could you please elaborate on how you ran `BenchmarkRunner.Run(typeof(Program).Assembly)`?

timcassell:
Just like this. No benchmark types.

```cs
public class Program
{
    {
        BenchmarkRunner.Run(typeof(Program).Assembly);
    }
}
```

Ran in Visual Studio, got no output at all.

AvishaiDotan:
You're right.
When running benchmarks based on an assembly, the `BenchmarkRunInfo[]` array is created like this:

```c#
assembly.GetRunnableBenchmarks()
    .Select(type => BenchmarkConverter.TypeToBenchmarks(type, config))
    .ToArray();
``` 

If `assembly.GetRunnableBenchmarks()` returns an empty collection, `TypeToBenchmarks` is never called, which means no `BenchmarkRunInfo` instances are created.

The key point is: when `TypeToBenchmarks` is called, it sets up the `BenchmarkRunInfo` with a default logger — `BenchmarkDotNet.Loggers.ConsoleLogger`. This logger is responsible for displaying validation errors in the console.

However, in the case of an empty assembly (i.e., no runnable benchmarks), `TypeToBenchmarks` is skipped entirely, and an empty array of b`enchmarkRunInfos` is returned — meaning no logger is configured, and validation error are not being logged.

**Possible Solution**
I was thinking of adding this check inside `CreateCompositeLogger`:

```c#
if (benchmarkRunInfos.Length == 0)
{
    AddLogger(new ConsoleLogger());
}
``` 
This ensures a console logger is present even if no benchmarks are found, so validation errors still get reported.
That said, it feels a bit messy - I Would love your input on this.

timcassell:
I haven't looked into the implementation details, but I did see that `BenchmarkSwitcher` does log the error if there are no benchmarks, so perhaps you could check that for how to handle it in `BenchmarkRunner`?

timcassell (approved):
Thanks @AvishaiDotan!

PR #2676, opened 2024-12-14
Use ArtifactsPath

timcassell:
Fixes #2425
Fixes #2664

We set `ArtifactsPath` if the dotnet sdk used supports it (v8.0 or higher). Older sdks still use the same build logic as previous, and as far as I could find out, it's the best we can do for those older sdks.
Passing `ArtifactsPath` also makes MSBuild output publish directory differently by default, so we pass `PublishDir` also to work for both old and new sdks.

PR #2725, opened 2025-04-27
feat: Add option to disable parallel build

### LINKED ISSUE #2721: Feature Request: Add options to disable build in parallel.
When running benchmark with multiple Jobs that use `WithCustomBuildConfiguration`.
Parallel build is failed. and switching to sequential build.

https://github.com/dotnet/BenchmarkDotNet/blob/f4bfcd67b4ed44996c67785bbc5a1340db6f88ca/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L365-L393

I though parallel build failure is caused because `obj\project.assets.json` is shared between build configurations.
And wrong dependencies are used for parallel build.

Is is able to add options to disable parallel build?
And If it can disable parallel build. It will be a workaround for #2425 also.

filzrev:
This PR intended to fix #2721

**What's changed in this PR**
1. Add `DisableParallelBuild` flag to `ConfigOptions`.
2. Modify `BenchmarkRunnerClean::Run`.
2.1. Add `BuildSequential` method that build project in sequentially.
2.2. Add logics to check `DisableParallelBuild` option.  If any of benchmarks contains `DisableParallelBuild` flag. Use sequential build path instead of parallel build path.

**Test** 
I've manually confirmed that `sequential build` path is called when specifying `WithOptions(ConfigOptions.DisableParallelBuild);` on benchmark config.

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
@@ -86,13 +86,32 @@ namespace BenchmarkDotNet.Running
                 var globalChronometer = Chronometer.Start();
 
 
-                var buildPartitions = BenchmarkPartitioner.CreateForBuild(supportedBenchmarks, resolver);
-                eventProcessor.OnStartBuildStage(buildPartitions);
+                var parallelBuildBenchmarks = supportedBenchmarks.Where(x => !x.Config.Options.IsSet(ConfigOptions.DisableParallelBuild)).ToArray();
+                var parallelBuildPartitions = BenchmarkPartitioner.CreateForBuild(parallelBuildBenchmarks, resolver);
 
-                bool useParallelBuild = !supportedBenchmarks.Any(x => x.Config.Options.IsSet(ConfigOptions.DisableParallelBuild));
-                var buildResults = useParallelBuild
-                    ? BuildInParallel(compositeLogger, rootArtifactsFolderPath, buildPartitions, in globalChronometer, eventProcessor)
-                    : BuildSequential(compositeLogger, rootArtifactsFolderPath, buildPartitions, in globalChronometer, eventProcessor);
+                var sequentialBuildBenchmarks = supportedBenchmarks.Where(x => x.Config.Options.IsSet(ConfigOptions.DisableParallelBuild)).ToArray();
+                var sequentialBuildPartitions = BenchmarkPartitioner.CreateForBuild(sequentialBuildBenchmarks, resolver);
+
+                eventProcessor.OnStartBuildStage([.. parallelBuildPartitions, .. sequentialBuildPartitions]);
+
+                var buildResults = new Dictionary<BuildPartition, BuildResult>();
+                if (parallelBuildBenchmarks.Length > 0)
+                {
+                    var results = BuildInParallel(compositeLogger, rootArtifactsFolderPath, parallelBuildPartitions, in globalChronometer, eventProcessor);
+                    foreach (var kvp in results)
+                    {
+                        buildResults.Add(kvp.Key, kvp.Value);
+                    }
+                }
+
+                if (sequentialBuildBenchmarks.Length > 0)
+                {
+                    var results = BuildSequential(compositeLogger, rootArtifactsFolderPath, sequentialBuildPartitions, in globalChronometer, eventProcessor);
+                    foreach (var kvp in results)
+                    {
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -413,6 +417,31 @@ static string GetFormattedDifference(ClockSpan before, ClockSpan after)
                 => (after.GetTimeSpan() - before.GetTimeSpan()).ToFormattedTotalTime(DefaultCultureInfo.Instance);
         }
 
+        private static Dictionary<BuildPartition, BuildResult> BuildSequential(ILogger logger, string rootArtifactsFolderPath, BuildPartition[] buildPartitions, in StartedClock globalChronometer, EventProcessor eventProcessor)
+        {
+            logger.WriteLineHeader($"// ***** Building {buildPartitions.Length} exe(s): Start   *****");
+
+            var buildLogger = buildPartitions.Length == 1 ? logger : NullLogger.Instance; // when we have just one partition we can print to std out
```
timcassell, line 424: Just use `logger` directly since it's sequential.

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -413,6 +417,31 @@ static string GetFormattedDifference(ClockSpan before, ClockSpan after)
...
+                eventProcessor.OnBuildComplete(buildPartition, buildResults[buildPartition]);
+            }
+
+            var afterBuild = globalChronometer.GetElapsed();
+
+            logger.WriteLineHeader($"// ***** Done, took {GetFormattedDifference(beforeBuild, afterBuild)}   *****");
+
+            return buildResults;
+
+            static string GetFormattedDifference(ClockSpan before, ClockSpan after)
```
timcassell, line 441: Pull this local function out so both build functions can use it instead of copy-paste.

timcassell (changes requested):

timcassell:
@AndreyAkinshin @adamsitnik It's ok to use `ConfigOptions` enum for this? Or another method is preferred?

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
```diff
@@ -88,7 +88,11 @@ internal static Summary[] Run(BenchmarkRunInfo[] benchmarkRunInfos)
 
                 var buildPartitions = BenchmarkPartitioner.CreateForBuild(supportedBenchmarks, resolver);
                 eventProcessor.OnStartBuildStage(buildPartitions);
-                var buildResults = BuildInParallel(compositeLogger, rootArtifactsFolderPath, buildPartitions, in globalChronometer, eventProcessor);
+
+                bool useParallelBuild = !supportedBenchmarks.Any(x => x.Config.Options.IsSet(ConfigOptions.DisableParallelBuild));
```
timcassell, line 92: It might be better to split the parallelizable builds from the sequential builds and run them both.

```cs
var parallelBenchmarks = supportedBenchmarks.Where(x => !x.Config.Options.IsSet(ConfigOptions.DisableParallelBuild));
var sequentialBenchmarks = supportedBenchmarks.Where(x => x.Config.Options.IsSet(ConfigOptions.DisableParallelBuild));
```

That will also make it easier for the validation step (#2676).

AndreyAkinshin:
> It's ok to use `ConfigOptions` enum for this?

@timcassell it's OK

src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs
filzrev: On commit(https://github.com/dotnet/BenchmarkDotNet/pull/2725/commits/409425e4ccd13bf16c3a0a446e866053736073c0).
I've added code to run parallel build if possible.

timcassell (approved):
It looks like MacOS integration tests are failing by timeout on the IsWasmSupported test. It's unrelated to these changes.

Thanks @filzrev!

filzrev:
CI test `WasmIsSupported` failed with following error on mac-os.
> `The configured timeout 00:02:00 was reached!`

It might needs to increase timeout for Wasm project build that similar to following code. 
https://github.com/dotnet/BenchmarkDotNet/blob/924ca6d51460b9f7a8f6ae2c38ea7580fa224042/src/BenchmarkDotNet/Running/BuildPartition.cs#L73-L75

Issue #2719, opened 2025-04-23
Feature Request: Add MsBuildArgument supports for MSBuild special characters


filzrev:
Currently [MsBuildArgument constructor](https://github.com/dotnet/BenchmarkDotNet/blob/f4bfcd67b4ed44996c67785bbc5a1340db6f88ca/src/BenchmarkDotNet/Jobs/Argument.cs#L50) accept single string argument.
And this string value is passed to MSBuild command line parameter as is.

So when specified string contains [MSBuild  special characters](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-special-characters?view=vs-2022) 
It needs manually escape these characters.
(e.g. When passing list values. It need to escape semicolon(`;`) char to `%3B`)

Is it able to handle these special characters inside MsBuildArgument class?
or add  [MsBuildArgument constructor] that accept string params to handle multiple values?

**Test Code**
```csharp
    internal class Program
    {
        static void Main(string[] args)
        {
            BenchmarkRunner.Run<Benchmarks>();
        }
    }

    [Config(typeof(CustomConfig))]
    public class Benchmarks
    {
        [Benchmark]
        public void Benchmark01()
        {
#if TEST1 && TEST2
            // DefineConstants symbols are defined as expected.
#else
            throw new Exception("DefineConstants value is unexpected value!");
#endif
        }

        public class CustomConfig : ManualConfig
        {
            public CustomConfig()
            {
                AddLogger(ConsoleLogger.Default);

                AddColumnProvider(DefaultConfig.Instance.GetColumnProviders().ToArray());
              
                AddJob(Job.ShortRun
                    .WithStrategy(RunStrategy.Monitoring)
                    .WithArguments(
                    [
                       // new MsBuildArgument("/p:DefineConstants=TEST1;TEST2"),     // Error: MSBUILD : error MSB1006: Property is not valid.
                       // new MsBuildArgument("/p:DefineConstants=\"TEST1;TEST2\""), // Double quote does not works.
                       new MsBuildArgument("/p:DefineConstants=TEST1%3BTEST2"),      // It works when using MSBuild special character
                    ])
                    .WithId("CustomJob"));

                WithOption(ConfigOptions.KeepBenchmarkFiles, true); // Use this settings for output to JobId folder
            }
        }
    }
```

timcassell:
I think it's reasonable to escape those characters in the `MsBuildArgument` ctor.

filzrev:
[...]

PR #2729, opened 2025-05-13
Auto-escape MSBuild special characters in MsBuildArgument

Gal-lankri:
### Summary

This PR adds automatic escaping of MSBuild-reserved characters in the `MsBuildArgument` class (e.g., `;`, `%`, `$`, `@`, etc.), so users no longer need to manually encode these characters when passing MSBuild parameters.

### Why

Previously, developers had to encode characters like semicolons manually (`;` → `%3B`). This PR improves usability and correctness when passing multiple values to MSBuild properties such as `DefineConstants`.

### Changes

- Updated `MsBuildArgument` constructor to auto-escape special characters
- Added a `[Theory]`-based test (`EscapesSpecialCharactersCorrectly`) to cover:
  - Semicolons
  - Percent
  - Dollar sign
  - At-symbol
  - Parentheses
  - Asterisk
  - Normal input (no escape)

### Example

```csharp
new MsBuildArgument("/p:DefineConstants=DEBUG;TRACE")
// becomes:
"/p:DefineConstants=DEBUG%3BTRACE"

### DIFF
--- a/src/BenchmarkDotNet/Jobs/Argument.cs
+++ b/src/BenchmarkDotNet/Jobs/Argument.cs
@@ -62,50 +62,30 @@ namespace BenchmarkDotNet.Jobs
             { '*', "%2A" }
         };
 
-        // This lets us check if a %xx is a known escape
-        private static readonly HashSet<string> KnownEscapeValues = [.. MsBuildEscapes.Values];
-
         private static string EscapeMsBuildSpecialChars(string value)
         {
             if (string.IsNullOrEmpty(value))
                 return value;
 
             var sb = new StringBuilder(value.Length);
-            int i = 0;
-
-            while (i < value.Length)
+            foreach (char c in value)
             {
-                char c = value[i];
-
-                // Check for a known %xx escape
-                if (c == '%' && i + 2 < value.Length)
-                {
-                    string candidate = value.Substring(i, 3); // e.g., "%3B"
-
-                    if (KnownEscapeValues.Contains(candidate))
-                    {
-                        sb.Append(candidate);
-                        i += 3;
-                        continue;
-                    }
-                }
-
-                // Escape only if it's a special MSBuild character
                 if (MsBuildEscapes.TryGetValue(c, out var escaped))
-                {
                     sb.Append(escaped);
-                }
                 else
-                {
                     sb.Append(c);
-                }
-
-                i++;
             }
 
             return sb.ToString();
         }
-
-        public MsBuildArgument(string value) : base(EscapeMsBuildSpecialChars(value)) { }
+        /// <summary>
+        /// Represents an MSBuild command-line argument.
+        /// <param name="value">The raw or escaped argument value (e.g., "/p:DefineConstants=TEST1;TEST2").</param>
+        /// <param name="escapeSpecialChars">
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Gal-lankri** [comment]
@dotnet-policy-service agree

**Comment by timcassell** [comment]
@Gal-lankri Can you please pull latest upstream master to your fork and rebase your branch? Github is picking up a lot of unrelated changes.

File: src/BenchmarkDotNet/Jobs/Argument.cs [hunk 1]
```diff
@@ -47,6 +49,40 @@ public MonoArgument(string value) : base(value)
     [PublicAPI]
     public class MsBuildArgument : Argument
     {
-        public MsBuildArgument(string value) : base(value) { }
+        private static readonly Dictionary<char, string> MsBuildEscapes = new ()
+        {
+            { '%', "%25" },
```
timcassell, line 54: If anyone has already escaped their arguments, this will break them. Is there a way to avoid any breaks?

timcassell: It can get weird if someone puts `100%25`, it's ambiguous to us whether it should be `100%` or `100%25`. I think it would be better to add a `bool escapeSpecialChars = true` param, and don't try to parse for existing escapes.

Gal-lankri: Yes, good point, thank you!

You're right, there's no reliable way to guess the user's intent in edge cases like `100%25`. Maybe it's smarter to let the user to decide. 

I've updated the implementation to:
- Add a `bool escapeSpecialChars = true` parameter to the `MsBuildArgument` constructor
- If `true`, the value is escaped
- If `false`, the value is passed through exactly as provided (for pre-escaped input)

This avoids ambiguity and doesn't change the default behavior for existing users.

filzrev:
This issue is tracked by #2719 also.

I though it's better to separate MSBuildArgument's `TextRepresentation` with **argument that is used inside `.bat`/`.sh` files**.
https://github.com/dotnet/BenchmarkDotNet/blob/v0.14.0/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs#L203C72-L203C90.

Because `TextRepresentation` is shown on `Argument` column on benchmark report and it's hard to read if contains escape chars.

Additionally, It seems escape is needed when passing MSBuild parameter with `-p:{key}={value}`  and `-rp:{key}={value}` format.
As far as I've confirmed, Other msbuild command argument works without escaping  (e.g. `-getProperty:Prop1;Prop2`)

timcassell:
> Because `TextRepresentation` is shown on `Argument` column on benchmark report and it's hard to read if contains escape chars.

That's a good point. `MsBuildArgument` can override `ToString` to display the original value.

> Additionally, It seems escape is needed when passing MSBuild parameter with `-p:{key}={value}` and `-rp:{key}={value}` format. As far as I've confirmed, Other msbuild command argument works without escaping (e.g. `-getProperty:Prop1;Prop2`)

Seems like a good reason to not escape by default. Users should opt-in rather than opt-out.

timcassell:
Superseded by #3055

PR #2732, opened 2025-05-21
Add MsBuildArgument constructor supporting multiple values

jespernapp:
Added MsBuildArgument(string key, params string[] values) constructor.
Implemented escaping of semicolons in argument values.

### DIFF
--- a/src/BenchmarkDotNet/Jobs/Argument.cs
+++ b/src/BenchmarkDotNet/Jobs/Argument.cs
@@ -48,5 +48,18 @@ namespace BenchmarkDotNet.Jobs
     public class MsBuildArgument : Argument
     {
         public MsBuildArgument(string value) : base(value) { }
+
+        public MsBuildArgument(string key, params string[] values)
+            : base($"/p:{key}={EscapeAndJoin(values)}") { }
+
+        private static string EscapeAndJoin(string[] values)
+        {
+            return string.Join("%3b", values.Select(EscapeSpecialChars));
+        }
+
+        private static string EscapeSpecialChars(string input)
+        {
+            return input.Replace(";", "%3B");
+        }
     }
 }
--- a/tests/BenchmarkDotNet.IntegrationTests.ManualRunning/MsBuildArgumentTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests.ManualRunning/MsBuildArgumentTests.cs
@@ -25,6 +25,13 @@ namespace BenchmarkDotNet.IntegrationTests.ManualRunning
         }
 
         [Fact]
+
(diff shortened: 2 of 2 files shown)


jespernapp:
@dotnet-policy-service agree

timcassell:
I would rather something like this be implemented as a new type `public class MsBuildProperty : MsBuildArgument`. Also, it seems that wrapping the values in quotes is sufficient without escaping the semicolons (https://github.com/dotnet/BenchmarkDotNet/pull/2730#issuecomment-2885393999).

src/BenchmarkDotNet/Jobs/Argument.cs
```diff
@@ -45,8 +45,34 @@ public MonoArgument(string value) : base(value)
...
+
+        private static string EscapeAndJoin(string[] values)
+        {
+            return string.Join("%3b", values.Select(EscapeSpecialChars));
+        }
+
+        private static string EscapeSpecialChars(string input)
+        {
+            return input.Replace(";", "%3B");
+        }
```
timcassell, line 76:
```suggestion
```

src/BenchmarkDotNet/Jobs/Argument.cs
```diff
@@ -45,8 +45,34 @@ public MonoArgument(string value) : base(value)
...
     /// </summary>
     [PublicAPI]
+
+    /// <summary>
+    /// A specialized MSBuild argument that sets a property using a quoted, semicolon-separated list of values.
+    /// </summary>
+    public class MsBuildProperty : MsBuildArgument
+    {
+        public MsBuildProperty(string key, params string[] values)
+            : base($"/p:{key}=\"{string.Join(";", values)}")
```
timcassell, line 55: It seems that quotes need to be escaped to work cross-platform.
```suggestion
            : base($"/p:{key}=\\\"{string.Join(";", values)}\\\"")
```

timcassell:
Please add an integration test.

tests/BenchmarkDotNet.IntegrationTests.ManualRunning/MsBuildArgumentTests.cs
```diff
@@ -24,6 +24,13 @@ public void ProcessIsBuiltWithCustomProperty(bool setCustomProperty)
             CanExecute<PropertyDefine>(config);
         }
 
+        [Fact]
+
+        public void MsBuildProperty_ShouldWrapMultipleValuesInQuotes()
```
timcassell, line 29: Please move this test to `BenchmarkDotNet.Tests` project, and add a real integration test in `BenchmarkDotNet.IntegrationTests` project.

timcassell:
Superseded by #3055

Issue #2744, opened 2025-05-31
Allow ArgumentsSource to reference methods in other types


JimmyCushnie:
According to [the docs for `ArgumentsSource`](https://benchmarkdotnet.org/articles/features/parameterization.html#sample-introargumentssource):

> The source must be within benchmarked type!

It would be nice to lift this restriction. The attribute constructor could have an overload that takes a `Type` parameter.

For example, NUnit's [`TestCaseSource`](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html) does this.

A common pattern in my code is that I have one class which defines all the different cases that a method might encounter. I feed the results of this class into both my unit tests and my benchmarks, to test each case for both correctness and performance. Currently, my benchmark code is bloated with methods that just return immediately with the result of a method in the cases class, which I need to do because `ArgumentsSource` can't directly access the method I want to use for the arguments. I would love to be able to remove all those methods.

Example code as it currently is:

```csharp
[Benchmark]
[ArgumentsSource(nameof(MethodACases))]
public MethodAResults BenchmarkMethodA(MethodACase @case)
    => CodeBeingBenchmarked.MethodA(@case.Param1, @case.Param2, @case.Param3);

public static IEnumerable<MethodACase> MethodACases() => DiagnosticsCases.MethodACases();
```

Example code as I propose it ought to be:

```csharp
[Benchmark]
[ArgumentsSource(typeof(DiagnosticsCases), nameof(DiagnosticsCases.MethodACases))]
public MethodAResults BenchmarkMethodA(MethodACase @case)
    => CodeBeingBenchmarked.MethodA(@case.Param1, @case.Param2, @case.Param3);
```

JimmyCushnie:
> The source method would probably have to be static if it's in a different type, though

Looking at the code, I don't believe this is the case. The current implementation just creates a new instance of the target type, and we can do the same with a custom type. It's not like it's using the same instance that the benchmarks are being run on.

https://github.com/dotnet/BenchmarkDotNet/blob/93d12c42c0adbc33702cdc42e1fc494d2c919208/src/BenchmarkDotNet/Running/BenchmarkConverter.cs#L310-L314

timcassell:
[...]

PR #2748, opened 2025-06-03
Allow ArgumentsSource to reference methods in other types

JimmyCushnie:
Implements #2744

I added a new constructor for `ArgumentsSourceAttribute` which takes a custom type. I implemented the logic in `BenchmarkConverter` so that the custom type can be used. The behavior of the existing constructor is preserved.

I updated the docs to explain the new feature.

I renamed a few internal method parameters for clarity.

This is my first contribution to BenchmarkDotNet. Please let me know if I did anything wrong!

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
@@ -770,6 +770,42 @@ namespace BenchmarkDotNet.IntegrationTests
         }
 
         [Theory, MemberData(nameof(GetToolchains))]
+        public void MethodsAndPropertiesFromAnotherClassCanBeUsedAsSources(IToolchain toolchain)
+            => CanExecute<ParamsSourcePointingToAnotherClass>(toolchain);
+
+        public class ParamsSourcePointingToAnotherClass
+        {
+            [ParamsSource(typeof(ExternalClassWithParamsSource), nameof(ExternalClassWithParamsSource.Method))]
+            public int ParamOne { get; set; }
+
+            [ParamsSource(typeof(ExternalClassWithParamsSource), nameof(ExternalClassWithParamsSource.Property))]
+            public int ParamTwo { get; set; }
+
+            [Benchmark]
+            public void Test()
+            {
+                if (ParamOne != 123)
+                    throw new ArgumentException("The ParamOne value is incorrect!");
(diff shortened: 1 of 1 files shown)


timcassell:
Maybe do the same thing for `ParamsSource`?

JimmyCushnie:
Makes sense. Pushed!

timcassell:
Can you also please add or update integration tests?

JimmyCushnie:
I'm having trouble grokking the tests in this repo. I could use some help. Where would be the appropriate place for the new/modified tests? I can't find any existing tests that involve `ArgumentsSource` or `ParamsSource`. The tests for other attributes seem unrelated.

timcassell:
https://github.com/dotnet/BenchmarkDotNet/blob/93d12c42c0adbc33702cdc42e1fc494d2c919208/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
https://github.com/dotnet/BenchmarkDotNet/blob/93d12c42c0adbc33702cdc42e1fc494d2c919208/tests/BenchmarkDotNet.IntegrationTests/ParamsTests.cs

JimmyCushnie:
Thanks very much. Added two new tests to verify the newly-added behavior. All tests are passing locally.

Edit: and I see they're also passing on the CI, nice

tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
```diff
@@ -747,6 +769,42 @@ public void TestProperty(int argument)
             }
         }
 
+        [Theory, MemberData(nameof(GetToolchains))]
+        public void MethodsAndPropertiesFromAnotherClassCanBeUsedAsSources(IToolchain toolchain)
```
timcassell, line 773: Can you please move this to ParamsTest.cs?

JimmyCushnie: Are you sure? All existing tests for `ParamsSource` are in `ArgumentsTests.cs`. `ParamsTest.cs` seems to be used only for `Params`, which is different.

timcassell: Oh I didn't realize. Those tests are all in the wrong place, but it's not a concern for this PR.

PR #2749, opened 2025-06-04
Fix "Unknown Processor" on Windows if WMIC is not present

### LINKED ISSUE #2747: "Unknown processor" on Windows
Hi,
When I run any Benchmark on my Laptop on Windows the Processor Model is not detected by BenchmarkDotNet.

```
BenchmarkDotNet v0.15.0, Windows 11 (10.0.26100.4061/24H2/2024Update/HudsonValley)
Unknown processor
.NET SDK 9.0.300
  [Host]   : .NET 8.0.16 (8.0.1625.21506), X64 RyuJIT AVX2
  .NET 9.0 : .NET 9.0.5 (9.0.525.21509), X64 RyuJIT AVX2

Job=.NET 9.0  Runtime=.NET 9.0
```

0.13.12  behaved the same when I used it before updating to 0.15.0 . 

The processor is: 12th Gen Intel(R) Core(TM) i7-1270P

The Number of Cores is 12
The Number of Logical Processors is 16.

I saw in a previous bug report a wmic output was requested. I couldn't find wmic in the suggested file path: ``C:\Windows\System32\wbem`` and couldn't find any environment variable corresponding to wmic.

Thanks in advance.

alastairlundy:
This PR fixes #2747  by adding the ``WmiLightCpuDetector`` detector class that detects CPU info using the WmiLight library.

This PR also updates ``WmicCpuDetector`` to also check for the presence of WMIC.exe at the expected file path to check for applicability so that it isn't used on systems without WMIC.exe .

The output is as expected when tested. For testing I used the BenchmarkDotNet Samples project.
```
// * Summary *

BenchmarkDotNet v0.15.1-develop (2025-06-04), Windows 11 (10.0.26100.4061/24H2/2024Update/HudsonValley)
12th Gen Intel Core i7-1270P 2.20GHz, 1 CPU, 16 logical and 12 physical cores
.NET SDK 9.0.300
  [Host]     : .NET 8.0.16 (8.0.1625.21506), X64 RyuJIT AVX2 [AttachedDebugger]
  DefaultJob : .NET 8.0.16 (8.0.1625.21506), X64 RyuJIT AVX2
```


alastairlundy:
@dotnet-policy-service agree

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmiLightCpuDetector.cs
```diff
@@ -0,0 +1,70 @@
...
+            };
+        }
+
+#if NET6_0_OR_GREATER
+    [System.Runtime.Versioning.SupportedOSPlatform("windows")]
+#endif
+        public bool IsApplicable()
+        {
+            return OsDetector.IsWindows() && (RuntimeInformation.IsNetCore ||
+                RuntimeInformation.IsFullFramework) && !RuntimeInformation.IsMono;
```
timcassell: Why can't it be used in Mono or other runtimes like NativeAOT?

alastairlundy: > Why can't it be used in Mono or other runtimes like NativeAOT?

For Mono I just assumed it wouldn't work as ``MosCpuDetector`` also has Mono support disabled. I'll enable NativeAOT

alastairlundy: @timcassell I suppose in theory Mono might be able to work. How does BenchmarkDotNet test code against Mono as the Host though?

I tried to run a few BenchmarkDotNet Samples using an exe targetting .NET 4.6.2 with Mono 6.12.0 and kept getting the error:
```
// [BenchmarkSampleName]: DefaultJob
//    * The Roslyn toolchain is only supported on .NET Framework

```

timcassell: Hm, I thought Roslyn toolchain supported Mono host. That might be another bug we could address. You can try it with `[InProcess]` at least.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmiLightCpuDetector.cs
```diff
@@ -0,0 +1,69 @@
...
+                PhysicalProcessorCount = processorsCount > 0 ? processorsCount : null,
+                PhysicalCoreCount = physicalCoreCount > 0 ? physicalCoreCount : null,
+                LogicalCoreCount = logicalCoreCount > 0 ? logicalCoreCount : null,
+                NominalFrequencyHz = maxFrequency?.Hertz.RoundToLong(),
+                MaxFrequencyHz = maxFrequency?.Hertz.RoundToLong()
+            };
+        }
+
+#if NET6_0_OR_GREATER
+    [System.Runtime.Versioning.SupportedOSPlatform("windows")]
```
timcassell:
```suggestion
        [System.Runtime.Versioning.SupportedOSPlatform("windows")]
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmiLightCpuDetector.cs
alastairlundy: > Hm, I thought Roslyn toolchain supported Mono host. That might be another bug we could address. You can try it with `[InProcess]` at least.

[InProcess] does work.

But WmiLight threw an exception:

```
// * Summary *

BenchmarkDotNet v0.15.1-develop (2025-06-05), Windows 11 (10.0.26100.4061/24H2/2024Update/HudsonValley)

Unhandled Exception:
System.DllNotFoundException: WmiLight.Native.dll assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) WmiLight.NativeMethods.CreateWbemLocator(intptr&)
  at WmiLight.Wbem.WbemLocator.CreateWbemLocator () [0x00000] in <5a47e471c51e48a496a00321e7b88e62>:0
  at WmiLight.Wbem.WbemLocator..ctor () [0x00000] in <5a47e471c51e48a496a00321e7b88e62>:0
  at WmiLight.WmiConnection.Open () [0x0001e] in <5a47e471c51e48a496a00321e7b88e62>:0
  at WmiLight.WmiConnection.ExecuteQuery (WmiLight.WmiQuery query) [0x00021] in <5a47e471c51e48a496a00321e7b88e62>:0
  at WmiLight.WmiQuery.GetEnumerator () [0x00000] in <5a47e471c51e48a496a00321e7b88e62>:0
  at BenchmarkDotNet.Detectors.Cpu.Windows.WmiLightCpuDetector.Detect () [0x0002c] in <f4b5039b178b4ab1bd97c724139ac640>:0
  at BenchmarkDotNet.Detectors.CpuDetector+<>c.<Detect>b__8_1 (BenchmarkDotNet.Detectors.Cpu.ICpuDetector loader) [0x00000] in <f4b5039b178b4ab1bd97c724139ac640>:0
  at System.Linq.Enumerable+WhereSelectArrayIterator`2[TSource,TResult].MoveNext () [0x0003a] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at System.Linq.Enumerable+WhereEnumerableIterator`1[TSource].MoveNext () [0x0004e] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at System.Linq.Enumerable.TryGetFirst[TSource] (System.Collections.Generic.IEnumerable`1[T] source, System.Boolean& found) [0x00045] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at System.Linq.Enumerable.FirstOrDefault[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00000] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at BenchmarkDotNet.Detectors.CpuDetector.Detect () [0x00053] in <f4b5039b178b4ab1bd97c724139ac640>:0
  at BenchmarkDotNet.Detectors.CpuDetector+<>c.<Detect>b__8_1 (BenchmarkDotNet.Detectors.Cpu.ICpuDetector loader) [0x00000] in <f4b5039b178b4ab1bd97c724139ac640>:0
  at System.Linq.Enumerable+WhereSelectArrayIterator`2[TSource,TResult].MoveNext () [0x0003a] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at System.Linq.Enumerable+WhereEnumerableIterator`1[TSource].MoveNext () [0x0004e] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at System.Linq.Enumerable.TryGetFirst[TSource] (System.Collections.Generic.IEnumerable`1[T] source, System.Boolean& found) [0x00045] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at System.Linq.Enumerable.FirstOrDefault[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00000] in <54b49a39d6c94f36adb89bde3c275b44>:0
  at BenchmarkDotNet.Detectors.CpuDetector.Detect () [0x00053] in <f4b5039b178b4ab1bd97c724139ac640>:0
  at BenchmarkDotNet.Environments.HostEnvironmentInfo+<>c.<.ctor>b__40_1 () [0x00005] in <f4b5039b178b4ab1bd97c724139ac640>:0
  at System.Lazy`1[T].ViaFactory (System.Threading.LazyThreadSafetyMode mode) [0x0001c] in <b9aa46ca7e704286b432955093e83a10>:0
--- End of stack trace from previous location where exception was thrown ---
```

I didn't find a WmiLight.Native.Dll file in either the .NET 8 Release folder or the .NET 4.6.2 folder.

I'm re-running that test with [InProcess] on .NET 8 to see if that's the culprit.

alastairlundy: > I didn't find a WmiLight.Native.Dll file in either the .NET 8 Release folder or the .NET 4.6.2 folder.
> 
> I'm re-running that test with [InProcess] on .NET 8 to see if that's the culprit.

That same test worked fine on .NET 8, so I think WmiLight might just have an issue with Mono. 
```
// * Summary *

BenchmarkDotNet v0.15.1-develop (2025-06-05), Windows 11 (10.0.26100.4061/24H2/2024Update/HudsonValley)
12th Gen Intel Core i7-1270P 2.20GHz, 1 CPU, 16 logical and 12 physical cores
.NET SDK 9.0.300
  [Host] : .NET 8.0.16 (8.0.1625.21506), X64 RyuJIT AVX2 [AttachedDebugger]

Job=InProcess  Toolchain=InProcessEmitToolchain

| Method                  | Mean     | Error    | StdDev    | Median   |
|------------------------ |---------:|---------:|----------:|---------:|
| 'new byte[10kB]'        | 642.6 ns | 39.14 ns | 111.66 ns | 657.6 ns |
| 'stackalloc byte[10kB]' | 131.4 ns |  7.36 ns |  20.75 ns | 137.2 ns |
```

alastairlundy: @timcassell  I've opened [an issue](https://github.com/MartinKuschnik/WmiLight/issues/54) in WmiLight to see if WmiLight has an issue with Mono. In the meantime I think it's worth keeping Mono support disabled for ``WmiLightCpuDetector``and possibly re-enabling it at a future date if it works then.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
```diff
@@ -22,6 +22,7 @@
     <PackageReference Include="Perfolizer" Version="[0.5.3]" />
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.1.21" PrivateAssets="contentfiles;analyzers" />
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="3.1.6" />
+    <PackageReference Include="WmiLight" Version="6.13.0" />
```
AndreyAkinshin: The approach is generally reasonable, but I’m concerned about introducing a new dependency. We strive to keep the dependency footprint minimal to avoid potential conflicts with user applications that might reference a different version of the same package. While adding a dependency is acceptable when strictly necessary, it’s still something we try to avoid.

In this case, the benefit seems marginal: we’re addressing a rare scenario (Windows systems without wmic) by pulling in a third-party library. That feels excessive. I’d prefer a simpler solution that invokes a command-line tool directly, without adding any dependencies.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuDetector.cs
```diff
@@ -13,7 +13,7 @@ internal class WmicCpuDetector : ICpuDetector
 {
     private const string DefaultWmicPath = @"C:\Windows\System32\wbem\WMIC.exe";
 
-    public bool IsApplicable() => OsDetector.IsWindows();
+    public bool IsApplicable() => OsDetector.IsWindows() && File.Exists(DefaultWmicPath);
```
AndreyAkinshin: This check is too restrictive. `wmic` is not guaranteed to reside at the default path — it can be relocated or made available via `PATH`. By hardcoding the path check, we lose the ability to resolve `wmic` through the environment, which is how it currently works. This change would cause false negatives on systems where `wmic` is available but not at the default location.

alastairlundy: > This check is too restrictive. `wmic` is not guaranteed to reside at the default path — it can be relocated or made available via `PATH`. By hardcoding the path check, we lose the ability to resolve `wmic` through the environment, which is how it currently works. 

Okay. I've reverted this change. Can the robustness of the checks be improved some other way? It currently resolves as the detection mechanism that can be used even if Wmic is not installed thus still leading to "Unknown Processor" being detected unless another detector resolves as applicable before it.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
alastairlundy: > The approach is generally reasonable, but I’m concerned about introducing a new dependency. We strive to keep the dependency footprint minimal to avoid potential conflicts with user applications that might reference a different version of the same package. While adding a dependency is acceptable when strictly necessary, it’s still something we try to avoid.
> 
> In this case, the benefit seems marginal: we’re addressing a rare scenario (Windows systems without wmic) by pulling in a third-party library

I don't believe Windows still ships with WMIC by default, and in any case [WMIC is deprecated](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic) and will likely be removed entirely in a future version of Windows.

>  That feels excessive. I’d prefer a simpler solution that invokes a command-line tool directly, without adding any dependencies.

Fair enough. Since Microsoft suggests using Powershell for WMI interaction I'll work on a detector that gets the information through Powershell and scrap the WmiLight related code.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuDetector.cs
AndreyAkinshin: I suspect this unexpected behavior occurs because `WmicCpuDetector` attempts to always return a non-nullable result, even when `wmic` is not installed. By convention, it should return null in case of failure. I believe that returning null when `wmicOutput` is null will resolve the issue.

src/BenchmarkDotNet/BenchmarkDotNet.csproj
AndreyAkinshin: > I don't believe Windows still ships with WMIC by default, and in any case [WMIC is deprecated](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic) and will likely be removed entirely in a future version of Windows.

I wasn't aware of it; thanks for the information. Could you please add a brief note about the current and future state of WMIC to the comments in `WmicCpuDetector`?

alastairlundy: > > I don't believe Windows still ships with WMIC by default, and in any case [WMIC is deprecated](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic) and will likely be removed entirely in a future version of Windows.
> 
> I wasn't aware of it; thanks for the information. Could you please add a brief note about the current and future state of WMIC to the comments in `WmicCpuDetector`?

I've added the note as remarks in the ``WmicCpuDetector`` xml doc comments.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
```diff
@@ -52,6 +52,9 @@ internal static CpuInfo Parse(string? wmicOutput)
             ? Frequency.FromMHz(sumMaxFrequency / processorsCount)
             : null;
 
+        if (string.IsNullOrEmpty(processorName) || processorName.ToLower().Equals("unknown processor"))
```
AndreyAkinshin: There is no need to check for "unknown processor" here — this string never occurs during CPU detection. It's an artifact of post-formatting that appears when no processor name is provided.

alastairlundy: Fixed

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
```diff
@@ -52,6 +52,9 @@ internal static CpuInfo Parse(string? wmicOutput)
             ? Frequency.FromMHz(sumMaxFrequency / processorsCount)
             : null;
 
+        if (string.IsNullOrEmpty(processorName))
```
AndreyAkinshin: Theoretically, a situation where we fail to determine the processorName but successfully identify other properties is valid. It is preferable to return null if `wmicOutput` is null or empty.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
```diff
@@ -0,0 +1,46 @@
...
+            powershell7PlusPath = $"{subDirectory}{Path.DirectorySeparatorChar}pwsh.exe";
+        }
+        const string argList = $"{WmicCpuInfoKeyNames.Name}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfCores}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfLogicalProcessors}, " +
+                               $"{WmicCpuInfoKeyNames.MaxClockSpeed}";
+
+        // Optimistically, use Cross-platform new PowerShell when available but fallback to Windows PowerShell if not available.
+        string powershellPath = File.Exists(windowsPowershellPath) ? powershell7PlusPath : windowsPowershellPath;
+        string output = ProcessHelper.RunAndReadOutput(powershellPath, "Get-CimInstance Win32_Processor -Property " + argList);
```
timcassell: I think this needs to check if the file exists and bail out in case neither PowerShell is installed.

alastairlundy: > I think this needs to check if the file exists and bail out in case neither PowerShell is installed.

I've added a check for this but Windows Powershell should be installed on all copies of Windows except Windows Nano Server. But if it's not installed for some reason, it'll return null now.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            powershell7PlusPath = $"{subDirectory}{Path.DirectorySeparatorChar}pwsh.exe";
+        }
+        const string argList = $"{WmicCpuInfoKeyNames.Name}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfCores}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfLogicalProcessors}, " +
+                               $"{WmicCpuInfoKeyNames.MaxClockSpeed}";
+
+        // Optimistically, use Cross-platform new PowerShell when available but fallback to Windows PowerShell if not available.
+        string powershellPath = File.Exists(windowsPowershellPath) ? powershell7PlusPath : windowsPowershellPath;
```
timcassell: The check is backwards
```suggestion
        string powershellPath = File.Exists(powershell7PlusPath) ? powershell7PlusPath : windowsPowershellPath;
```

timcassell:
Implementation is incomplete. Mark it ready for review when it's ready.

alastairlundy:
@timcassell It's ready for review but I've noticed in the ``WmicCpuDetector`` parser something. Should the Max CPU Frequency be added to, with += rather than assignment, if multiple processors are detected? I'm inclined to say no but wanted to check.

```csharp

            if (processor.TryGetValue(WmicCpuInfoKeyNames.MaxClockSpeed, out string frequencyValue)
                && int.TryParse(frequencyValue, out int frequency)
                && frequency > 0)
            {
                sumMaxFrequency += frequency;
            }
```

timcassell:
It looks like that is used to calculate an "average" across CPUs `sumMaxFrequency / processorsCount`. I don't understand what the purpose of that actually is, as it seems like a meaningless value to me. But @AndreyAkinshin probably has a better grasp of it.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
```diff
@@ -12,8 +12,11 @@ internal static class WmicCpuInfoParser
     /// Parses wmic output and returns <see cref="CpuInfo"/>
     /// </summary>
     /// <param name="wmicOutput">Output of `wmic cpu get Name, NumberOfCores, NumberOfLogicalProcessors /Format:List`</param>
-    internal static CpuInfo Parse(string? wmicOutput)
+    internal static CpuInfo? Parse(string? wmicOutput)
     {
+        if (string.IsNullOrEmpty(wmicOutput))
+            return null;
```
timcassell: It's probably better to move the check to the WmicCpuDetector to check if wmic exists before starting the process (same as checking if powershell exists). [Edit] Can't hurt to have both checks.

AndreyAkinshin:
Estimating the CPU's maximum frequency is not always straightforward. One of the main challenges is that some modern CPUs consist of heterogeneous cores — that is, cores with different nominal and maximum frequencies. To represent this accurately, we would technically need to track a full list of cores with individual frequency metrics. However, this seems unnecessarily complex for our use case.

First, such a detailed model would usually result in a verbose and redundant dataset — in most real-world scenarios, many cores share identical characteristics. Second, we’d still need to reduce this data to a single representative number for downstream analyses (e.g., in `ZeroMeasurementAnalyzer`), which defeats the purpose of maintaining fine-grained data.

After some consideration, I propose the following pragmatic approach:

- **MaxFrequency** should be defined as the _maximum_ of all reported per-core frequencies. This may not accurately reflect the maximum frequency of all cores (especially low-power ones in heterogeneous setups), but it aligns with our goal of capturing the upper bound of the CPU’s capabilities.

- **NominalFrequency** should be defined as the _minimum_ of all reported frequencies (excluding non-positive or clearly incorrect values). This gives us a baseline that roughly reflects the “typical” operating frequency. When different from `MaxFrequency`, it provides a useful range to understand performance scaling behavior.

What do you think?

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
```diff
@@ -0,0 +1,82 @@
...
+        if (checkForPowershell7Plus)
+        {
+            powershellPath = File.Exists(powershell7PlusPath) ? powershell7PlusPath : windowsPowershellPath;
+        }
+        else
+        {
+            powershellPath = windowsPowershellPath;
+        }
+
+        if (File.Exists(powershellPath) == false)
```
AndreyAkinshin: Knowing the exact path to PowerShell can be beneficial, but we shouldn't assume that we can reliably determine it for all existing and future versions of Windows. If PowerShell isn't found in any of the known locations, we should fall back to invoking it simply as `"PowerShell"`, relying on the system to resolve it via the `PATH` environment variable. This provides a reasonable fallback mechanism without hardcoding assumptions that may not hold across different environments.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
alastairlundy: > It's probably better to move the check to the WmicCpuDetector to check if wmic exists before starting the process (same as checking if powershell exists). [Edit] Can't hurt to have both checks.

If the check is added before the process starts then it doesn't give a chance for the Process Helper to try to resolve the file path from environment variables. Don't we want to have the check after we get the Process output if it's null?

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
alastairlundy: > Knowing the exact path to PowerShell can be beneficial, but we shouldn't assume that we can reliably determine it for all existing and future versions of Windows. If PowerShell isn't found in any of the known locations, we should fall back to invoking it simply as `"PowerShell"`, relying on the system to resolve it via the `PATH` environment variable. This provides a reasonable fallback mechanism without hardcoding assumptions that may not hold across different environments.

Done.

alastairlundy:
> After some consideration, I propose the following pragmatic approach:
> 
>     * **MaxFrequency** should be defined as the _maximum_ of all reported per-core frequencies. This may not accurately reflect the maximum frequency of all cores (especially low-power ones in heterogeneous setups), but it aligns with our goal of capturing the upper bound of the CPU’s capabilities.
> 
>     * **NominalFrequency** should be defined as the _minimum_ of all reported frequencies (excluding non-positive or clearly incorrect values). This gives us a baseline that roughly reflects the “typical” operating frequency. When different from `MaxFrequency`, it provides a useful range to understand performance scaling behavior.
> 
> 
> What do you think?

Makes sense. Should this be worked on in this PR or a new one? And how do we want to format the resulting clockspeeds? Something like: CPU Name, Number of Processors, Logical Processors, Nominal Frequency GHz, Maximum Frequency GHz  ?

AndreyAkinshin:
> Should this be worked on in this PR or a new one? 

Since we are introducing a new detector, it makes sense to implement a solution that addresses this concern. Let's do it in this PR.

> And how do we want to format the resulting clockspeeds?

The detector should not be concerned about the formatting. Its purpose is solely to detect the CPU properties and return a data model.

alastairlundy:
> The detector should not be concerned about the formatting. Its purpose is solely to detect the CPU properties and return a data model.

Understood.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
alastairlundy: I've addressed everything else in this PR. Just waiting on this.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
```diff
@@ -0,0 +1,77 @@
...
+        $"{Environment.SystemDirectory}{Path.DirectorySeparatorChar}WindowsPowerShell{Path.DirectorySeparatorChar}" +
+        $"v1.0{Path.DirectorySeparatorChar}powershell.exe";
+
+    public bool IsApplicable() => OsDetector.IsWindows();
+
+    public CpuInfo? Detect()
+    {
+        if (!IsApplicable()) return null;
+
+        string programFiles = Environment.Is64BitOperatingSystem ? Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles) : Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86);
```
AndreyAkinshin: The logic for determining the PowerShell location deserves its own method, or even a class, if we want to call it from another location.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                processorModelNames.Add(name);
+                processorsCount++;
+            }
+
+            if (processor.TryGetValue(WmicCpuInfoKeyNames.MaxClockSpeed, out string frequencyValue)
+                && int.TryParse(frequencyValue, out int frequency)
+                && frequency > 0)
+            {
+                if (frequency > maxFrequency)
```
AndreyAkinshin: These checks seem too verbose. I believe we can create a much simpler design with something like

```cs
double nominalFrequency = 0;
double maxFrequency = 0;

// frequency update
nominalFrequency = nominalFrequency == 0 ? frequency : Math.Min(nominalFrequency, frequency);
maxFrequency = Math.Max(maxFrequency, frequency);
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
AndreyAkinshin: > If the check is added before the process starts then it doesn't give a chance for the Process Helper to try to resolve the file path from environment variables.

This check can be placed in the `Detect()` method just right after obtaining `wmicOutput` from `ProcessHelper.RunAndReadOutput`

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System.Collections.Generic;
+using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Helpers;
+using Perfolizer.Horology;
+using Perfolizer.Models;
+
+namespace BenchmarkDotNet.Detectors.Cpu.Windows;
+
+internal static class PowershellWmiCpuInfoParser
```
AndreyAkinshin, line 10: Please, add tests like `WmicCpuInfoParserTests`

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
AndreyAkinshin: It's also worth wrapping the entire logic in a try-catch block in case we encounter exceptions due to permission issues with system directories.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        string powershell7PlusPath = $"{programFiles}{Path.DirectorySeparatorChar}Powershell{Path.DirectorySeparatorChar}";
+
+        bool checkForPowershell7Plus = true;
+
+        if (Directory.Exists(powershell7PlusPath))
+        {
+            //Use .Last so that we get the newest major PowerShell version
```
AndreyAkinshin: The comment is inconsistent with the code below, where we take the first element. It is also worth adding an example of the exact names of the directories we can expect within this directory.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
timcassell: Good point, I didn't think about resolving through PATH.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
timcassell: Agreed on using Math.Min/Max. Additionally should check the min frequency is > 0 before overwrite.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
alastairlundy: Updated to 
```
//Use the highest number string directory for PowerShell so that we get the newest major PowerShell version
// Example version directories are 6, 7, and in the future 8.
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
alastairlundy: Done

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
alastairlundy: > Agreed on using Math.Min/Max. Additionally should check the min frequency is > 0 before overwrite.

It is already checked in the if statement above. Did you mean elsewhere?
``` && frequency > 0)```

File: src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs (hunk 13 above)
Comment (alastairlundy) [reply src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs:10]: I've adapted ``WmicCpuInfoParserTests``'s tests based on the format of the Powershell WMI output for PowershellWmiCpuInfoParserTests.

**Comment by AndreyAkinshin** [comment]
LGTM

@alastairlundy could you please resolve the conflict with the master branch

File: src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs [hunk 15]
```diff
@@ -0,0 +1,40 @@
...
+        if (!IsApplicable()) return null;
+
+        const string argList = $"{WmicCpuInfoKeyNames.Name}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfCores}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfLogicalProcessors}, " +
+                               $"{WmicCpuInfoKeyNames.MaxClockSpeed}";
+
+        string output = ProcessHelper.RunAndReadOutput(PowerShellLocator.LocateOnWindows() ?? "PowerShell",
+            "Get-CimInstance Win32_Processor -Property " + argList);
+        return PowershellWmiCpuInfoParser.Parse(output);
```
timcassell, line 42: I think this needs a NullOrEmpty check here similar to Wmic.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    internal static CpuInfo? Parse(string? powershellWmiOutput)
+    {
+        if (string.IsNullOrEmpty(powershellWmiOutput))
+            return CpuInfo.Unknown;
```
timcassell: Not sure why return CpuInfo.Unknown here, others return null?

In any case, with the check moved to the caller, the method signature can be changed to non-null.
```cs
internal static CpuInfo Parse(string powershellWmiOutput)
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
```diff
@@ -12,15 +13,19 @@ internal static class WmicCpuInfoParser
     /// Parses wmic output and returns <see cref="CpuInfo"/>
     /// </summary>
     /// <param name="wmicOutput">Output of `wmic cpu get Name, NumberOfCores, NumberOfLogicalProcessors /Format:List`</param>
-    internal static CpuInfo Parse(string? wmicOutput)
+    internal static CpuInfo? Parse(string? wmicOutput)
     {
-        var processorModelNames = new HashSet<string>();
+        if (string.IsNullOrEmpty(wmicOutput))
+            return CpuInfo.Unknown;
```
timcassell: Same as other, the caller contains the NullOrEmpty check, so can remove it here and change the signature.
```cs
internal static CpuInfo Parse(string wmicOutput)
```

tests/BenchmarkDotNet.Tests/Detectors/Cpu/PowershellWmiCpuInfoParserTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        CpuInfo? actual = PowershellWmiCpuInfoParser
+            .Parse("malformedkey=malformedvalue\n\nmalformedkey2=malformedvalue2");
+        CpuInfo expected = new CpuInfo();
+        Output.AssertEqual(expected, actual);
+    }
+
+    [Fact]
+    public void RealTwoProcessorEightCoresTest()
+    {
+        const string cpuInfo = @"
```
timcassell: Use `"""` for better formatting.

tests/BenchmarkDotNet.Tests/Detectors/Cpu/PowershellWmiCpuInfoParserTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            MaxFrequencyHz = 2_400_000_000,
+        };
+
+        Output.AssertEqual(expected, actual);
+    }
+
+    [Fact]
+    public void RealTwoProcessorEightCoresWithWmicBugTest()
+    {
+        const string cpuInfo =
```
timcassell, line 66: Why is this string formatted differently than the others?

tests/BenchmarkDotNet.Tests/Detectors/Cpu/PowershellWmiCpuInfoParserTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            MaxFrequencyHz = 3_111_000_000,
+        };
+
+        Output.AssertEqual(expected, actual);
+    }
+
+    [Fact]
+    public void RealOneProcessorFourCoresTest()
+    {
+        const string cpuInfo = @"
```
timcassell: `"""`

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
AndreyAkinshin: I agree, it's a bug. The value of `CpuInfo.Unknown` will prevent other detectors from executing. The current logic assumes that if we fail to detect the CPU, we should return `null`.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
```diff
@@ -0,0 +1,71 @@
...
+                processorModelNames.Add(name);
+                processorCount++;
+            }
+
+            if (processor.TryGetValue(WmicCpuInfoKeyNames.MaxClockSpeed, out string frequencyValue)
+                && int.TryParse(frequencyValue, out int frequency)
+                && frequency > 0)
+            {
+               nominalFrequency = nominalFrequency == 0 ? frequency : Math.Min(nominalFrequency, frequency);
+               maxFrequency = maxFrequency == 0 ? frequency : Math.Max(maxFrequency, nominalFrequency);
```
timcassell:
```suggestion
               maxFrequency = Math.Max(maxFrequency, frequency);
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
alastairlundy: > Not sure why return CpuInfo.Unknown here, others return null?

The Unit test for ``EmptyTest`` threw an exception if the null check returned null. Though I can change the behaviour back and edit the unit test code to not throw if that's preferable.

timcassell:
Since the max/nominal frequency calculations are changed here, all the other cpu detectors need to be updated as well.

tests/BenchmarkDotNet.Tests/Detectors/Cpu/PowershellWmiCpuInfoParserTests.cs
alastairlundy: It's how the WmicParserTest formatted the string. I ported it to PowershellWmi parser test.

tests/BenchmarkDotNet.Tests/Detectors/Cpu/PowershellWmiCpuInfoParserTests.cs
alastairlundy: > `"""`

Changed it.

tests/BenchmarkDotNet.Tests/Detectors/Cpu/PowershellWmiCpuInfoParserTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                               NumberOfCores:8
+                               NumberOfLogicalProcessors:16
+
+
+                               MaxClockSpeed:2400
+                               Name:Intel(R) Xeon(R) CPU E5-2630 v3 @ 2.40GHz
+                               NumberOfCores:8
+                               NumberOfLogicalProcessors:16
+
+                               """;
```
timcassell: """ strings match the indentation of the closing """, so you don't need this many indents.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
timcassell: I don't think the test needs to change. The parser should be able to parse unexpected (non-null) strings. The existing wmic test does that.

tests/BenchmarkDotNet.Tests/Detectors/Cpu/PowershellWmiCpuInfoParserTests.cs
alastairlundy: Fixed

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
timcassell: What I'm saying is don't return null here, just remove the check and put it in the caller instead.

alastairlundy: > What I'm saying is don't return null here, just remove the check and put it in the caller instead.

Done.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
alastairlundy: Moved NullOrEmpty check from parser to caller

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
```diff
@@ -0,0 +1,44 @@
...
+        const string argList = $"{WmicCpuInfoKeyNames.Name}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfCores}, " +
+                               $"{WmicCpuInfoKeyNames.NumberOfLogicalProcessors}, " +
+                               $"{WmicCpuInfoKeyNames.MaxClockSpeed}";
+
+        string output = ProcessHelper.RunAndReadOutput(PowerShellLocator.LocateOnWindows() ?? "PowerShell",
+            "Get-CimInstance Win32_Processor -Property " + argList);
+
+        if (string.IsNullOrEmpty(output))
+            return CpuInfo.Unknown;
```
timcassell:
```suggestion
            return null;
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuInfoParser.cs
```diff
@@ -0,0 +1,68 @@
...
+using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Helpers;
+using Perfolizer.Horology;
+using Perfolizer.Models;
+
+namespace BenchmarkDotNet.Detectors.Cpu.Windows;
+
+internal static class PowershellWmiCpuInfoParser
+{
+    internal static CpuInfo? Parse(string? powershellWmiOutput)
```
timcassell:
```suggestion
    internal static CpuInfo Parse(string powershellWmiOutput)
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/MosCpuDetector.cs
```diff
@@ -41,14 +43,20 @@ public bool IsApplicable() => OsDetector.IsWindows() &&
                     processorsCount++;
                     physicalCoreCount += (int)(uint)moProcessor[WmicCpuInfoKeyNames.NumberOfCores];
                     logicalCoreCount += (int)(uint)moProcessor[WmicCpuInfoKeyNames.NumberOfLogicalProcessors];
-                    sumMaxFrequency = (int)(uint)moProcessor[WmicCpuInfoKeyNames.MaxClockSpeed];
+                    double tempMaxFrequency = (double)moProcessor[WmicCpuInfoKeyNames.MaxClockSpeed];
+
+                    nominalFrequency = nominalFrequency == 0 ? maxFrequency : Math.Min(nominalFrequency, maxFrequency);
```
timcassell:
```suggestion
                    if (tempMaxFrequency > 0)
                    {
                        nominalFrequency = nominalFrequency == 0 ? tempMaxFrequency : Math.Min(nominalFrequency, tempMaxFrequency);
                    }
```

alastairlundy:
> Since the max/nominal frequency calculations are changed here, all the other cpu detectors need to be updated as well.

The macOS detector & parser are able to get the nominal frequency already without any averaging so I think that's fine to leave as is.
I copied the Min/Max of nominal and max frequency from ``PowershellWmiCpuDetector``  to ``MosCpuDetector`` just now, and added it before to ``WmicCpuDetector``. So all of the Windows supported Cpu detectors now behave the same (for detecting nominal and max frequency).

The LinuxCpuDetector however seems to follow a completely different methodology for getting nominal clockspeed.

@timcassell Do you want me to update the LinuxCpuDetector to also get nominal clockspeed the same way?

timcassell:
Yeah the Linux one seems strange, but I think it should also be updated to match behavior.

src/BenchmarkDotNet/Detectors/Cpu/Windows/WmicCpuInfoParser.cs
alastairlundy: Done.

src/BenchmarkDotNet/Helpers/PowerShellLocator.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.Versioning;
+using System.Text.RegularExpressions;
+using BenchmarkDotNet.Detectors;
+
+namespace BenchmarkDotNet.Helpers
```
AndreyAkinshin: nit: please, use a file-scoped namespace declaration for new files.

src/BenchmarkDotNet/Detectors/Cpu/Linux/LinuxCpuDetector.cs
```diff
@@ -24,8 +24,12 @@ internal class LinuxCpuDetector : ICpuDetector
             ["LANGUAGE"] = "C"
         };
 
-        string cpuInfo = ProcessHelper.RunAndReadOutput("cat", "/proc/cpuinfo") ?? "";
-        string lscpu = ProcessHelper.RunAndReadOutput("/bin/bash", "-c \"lscpu\"", environmentVariables: languageInvariantEnvironment);
+        string? cpuInfo = ProcessHelper.RunAndReadOutput("cat", "/proc/cpuinfo");
+        string? lscpu = ProcessHelper.RunAndReadOutput("/bin/bash", "-c \"lscpu\"", environmentVariables: languageInvariantEnvironment);
+
+        if (cpuInfo is null || lscpu is null)
+            return null;
```
timcassell, line 31: Not sure if this change is correct. LinuxCpuInfoParser looks like it handles if either is null. Maybe this should be `&&` rather than `||`. @AndreyAkinshin

alastairlundy: > Not sure if this change is correct. LinuxCpuInfoParser looks like it handles if either is null. Maybe this should be `&&` rather than `||`. @AndreyAkinshin

I now think so too. I've changed it to && with a fallback to Empty String if only one is null to allow the parser to not accept null strings.

src/BenchmarkDotNet/Detectors/Cpu/Linux/LinuxCpuInfoParser.cs
```diff
@@ -51,14 +54,22 @@ internal static CpuInfo Parse(string? cpuInfo, string? lscpu)
...
             {
-                maxFrequency = maxCpuFreq;
+                maxFrequency = Math.Max(maxFrequency, maxCpuFreq);
+            }
+
+            if (logicalCore.TryGetValue(ProcCpu.NominalFrequency, out string nominalFreqValue) &&
+                double.TryParse(nominalFreqValue, out double nominalCpuFreq))
+            {
+                if (nominalCpuFreq > 0)
+                    nominalFrequency = Math.Min(nominalFrequency, nominalCpuFreq);
```
timcassell:
```suggestion
            if (logicalCore.TryGetValue(ProcCpu.NominalFrequency, out string nominalFreqValue) &&
                double.TryParse(nominalFreqValue, out double nominalCpuFreq)
                && nominalCpuFreq > 0)
            {
                nominalFrequency = nominalFrequency == 0 ? nominalCpuFreq : Math.Min(nominalFrequency, nominalCpuFreq);
```

timcassell:
Maybe MacOS detector could use a null check, but other than that LGTM.

timcassell:
@alastairlundy Please fix failing tests.

alastairlundy:
> @alastairlundy Please fix failing tests.

The CPU detection and parsing tests pass now.

src/BenchmarkDotNet/Detectors/Cpu/Windows/MosCpuDetector.cs
```diff
@@ -41,14 +43,23 @@ public bool IsApplicable() => OsDetector.IsWindows() &&
                     processorsCount++;
                     physicalCoreCount += (int)(uint)moProcessor[WmicCpuInfoKeyNames.NumberOfCores];
                     logicalCoreCount += (int)(uint)moProcessor[WmicCpuInfoKeyNames.NumberOfLogicalProcessors];
-                    sumMaxFrequency = (int)(uint)moProcessor[WmicCpuInfoKeyNames.MaxClockSpeed];
+                    double tempMaxFrequency = Convert.ToDouble((uint)moProcessor[WmicCpuInfoKeyNames.MaxClockSpeed]);
```
timcassell: Why is `Convert` necessary here? Isn't `uint` implicitly convertible to `double`?

src/BenchmarkDotNet/Detectors/Cpu/Windows/MosCpuDetector.cs
```diff
@@ -41,14 +43,23 @@ public bool IsApplicable() => OsDetector.IsWindows() &&
...
+
+                    if (tempMaxFrequency > 0)
+                    {
+                        nominalFrequency = nominalFrequency == 0 ? Convert.ToDouble(tempMaxFrequency) : Math.Min(nominalFrequency, tempMaxFrequency);
```
timcassell: `Convert` also looks unnecessary here.

alastairlundy: Replaced it with an implicit conversion.

src/BenchmarkDotNet/Detectors/Cpu/Windows/MosCpuDetector.cs
alastairlundy: I've changed it to use implicit conversion

src/BenchmarkDotNet/Detectors/Cpu/Linux/LinuxCpuInfoParser.cs
```diff
@@ -28,12 +31,13 @@ private static class Lscpu
 
     /// <param name="cpuInfo">Output of `cat /proc/cpuinfo`</param>
     /// <param name="lscpu">Output of `lscpu`</param>
-    internal static CpuInfo Parse(string? cpuInfo, string? lscpu)
+    internal static CpuInfo Parse(string cpuInfo, string lscpu)
```
AndreyAkinshin, line 34: This change appears to be incorrect. It is valid for one of these arguments to be null while the other is not. In `LinuxCpuDetector`, we only check the case when both values are null. As a result, we could end up with a null value in a non-nullable string.

AndreyAkinshin:
LGTM. @alastairlundy, please resolve the conflicts with the latest master, and we will be ready for the merge.

alastairlundy:
> LGTM. @alastairlundy, please resolve the conflicts with the latest master, and we will be ready for the merge.

It's resolved

AndreyAkinshin:
@alastairlundy GitHub indicates that there are still conflicts preventing an automatic merge. The easiest way to resolve this issue is to squash your commits, rebase onto the latest master, and force-push your branch.

timcassell:
No problem with squash and merge

PR #2752, opened 2025-06-05
chore: Fix the line endings in the files to LF.

filzrev:
This PR modify files that are using `CRLF` (or `CRLF`/`LF` mixed line endings) to use `LF` line endings only.

**Background**

Files that are committed to git should use only `LF` line endings.
But currently some files seems contains `CRLF` line endings.
And cause diffs when code formatter is applied.


AndreyAkinshin:
@filzrev thanks for the PR! Could you please resolve conflicts with the latest master?

Issue #2745, opened 2025-05-31
Running benchmark with BenchmarkDotNet 15.0 fails with Fatal error. 0xC0000005, WakeLock


christiannagel:
Using version 14.0, the benchmark runs successfully. Updating to version 15, running the benchmark fails with a fatal error:

```
Build succeeded in 3,4s
Fatal error. 0xC0000005
   at BenchmarkDotNet.Running.WakeLock+PInvoke.PowerCreateRequest(REASON_CONTEXT)
   at BenchmarkDotNet.Running.WakeLock+PInvoke.PowerCreateRequest(System.String)
   at BenchmarkDotNet.Running.WakeLock+WakeLockSentinel..ctor(BenchmarkDotNet.Configs.WakeLockType, System.String, BenchmarkDotNet.Loggers.ILogger)
   at BenchmarkDotNet.Running.BenchmarkRunnerClean.Run(BenchmarkDotNet.Running.BenchmarkRunInfo[])
   at BenchmarkDotNet.Running.BenchmarkRunner.RunWithDirtyAssemblyResolveHelper(System.Type, BenchmarkDotNet.Configs.IConfig, System.String[])
   at BenchmarkDotNet.Running.BenchmarkRunner+<>c__DisplayClass0_0`1[[System.__Canon, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].<Run>b__0()
   at BenchmarkDotNet.Running.BenchmarkRunner.RunWithExceptionHandling(System.Func`1<BenchmarkDotNet.Reports.Summary>)
   at BenchmarkDotNet.Running.BenchmarkRunner.Run[[System.__Canon, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](BenchmarkDotNet.Configs.IConfig, System.String[])
   at Program.<Main>$(System.String[])
```

The reason might be I'm using a Surface Copilot device with ARM CPU (Snapdragon X 12-Core X1E80100 @3.40 GHz).

timcassell:
@christiannagel Since you have the device to repro this issue, can you try to fix it and send a PR? https://github.com/dotnet/BenchmarkDotNet/blob/93d12c42c0adbc33702cdc42e1fc494d2c919208/src/BenchmarkDotNet/Running/WakeLock.cs

christiannagel:
[...]

PR #2756, opened 2025-06-07
Refactor WakeLock.PInvoke for compatibility with the ARM CPU #2745

christiannagel:
Refactor `WakeLock.PInvoke` for compatibility with the ARM CPU #2745

Replaced `SimpleReasonString` field from `REASON_CONTEXT` with a union, `REASON_CONTEXT_UNION`.
The ARM CPU is more strict with unions.
Within unions, strings are not allowed, thus using a pointer to a string, with updated allocation within `PowerCreateRequest`.


christiannagel:
@dotnet-policy-service agree

timcassell (approved):
Thanks @christiannagel!

PR #2760, opened 2025-06-08
[build] Add windows-11-arm and ubuntu-24.04-arm to CI run-tests.yaml

nietras:
cc @timcassell

PR #2761, opened 2025-06-08
[build] Add ubuntu-24.04-arm to CI run-tests.yaml

nietras:
cc @timcassell might as well add linux arm too


timcassell:
You can add this to #2760.

nietras:
Added to #2760

PR #2766, opened 2025-06-10
chore: Fix benchmark TestAdapter related issues

filzrev:
This PR intended to fix `BenchmarkDotNet.TestAdapter` related issues.

**What's changed in this PR**
1. Modify `System.Collections.Immutable` baseline version from `9.0.3` -> `9.0.0` and modify related benchmark.
2. Add `AppDomain.CurrentDomain.AssemblyResolver` for .NET 4.6.2 on TestAdapter project to fix assembly loading issue.
3. Remove explicit PackageReference to `System.Memory`  (Use transitively referenced package version instead)

**Background**
`System.Collections.Immutable` is transitively referenced by other packages.
And it cause version conflicts when using TestAdapter. So it need to adjust minimum version to other packages.
(Normally it's not cause issue Because newer version is loaded first. But when running with TestAdapter. It seems `9.0.0` version is loaded and cause conflicts)

And when using TestAdapter on .NET 4.6.2.
It failed to load `BenchmarkDotNet.dll`.
So I've added custom  AssemblyResolve for net462.

I've confirmed this issue on following condigitons.
```
1. Run BenchmarkDotNet.Samples project with following command.
    > dotnet test -c Release --list-tests --framework net462 -tl:off
2. When using `BenchmarkDotNet.TestAdapter` package and targeting .NET Framework.
```

### DIFF
--- a/samples/BenchmarkDotNet.Samples/BenchmarkDotNet.Samples.csproj
+++ b/samples/BenchmarkDotNet.Samples/BenchmarkDotNet.Samples.csproj
@@ -16,11 +16,10 @@
   </PropertyGroup>
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
     <Reference Include="System.Reflection" />
-    <PackageReference Include="System.Memory" Version="4.6.3" />
   </ItemGroup>
   <ItemGroup>
-    <!-- Use v9.0.3 as baseline package for WithNuGet tests -->
-    <PackageReference Include="System.Collections.Immutable" Version="9.0.3" />
+    <!-- Use v9.0.0 as baseline package for WithNuGet tests -->
+    <PackageReference Include="System.Collections.Immutable" Version="9.0.0" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="System.Drawing.Common" Version="9.0.5" />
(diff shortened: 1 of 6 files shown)


timcassell (approved):
We may want to refactor the test adapter to use the new `Microsoft.Testing.Platform` instead of the old `VSTest` at some point.

filzrev:
> We may want to refactor the test adapter to use the new Microsoft.Testing.Platform instead of the old VSTest at some point.

To migrate from `VSTest` to `Microsoft.Testing.Platform`.
It need to migrate to use `xunit.v3` packages.

And  `xunit.v3` 2.x or later [supports following runtimes](https://xunit.net/docs/getting-started/v3/migration) as minimum requirements.

```
.NET Framework 4.7.2 (or later)
.NET 8 (or later)
Our target frameworks for v3 currently are:

.NET Standard 2.0 (netstandard2.0)
.NET Framework 4.7.2 (net472)
.NET 8 (net8.0)
Also new for v3: Mono is officially supported on Linux and macOS for .NET Framework test projects. While it did often work with v1 and v2, we do officially test and verify with it now.
```

So it need to drop legacy runtimes supports when migrating to `Microsoft.Testing.Platform`.
Personally, I think it is reasonable to periodically drop out-of-support runtimes supports for future maintenance cost.

timcassell:
I don't understand. TestAdapter doesn't use xunit, and `Microsoft.Testing.Platform` supports netstandard2.0.

filzrev:
Thanks for your confirmations.
I misunderstood that TestAdapter depends on the xUnit discovery feature.

PR #2769, opened 2025-06-12
chore: Suppress xunit non-serializable data warnings

filzrev:
This PR intended to fix issue that following xUnit warnings are recorded on test discovery.

> BenchmarkDotNet.IntegrationTests: Non-serializable data ('System.Object[]') found for 'BenchmarkDotNet.IntegrationTests.ValuesReturnedByBenchmarkTest.AnyValueCanBeReturned'; falling back to single test case.

This warning is occurred when Theory data contains non serialize/deserialize data by xunit (`IToolchain`, `Jit`, and `Runtime).

It's not needed to display individual test cases on TestExplorer.
So suppress these warnings by specifing `DisableDiscoveryEnumeration=true` parameter.

**Other changes**
~~Add **condition** to reference TestAdapter project when using `Release` configuration only.~~
~~It's required because it failed to load BenchmarkDotNet assembly when build project with `Debug` configuration.~~
~~It's caused by `strong-named assemblies` are required to load assembly.~~
~~(These log can be confirmed by `Tests` log on VS output window)~~

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
@@ -24,7 +24,7 @@ namespace BenchmarkDotNet.IntegrationTests
 
         public ArgumentsTests(ITestOutputHelper output) : base(output) { }
 
-        [Theory, MemberData(nameof(GetToolchains))]
+        [Theory, MemberData(nameof(GetToolchains), DisableDiscoveryEnumeration = true)]
         public void ArgumentsArePassedToBenchmarks(IToolchain toolchain) => CanExecute<WithArguments>(toolchain);
 
         public class WithArguments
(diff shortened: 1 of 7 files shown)


timcassell:
TestAdapter should not have to be removed in Debug configuration. We should rather fix whatever the issue is.

filzrev:
> TestAdapter should not have to be removed in Debug configuration. We should rather fix whatever the issue is.

I've removed `Condition` from this PR.

To fix `Debug` configuration issue.
It need to override  [common.props](https://github.com/filzrev/BenchmarkDotNet/blob/master/build/common.props#L34-L38]) setting for TestAdapter project.

timcassell:
> To fix `Debug` configuration issue. It need to override [common.props](https://github.com/filzrev/BenchmarkDotNet/blob/master/build/common.props#L34-L38%5D) setting for TestAdapter project.

Just set `<SignAssembly>false</SignAssembly>` and leave a comment why, should be fine.

filzrev:
I've removed Condition from this PR. (It's not not pushed)

> Just set <SignAssembly>false</SignAssembly> and leave a comment why, should be fine.

When TestAdapter assembly is loaded by TestExplorer.
Assembly must be signed.
Without assembly signing. It failed to load assembly (It's occurred on .NET 4.6.2 target.)
Error log is recorded on output window's `Tests` pane.

I'll try to create another PR that enable SignAssembly for `Debug` build also.

Issue #2609, opened 2024-08-07
When benchmarking against multiple TFMs including a Native AOT one, BenchmarkDotNet gets confused about framework identifiers


akamsteeg:
(I'm using BenchmarkDotNet v0.14. The issue occurs on Windows, I haven't tested it on Linux or Mac OS.)

I have a `Program.cs` inside a benchmark project that looks like this:
```csharp
public static class Program
{
  public static void Main(string[] args)
  {
    var config = GetConfig();
    BenchmarkSwitcher
      .FromAssembly(Assembly.GetExecutingAssembly())
      .Run(args, config);
  }

  private static ManualConfig GetConfig()
  {
    var job = Job.Default;

    var config = ManualConfig.Create(DefaultConfig.Instance)
      .AddJob(
        job.WithToolchain(CsProjCoreToolchain.NetCoreApp80).AsBaseline(),
        Job.Default.WithRuntime(NativeAotRuntime.Net80),
        job.WithToolchain(CsProjCoreToolchain.NetCoreApp60)
        )
      .AddDiagnoser(MemoryDiagnoser.Default);

    if (Environment.OSVersion.Platform == PlatformID.Win32NT)
    {
      config.AddJob(job.WithToolchain(CsProjClassicNetToolchain.Net481));
    }

    config.SummaryStyle = SummaryStyle.Default
      .WithRatioStyle(RatioStyle.Percentage);

    config.AddValidator(JitOptimizationsValidator.FailOnError); // Fail when any of the referenced assemblies are not optimized

    config.WithOrderer(new DefaultOrderer(SummaryOrderPolicy.FastestToSlowest));

    return config;
  }
}
```

So I'm running benchmarks against .NET 8 as my baseline, .NET 8 with Native AOT, .NET 6 and .NET 4.8.1 when on Windows. When running this using `dotnet build -c Release && dotnet run -c Release -f net8.0` the final report in the console shows this:

```
// * Summary *

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22631.3958/23H2/2023Update/SunValley3)
AMD Ryzen 7 7800X3D, 1 CPU, 16 logical and 8 physical cores
.NET SDK 9.0.100-preview.6.24328.19
  [Host]     : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
[...]

PR #2771, opened 2025-06-12
feat: Add validator for benchmarks that contains null runtime

filzrev:
This PR add validator to handle issue that is reported #2609.

**What's changed in this PR**

1. Add `RuntimeValidator.cs` for benchmarks that have multiple runtimes and contains null.
2. Add RuntimeValidator to DefaultConfig's validator
3. Add unit tests.

**Background**
When benchmark don't contains `Runtime` characteristic. (e.g. When specify `WithToolChain() only`)
[BenchmarkCase::GetRuntime](https://github.com/dotnet/BenchmarkDotNet/blob/c84de9a0d4af419309743c5f86e765de8d490a90/src/BenchmarkDotNet/Running/BenchmarkCase.cs#L31-L33) returns current runtime instead of actual runtime that used by benchmark.

Normally, `Runtime` column is hidden on table because all benchmarks have same runtime value.
So it's not a problem. But when mixed `WithRuntime` settings. it display wrong runtime.

So it need to add validator to check benchmarks that contains multiple runtimes and contains null.

**RuntimeValidator's error message**
> //    * There are benchmarks that job don't have a Runtime characteristic. It's recommended explicitly specify runtime by `WithRuntime`

### DIFF
--- a/src/BenchmarkDotNet/Validators/RuntimeValidator.cs
+++ b/src/BenchmarkDotNet/Validators/RuntimeValidator.cs
@@ -1,7 +1,6 @@
 ﻿using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
-using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Characteristics;
 
 namespace BenchmarkDotNet.Validators;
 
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Validators/RuntimeValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    public IEnumerable<ValidationError> Validate(ValidationParameters input)
+    {
+        var allBenchmarks = input.Benchmarks.ToArray();
+
+        var runtimes = allBenchmarks.Select(x => x.Job.Environment.Runtime)
+                                    .Distinct()
+                                    .ToArray();
+
+        if (runtimes.Length > 1 && runtimes.Contains(null))
```
timcassell: I think we should check if any job contains non-null runtime, and any job contains null runtime. And update the message accordingly.

timcassell:
I think the UX needs to be improved around this, but this is ok as a stop-gap for now.

timcassell (changes requested):
See comment

filzrev:
I've updated runtime validation logics (https://github.com/dotnet/BenchmarkDotNet/pull/2771/commits/457330d655e3c688f14beb34983bb7f32ca0bb05)
Additionally, I've added JobId (or Job text representation) to validation error message.

**Example validation error messages**
```
// Validating benchmarks:
//    * Job(Toolchain=.NET 10.0) don't have a Runtime characteristic. It's recommended to specify runtime by using WithRuntime explicitly.

//    * Job(Dry) don't have a Runtime characteristic. It's recommended to specify runtime by using WithRuntime explicitly.

//    * Job(abc) don't have a Runtime characteristic. It's recommended to specify runtime by using WithRuntime explicitly.
```

src/BenchmarkDotNet/Validators/RuntimeValidator.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        var errors = new List<ValidationError>();
+        foreach (var benchmark in nullRuntimeBenchmarks)
+        {
+            var job = benchmark.Job;
+            var jobText = job.HasValue(CharacteristicObject.IdCharacteristic)
+                ? job.Id
+                : CharacteristicSetPresenter.Display.ToPresentation(job); // Use job text representation instead for auto generated JobId.
+
+            var message = $"Job({jobText}) don't have a Runtime characteristic. It's recommended to specify runtime by using WithRuntime explicitly.";
```
timcassell:
```suggestion
            var message = $"Job({jobText}) doesn't have a Runtime characteristic. It's recommended to specify runtime by using WithRuntime explicitly.";
```

filzrev:
It's not related to this PR thought. `AllocationQuantumIsNotAnIssueForNetCore21Plus` test seems flaky on CI.
Currently it's confirmed on `mac-os (ARM64)`.

```
2025-06-15T04:21:35.2938130Z [xUnit.net 00:09:12.36]     AllocationQuantumIsNotAnIssueForNetCore21Plus(toolchain: .NET 8.0) [FAIL]
2025-06-15T04:21:35.2938900Z [xUnit.net 00:09:12.36]       Assert.Equal() Failure: Values differ
2025-06-15T04:21:35.2939300Z [xUnit.net 00:09:12.36]       Expected: 88
2025-06-15T04:21:35.2939630Z [xUnit.net 00:09:12.36]       Actual:   760
```

timcassell (approved):

timcassell:
Just thought about in-process toolchains after I merged. The current runtime behavior is correct on those toolchains, so they should not be included as part of the validation. @filzrev

PR #2774, opened 2025-06-13
chore: Enable assembly signing for debug build

filzrev:
This PR enable assembly signing for `Debug` build also.

**Background**
When using `BenchmarkDotNet.TestAdapter` and project contains `net462` target.
`net462` benchmarks are not shown on TestExplorer with `Debug` configuration.
And following error log is recorded on VS output window's `Tests` pane.

> System.IO.FileLoadException: A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)'

Currently BenchmarkDotNet sign assembly on `Release` build only.
This PR change this behavior to sign `Debug` build assembly also.

**What's Tested**
- `BenchmarkDotNet.Samples (net462)` benchmarks are shown on TestExplorer 
- It can display benchmarks on TestExplorer when benchmark project don't use assembly-signing.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -31,9 +31,12 @@
     <None Include="$(MSBuildThisFileDirectory)package-icon.png" Pack="True" PackagePath=""/>
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+  <PropertyGroup>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)strongNameKey.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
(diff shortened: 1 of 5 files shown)


AndreyAkinshin:
The current behavior of non-signed Debug builds was introduced in commit 5876f573b1d566d4a9048ef2b78cb8e196eea6b6 on April 12, 2016, by @adamsitnik. I am open to reconsidering this, but I would like to understand the use case better. Why do you want to use the Debug build of BenchmarkDotNet through VS TestExplorer?

P.S. BenchmarkDotNet intentionally creates inconveniences to prevent users from running benchmarks with non-optimized assemblies. Therefore, the absence of discovered tests with the Debug version of BenchmarkDotNet seems to be the desired behavior. 🙂

adamsitnik (approved):
In general the problem you have described is specific to our own solution, as all our customers use the `Release` builds we ship to nuget.org. But I don't mind improving our dev loop experience of course. Thanks for providing the fix @filzrev!

PR #2738, opened 2025-05-25
chore: Update run-tests.yaml

filzrev:
## What's changed in this PR

1. Update third-party GitHub Actions workflows versions and fix version with commit hash.
3. Update Node.js version to `22` (Latest LTS version)
4. Update macOS image version to `macos-15`(x64)

PR #2775, opened 2025-06-13
chore: Modify macos runner image for CI

filzrev:
This PR fix macos runner image issue that unintentionally changed image from x64 to ARM64. (#2738)
`macos15` using ARM64 architecture. It need to use `macos13` image to tests on x64. 
https://github.com/actions/runner-images?tab=readme-ov-file#available-images

**What's changed in this PR**
1. Modify `test-macos` job to use both ARM64/x64 images with matrix.
2. Modify `BenchmarkDotNet.Build.csproj` to output trx log with unit file name.

~~Note:~~ 
~~Artifact zip archive name is outputted as following.~~
~~(It's because GitHub Action don't have build-in functions for toLowerCase)~~  
~~- `test-macos(ARM64)-trx-15629555566`~~
~~- `test-macos(X64)-trx-15629555566`~~

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -91,7 +91,13 @@ jobs:
           path: "**/*.trx"
 
   test-macos:
-    runs-on: macos-15
+    runs-on: ${{ matrix.os.runs-on }}
+    strategy:
+      matrix:
+        os: [
+          { runs-on: 'macos-latest', jsvu-os: 'mac64arm' },
+          { runs-on: 'macos-13',     jsvu-os: 'mac64' },
+        ]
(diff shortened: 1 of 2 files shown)


.github/workflows/run-tests.yaml
```diff
@@ -91,15 +91,25 @@ jobs:
           path: "**/*.trx"
 
   test-macos:
-    runs-on: macos-15
+    name: test-macos(${{ matrix.os.runs-on }})
```
timcassell:
```suggestion
    name: test-macos (${{ matrix.os.arch }})
```

timcassell (approved):

PR #2781, opened 2025-06-15
feat: Add disassembler support for macos

filzrev:
This PR intended to add DisassemblyDiagnoser supports for macos.

Currently DisassemblyDiagnoser is not supported on macos.
This limitation appears to be coming from the ClrMD library.
And It seems to be resolved in the latest version.

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs
@@ -37,18 +37,10 @@ namespace BenchmarkDotNet.IntegrationTests
                 {
                     yield return new object[] { Jit.RyuJit, Platform.X64, CoreRuntime.Core80 }; // .NET Core x64
                 }
-                else if (RuntimeInformation.GetCurrentPlatform() is Platform.Arm64 && OsDetector.IsLinux())
+                else if (RuntimeInformation.GetCurrentPlatform() is Platform.Arm64)
                 {
                     yield return new object[] { Jit.RyuJit, Platform.Arm64, CoreRuntime.Core80 }; // .NET Core arm64
                 }
-            }
-            if (OsDetector.IsMacOS())
-            {
-                // This scope of tests is not supported on macOS
-                // However, when the MemberData method provides no data, xUnit throws an "No data found" InvalidOperationException
-                // In order to fix the problem, we should provide at least one input data set
-                // All the tests check the OS on the first line and stop the test if it's macOS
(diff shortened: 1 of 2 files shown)


timcassell:
You tested it? The issues on ClrMD are still open. https://github.com/microsoft/clrmd/issues/1034 https://github.com/microsoft/clrmd/issues/1230

filzrev:
I've not noticed issues listed above. I confirmed only `https://github.com/microsoft/clrmd/issues/888`

And I'm only confirmed unit tests passed on CI. (Because I'm not have mac)
If it's still not works on macos. Please close this PR.

tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs
```diff
@@ -37,19 +37,11 @@ public static IEnumerable<object[]> GetAllJits()
                 {
                     yield return new object[] { Jit.RyuJit, Platform.X64, CoreRuntime.Core80 }; // .NET Core x64
                 }
-                else if (RuntimeInformation.GetCurrentPlatform() is Platform.Arm64 && OsDetector.IsLinux())
+                else if (RuntimeInformation.GetCurrentPlatform() is Platform.Arm64)
```
filzrev, line 40: Note: This changes affects `windows-11-arm` tests also.

filzrev:
`DisassemblyDiagnoserTests` seems successfully passed on both macos(x64/arm64)

It would be helpful if someone could confirm DisassemblyDiagnoser's behaviors with an actual mac device.
In particular, it would be helpful to verify that the following code paths are working on macOS.
https://github.com/dotnet/BenchmarkDotNet/blob/ed5316b8309c373204a471c57a65e363d399139c/src/BenchmarkDotNet/Disassemblers/ClrMdV3Disassembler.cs#L48-L50

filzrev:
It seems `ShouldUseSameArchitectureDisassembler` method also need to be modified.

https://github.com/dotnet/BenchmarkDotNet/blob/f8390f8ff1af22928d094c58773723fb8b099019/src/BenchmarkDotNet/Disassemblers/DisassemblyDiagnoser.cs#L149-L165

Because macos (on ARM64) support running x86/x64 binary by using Rosetta2.
Rosetta2 is optional feature and it will be not supported in future releases.

So when specified binary has different architecture on macos.
It should be skipped same as previous behavior.

timcassell:
I tested on my MacBook Air M3, the tests passed, but when I tried it with Samples, it just says `No benchmarks were disassembled`. I can't figure why.

timcassell:
Superseded by #2843.

PR #2790, opened 2025-06-22
fix: Console logs are outputted twice when using TestAdapter

filzrev:
This PR intended to fix issue that console and VS Output Window logs are outputted twice when running benchmarks with TestAdapter/ `dotnet test` command.

**What's Tested**

1. When running benchmark from Visual Studio TestExplorer.  
    No duplicated logs outputted on Output Window's `Tests` pane. 
3. When running following command.   
   No duplicated logs outputted. 
    > dotnet test BenchmarkDotNet.sln -c Release --filter Category=Fast -tl:off --framework net8.0 --logger:"console;verbosity=normal"

### DIFF
--- a/src/BenchmarkDotNet/Configs/ManualConfig.cs
+++ b/src/BenchmarkDotNet/Configs/ManualConfig.cs
@@ -327,6 +327,14 @@ namespace BenchmarkDotNet.Configs
             return manualConfig;
         }
 
+        internal ManualConfig RemoveLogger<T>()
+        {
+            var filterdLoggers = loggers.Where(logger => logger is not T).ToArray();
+            loggers.Clear();
+            loggers.AddRange(filterdLoggers);
+            return this;
+        }
+
         internal void RemoveAllJobs() => jobs.Clear();
 
         internal void RemoveAllDiagnosers() => diagnosers.Clear();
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Configs/ManualConfig.cs
```diff
@@ -327,6 +327,14 @@ public static ManualConfig Union(IConfig globalConfig, IConfig localConfig)
             return manualConfig;
         }
 
+        internal ManualConfig RemoveLogger<T>()
+        {
+            var filterdLoggers = loggers.Where(logger => logger is not T).ToArray();
+            loggers.Clear();
+            loggers.AddRange(filterdLoggers);
```
timcassell:
```suggestion
            loggers.RemoveAll(logger => logger is T);
```

src/BenchmarkDotNet/Configs/ManualConfig.cs
```diff
@@ -327,6 +327,14 @@ public static ManualConfig Union(IConfig globalConfig, IConfig localConfig)
             return manualConfig;
         }
 
+        internal ManualConfig RemoveLogger<T>()
```
timcassell:
```suggestion
        internal ManualConfig RemoveLoggersOfType<T>()
```

Issue #2708, opened 2025-03-13
Compilation fails with Literal of type double cannot be implicitly converted.


redknightlois:
After updating to the latest .Net 9.0 SDK and updating Visual Studio we started having issues compiling the solutions. Since we had similar issues on our codebase, maybe it is related to the changes in overload resolution. 

```
S:\Src\ravendb-60-git\bench\Micro.Benchmark\bin\Release\net9.0\e60f7001-ab2b-4959-8145-809b29eb272a\e60f7001-ab2b-4959-8145-809b29eb272a.notcs(17761,69): error CS0664: Literal of type double cannot be implicitly converted to type 'float'; use an 'F' suffix to create a literal of this type [S:\Src\ravendb-60-git\bench\Micro.Benchmark\bin\Release\net9.0\e60f7001-ab2b-4959-8145-809b29eb272a\BenchmarkDotNet.Autogenerated.csproj]
S:\Src\ravendb-60-git\bench\Micro.Benchmark\bin\Release\net9.0\e60f7001-ab2b-4959-8145-809b29eb272a\e60f7001-ab2b-4959-8145-809b29eb272a.notcs(16375,67): error CS0664: Literal of type double cannot be implicitly converted to type 'float'; use an 'F' suffix to create a literal of this type [S:\Src\ravendb-60-git\bench\Micro.Benchmark\bin\Release\net9.0\e60f7001-ab2b-4959-8145-809b29eb272a\BenchmarkDotNet.Autogenerated.csproj]
S:\Src\ravendb-60-git\bench\Micro.Benchmark\bin\Release\net9.0\e60f7001-ab2b-4959-8145-809b29eb272a\e60f7001-ab2b-4959-8145-809b29eb272a.notcs(17167,69): error CS0664: Literal of type double cannot be implicitly converted to type 'float'; use an 'F' suffix to create a literal of this type [S:\Src\ravendb-60-git\bench\Micro.Benchmark\bin\Release\net9.0\e60f7001-ab2b-4959-8145-809b29eb272a\BenchmarkDotNet.Autogenerated.csproj]
[...]

PR #2791, opened 2025-06-22
Cast value if it does not match the ParameterType

pr0s3q:
Fix the problem if the build fails due to implicit casting of Params
https://github.com/dotnet/BenchmarkDotNet/issues/2708

### DIFF
--- a/src/BenchmarkDotNet/Validators/ParamsAllValuesValidator.cs
+++ b/src/BenchmarkDotNet/Validators/ParamsAllValuesValidator.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Reflection;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Extensions;
+using BenchmarkDotNet.Parameters;
 
 namespace BenchmarkDotNet.Validators
 {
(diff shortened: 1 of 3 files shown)


pr0s3q:
@dotnet-policy-service agree

src/BenchmarkDotNet/Validators/ParamsAllValuesValidator.cs
```diff
@@ -60,5 +67,15 @@ private ValidationError GetErrorOrDefault(Type parameterType)
                         $"Type {parameterType.Name} cannot be used with [ParamsAllValues], allowed types are: bool, enum types and nullable type for another allowed type.");
             }
         }
+
+        private ValidationError GetErrorOrDefault(ParameterInstance instance)
+        {
+            if (!instance.Definition.ParameterType.IsInstanceOfType(instance.Value))
```
timcassell, line 73: I'm not convinced that this is sufficient to determine assignability. Test with a property of type `byte`.

pr0s3q: More tests added. All of them are passing the validation check - throwing error if the types are mixed

timcassell: That's a bad test. This should succeed.

```cs
[Params(0, 1, 5, 9, 10)]
public byte ParamProperty { get; set; }
```

This should fail.

```cs
[Params(256, -65536)]
public byte ParamProperty { get; set; }
```

timcassell: I'm thinking the logic to get it correct is probably too complicated for our validators, and we should probably just do nothing here and let the compiler error.

pr0s3q: Your call.
I guess then, the issue #2708 can be closed as designed

PR #2795, opened 2025-06-25
Lowercase programName in NativeMemoryLogParser.cs

### LINKED ISSUE #2794: NativeMemoryProfiler no longer works as of 0.15
In BenchmarkDotNet 0.15, the `Allocated native memory` and `Native memory leak` columns of the NativeMemoryProfiler are always blank. They work correctly in BenchmarkDotNet 0.14.

The issue is that `NativeMemoryLogParser.IsCallStackIn()` performs a case-sensitive search for the name of the EXE in the frame stack, but the module names in the frame stack are always lowercased. Prior to #2423, the name of the EXE was a GUID, so case-sensitivity didn't matter because there were no letters in the EXE name. After that commit, the EXE name now contains some uppercase characters, so `IsCallStackIn()` always returns false.

mriehm:
Fixes #2794


timcassell (approved):
Thanks @mriehm

Issue #2383, opened 2023-07-27
`DisassemblyDiagnoser` does not work in `InProcessEmitToolchain`


timcassell:
Trying to use `DisassemblyDiagnoser` with `InProcessEmitToolchain`, I get an error:

```cmd
// ! InProcessEmitExecutor, exception: System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.ThrowHelper.ThrowNoMatchException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at BenchmarkDotNet.Disassemblers.ClrMdV2Disassembler.AttachAndDisassemble(Settings settings)
   at BenchmarkDotNet.Disassemblers.SameArchitectureDisassembler.Disassemble(DiagnoserActionParameters parameters)
   at BenchmarkDotNet.Diagnosers.DisassemblyDiagnoser.Handle(HostSignal signal, DiagnoserActionParameters parameters)
   at BenchmarkDotNet.Diagnosers.CompositeDiagnoser.Handle(HostSignal signal, DiagnoserActionParameters parameters)
   at BenchmarkDotNet.Toolchains.InProcess.InProcessHost.SendSignal(HostSignal hostSignal)
   at BenchmarkDotNet.Engines.HostExtensions.AfterAll(IHost host)
   at BenchmarkDotNet.Toolchains.InProcess.Emit.Implementation.RunnableProgram.Run(BenchmarkId benchmarkId, Assembly partitionAssembly, BenchmarkCase benchmarkCase, IHost host)
   at BenchmarkDotNet.Toolchains.InProcess.Emit.InProcessEmitExecutor.ExecuteCore(IHost host, ExecuteParameters parameters)
Executable C:\Users\Tim\Documents\git\ProtoPromiseBenchmarks\AsynchronousBenchmarks\bin\Release\net6.0\001a3848-5fde-4a5d-a284-00d6519a2aaeEmitted.dll not found
ExitCode != 0 and no results reported
No Workload Results were obtained from the run.
```

The last version where it worked was 0.13.1. It seems `ClrMdV2Disassembler` broke this case in 0.13.2. @adamsitnik

adamsitnik:
I am surprised that it ever worked (in the past we were attaching the debugger in a way it was performing full suspend of the debugged process, suspending yourself == deadlock).

We even have a check for that, but it checks only one of the InProcessToolchains:

[...]

PR #2799, opened 2025-06-29
Fix `IsNetCore` and `IsNativeAOT` for single-file apps without AOT

timcassell:
Fixes https://github.com/dotnet/BenchmarkDotNet/issues/2383#issuecomment-2904791572

### DIFF
--- a/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
@@ -2,22 +2,16 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
-using System.IO;
 using System.Linq;
 using System.Management;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using BenchmarkDotNet.Detectors;
-using BenchmarkDotNet.Detectors.Cpu;
 using BenchmarkDotNet.Environments;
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -173,10 +168,21 @@ private static string GetNetCoreVersion()
...
+                {
+                    return CoreRuntime.TryGetVersion(out var ver) && ver.Major >= 5
+                        ? $".NET {ver} (assembly location unavailable)"
+                        : $".NET Core {ver?.ToString() ?? Unknown} (assembly location unavailable)";
+                }
+
+                var coreclrAssemblyInfo = FileVersionInfo.GetVersionInfo(coreclrLocation);
+                var corefxAssemblyInfo = FileVersionInfo.GetVersionInfo(corefxLocation);
+
+                if (CoreRuntime.TryGetVersion(out var version) && version.Major >= 5)
```
jkotas: .NET Core 3.x is out support for several years. Should these special cases be dropped to make the logic simpler?

timcassell: For now BDN still supports back to .Net Core 2.0. This doesn't seem too complex to warrant dropping support by itself (and I'm not adding complexity here, this already exists).

jkotas: I meant just drop this special code that tries to differentiate between coreclr and corefx versions.

Issue #2800, opened 2025-07-02
Feature Request: Add OpenMetrics Exporter for GitLab CI Metrics Integration


DefinitelyADev:
Hi BenchmarkDotNet team,

I'd like to propose and contribute a new **OpenMetrics exporter** to BenchmarkDotNet. This would allow users to generate performance benchmark output in the [OpenMetrics format](https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md), which is compatible with GitLab's built-in [Metrics Reports feature](https://docs.gitlab.com/ci/testing/metrics_reports/).

### Motivation

GitLab supports ingesting metrics in the OpenMetrics format via a `metrics.txt` artifact. This opens the door for integrating performance benchmarks directly into CI pipelines, making performance regressions first-class citizens of the CI/CD process. BenchmarkDotNet already supports various exporters (CSV, Markdown, HTML, etc.), but not yet OpenMetrics.

Adding an `OpenMetricsExporter` would allow:

* Comparing benchmark results between commits/MRs in GitLab pipelines
* Alerting on performance regressions directly from the GitLab UI
* Seamless integration with DevOps workflows that rely on GitLab CI

### Proposed Scope

* A new `OpenMetricsExporter` implementation
* Support for writing `metrics.txt` files in the OpenMetrics format
* Mapping relevant benchmark measurements (e.g., mean, error, stdev, ops/sec) to OpenMetrics-compatible metric types (e.g., `gauge`)
* Optional: labels for benchmark class/method/config to support filtering and grouping

### Example output (simplified)

```text
# TYPE benchmark_duration_seconds gauge
benchmark_duration_seconds{benchmark="MyBenchmark.MethodA",job="Default"} 0.000123

# TYPE benchmark_ops_per_second gauge
benchmark_ops_per_second{benchmark="MyBenchmark.MethodA",job="Default"} 8120.0
```

### Benefits

* Enables performance regression detection as part of the CI process
* Aligns with modern DevOps practices
* Bridges BenchmarkDotNet with GitLab-native tooling

I'm happy to contribute the implementation and open a PR if the maintainers agree with the direction.

Thanks for your amazing work on this project!

adamsitnik:
Sounds great to me! It sounds like you would not need to introduce any new NuGet dependencies?

DefinitelyADev:
[...]

PR #2801, opened 2025-07-02
feat: add OpenMetrics exporter support and update related tests

DefinitelyADev:
This PR resolves https://github.com/dotnet/BenchmarkDotNet/issues/2800

It adds a new exporter according to the [OpenMetrics](https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md) spec.


DefinitelyADev:
@dotnet-policy-service agree

src/BenchmarkDotNet/Exporters/OpenMetrics/OpenMetricsExporter.cs
```diff
@@ -0,0 +1,305 @@
...
+                .ToArray());
+            return normalized;
+        }
+
+        private static ImmutableSortedDictionary<string, string> BuildLabelDict(Descriptor descriptor, ParameterInstances parameters)
+        {
+            var dict = new SortedDictionary<string, string>
+            {
+                ["method"] = descriptor.WorkloadMethod.Name,
+                ["type"] = descriptor.Type.Name
```
filzrev: There is a case that benchmark type is defined as generics. (https://benchmarkdotnet.org/articles/samples/IntroGenericTypeArguments.html)

So I thought it's better to use [`Descriptor::TypeInfo` property](https://github.com/dotnet/BenchmarkDotNet/blob/4f646d3663a5947cddee2ace92219b1755dee962/src/BenchmarkDotNet/Running/Descriptor.cs#L25) instead of type name.

DefinitelyADev: I fixed this. I also added support for `# UNIT` lines and did some naming improvements.

src/BenchmarkDotNet/Exporters/OpenMetrics/OpenMetricsExporter.cs
```diff
@@ -0,0 +1,322 @@
...
+
+            return hash.ToHashCode();
+        }
+
+        public override string ToString()
+        {
+            string labelStr = labels.Count > 0
+                ? $"{{{string.Join(", ", labels.Select(kvp => $"{kvp.Key}=\"{kvp.Value}\""))}}}"
+                : string.Empty;
+            return $"{Name}{labelStr} {value}";
```
14156686-35d1-4e73-bb5a-fc7b4f32e4ef: Converting `value`  to string this way will create a invalid metrics file if the current CultureInfo is set to a country that uses comma instead of a period as a decimal separator.

DefinitelyADev: Should be ok now. I also added a test case.

timcassell (approved):
Thanks @DefinitelyADev

Issue #2787, opened 2025-06-18
MinIterationTime warning occurred when using `IterationTime:250ms` setting.


filzrev:
I'm trying to using `dotnet/performance` repository's [RecommendedConfig](https://github.com/dotnet/performance/blob/0396fc87beea5ab8c04a22c1946b3e8901544b22/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L35-L40) settings to reduce benchmark times.

But when using `IterationTime:250ms` setting.
Following warning are randomly raised for some benchmarks.

```
// * Warnings *
MinIterationTime
  DummyBenchmark.RunDummyBenchmark: Toolchain=.NET 9.0, IterationTime=250ms, LaunchCount=1, MaxIterationCount=20, MinIterationCount=15, RunStrategy=Throughput, WarmupCount=25 -> The minimum observed iteration time is 74.728ms which is very small. It's recommended to increase it to at least 100ms using more operations.
```

This warning is raised, because iteration time is expected to be around `250ms`.
But it actual execution time is `74.728ms` and it's lower than `MinIterationTimeAnalyser`'s threshold (`100ms`)

**Problem Details**
`IterationTime` setting is used by `WorkloadPilot` phase to [estimate `PerfectInvocationCount`](https://benchmarkdotnet.org/articles/guides/how-it-works.html#pseudocode) 
But from benchmark logs. It seems JIT process is not sufficient and underestimate `PerfectInvocationCount`.   
(It expected to be around `250ms` with 80ops but `75ms` on actual run) 

- It's occurred when using .NET 9 toolchain. And it's not occurred with .NET8 toolchain.
- Increasing warmup count is not works because warmup is executed after `WorkloadPilot` phase

**Benchmark Logs**
> IterationTime=250ms, LaunchCount=1, MaxIterationCount=20, MinIterationCount=15, RunStrategy=Throughput, WarmupCount=25

```
OverheadJitting  1: 1 op, 110200.00 ns, 110.2000 us/op
WorkloadJitting  1: 1 op, 5121600.00 ns, 5.1216 ms/op

OverheadJitting  2: 16 op, 204700.00 ns, 12.7938 us/op
WorkloadJitting  2: 16 op, 61949900.00 ns, 3.8719 ms/op

WorkloadPilot    1: 16 op, 61489600.00 ns, 3.8431 ms/op
WorkloadPilot    2: 80 op, 303822700.00 ns, 3.7978 ms/op

[...]

Issue #1993, opened 2022-04-15
MinIterationTime, WarmupCount, and tiered JIT


AndreyAkinshin:
Currently, I'm working on reducing the default value for `MinIterationTime` (the current value is 500 ms). I believe that we could achieve the same level of accuracy with a lower value of `MinIterationTime` that will allow us to significantly reduce the total benchmarking time. The only issue I observe right now is about the tiered JIT. Let's consider the following benchmark:
```cs
[Config(typeof(MyConfig))]
public class TieredJitIssue
{
    public class MyConfig : ManualConfig
    {
        public MyConfig()
        {
            var baseJob = Job.Default.WithMinIterationTime(TimeInterval.Millisecond).WithId("Base");
            AddJob(baseJob);
            foreach (var warmupCount in new[] {10, 200})
                AddJob(baseJob.WithWarmupCount(warmupCount).WithId("W" + warmupCount)); ;
        }
    }
    
    [Params(1)]
    public int N { get; set; }
    
    [Benchmark]
    public int Foo() => DoIt(N);

    [MethodImpl(MethodImplOptions.NoInlining)]
    private static int DoIt(int m)
    {
        var sum = 0;
        for (var i = 0; i < m; i++)
            sum += i * i;
        return sum;
    }
}
```

Here are the results on my machine:
```
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK=6.0.200
  [Host] : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  Base   : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  W10    : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  W200   : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT

MinIterationTime=1.0000 ms  

| Method |  Job | WarmupCount | N |      Mean |     Error |    StdDev |    Median |
|------- |----- |------------ |-- |----------:|----------:|----------:|----------:|
|    Foo | Base |     Default | 1 | 2.2980 ns | 0.0672 ns | 0.1611 ns | 2.2162 ns |
|    Foo |  W10 |          10 | 1 | 2.4105 ns | 0.0257 ns | 0.0215 ns | 2.4178 ns |
[...]

Issue #1210, opened 2019-07-23
Adjust invocation count during warmup stage


AndreyAkinshin:
Suggested by @kouvel (see https://github.com/dotnet/coreclr/issues/25680#issuecomment-514024669):
> The pilot phase runs early before warmup and perf could be very different in some cases during the pilot phase and after warmup when tiering is enabled. If piloting completes before the main parts of the benchmark are tiered up, the invocation count that is determined could be much lower compared with tiering disabled. Some benchmarks may perform differently at different invocation counts. Perhaps the warmup phase could also adjust invocation counts, and overhead could be measured after warmup (if overhead measurement is dependent on the invocation count)?

PR #2806, opened 2025-07-09
Refactor engine JIT stage

### LINKED ISSUE #2004: Wildly different results from simply swapping the order of benchmarks
Faced with seemingly inconsistent and inscrutable benchmark results, I followed a hunch and swapped the order of appearance of my two benchmark methods.

The following results are consistently repeatable.

#### A then B:
```
|           Method |     Mean |     Error |    StdDev | Ratio |
|----------------- |---------:|----------:|----------:|------:|
|          MethodA | 1.541 ns | 0.0033 ns | 0.0029 ns |  1.00 |
|          MethodB | 3.451 ns | 0.0079 ns | 0.0074 ns |  2.24 |
```
#### Swap the order:
```
|           Method |      Mean |     Error |    StdDev | Ratio | RatioSD |
|----------------- |----------:|----------:|----------:|------:|--------:|
|          MethodB | 4.9731 ns | 0.0951 ns | 0.1057 ns | 21.52 |    0.64 |
|          MethodA | 0.2302 ns | 0.0024 ns | 0.0020 ns |  1.00 |    0.00 |
```

Nothing has changed between these two runs, aside from swapping the order of the two benchmark methods on the benchmark class.

How should I diagnose what is going on here?

<details>
<summary>Environment info:</summary>

```
BenchmarkDotNet=v0.13.1
AMD Ryzen 9 5900HX with Radeon Graphics, 1 CPU, 16 logical and 8 physical cores
.NET SDK=6.0.202
  [Host]     : .NET 6.0.4 (6.0.422.16404), X64 RyuJIT
  Job-MPHSCA : .NET 6.0.4 (6.0.422.16404), X64 RyuJIT

Runtime=.NET 6.0
```
I'm running Windows 11, but the output reports Windows 10, so I have stripped that.
</details>

timcassell:
Fixes #2004
Contributes to #2787, #1993, #1780, #1210

- Moved JIT stage from `EngineFactory` to a proper `EngineStage`.
- Moved heuristic from `EngineFactory` to a new pilot stage.
- Fixed the heuristic to never include the first invocation.
- Cleanup around the `Engine` (breaking changes).


src/BenchmarkDotNet/Engines/EngineJitStage.cs
```diff
@@ -0,0 +1,246 @@
...
+        // This should succeed for 99%+ of microbenchmarks. For any sufficiently short benchmarks where this fails,
+        // the following stages (Pilot and Warmup) will likely take it the rest of the way. Long-running benchmarks may never fully reach tier1.
+        private IEnumerator<IterationData> EnumerateIterations()
+        {
+            ++iterationIndex;
+            yield return GetDummyIterationData(dummy1Action);
+            yield return GetOverheadNoUnrollIterationData();
+            yield return GetDummyIterationData(dummy2Action);
+            yield return GetWorkloadNoUnrollIterationData();
+            yield return GetDummyIterationData(dummy3Action);
```
timcassell: @AndreyAkinshin You added dummy actions in 2017. I don't know what they are for. Do we still need them?

timcassell:
cc @AndyAyersMS @egorbo

EgorBo:
> JIT stage now attempts to push the benchmarked method through all JIT tiers.
> Set environment variable for the runtime to enable aggressive tiering by default.

Honestly, I think you shouldn't use `TC_AggressiveTiering`, just 1 iteration to promote to Tier1 is mostly just for internal testing. I think `CallCountingDelayMs=0` should be enough.

timcassell:
> Honestly, I think you shouldn't use `TC_AggressiveTiering`, just 1 iteration to promote to Tier1 is mostly just for internal testing. I think `CallCountingDelayMs=0` should be enough.

Can you elaborate on that? Why would we need more than 1 invocation per tier for throughput benchmarks? 30 invocations is too much for the stage to complete in a timely manner for long-running benchmarks.

Also, I tried `CallCountingDelayMs=0`, but it breaks the disassembler (https://github.com/dotnet/runtime/issues/117339).

EgorBo:
> Can you elaborate on that? 

I think the profile will not be representable (a benchmark may invoke the same method from different places and we don't have context-sensitive profiling yet) + we have optimizations like we intentionally make call counting for some methods smaller so their callers are guaranteed to be promoted later (it's for some internal calls so we can bake final addresses of their Tier1 code versions directly instead of having indirect calls), although, I am mostly concerned about PGO quality.

timcassell:
Thanks, that makes sense. I guess I can remove that env var and just run the jit stage with a timeout, and if it doesn't fully reach tier1, we can allow the pilot/warmup stages to handle it later (#1210).

Can you also verify the logic in `JitInfo.cs`?

EgorBo:
> Thanks, that makes sense. I guess I can remove that env var and just run the jit stage with a timeout, and if it doesn't fully reach tier1

How do you check that? I don't think there is a way to check whether a benchmark and all of its callees are fully warmed up

timcassell:
> > Thanks, that makes sense. I guess I can remove that env var and just run the jit stage with a timeout, and if it doesn't fully reach tier1
> 
> How do you check that? I don't think there is a way to check whether a benchmark and all of its callees are fully warmed up

We don't. We just run a number of invocations based on the configured values retrieved from `JitInfo` and hope for the best. The pilot/warmup stages will have to work with some sort of heuristic to try to determine if tiering caused the measured time to significantly drop.

timcassell:
https://github.com/dotnet/runtime/issues/117787#issuecomment-3085668362
> The "third tier" you see may be OSR, since your method loops a lot and isn't called often.

@AndyAyersMS (to not derail that issue), how can we account for OSR in the jit stage here?

EgorBo:
> [dotnet/runtime#117787 (comment)](https://github.com/dotnet/runtime/issues/117787#issuecomment-3085668362)
> 
> > The "third tier" you see may be OSR, since your method loops a lot and isn't called often.
> 
> @AndyAyersMS (to not derail that issue), how can we account for OSR in the jit stage here?

I think for BDN specifically OSR is just some intermediate tier it doesn't have to care about, it shouldn't impact the Tier0->Tier1 promotion velocity. Since the method is too slow, I guess BDN decided not too call it too many times?

timcassell:
> Since the method is too slow, I guess BDN decided not too call it too many times?

This is purely for the jit stage, where the number of invocations are fixed (in an attempt to push it through all tiers). I'm not sure what the jit thinks is not called enough times. Perhaps because of how the stages work, it only invokes once per iteration, and the jit can't see that the iterations are being ran multiple times? If we called it through the `WorkloadUnroll` method (with `unrollFactor = 16`), the jit would skip the OSR?

timcassell:
> I think for BDN specifically OSR is just some intermediate tier it doesn't have to care about, it shouldn't impact the Tier0->Tier1 promotion velocity.

That's what I thought, but the evidence shows otherwise. It took 60 invocations to fully reach tier1, instead of 30 (DPGO disabled).

AndyAyersMS:
Did you try profiling the example from https://github.com/dotnet/runtime/issues/117787? If not, I can do it soonish.

timcassell:
> Did you try profiling the example from [dotnet/runtime#117787](https://github.com/dotnet/runtime/issues/117787)? If not, I can do it soonish.

Nope, I don't have much experience to know what to look for. If you're going to do it from this branch, add +2 to `remainingTiers` in the jit stage to see results of all tiers. Appreciate it.

src/BenchmarkDotNet/Engines/EngineJitStage.cs
timcassell: I think it is no longer needed with the new implementation, so I removed them. If you disagree, I can revert that commit @AndreyAkinshin.

PR #2808, opened 2025-07-11
Fix EtwProfiler for file paths slightly under 260 chars

### LINKED ISSUE #2807: EtwProfiler traces fail to merge if path approaches 260 chars
When [EtwProfiler] is used, the intended behavior is that, initially, an *.etl file and a *.kernel.etl file are generated and then, upon completion, they are merged into a single *.etl file.

However, if the original *.etl file is between 242 to 248 characters (inclusive), then the *.kernel.etl file is not included in the final merged *.etl file.

This occurs because ArtifactFileNameHelper.GetFilePath() uses an alternative base file name for any of the original *.etl file names that exceed 248 characters (MAX_PATH - "userheap.etl".Length). Since "kernel.etl" is longer than "etl", there is a small range of file name lengths where the original *.kernel.etl file name exceeds 248 characters and so uses the alternative base file name while the original *.etl file name doesn't exceed 248 characters and so doesn't use the alternative base file name. That causes the ETL merge logic to not find the *.kernel.etl file because it assumes that all source *.etl files to be included in the merge will have the same base file name as the original *.etl file.

mriehm:
Fixes #2807


timcassell (approved):
Thanks @mriehm

PR #2814, opened 2025-07-21
Use `--nodeReuse:false`

### LINKED ISSUE #2693: Feature: Add extensibility to the  DotnetCliCommand
It would be very useful if all methods of the DotnetCliCommand were virtual.

Currently, it's impossible to customize the build/restore behaviour of the CsProjCoreToolchain.

timcassell:
Fixes #2693.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
+++ b/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
@@ -218,7 +218,7 @@ namespace BenchmarkDotNet.Toolchains.DotNetCli
         {
             // we use these settings to make sure that MSBuild does the job and simply quits without spawning any long living processes
             // we want to avoid "file in use" and "zombie processes" issues
-            const string NoMsBuildZombieProcesses = "/p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 /p:Deterministic=true";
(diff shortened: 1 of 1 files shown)


src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -218,7 +218,7 @@ private static string GetMandatoryMsBuildSettings(string buildConfiguration)
         {
             // we use these settings to make sure that MSBuild does the job and simply quits without spawning any long living processes
             // we want to avoid "file in use" and "zombie processes" issues
-            const string NoMsBuildZombieProcesses = "/p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 /p:Deterministic=true";
```
adamsitnik, line 221: Is this flag supported by older MSBuild versions?

I just wonder what is going to happen for users who are still using .NET 6 SDK etc.

timcassell: I'm not sure which version added it, but it's definitely supported by sdk 8, and your comment in that issue said not to worry about older unsupported sdks.

timcassell: According to Google it was added in .Net 5.

PR #2828, opened 2025-09-04
Update the naot instruction set support for .NET 10+

tannergooding:
The instruction set support was simplified in the JIT for .NET 10 and so this updates the support to better match said updates while still trying to remain backwards compatible with .NET 6-9.

This resolves https://github.com/dotnet/runtime/issues/119353

### DIFF
--- a/src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
+++ b/src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
@@ -11,38 +11,40 @@ using System.Runtime.Intrinsics.Arm;
 
 namespace BenchmarkDotNet.Detectors.Cpu
 {
+    // based on https://github.com/dotnet/runtime/blob/main/src/coreclr/tools/Common/JitInterface/ThunkGenerator/InstructionSetDesc.txt
     internal static class HardwareIntrinsics
     {
         internal static string GetVectorSize() => Vector.IsHardwareAccelerated ? $"VectorSize={Vector<byte>.Count * 8}" : string.Empty;
 
         internal static string GetShortInfo()
         {
-            if (IsX86Avx512FSupported)
-                return GetShortAvx512Representation();
-            if (IsX86Avx2Supported)
-                return "AVX2";
-            else if (IsX86AvxSupported)
-                return "AVX";
-            else if (IsX86Sse42Supported)
-                return "SSE4.2";
-            else if (IsX86Sse41Supported)
-                return "SSE4.1";
-            else if (IsX86Ssse3Supported)
-                return "SSSE3";
-            else if (IsX86Sse3Supported)
-                return "SSE3";
-            else if (IsX86Sse2Supported)
-                return "SSE2";
-            else if (IsX86SseSupported)
-                return "SSE";
-            else if (IsX86BaseSupported)
-                return "X86Base";
-            else if (IsArmAdvSimdSupported)
-                return "AdvSIMD";
+            if (IsX86BaseSupported)
+            {
+                if (IsX86Avx512Supported)
+                {
+                    return "x86-64-v4";
+                }
+                else if (IsX86Avx2Supported)
+                {
+                    return "x86-64-v3";
+                }
+                else if (IsX86Sse42Supported)
+                {
+                    return "x86-64-v2";
+                }
+                else
+                {
+                    return "x86-64-v1";
+                }
+            }
             else if (IsArmBaseSupported)
-                return "ArmBase";
+            {
+                return "armv8.0-a";
+            }
             else
+            {
                 return GetVectorSize(); // Runtimes prior to .NET Core 3.0 (APIs did not exist so we print non-exact Vector info)
+            }
         }
 
         internal static string GetFullInfo(Platform platform)
@@ -55,32 +57,32 @@ namespace BenchmarkDotNet.Detectors.Cpu
                 {
                     case Platform.X86:
                     case Platform.X64:
-
-                        if (IsX86Avx512FSupported) yield return GetShortAvx512Representation();
-                        else if (IsX86Avx2Supported) yield return "AVX2";
-                        else if (IsX86AvxSupported) yield return "AVX";
-                        else if (IsX86Sse42Supported) yield return "SSE4.2";
-                        else if (IsX86Sse41Supported) yield return "SSE4.1";
-                        else if (IsX86Ssse3Supported) yield return "SSSE3";
-                        else if (IsX86Sse3Supported) yield return "SSE3";
-                        else if (IsX86Sse2Supported) yield return "SSE2";
-                        else if (IsX86SseSupported) yield return "SSE";
-                        else if (IsX86BaseSupported) yield return "X86Base";
-
-                        if (IsX86AesSupported) yield return "AES";
-                        if (IsX86Bmi1Supported) yield return "BMI1";
-                        if (IsX86Bmi2Supported) yield return "BMI2";
-                        if (IsX86FmaSupported) yield return "FMA";
-                        if (IsX86LzcntSupported) yield return "LZCNT";
-                        if (IsX86PclmulqdqSupported) yield return "PCLMUL";
-                        if (IsX86PopcntSupported) yield return "POPCNT";
+                    {
+                        if (IsX86Avx10v2Supported) yield return "AVX10v2";
+                        if (IsX86Avx10v1Supported)
+                        {
+                            yield return "AVX10v1";
+                            yield return "AVX512 BF16+FP16"
+                        }
+                        if (IsX86Avx512v3Supported) yield return "AVX512 BITALG+VBMI2+VNNI+VPOPCNTDQ"
+                        if (IsX86Avx512v2Supported) yield return "AVX512 IFMA+VBMI"
+                        if (IsX86Avx512Supported) yield return "AVX512 F+BW+CD+DQ+VL";
+                        if (IsX86Avx2Supported) yield return "AVX2+BMI1+BMI2+F16C+FMA+LZCNT+MOVBE";
+                        if (IsX86AvxSupported) yield return "AVX";
+                        if (IsX86Sse42Supported) yield return "SSE3+SSSE3+SSE4.1+SSE4.2+POPCNT";
+                        if (IsX86BaseSupported) yield return "X86Base+SSE+SSE2";
+                        if (IsX86AesSupported) yield return "AES+PCLMUL";
                         if (IsX86AvxVnniSupported) yield return "AvxVnni";
                         if (IsX86SerializeSupported) yield return "SERIALIZE";
-                        // TODO: Add MOVBE when API is added.
                         break;
+                    }
+                    
                     case Platform.Arm64:
-                        if (IsArmAdvSimdSupported) yield return "AdvSIMD";
-                        else if (IsArmBaseSupported) yield return "ArmBase";
+                    {
+                        if (IsArmBaseSupported)
+                        {
+                            yield return "ArmBase+AdvSimd";
+                        }
 
                         if (IsArmAesSupported) yield return "AES";
                         if (IsArmCrc32Supported) yield return "CRC32";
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
```diff
@@ -11,38 +11,40 @@
...
-                return "SSE";
-            else if (IsX86BaseSupported)
-                return "X86Base";
-            else if (IsArmAdvSimdSupported)
-                return "AdvSIMD";
+            if (IsX86BaseSupported)
+            {
+                if (IsX86Avx512Supported)
+                {
+                    return "x86-64-v4";
```
tannergooding, line 25: These are the official short names for the ISA groupings.

src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
```diff
@@ -55,32 +57,32 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
...
-
-                        if (IsX86AesSupported) yield return "AES";
-                        if (IsX86Bmi1Supported) yield return "BMI1";
-                        if (IsX86Bmi2Supported) yield return "BMI2";
-                        if (IsX86FmaSupported) yield return "FMA";
-                        if (IsX86LzcntSupported) yield return "LZCNT";
-                        if (IsX86PclmulqdqSupported) yield return "PCLMUL";
-                        if (IsX86PopcntSupported) yield return "POPCNT";
+                    {
+                        if (IsX86Avx10v2Supported) yield return "AVX10v2";
```
tannergooding, line 61: This prints the full set of ISAs supported based on the way we group them in RyuJIT

src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
```diff
@@ -89,71 +91,39 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
...
-            if (IsX86Avx512BWSupported) avx512.Append("+BW");
-            if (IsX86Avx512DQSupported) avx512.Append("+DQ");
-            if (IsX86Avx512FVLSupported) avx512.Append("+VL");
-            if (IsX86Avx512VbmiSupported) avx512.Append("+VBMI");
-
-            return avx512.ToString();
-        }
-
+#pragma warning disable CA2252 // Some APIs require opting into preview features
         internal static bool IsX86BaseSupported =>
```
tannergooding, line 102: These are grouped based on the `DOTNET_Enable*` flags we expose so that we're only querying what can actually be toggled on/off.

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -256,39 +256,77 @@ private IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
...
-                    if (HardwareIntrinsics.IsX86SseSupported) yield return "sse";
-                    if (HardwareIntrinsics.IsX86Sse2Supported) yield return "sse2";
-                    if (HardwareIntrinsics.IsX86Sse3Supported) yield return "sse3";
-                    if (HardwareIntrinsics.IsX86Ssse3Supported) yield return "ssse3";
-                    if (HardwareIntrinsics.IsX86Sse41Supported) yield return "sse4.1";
-                    if (HardwareIntrinsics.IsX86Sse42Supported) yield return "sse4.2";
+                    if (HardwareIntrinsics.IsX86Sse42Supported)
+                    {
+                        yield return "sse4.2";
+                        if (runtimeMoniker <= RuntimeMoniker.NativeAot90) yield return "popcnt";
```
tannergooding: This will need another update for .NET 11, as `sse4.2` is becoming the baseline and so becomes part of "base".

It could likely just be the following:
```suggestion
                        if (runtimeMoniker <= RuntimeMoniker.NativeAot10_0) yield return "sse4.2";
                        if (runtimeMoniker <= RuntimeMoniker.NativeAot90) yield return "popcnt";
```

src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
```diff
@@ -11,38 +11,40 @@
 
 namespace BenchmarkDotNet.Detectors.Cpu
 {
+    // based on https://github.com/dotnet/runtime/blob/main/src/coreclr/tools/Common/JitInterface/ThunkGenerator/InstructionSetDesc.txt
```
timcassell: Permanent links are better.

tannergooding: If this waits a few more days, it can link to the tag for the release, where the code is considered "production ready". For now, this is the closest https://github.com/dotnet/runtime/blob/v10.0.0-preview.7.25380.108/src/coreclr/tools/Common/JitInterface/ThunkGenerator/InstructionSetDesc.txt

A link to a specific sha loses a lot of important/relevant information and doesn't really provide that much additional benefit in comparison.

tannergooding:
```suggestion
    // based on https://github.com/dotnet/runtime/tree/v10.0.0-rc.1.25451.107/src/coreclr/tools/Common/JitInterface/ThunkGenerator/InstructionSetDesc.txt
```

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -236,7 +236,7 @@ private void GenerateReflectionFile(ArtifactsPaths artifactsPaths)
         private string GetCurrentInstructionSet(Platform platform)
             => string.Join(",", GetCurrentProcessInstructionSets(platform));
 
-        // based on https://github.com/dotnet/runtime/blob/ce61c09a5f6fc71d8f717d3fc4562f42171869a0/src/coreclr/tools/Common/JitInterface/CorInfoInstructionSet.cs#L727
+        // based on https://github.com/dotnet/runtime/blob/main/src/coreclr/tools/Common/JitInterface/ThunkGenerator/InstructionSetDesc.txt
```
tannergooding:
```suggestion
        // based on https://github.com/dotnet/runtime/tree/v10.0.0-rc.1.25451.107/src/coreclr/tools/Common/JitInterface/ThunkGenerator/InstructionSetDesc.txt
```

src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
```diff
@@ -55,32 +57,32 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
...
-                        if (IsX86PopcntSupported) yield return "POPCNT";
+                    {
+                        if (IsX86Avx10v2Supported) yield return "AVX10v2";
+                        if (IsX86Avx10v1Supported)
+                        {
+                            yield return "AVX10v1";
+                            yield return "AVX512 BF16+FP16"
+                        }
+                        if (IsX86Avx512v3Supported) yield return "AVX512 BITALG+VBMI2+VNNI+VPOPCNTDQ"
+                        if (IsX86Avx512v2Supported) yield return "AVX512 IFMA+VBMI"
```
timcassell:
```suggestion
                            yield return "AVX512 BF16+FP16";
                        }
                        if (IsX86Avx512v3Supported) yield return "AVX512 BITALG+VBMI2+VNNI+VPOPCNTDQ";
                        if (IsX86Avx512v2Supported) yield return "AVX512 IFMA+VBMI";
```

src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
```diff
@@ -89,71 +91,39 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
...
-
-        internal static bool IsX86Sse2Supported =>
-#if NET6_0_OR_GREATER
+            X86Base.IsSupported &&
+            Sse.IsSupported &&
             Sse2.IsSupported;
 #elif NETSTANDARD
+            GetIsSupported("System.Runtime.Intrinsics.X86.X86Base") &&
+            GetIsSupported("System.Runtime.Intrinsics.X86.Sse")
             GetIsSupported("System.Runtime.Intrinsics.X86.Sse2");
```
timcassell:
```suggestion
            GetIsSupported("System.Runtime.Intrinsics.X86.X86Base") &&
            GetIsSupported("System.Runtime.Intrinsics.X86.Sse") &&
            GetIsSupported("System.Runtime.Intrinsics.X86.Sse2");
```

timcassell:
Still getting compile errors. PTAL

src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
```diff
@@ -55,32 +57,32 @@ static IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
...
+                        if (IsX86AvxSupported) yield return "AVX";
+                        if (IsX86Sse42Supported) yield return "SSE3+SSSE3+SSE4.1+SSE4.2+POPCNT";
+                        if (IsX86BaseSupported) yield return "X86Base+SSE+SSE2";
+                        if (IsX86AesSupported) yield return "AES+PCLMUL";
                         if (IsX86AvxVnniSupported) yield return "AvxVnni";
                         if (IsX86SerializeSupported) yield return "SERIALIZE";
-                        // TODO: Add MOVBE when API is added.
                         break;
+                    }
+
```
tannergooding:
```suggestion

```

src/BenchmarkDotNet/Detectors/Cpu/HardwareIntrinsics.cs
adamsitnik: While I am very happy about the fix and other improvements, I find myself hesitant about the new "display name":

- most engineers know what AVX is, I am not sure about the ISA groupings (at least I was not familiar with those names so far)
- the ISA groupings contain information that is already displayed on the screen: the architecture

```diff
BenchmarkDotNet v0.15.3-develop (2025-09-11), Windows 11 (10.0.26100.6584/24H2/2024Update/HudsonValley)
AMD Ryzen Threadripper PRO 3945WX 12-Cores 3.99GHz, 1 CPU, 24 logical and 12 physical cores
.NET SDK 10.0.100-preview.6.25358.103
-  [Host] : .NET 8.0.20 (8.0.20, 8.0.2025.41914), X64 RyuJIT AVX2
+  [Host] : .NET 8.0.20 (8.0.20, 8.0.2025.41914), X64 RyuJIT x86-64-v3
```

@AndreyAkinshin thoughts?

AndreyAkinshin: I have the same concern: AVX2 is much more convinient for the users unlike x86-64-v3. Since we have enough space in this line, I'm suggesting to print both options.

tannergooding: I'm in a bit of a disagreement here.

The names were formalized several years ago by Intel, AMD, LLVM, and others explicitly because it was becoming too unwieldy. Both to list all the options and because of the rapidly exploding number of combinations that could be specified. The manufacturers largely do not want developers thinking in terms of individual ISAs so much anymore and want them to essentially target "profiles" instead, for simplicity.

This also follows the general model that Arm64 and other manufacturers use as well. You declare your base profile (i.e. `armv8.0-a`) and then optionally list any key extensions on top (i.e. `armv8.0-a + lse`). So the intent is that you squash the complex profiles together (i.e. `x86-64-v4`) and then list key important extensions that aren't part of a defined profile on top (i.e. `x86-64-v4 + APX`).

This is likewise how we've changed the JIT to largely support the feature sets, how the JIT Dumps and JIT Disasm now print the information, and how the configuration knobs for manual disablement work.

We explicitly want developers to stop thinking in terms of "AVX2" or "AVX512" in the common cases, but they still have the ability to print the full ISA list if desired (via `GetFullInfo`)

AndreyAkinshin: I have no objections to using the new notation. However, I want us to maintain backward compatibility with user habits.

> We explicitly want developers to stop thinking in terms of "AVX2" or "AVX512" in the common cases

BenchmarkDotNet does not have this goal. In this project, we want to provide a convenient user experience so that people can quickly gain insights from the information presented based on their existing knowledge. If we decide that we want to promote the new notation in addition to a convenient user experience, it makes even more sense to display both versions. This approach will naturally help people create a mental map without needing to search for details when they encounter unfamiliar labels. We can use the new notation in the first place, but then explicitly specify instruction sets inside.

tannergooding: Isn’t it fairly counterintuitive to go against what the hardware manufacturers, compiler developers, and underlying platform have all agreed is the best for long term UX?

This wasn’t some arbitrary decision, but the result of extensive discussion based around real world pain points and confusion caused by the rapidly expanding complexity of ISAs

———-

Part of the consideration is that typical users don’t understand what things like AVX2 are and those that roughly know often get it confused for meaning the wrong thing. It’s only the people that directly work with intrinsics that actually understand

This is very similar in concept to people understanding the brand name “Windows 8” or “Windows 11” but not understanding how that maps to version numbers (6.2 and 10.0.22000)

The profiles are meant to be explicitly clear and give a point that shows something as being meaningfully greater without getting into the technical nuance that is irrelevant to the typical case — I.e listing things that won’t actually impact codegen or make a meaningful difference; which is an actual “problem” with how BDN lists ISAs today, surfacing info that likely isn’t causing profiling differences by default to typical devs

Particularly with how .NET actually works, this also applies here and simplifies what devs should be considering without removing pertinent info. They see the different baselines we support and do codegen differentiation against, nothing more unless they ask for details

———-

I strongly expect that the distaste here is more an initial visceral reaction to the “cheese being moved” from devs that do have more technical knowledge of the space. Such a reaction would likely be less strong if you had less knowledge of the space or were more ingrained with the space and kept up to date with the best practices and recommendations for dealing with hardware profiles as they’ve evolved over the past 5-10 years (not just for x64 but also for Arm64 and other major cpu vendors who are all unifying on such approaches)

AndreyAkinshin: OK, you convinced me; let's try your way. =)
I still expect that it will cause confusion for some of our users, but it's probably not a big problem.

PR #2832, opened 2025-09-19
Allow override method/property for ParamsSource

neuecc:
Currently, when specifying abstract properties or methods with ParamsSource, an exception occurs with SingleOrDefault:
`System.InvalidOperationException: Sequence contains more than one matching element`

```csharp
BenchmarkRunner.Run<Bench>();

public abstract class BenchBase
{
    public abstract int[] Levels { get; }

    [ParamsSource(nameof(Levels))]
    public int Level { get; set; }

    [Benchmark]
    public void Foo()
    {
    }
}

public class Bench : BenchBase
{
    public override int[] Levels => [1, 2, 3];

    [Benchmark]
    public void Bar()
    {
    }
}
```

This is because `sourceType.GetAllMethods()` and `sourceType.GetAllProperties()` traverse up the parent class hierarchy, enumerating both the overridden property and abstract property with the same name.
And since `GetValidValuesForParamsSource` processes with `SingleOrDefault`, the exception occurs.

Since the enumeration starts from sourceType in order, this should be safe to change to `FirstOrDefault` as the best match.

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkConverter.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkConverter.cs
@@ -308,7 +308,7 @@ namespace BenchmarkDotNet.Running
 
         private static (MemberInfo source, object[] values) GetValidValuesForParamsSource(Type sourceType, string sourceName)
         {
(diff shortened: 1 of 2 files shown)


timcassell:
Can you add an integration test for that?

neuecc:
Thank you for checking quickly!
I've added tests.

Issue #389, opened 2017-03-13
Analyzers for Framework


JasonBock:
I wanted to propose the idea of adding analyzers for Benchmark.NET. There's a couple of rules I was thinking of for this library:

* Make sure any classes that contain `[Benchmark]` methods are not sealed
* Make sure that any `[Benchmark]` methods return a value (see the "Avoid Dead Code Elimination" section here (http://benchmarkdotnet.org/RulesOfBenchmarking.htm) for the reason why you want to do this)
* `[Benchmark methods should not have any parameters

There may be others. My point is, I'd like to see these analyzers for this framework. I'd also be interested in writing them as well.

AndreyAkinshin:
Hey @JasonBock, I like the idea in general, but I have some concerns about the suggested analyzers:

> Make sure any classes that contain `[Benchmark]` methods are not sealed

It's true for the default toolchain, but now we also have `InProcessToolchain` which supports not-sealed classes. And it's impossible to say which toolchain will be used in advance.

> Make sure that any `[Benchmark]` methods return a value 

BenchmarkDotNet supports void methods, and it's not a mistake. Dead code elimination could be a problem only in some cases and I'm not sure that it's possible to write an analyzer which detects all such cases correctly (without huge amount of false-positive results).

> `[Benchmark]` methods should not have any parameters

Yeah, we don't support such methods for now, but I'm planning to support them in 1.0 (see also https://github.com/dotnet/BenchmarkDotNet/issues/60).

adamsitnik:
Hi @JasonBock 

We have something that you could extend to get what you want: it's called [ExecutionValidator](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet.Core/Validators/ExecutionValidator.cs). Validator validates the input (config & types with benchmarks) before running the benchmarks.  It's not enabled by default, you have to configure it on your own. More info can be found [here](http://benchmarkdotnet.org/Configs/Validators.htm)

> Make sure any classes that contain [Benchmark] methods are not sealed

It would be nice to have it

> `[Benchmark methods should not have any parameters

@AndreyAkinshin I believe that Jason could implement it now and once we support methods with parameters we can delete this rule.

ig-sinicyn:
@AndreyAkinshin 

>> Make sure any classes that contain [Benchmark] methods are not sealed

[...]

PR #2837, opened 2025-10-10
Add Roslyn analyzers to detect incorrect usage of BenchmarkDotNet

silkfire:
This PR introduces an extensive set of analyzers that warns the user of incorrect usage of BenchmarkDotNet. This is something that has been asked since 2017 but has yet to be included as of this date. BDN has a set of validators that use reflection to detect errors but they are only triggered after the benchmark code has been compiled and is about to run.

I had the idea to implement this in 2022 but the testing framework back then wasn't trivial to use so I gave up in the end. Today, the Roslyn analyzer testing is completely testing framework-agnostic, making things considerably easier. It's also trivial to add multiple source files, references and framework assemblies in order to test your analyzer precisely the way you want.
All unit tests are implemented using xUnit v2.

With these analyzers, developers can detect errors early and solve them immediately. The descriptions are very clear and succinct, guiding the user and explaining the reasoning behind the specific rule.

Here's a list of currently implemented analyzers. There are still some remaining but I believe this is a good start and covers most common usage errors. The rest is up for grabs and can be added along the way.

| Rule ID | Category | Severity | Title | Description | Comment
-- | -- | -- | -- | -- | --
**BDN1000** | Usage | Error | Benchmark class (or any of its ancestors) has no annotated method(s) | The referenced benchmark class (or any of its inherited classes) must have at least one method annotated with the [Benchmark] attribute | Triggers when invoking `BenchmarkRunner.Run<BenchmarkClass>()` and the benchmark class `BenchmarkClass` (or any of its inherited classes) has no public methods  marked with the `[Benchmark]` attribute
**BDN1001** | Usage | Error | Benchmark classes must be public | A benchmark class referenced in the BenchmarkRunner.Run method must be public
**BDN1002** | Usage | Error | Benchmark classes must be unsealed | A benchmark class referenced in the BenchmarkRunner.Run method must be unsealed
**BDN1003** | Usage | Error | Benchmark classes must be non-abstract | A benchmark class referenced in the BenchmarkRunner.Run method must be non-abstract
**BDN1004** | Usage | Error | Generic benchmark classes must be annotated with at least one [GenericTypeArguments] attribute | A generic benchmark class referenced in the BenchmarkRunner.Run method must be annotated with at least one [GenericTypeArguments] attribute
**BDN1100** | Usage | Error | Benchmark classes annotated with a [GenericTypeArguments] attribute must be non-abstract | A benchmark class annotated with a [GenericTypeArguments] attribute must be non-abstract
**BDN1101** | Usage | Error | Benchmark classes annotated with a [GenericTypeArguments] attribute must be generic | A benchmark class annotated with a [GenericTypeArguments] attribute must be generic, having between one to three type parameters
**BDN1102** | Usage | Error | Number of type arguments passed to a [GenericTypeArguments] attribute must match the number of type parameters on the targeted benchmark class | The number of type arguments passed to a [GenericTypeArguments] attribute must match the number of type parameters on the targeted benchmark class
**BDN1103** | Usage | Error | Benchmark methods must be public | A method annotated with the [Benchmark] attribute must be public
**BDN1104** | Usage | Error | Benchmark methods must be non-generic | A method annotated with the [Benchmark] attribute must be non-generic
**BDN1105** | Usage | Error | Benchmark classes must be non-static | A benchmark class must be an instance class
**BDN1106** | Usage | Error | Single null argument to the `[BenchmarkCategory]` attribute results in unintended null array | 
**BDN1107** | Usage | Error | Only one benchmark method can be baseline per class | 
**BDN1108** | Usage | Warning | Only one benchmark method can be baseline per class and category | 
**BDN1200** | Usage | Error | Only one parameter attribute can be applied to a field | Parameter attributes are mutually exclusive; only one of the attributes [Params], [ParamsSource] or [ParamsAllValues] can be applied to a field at any one time
**BDN1201** | Usage | Error | Only one parameter attribute can be applied to a property | Parameter attributes are mutually exclusive; only one of the attributes [Params], [ParamsSource] or [ParamsAllValues] can be applied to a property at any one time
**BDN1202** | Usage | Error | Fields annotated with a parameter attribute must be public | A field annotated with a parameter attribute must be public
**BDN1203** | Usage | Error | Properties annotated with a parameter attribute must be public | A property annotated with a parameter attribute must be public
**BDN1204** | Usage | Error | Fields annotated with a parameter attribute cannot be read-only | Parameter attributes are not valid on fields with a readonly modifier
**BDN1205** | Usage | Error | Parameter attributes are not valid on constant field declarations | Parameter attributes are not valid on constant field declarations
**BDN1206** | Usage | Error | Properties annotated with a parameter attribute cannot have an init-only setter | A property annotated with a parameter attribute must have a public, assignable setter i.e. { set; }
**BDN1207** | Usage | Error | Properties annotated with a parameter attribute must have a public setter | A property annotated with a parameter attribute must have a public setter; make sure that the access modifier of the setter is empty and that the property is not an auto-property or an expression-bodied property.
**BDN1300** | Usage | Error | The [Params] attribute must include at least one value | The [Params] attribute requires at least one value. No values were provided, or an empty array was specified.
**BDN1301** | Usage | Error | Type of all value(s) passed to the [Params] attribute must match the type of (or be implicitly convertible to) the annotated field or property | The type of each value provided to the [Params] attribute must match the type of (or be implicitly convertible to) the field or property it is applied to
**BDN1302** | Usage | Information | Unnecessary single value passed to [Params] attribute | Providing a single value to the [Params] attribute is unnecessary. This attribute is only useful when provided two or more values.
**BDN1303** | Usage | Error | The [ParamsAllValues] attribute cannot be applied to fields or properties of enum types marked with [Flags] | The [ParamsAllValues] attribute cannot be applied to a field or property of an enum type marked with the [Flags] attribute. Use this attribute only with non-flags enum types, as [Flags] enums support bitwise combinations that cannot be exhaustively enumerated.
**BDN1304** | Usage | Error | The [ParamsAllValues] attribute is only valid on fields or properties of enum or bool type and nullable type for another allowed type | The [ParamsAllValues] attribute can only be applied to a field or property of enum or bool type (or nullable of these types)
**BDN1400** | Usage | Error | Benchmark methods without an [ArgumentsSource] or [Arguments] attribute(s) cannot declare parameters | This method declares one or more parameters but is not annotated with either an [ArgumentsSource] attribute or one or more [Arguments] attributes. To ensure correct argument binding, methods with parameters must explicitly be annotated with an [ArgumentsSource] attribute or one or more [Arguments] attributes. Either add the [ArgumentsSource] or [Arguments] attribute(s) or remove the parameters.
**BDN1500** | Usage | Error | [Arguments] attribute can only be used on methods annotated with the [Benchmark] attribute | The [Arguments] attribute can only be used on methods annotated with the [Benchmark] attribute
**BDN1501** | Usage | Error | Number of values passed to an [Arguments] attribute must match the number of parameters declared in the targeted benchmark method | The number of values passed to an [Arguments] attribute must match the number of parameters declared in the targeted benchmark method
**BDN1502** | Usage | Error | Values passed to an [Arguments] attribute must match exactly the parameters declared in the targeted benchmark method in both type (or be implicitly convertible to) and order | The values passed to an [Arguments] attribute must match the parameters declared in the targeted benchmark method in both type (or be implicitly convertible to) and order

### TODO

* Check that `[Arguments]` and `[ArgumentsSource]` attribute usages are mutually exclusive
* Check that the argument to `[ArgumentsSource]` points to a valid method
* Check that the argument to `[ParamsSource]` points to a valid method

See https://github.com/dotnet/BenchmarkDotNet/discussions/2666 for discussion as well as https://github.com/dotnet/BenchmarkDotNet/issues/389.

### DIFF
--- a/src/BenchmarkDotNet.Analyzers/Attributes/ArgumentsAttributeAnalyzer.cs
+++ b/src/BenchmarkDotNet.Analyzers/Attributes/ArgumentsAttributeAnalyzer.cs
@@ -1,333 +1,335 @@
-﻿namespace BenchmarkDotNet.Analyzers.Attributes
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+using System;
+using System.Collections.Immutable;
+
+namespace BenchmarkDotNet.Analyzers.Attributes;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class ArgumentsAttributeAnalyzer : DiagnosticAnalyzer
 {
-    using Microsoft.CodeAnalysis.CSharp;
-    using Microsoft.CodeAnalysis.CSharp.Syntax;
-    using Microsoft.CodeAnalysis;
-    using Microsoft.CodeAnalysis.Diagnostics;
-    using Microsoft.CodeAnalysis.Text;
+    internal static readonly DiagnosticDescriptor RequiresBenchmarkAttributeRule = new(
+        DiagnosticIds.Attributes_ArgumentsAttribute_RequiresBenchmarkAttribute,
+        AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_RequiresBenchmarkAttribute_Title)),
+        AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_RequiresBenchmarkAttribute_MessageFormat)),
+        "Usage",
+        DiagnosticSeverity.Error,
+        isEnabledByDefault: true);
 
-    using System;
-    using System.Collections.Immutable;
+    internal static readonly DiagnosticDescriptor MustHaveMatchingValueCountRule = new(
+        DiagnosticIds.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount,
+        AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount_Title)),
+        AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount_MessageFormat)),
+        "Usage",
+        DiagnosticSeverity.Error,
+        isEnabledByDefault: true,
+        description: AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount_Description)));
 
-    [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    public class ArgumentsAttributeAnalyzer : DiagnosticAnalyzer
+    internal static readonly DiagnosticDescriptor MustHaveMatchingValueTypeRule = new(
+        DiagnosticIds.Attributes_ArgumentsAttribute_MustHaveMatchingValueType,
+        AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueType_Title)),
+        AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueType_MessageFormat)),
+        "Usage",
+        DiagnosticSeverity.Error,
+        isEnabledByDefault: true,
+        description: AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueType_Description)));
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
+    [
+        RequiresBenchmarkAttributeRule,
+        MustHaveMatchingValueCountRule,
+        MustHaveMatchingValueTypeRule
+    ];
+
+    public override void Initialize(AnalysisContext analysisContext)
     {
-        internal static readonly DiagnosticDescriptor RequiresBenchmarkAttributeRule = new DiagnosticDescriptor(DiagnosticIds.Attributes_ArgumentsAttribute_RequiresBenchmarkAttribute,
-                                                                                                                AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_RequiresBenchmarkAttribute_Title)),
-                                                                                                                AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_RequiresBenchmarkAttribute_MessageFormat)),
-                                                                                                                "Usage",
-                                                                                                                DiagnosticSeverity.Error,
-                                                                                                                isEnabledByDefault: true);
+        analysisContext.EnableConcurrentExecution();
+        analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
 
-        internal static readonly DiagnosticDescriptor MustHaveMatchingValueCountRule = new DiagnosticDescriptor(DiagnosticIds.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount,
-                                                                                                                AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount_Title)),
-                                                                                                                AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount_MessageFormat)),
-                                                                                                                "Usage",
-                                                                                                                DiagnosticSeverity.Error,
-                                                                                                                isEnabledByDefault: true,
-                                                                                                                description: AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueCount_Description)));
-
-        internal static readonly DiagnosticDescriptor MustHaveMatchingValueTypeRule = new DiagnosticDescriptor(DiagnosticIds.Attributes_ArgumentsAttribute_MustHaveMatchingValueType,
-                                                                                                               AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueType_Title)),
-                                                                                                               AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueType_MessageFormat)),
-                                                                                                               "Usage",
-                                                                                                               DiagnosticSeverity.Error,
-                                                                                                               isEnabledByDefault: true,
-                                                                                                               description: AnalyzerHelper.GetResourceString(nameof(BenchmarkDotNetAnalyzerResources.Attributes_ArgumentsAttribute_MustHaveMatchingValueType_Description)));
-
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
-        [
-            RequiresBenchmarkAttributeRule,
-            MustHaveMatchingValueCountRule,
-            MustHaveMatchingValueTypeRule
-        ];
-
-        public override void Initialize(AnalysisContext analysisContext)
+        analysisContext.RegisterCompilationStartAction(ctx =>
         {
-            analysisContext.EnableConcurrentExecution();
-            analysisContext.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
-
-            analysisContext.RegisterCompilationStartAction(ctx =>
+            // Only run if BenchmarkDotNet.Annotations is referenced
+            var benchmarkAttributeTypeSymbol = AnalyzerHelper.GetBenchmarkAttributeTypeSymbol(ctx.Compilation);
+            if (benchmarkAttributeTypeSymbol == null)
             {
-                // Only run if BenchmarkDotNet.Annotations is referenced
-                var benchmarkAttributeTypeSymbol = AnalyzerHelper.GetBenchmarkAttributeTypeSymbol(ctx.Compilation);
-                if (benchmarkAttributeTypeSymbol == null)
-                {
-                    return;
-                }
+                return;
+            }
 
-                ctx.RegisterSyntaxNodeAction(AnalyzeMethodDeclaration, SyntaxKind.MethodDeclaration);
-            });
+            ctx.RegisterSyntaxNodeAction(AnalyzeMethodDeclaration, SyntaxKind.MethodDeclaration);
+        });
+    }
+
+    private static void AnalyzeMethodDeclaration(SyntaxNodeAnalysisContext context)
+    {
+        if (context.Node is not MethodDeclarationSyntax methodDeclarationSyntax)
+        {
+            return;
         }
 
-        private static void AnalyzeMethodDeclaration(SyntaxNodeAnalysisContext context)
+        var argumentsAttributeTypeSymbol = context.Compilation.GetTypeByMetadataName("BenchmarkDotNet.Attributes.ArgumentsAttribute");
+        var argumentsSourceAttributeTypeSymbol = context.Compilation.GetTypeByMetadataName("BenchmarkDotNet.Attributes.ArgumentsSourceAttribute");
+
+        if (argumentsAttributeTypeSymbol == null || argumentsSourceAttributeTypeSymbol == null)
         {
-            if (context.Node is not MethodDeclarationSyntax methodDeclarationSyntax)
+            return;
+        }
+
+        var hasBenchmarkAttribute = AnalyzerHelper.AttributeListsContainAttribute(AnalyzerHelper.GetBenchmarkAttributeTypeSymbol(context.Compilation), methodDeclarationSyntax.AttributeLists, context.SemanticModel);
+
+        var argumentsAttributes = AnalyzerHelper.GetAttributes(argumentsAttributeTypeSymbol, methodDeclarationSyntax.AttributeLists, context.SemanticModel);
+        if (argumentsAttributes.Length == 0)
+        {
+            return;
+        }
+
+        if (!hasBenchmarkAttribute)
+        {
+            foreach (var argumentsAttributeSyntax in argumentsAttributes)
             {
-                return;
+                context.ReportDiagnostic(Diagnostic.Create(RequiresBenchmarkAttributeRule, argumentsAttributeSyntax.GetLocation()));
             }
 
-            var argumentsAttributeTypeSymbol = context.Compilation.GetTypeByMetadataName("BenchmarkDotNet.Attributes.ArgumentsAttribute");
-            var argumentsSourceAttributeTypeSymbol = context.Compilation.GetTypeByMetadataName("BenchmarkDotNet.Attributes.ArgumentsSourceAttribute");
+            return;
+        }
 
-            if (argumentsAttributeTypeSymbol == null || argumentsSourceAttributeTypeSymbol == null)
+        var methodParameterTypeSymbolsBuilder = ImmutableArray.CreateBuilder<ITypeSymbol?>(methodDeclarationSyntax.ParameterList.Parameters.Count);
+
+        foreach (var parameterSyntax in methodDeclarationSyntax.ParameterList.Parameters)
+        {
+            if (parameterSyntax.Type != null)
             {
-                return;
-            }
-
-            var hasBenchmarkAttribute = AnalyzerHelper.AttributeListsContainAttribute(AnalyzerHelper.GetBenchmarkAttributeTypeSymbol(context.Compilation), methodDeclarationSyntax.AttributeLists, context.SemanticModel);
-
-            var argumentsAttributes = AnalyzerHelper.GetAttributes(argumentsAttributeTypeSymbol, methodDeclarationSyntax.AttributeLists, context.SemanticModel);
-            if (argumentsAttributes.Length == 0)
-            {
-                return;
-            }
-
-            if (!hasBenchmarkAttribute)
-            {
-                foreach (var argumentsAttributeSyntax in argumentsAttributes)
+                var expectedParameterTypeSymbol = context.SemanticModel.GetTypeInfo(parameterSyntax.Type).Type;
+                if (expectedParameterTypeSymbol != null && expectedParameterTypeSymbol.TypeKind != TypeKind.Error)
                 {
-                    context.ReportDiagnostic(Diagnostic.Create(RequiresBenchmarkAttributeRule, argumentsAttributeSyntax.GetLocation()));
+                    methodParameterTypeSymbolsBuilder.Add(expectedParameterTypeSymbol);
+
+                    continue;
                 }
 
-                return;
+                methodParameterTypeSymbolsBuilder.Add(null);
             }
+        }
 
-            var methodParameterTypeSymbolsBuilder = ImmutableArray.CreateBuilder<ITypeSymbol?>(methodDeclarationSyntax.ParameterList.Parameters.Count);
+        var methodParameterTypeSymbols = methodParameterTypeSymbolsBuilder.ToImmutable();
 
-            foreach (var parameterSyntax in methodDeclarationSyntax.ParameterList.Parameters)
+        foreach (var argumentsAttributeSyntax in argumentsAttributes)
+        {
+            if (argumentsAttributeSyntax.ArgumentList == null)
             {
-                if (parameterSyntax.Type != null)
+                if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
                 {
-                    var expectedParameterTypeSymbol = context.SemanticModel.GetTypeInfo(parameterSyntax.Type).Type;
-                    if (expectedParameterTypeSymbol != null && expectedParameterTypeSymbol.TypeKind != TypeKind.Error)
+                    ReportMustHaveMatchingValueCountDiagnostic(argumentsAttributeSyntax.GetLocation(), 0);
+                }
+            }
+            else if (!argumentsAttributeSyntax.ArgumentList.Arguments.Any())
+            {
+                if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
+                {
+                    ReportMustHaveMatchingValueCountDiagnostic(argumentsAttributeSyntax.ArgumentList.GetLocation(), 0);
+                }
+            }
+            else
+            {
+                // Check if this is an explicit params array creation
+
+                var attributeArgumentSyntax = argumentsAttributeSyntax.ArgumentList.Arguments.First();
+                if (attributeArgumentSyntax.NameEquals != null)
+                {
+                    // Ignore named arguments, e.g. Priority
+                    if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
                     {
-                        methodParameterTypeSymbolsBuilder.Add(expectedParameterTypeSymbol);
+                        ReportMustHaveMatchingValueCountDiagnostic(attributeArgumentSyntax.GetLocation(), 0);
+                    }
+                }
+
+                // Collection expression
+
+                else if (attributeArgumentSyntax.Expression is CollectionExpressionSyntax collectionExpressionSyntax)
+                {
+                    if (methodDeclarationSyntax.ParameterList.Parameters.Count != collectionExpressionSyntax.Elements.Count)
+                    {
+                        ReportMustHaveMatchingValueCountDiagnostic(
+                            collectionExpressionSyntax.Elements.Count == 0
+                                ? collectionExpressionSyntax.GetLocation()
+                                : Location.Create(context.FilterTree, collectionExpressionSyntax.Elements.Span),
+                            collectionExpressionSyntax.Elements.Count
+                        );
 
                         continue;
                     }
 
-                    methodParameterTypeSymbolsBuilder.Add(null);
+                    ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(i => collectionExpressionSyntax.Elements[i] is ExpressionElementSyntax expressionElementSyntax ? expressionElementSyntax.Expression : null);
                 }
-            }
 
-            var methodParameterTypeSymbols = methodParameterTypeSymbolsBuilder.ToImmutable();
-
-            foreach (var argumentsAttributeSyntax in argumentsAttributes)
-            {
-                if (argumentsAttributeSyntax.ArgumentList == null)
+                // Array creation expression
+                else
                 {
-                    if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
+                    var attributeArgumentSyntaxValueType = context.SemanticModel.GetTypeInfo(attributeArgumentSyntax.Expression).Type;
+                    if (attributeArgumentSyntaxValueType is IArrayTypeSymbol arrayTypeSymbol && arrayTypeSymbol.ElementType.SpecialType == SpecialType.System_Object)
                     {
-                        ReportMustHaveMatchingValueCountDiagnostic(argumentsAttributeSyntax.GetLocation(), 0);
+                        if (attributeArgumentSyntax.Expression is ArrayCreationExpressionSyntax arrayCreationExpressionSyntax)
+                        {
+                            if (arrayCreationExpressionSyntax.Initializer == null)
+                            {
+                                var rankSpecifierSizeSyntax = arrayCreationExpressionSyntax.Type.RankSpecifiers.First().Sizes.First();
+                                if (rankSpecifierSizeSyntax is LiteralExpressionSyntax literalExpressionSyntax && literalExpressionSyntax.IsKind(SyntaxKind.NumericLiteralExpression))
+                                {
+                                    if (literalExpressionSyntax.Token.Value is 0)
+                                    {
+                                        if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
+                                        {
+                                            ReportMustHaveMatchingValueCountDiagnostic(literalExpressionSyntax.GetLocation(), 0);
+                                        }
+                                    }
+                                }
+                            }
+                            else
+                            {
+                                if (methodDeclarationSyntax.ParameterList.Parameters.Count != arrayCreationExpressionSyntax.Initializer.Expressions.Count)
+                                {
+                                    ReportMustHaveMatchingValueCountDiagnostic(
+                                        arrayCreationExpressionSyntax.Initializer.Expressions.Count == 0
+                                            ? arrayCreationExpressionSyntax.Initializer.GetLocation()
+                                            : Location.Create(context.FilterTree, arrayCreationExpressionSyntax.Initializer.Expressions.Span),
+                                        arrayCreationExpressionSyntax.Initializer.Expressions.Count
+                                    );
+
+                                    continue;
+                                }
+
+                                // ReSharper disable once PossibleNullReferenceException
+                                ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(i => arrayCreationExpressionSyntax.Initializer.Expressions[i]);
+                            }
+                        }
+                    }
+                    else
+                    {
+                        // Params values
+
+                        var firstNamedArgumentIndex = IndexOfNamedArgument(argumentsAttributeSyntax.ArgumentList.Arguments);
+                        if (firstNamedArgumentIndex > 0)
+                        {
+                            if (methodDeclarationSyntax.ParameterList.Parameters.Count != firstNamedArgumentIndex.Value)
+                            {
+                                ReportMustHaveMatchingValueCountDiagnostic(
+                                    Location.Create(
+                                        context.FilterTree,
+                                        TextSpan.FromBounds(argumentsAttributeSyntax.ArgumentList.Arguments.Span.Start, argumentsAttributeSyntax.ArgumentList.Arguments[firstNamedArgumentIndex.Value - 1].Span.End)
+                                    ),
+                                    firstNamedArgumentIndex.Value
+                                );
+
+                                continue;
+                            }
+
+                            // ReSharper disable once PossibleNullReferenceException
+                            ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(i => argumentsAttributeSyntax.ArgumentList.Arguments[i].Expression);
+                        }
+                        else
+                        {
+                            if (methodDeclarationSyntax.ParameterList.Parameters.Count != argumentsAttributeSyntax.ArgumentList.Arguments.Count)
+                            {
+                                ReportMustHaveMatchingValueCountDiagnostic(
+                                    Location.Create(context.FilterTree, argumentsAttributeSyntax.ArgumentList.Arguments.Span),
+                                    argumentsAttributeSyntax.ArgumentList.Arguments.Count
+                                );
+
+                                continue;
+                            }
+
+                            // ReSharper disable once PossibleNullReferenceException
+                            ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(i => argumentsAttributeSyntax.ArgumentList.Arguments[i].Expression);
+                        }
                     }
                 }
-                else if (!argumentsAttributeSyntax.ArgumentList.Arguments.Any())
+            }
+        }
+
+        return;
+
+        void ReportMustHaveMatchingValueCountDiagnostic(Location diagnosticLocation, int valueCount)
+        {
+            context.ReportDiagnostic(Diagnostic.Create(MustHaveMatchingValueCountRule,
+                diagnosticLocation,
+                methodDeclarationSyntax.ParameterList.Parameters.Count,
+                methodDeclarationSyntax.ParameterList.Parameters.Count == 1 ? "" : "s",
+                methodDeclarationSyntax.Identifier.ToString(),
+                valueCount)
+            );
+        }
+
+        void ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(Func<int, ExpressionSyntax> valueExpressionSyntaxFunc)
+        {
+            for (var i = 0; i < methodParameterTypeSymbols.Length; i++)
+            {
+                var methodParameterTypeSymbol = methodParameterTypeSymbols[i];
+                if (methodParameterTypeSymbol == null)
                 {
-                    if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
+                    continue;
+                }
+
+                var valueExpressionSyntax = valueExpressionSyntaxFunc(i);
+                if (valueExpressionSyntax == null)
+                {
+                    continue;
+                }
+
+                var valueExpressionString = valueExpressionSyntax.ToString();
+
+                var constantValue = context.SemanticModel.GetConstantValue(valueExpressionSyntax);
+
+                var actualValueTypeSymbol = context.SemanticModel.GetTypeInfo(valueExpressionSyntax).Type;
+                if (actualValueTypeSymbol != null && actualValueTypeSymbol.TypeKind != TypeKind.Error)
+                {
+                    if (!AnalyzerHelper.IsAssignableToLocal(context.Compilation, methodParameterTypeSymbol, valueExpressionString, constantValue, actualValueTypeSymbol.ToString()))
                     {
-                        ReportMustHaveMatchingValueCountDiagnostic(argumentsAttributeSyntax.ArgumentList.GetLocation(), 0);
+                        ReportValueTypeMustBeImplicitlyConvertibleDiagnostic(
+                            valueExpressionSyntax.GetLocation(),
+                            valueExpressionSyntax.ToString(),
+                            methodParameterTypeSymbol.ToString(),
+                            actualValueTypeSymbol.ToString()
+                        );
                     }
                 }
                 else
                 {
-                    // Check if this is an explicit params array creation
-
-                    var attributeArgumentSyntax = argumentsAttributeSyntax.ArgumentList.Arguments.First();
-                    if (attributeArgumentSyntax.NameEquals != null)
+                    if (constantValue is { HasValue: true, Value: null }
+                        && !AnalyzerHelper.IsAssignableToLocal(context.Compilation, methodParameterTypeSymbol, valueExpressionString, constantValue, null))
                     {
-                        // Ignore named arguments, e.g. Priority
-                        if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
-                        {
-                            ReportMustHaveMatchingValueCountDiagnostic(attributeArgumentSyntax.GetLocation(), 0);
-                        }
-                    }
-
-                    // Collection expression
-
-                    else if (attributeArgumentSyntax.Expression is CollectionExpressionSyntax collectionExpressionSyntax)
-                    {
-                        if (methodDeclarationSyntax.ParameterList.Parameters.Count != collectionExpressionSyntax.Elements.Count)
-                        {
-                            ReportMustHaveMatchingValueCountDiagnostic(collectionExpressionSyntax.Elements.Count == 0
-                                                                          ? collectionExpressionSyntax.GetLocation()
-                                                                          : Location.Create(context.FilterTree, collectionExpressionSyntax.Elements.Span),
-                                                                       collectionExpressionSyntax.Elements.Count);
-
-                            continue;
-                        }
-
-                        ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(i => collectionExpressionSyntax.Elements[i] is ExpressionElementSyntax expressionElementSyntax ? expressionElementSyntax.Expression : null);
-                    }
-
-                    // Array creation expression
-                    else
-                    {
-                        var attributeArgumentSyntaxValueType = context.SemanticModel.GetTypeInfo(attributeArgumentSyntax.Expression).Type;
-                        if (attributeArgumentSyntaxValueType is IArrayTypeSymbol arrayTypeSymbol && arrayTypeSymbol.ElementType.SpecialType == SpecialType.System_Object)
-                        {
-                            if (attributeArgumentSyntax.Expression is ArrayCreationExpressionSyntax arrayCreationExpressionSyntax)
-                            {
-                                if (arrayCreationExpressionSyntax.Initializer == null)
-                                {
-                                    var rankSpecifierSizeSyntax = arrayCreationExpressionSyntax.Type.RankSpecifiers.First().Sizes.First();
-                                    if (rankSpecifierSizeSyntax is LiteralExpressionSyntax literalExpressionSyntax && literalExpressionSyntax.IsKind(SyntaxKind.NumericLiteralExpression))
-                                    {
-                                        if (literalExpressionSyntax.Token.Value is 0)
-                                        {
-                                            if (methodDeclarationSyntax.ParameterList.Parameters.Count > 0)
-                                            {
-                                                ReportMustHaveMatchingValueCountDiagnostic(literalExpressionSyntax.GetLocation(), 0);
-                                            }
-                                        }
-                                    }
-                                }
-                                else
-                                {
-                                    if (methodDeclarationSyntax.ParameterList.Parameters.Count != arrayCreationExpressionSyntax.Initializer.Expressions.Count)
-                                    {
-                                        ReportMustHaveMatchingValueCountDiagnostic(arrayCreationExpressionSyntax.Initializer.Expressions.Count == 0
-                                                                                      ? arrayCreationExpressionSyntax.Initializer.GetLocation()
-                                                                                      : Location.Create(context.FilterTree, arrayCreationExpressionSyntax.Initializer.Expressions.Span),
-                                                                                   arrayCreationExpressionSyntax.Initializer.Expressions.Count);
-
-                                        continue;
-                                    }
-
-                                    // ReSharper disable once PossibleNullReferenceException
-                                    ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(i => arrayCreationExpressionSyntax.Initializer.Expressions[i]);
-                                }
-                            }
-                        }
-                        else
-                        {
-                            // Params values
-
-                            var firstNamedArgumentIndex = IndexOfNamedArgument(argumentsAttributeSyntax.ArgumentList.Arguments);
-                            if (firstNamedArgumentIndex > 0)
-                            {
-                                if (methodDeclarationSyntax.ParameterList.Parameters.Count != firstNamedArgumentIndex.Value)
-                                {
-                                    ReportMustHaveMatchingValueCountDiagnostic(Location.Create(context.FilterTree, TextSpan.FromBounds(argumentsAttributeSyntax.ArgumentList.Arguments.Span.Start, argumentsAttributeSyntax.ArgumentList.Arguments[firstNamedArgumentIndex.Value - 1].Span.End)),
-                                                                               firstNamedArgumentIndex.Value);
-
-                                    continue;
-                                }
-
-                                // ReSharper disable once PossibleNullReferenceException
-                                ReportIfNotImplicitlyConvertibleValueTypeDiagnostic(i => argumentsAttributeSyntax.ArgumentList.Arguments[i].Expression);
-                            }
-                            else
-                            {
-                                if (methodDeclarationSyntax.ParameterList.Parameters.Count != argumentsAttributeSyntax.ArgumentList.Arguments.Count)
-                                {
-                                    ReportMustHaveMatchingValueCountDiagnostic(Location.Create(context.FilterTree, argumentsAttributeSyntax.ArgumentList.Arguments.Span),
-                                                                               argumentsAttributeSyntax.ArgumentList.Arguments.Count);
-
-                                    continue;
-                                }
-
-                                // ReSharper disable once PossibleNullReferenceException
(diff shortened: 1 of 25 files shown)


silkfire:
@dotnet-policy-service agree

src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/ParamsAllValuesAttributeAnalyzerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+    {
+        public class General : AnalyzerTestFixture<ParamsAllValuesAttributeAnalyzer>
+        {
+            [Theory, CombinatorialData]
+            public async Task A_field_or_property_not_annotated_with_the_paramsallvalues_attribute_should_not_trigger_diagnostic([CombinatorialValues("", "[Dummy]")] string missingParamsAttributeUsage,
+                                                                                                                                 [CombinatorialMemberData(nameof(FieldOrPropertyDeclarations))] string fieldOrPropertyDeclaration,
+                                                                                                                                 [CombinatorialMemberData(nameof(InvalidTypes))] string invalidType)
+            {
+                var testCode =
+/* lang=c#-test */ $@"using BenchmarkDotNet.Attributes;
```
timcassell: FYI you can use `"""` for nicer formatting.

silkfire: I used that in the beginning as I was internally only targeting .NET 9 for the test project. But when I expanded support to .NET Standard and .NET Framework 4.7.2, I was limited by that raw string literals weren't supported until C# 11 (.NET 7). I agree that it looks ugly.

timcassell: It should be framework agnostic, it's a compiler feature. Other test projects include `<Import Project="..\..\build\common.props" />` which sets the language version.

src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/ParamsAllValuesAttributeAnalyzerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+                                                                  "string",
+                                                                  "uint",
+                                                                  "ulong",
+                                                                  "ushort",
+                                                                                                       
+                                                                  "DummyEnumWithFlagsAttribute",
+                                                              
+                                                                  "object",
+                                                                  "System.Type"
+                                                              };
```
timcassell: We format left-aligned in this repo. And when you import the common.props mentioned in another comment, you can use collection literal syntax `[]`.
```suggestion
            public static IEnumerable<string> InvalidTypes =>
            [
                "byte",
                "char",
                "double",
                "float",
                "int",
                "long",
                "sbyte",
                "short",
                "string",
                "uint",
                "ulong",
                "ushort",
                "DummyEnumWithFlagsAttribute",
                "object",
                "System.Type"
            ];
```

src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.Tests/BenchmarkDotNet.Analyzers.Tests.csproj
```diff
@@ -0,0 +1,53 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net462;net6.0;net8.0</TargetFrameworks>
+    <Nullable Condition="'$(TargetFramework)' == 'net8.0' Or '$(TargetFramework)' == 'net6.0'">enable</Nullable>
```
timcassell: Condition is unnecessary after you import common.props.
```suggestion
    <Nullable>enable</Nullable>

```

src/BenchmarkDotNet.Disassembler.x64/BenchmarkDotNet.Disassembler.x64.csproj
```diff
@@ -1,21 +1,24 @@
...
+  <PropertyGroup>
+    <OutputPath>..\BenchmarkDotNet\Disassemblers</OutputPath>
+    <RootNamespace>BenchmarkDotNet.Disassembler</RootNamespace>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Iced" Version="1.21.0" />
+    <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="[1.1.142101]" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" />
```
timcassell: Why is this necessary? This project seems unrelated.

silkfire: I think it was because the package referenced in the props file wasn't the latest version

silkfire:
I'm not sure whether the analyzers should be automatically enabled with the base BenchmarkDotNet package or be opt-in via its own NuGet package, what do you think?

timcassell:
They should be enabled by default.

silkfire:
So maybe the VSIX package project can be removed then as the analyzer can be referenced through an analyzer project reference.

timcassell:
> So maybe the VSIX package project can be removed then as the analyzer can be referenced through an analyzer project reference.

I actually think the analyzer should be included directly into the annotations package. Otherwise, it was found that a separate analyzer package pulls in too many unnecessary dependencies. It's a bit complicated to set up the build to do it, though, so I can do it separately after this is merged if you want. [Edit] Or I can push to your branch after your changes are complete.

silkfire:
I'm getting `Referenced assembly 'BenchmarkDotNet.Analyzers, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' does not have a strong name.` when trying to compile the test project.

timcassell:
You need to import common.props in the analyzer project, too.

silkfire:
Solved it. I also needed to add the public key of the assembly to the InternalsVisibleTo attribute.

silkfire:
> > So maybe the VSIX package project can be removed then as the analyzer can be referenced through an analyzer project reference.
> 
> I actually think the analyzer should be included directly into the annotations package. Otherwise, it was found that a separate analyzer package pulls in too many unnecessary dependencies. It's a bit complicated to set up the build to do it, though, so I can do it separately after this is merged if you want. [Edit] Or I can push to your branch after your changes are complete.

Do I just reference the analyzer project from the annotations project? Or do I need to do something special for the analyzers to activate for the user? Mind that we of course don't want the analyzers to activate for the annotations project, but transitively for the user.

timcassell:
> Do I just reference the analyzer project from the annotations project?

Yes that's sufficient for now.

timcassell:
Also you should move the analyzers test project to under the tests/ directory (and move the analyzers project up 1 level).

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,35 @@
...
+
+Rule ID  | Category | Severity | Notes
+---------|----------|----------|--------------------
+BDN1000  |  Usage   | Error	   | BDN1000_BenchmarkRunner_Run_TypeArgumentClassMissingBenchmarkMethods
+BDN1001  |  Usage   | Error	   | BDN1001_General_BenchmarkClass_MethodMustBePublic
+BDN1002  |  Usage   | Error	   | BDN1002_General_BenchmarkClass_MethodMustBeNonGeneric
+BDN1003  |  Usage   | Error	   | BDN1003_General_BenchmarkClass_ClassMustBePublic
+BDN1004  |  Usage   | Error	   | BDN1004_General_BenchmarkClass_ClassMustBeNonStatic
+BDN1005  |  Usage   | Error	   | BDN1005_General_BenchmarkClass_ClassMustBeNonAbstract
+BDN1006  |  Usage   | Error	   | BDN1006_General_BenchmarkClass_ClassMustBeNonGeneric
```
timcassell: This is an incorrect error (or a misleading description). Benchmark classes may be generic, but if they are, they must have `[GenericTypeArguments]`.

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,35 @@
...
+### New Rules
+
+Rule ID  | Category | Severity | Notes
+---------|----------|----------|--------------------
+BDN1000  |  Usage   | Error	   | BDN1000_BenchmarkRunner_Run_TypeArgumentClassMissingBenchmarkMethods
+BDN1001  |  Usage   | Error	   | BDN1001_General_BenchmarkClass_MethodMustBePublic
+BDN1002  |  Usage   | Error	   | BDN1002_General_BenchmarkClass_MethodMustBeNonGeneric
+BDN1003  |  Usage   | Error	   | BDN1003_General_BenchmarkClass_ClassMustBePublic
+BDN1004  |  Usage   | Error	   | BDN1004_General_BenchmarkClass_ClassMustBeNonStatic
+BDN1005  |  Usage   | Error	   | BDN1005_General_BenchmarkClass_ClassMustBeNonAbstract
```
timcassell: I'm not sure if this rule is entirely correct. I think users may create an abstract base class with benchmarks, then create derived concrete classes.

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
timcassell: Also the attribute might not be required if it's abstract and a derived class supplies the generic types.

src/BenchmarkDotNet.Analyzers/General/BenchmarkClassAnalyzer.cs
```diff
@@ -0,0 +1,286 @@
...
+
+            var genericTypeArgumentsAttributes = AnalyzerHelper.GetAttributes("BenchmarkDotNet.Attributes.GenericTypeArgumentsAttribute", context.Compilation, classDeclarationSyntax.AttributeLists, context.SemanticModel);
+            if (genericTypeArgumentsAttributes.Length == 0)
+            {
+                if (classDeclarationSyntax.TypeParameterList != null)
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(ClassMustBeNonGenericRule, classDeclarationSyntax.TypeParameterList.GetLocation(), classDeclarationSyntax.Identifier.ToString()));
+                }
+            }
+            else if (genericTypeArgumentsAttributes.Length == 1)
```
timcassell: Multiple GenericTypeArgumentsAttributes are allowed. You should iterate over them all.

src/BenchmarkDotNet.Analyzers/General/BenchmarkClassAnalyzer.cs
```diff
@@ -0,0 +1,286 @@
...
+                    var benchmarkAttributes = AnalyzerHelper.GetAttributes(benchmarkAttributeSymbol, methodDeclarationSyntax.AttributeLists, context.SemanticModel);
+                    if (benchmarkAttributes.Length > 0)
+                    {
+                        benchmarkMethodsBuilder.Add((methodDeclarationSyntax, benchmarkAttributes.SelectMany(a => GetBaselineLocations(a)).ToImmutableArray()));
+                    }
+                }
+            }
+
+            var benchmarkMethods = benchmarkMethodsBuilder.ToImmutable();
+            if (benchmarkMethods.Length == 0)
```
timcassell: Base/derived classes without benchmarks may still apply other attributes like `[GlobalSetup]`, `[Params]`, etc. These types should be analyzed as well.

timcassell: I would split each attribute analysis to a separate method and call them all.

src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
```diff
@@ -14,4 +14,13 @@
   <ItemGroup>
     <Compile Include="..\BenchmarkDotNet\Helpers\CodeAnnotations.cs" Link="Attributes\CodeAnnotations.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="9.0.0">
```
timcassell: What is this added dependency for?

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
silkfire: To be honest, I'm not entirely sure where I got this rule from. Maybe we should just remove it completely.

timcassell: Perhaps it is only types with `[GenericTypeArguments]` that must be concrete.

src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
silkfire: That's strange, I have no recollection of actually adding it. Removed it.

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
silkfire: Perhaps? :)

I think I can add a rule that the referenced benchmark class in `BenchmarkRunner.Run<BenchmarkClass>` cannot be abstract.

silkfire:
Is `Baseline = true` unique per benchmark category or per benchmark class?

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
silkfire: If a non-generic class is annotated with `[GenericTypeArguments]`, should I loop infinitely over its ancestor classes until I find one that is generic?

timcassell: No, non-generic classes should not have `[GenericTypeArguments]`.

timcassell:
> Is `Baseline = true` unique per benchmark category or per benchmark class?

It's per category.

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
silkfire: > Also the attribute might not be required if it's abstract and a derived class supplies the generic types.

Ah I thought you meant like this, but this is then not valid?

```csharp
[GenericTypeArguments(typeof(int))]
public class BenchmarkClass : BenchmarkClassBase<int>
{
}

public abstract class BenchmarkClassBase<T1>
{
    [Benchmark]
    public void BenchmarkMethod()
    {

    }
}
```

timcassell: Right, the derived class supplies the type argument via `<int>`, so the attribute is useless.

src/BenchmarkDotNet.Analyzers/General/BenchmarkClassAnalyzer.cs
silkfire: Is it valid that the top class (the one referenced in the `BenchmarkRunner`) does not have any benchmark methods, but its ancestor classes do?

timcassell: Yes

src/BenchmarkDotNet.Analyzers/Attributes/ParamsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,283 @@
...
+
+            return;
+
+            void ReportIfUnexpectedValueTypeDiagnostic(ExpressionSyntax valueExpressionSyntax)
+            {
+                var actualValueTypeSymbol = context.SemanticModel.GetTypeInfo(valueExpressionSyntax).Type;
+                if (actualValueTypeSymbol != null && actualValueTypeSymbol.TypeKind != TypeKind.Error)
+                {
+                    var conversionSummary = context.Compilation.ClassifyConversion(actualValueTypeSymbol, expectedValueTypeSymbol);
+                    if (!conversionSummary.IsImplicit)
```
timcassell: Does this work for this?

```cs
[Params(1, 2)]
public byte num;
```

I think the compiler treats the numbers as `int`, and thus they are not implicitly convertible to `byte`, but it works today because we generate source with integer literals, which are implicitly convertible to `byte` (if they are in range).

Please add tests for it.

silkfire: Good point, I'll test it out tomorrow. I've only added a test for the other way around, byte to int.

I remember when I first started building this analyzer, I actually tested for the *exact* type, but then I inspected the source code that BDN generated and realized the requirement is implicit conversion so I adjusted the logic and added applicable tests.

silkfire: You were right, the conversion from `int` to `byte` (even with values within range) causes the diagnostic to trigger. It says that an explicit conversion exists, but it does that also for values outside the range. This will result in an overflow exception, do you have any suggestions on how to detect this or do I need special logic for numerical values?

EDIT: Okay I found a solution. I can attempt to convert the constant value to the target type and then check if it's within range. If an `OverflowException` or `InvalidCastException` is thrown then I assume that the conversion is invalid.

timcassell: I'm not very knowledgeable about roslyn analyzer APIs, so I asked copilot. Best suggestion is to create a dummy syntax tree:

```cs
public static bool IsConstantAssignableToType(
    Compilation compilation,
    ITypeSymbol targetType,
    object constantValue)
{
    // Create a dummy assignment expression: "T x = <constant>;"
    var code = $@"
class Dummy {{
    {targetType.ToDisplayString()} x = {FormatLiteral(constantValue)};
}}";

    var syntaxTree = CSharpSyntaxTree.ParseText(code);
    var semanticModel = compilation.AddSyntaxTrees(syntaxTree).GetSemanticModel(syntaxTree);

    // Find the variable declarator
    var root = syntaxTree.GetRoot();
    var variable = root.DescendantNodes()
        .OfType<VariableDeclaratorSyntax>()
        .FirstOrDefault();

    if (variable == null)
        return false;

    // Get the initializer expression
    var initializer = variable.Initializer?.Value;
    if (initializer == null)
        return false;

    // Ask the semantic model for the constant value
    var constant = semanticModel.GetConstantValue(initializer);
    return constant.HasValue;
}

private static string FormatLiteral(object value)
{
    return value switch
    {
        byte b => b.ToString(),
        sbyte sb => sb.ToString(),
        short s => s.ToString(),
        ushort us => us.ToString(),
        int i => i.ToString(),
        uint ui => ui.ToString() + "u",
        long l => l.ToString() + "L",
        ulong ul => ul.ToString() + "UL",
        float f => f.ToString(System.Globalization.CultureInfo.InvariantCulture) + "f",
        double d => d.ToString(System.Globalization.CultureInfo.InvariantCulture),
        decimal m => m.ToString(System.Globalization.CultureInfo.InvariantCulture) + "m",
        char c => $"'{c}'",
        bool b => b ? "true" : "false",
        string s => $"\"{s}\"",
        _ => throw new NotSupportedException($"Unsupported literal type: {value.GetType()}")
    };
}
```

timcassell: I think that would be even better to use the original expression instead of the constant value. `expression.ToFullString()` that way it will also work for any type, including `Type`, and you don't need that giant switch expression.

src/BenchmarkDotNet.Analyzers/General/BenchmarkClassAnalyzer.cs
silkfire: The only relevant attributes you're referring to that would be applicable to not require the `[Benchmark]` attribute is `[Params]`, `[ParamsAllValues]` and `[ParamsSource]`. I'm currently not analyzing `[GlobalSetup]`.

Also `[Params]` and `[ParamsAllValues]` has their own analyzers so I don't really understand the "split each attribute analysis to a separate method" suggestion here :)

timcassell: It's an old comment, looks like it's resolved already.

src/BenchmarkDotNet.Analyzers/Attributes/ParamsAttributeAnalyzer.cs
timcassell: Your implementation does a lot of special-casing. I think doing it with a dummy syntax tree will be more robust in the long-run, as we simply rely on the compiler behavior.

silkfire: I was able to shrink that method down to just a few lines. Instead of checking whether the constant value is actually valid I can just check if there is an initializer error. The syntax tree is also isolated which makes this method extremely precise.

```csharp
public static bool IsConstantAssignableToType(Compilation compilation, ITypeSymbol targetType, string valueExpression)
{
    var code = $$"""
                 file class Internal {
                     {{targetType}} x = {{valueExpression}};
                 }
                 """;

    var syntaxTree = CSharpSyntaxTree.ParseText(code);

    var diagnostics = compilation.AddSyntaxTrees(syntaxTree).GetSemanticModel(syntaxTree).GetMethodBodyDiagnostics();

    return diagnostics.Length == 0;
}
```

By using `GetMethodBodyDiagnostics` instead of `GetDiagnostics` I retrieve field initializer errors only, so if the user also happens to have a class called `Internal` in the global namespace then that error won't bubble up and conflict with our testing logic.

timcassell: Beautiful

src/BenchmarkDotNet.Analyzers/Attributes/ArgumentsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,331 @@
...
+                    {
+                        continue;
+                    }
+
+                    var valueExpressionString = valueExpressionSyntax.ToString();
+
+                    var actualValueTypeSymbol = context.SemanticModel.GetTypeInfo(valueExpressionSyntax).Type;
+                    if (actualValueTypeSymbol != null && actualValueTypeSymbol.TypeKind != TypeKind.Error)
+                    {
+                        if (!AnalyzerHelper.IsConstantAssignableToType(context.Compilation, methodParameterTypeSymbol, valueExpressionString))
```
timcassell: Please create a separate helper method to check arguments. We support implicit casts to ref structs for arguments. The generated code looks like this:

```cs
System.Span<System.Int32> arg0 =  __argField0;
```

silkfire: I'm not an expert on ref structs. Does this mean that strings are not allowed?

This gives me `error CS8345: Field or auto-implemented property cannot be of type 'Span<string>' unless it is an instance member of a ref struct.`:

```csharp
static readonly string x = "test";

static readonly System.Span<string> y = new[] { x };
```

timcassell: That's why it needs a separate implementation. Ref structs cannot be class fields, they may only live on the stack. So you need to put the assignment inside a method.

silkfire: Gotcha, thanks

timcassell: I would also rename the helper. `IsAssignableToField` and `IsAssignableToLocal`.

silkfire: Do you have an example of in which scenario a cast to ref struct would fail, despite the type being correct?

timcassell: Not sure what you mean by your question. Are you referring to how the existing analyzer fails?

silkfire: You're writing that for the `[Arguments]` attribute, the value in the generated source is implicitly cast to a `ref struct`. Does testing whether `System.Span<string> y = new[] { x };` compiles add any value in addition to the field/variable assignment done on the line before it, e.g. `string x = "test"`?

timcassell: We don't need to test the field assignment, it's fully generated. We need to test the local assignment. As an example, see https://github.com/dotnet/BenchmarkDotNet/blob/1c3faa21ab9aac8aab277e1fd19e151768c57cba/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs#L396-L397

silkfire: I just inspected the `notcs` of a benchmark and couldn't find any `Span`s in the file.

timcassell: From the `AnArrayCanBePassedToBenchmarkAsSpan` test, the generated file contains:

```cs
        public Runnable_0()
        {
            globalSetupAction = () => { };
            globalCleanupAction = () => { };
            iterationSetupAction = () => { };
            iterationCleanupAction = () => { };
            overheadDelegate = __Overhead;
            workloadDelegate = AcceptsSpan;
            __argField0 = new System.Int32[] { 0, 1, 2 };
        }
        
        // ...
        
        private System.Int32[] __argField0;
        
        // ...
        
        [System.Runtime.CompilerServices.MethodImpl(BenchmarkDotNet.Portability.CodeGenHelper.AggressiveOptimizationOption)]
        {
             System.Span<System.Int32> arg0 =  __argField0;
            {
                workloadDelegate( arg0);
            }
        }
```

silkfire: I finally understand what you actually meant originally. You wanted me to test the scenario where an array is implicitly converted to a `Span<T>`. I thought this was something that was done for all argument types. This made me discover that my array analyzing was incorrect. I need to scan whether the user supplied the `params` as an `object[]` or if an array is just a regular `params` value, and I realized that normal arrays weren't handled at all. Working on a fix for that.

timcassell: I don't think it needs to be that complicated. I think if you just adjust the dummy syntax tree to something like this, it should work.

```cs
$$"""
file class Internal
{
    public void M()
    {
       {{targetType}} x = {{valueExpression}};
    }
}
"""
```

timcassell:
Please also test these cases:

```cs
private const int x = 100;

[Params(x)]
public int num;

[Arguments(x)]
public void Benchmark(int i) { }
```

```cs
[Params(DifferentType.SomeConst)]
public int num;

[Arguments(DifferentType.SomeConst)]
public void Benchmark(int i) { }
```

silkfire:
> Please also test these cases:
> 
> ```cs
> private const int x = 100;
> 
> [Params(x)]
> public int num;
> 
> [Arguments(x)]
> public void Benchmark(int i) { }
> ```
> 
> ```cs
> [Params(DifferentType.SomeConst)]
> public int num;
> 
> [Arguments(DifferentType.SomeConst)]
> public void Benchmark(int i) { }
> ```

I had that in the pipeline too.

silkfire:
I assume we also support `null` as a valid constant value? I haven't yet built tests for that scenario.

timcassell:
Yes, `null` should work for both reference types and nullable value types.

silkfire:
> > Is `Baseline = true` unique per benchmark category or per benchmark class?
> 
> It's per category.

When it comes to category values, are these all considered the same?

- `null` value
- `""` (empty string value)
- No category specified
- `"<whitespace>"` (is whitespace trimmed to `""` and thus considered equal to `null`?)

tests/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/ArgumentsAttributeAnalyzerTests.cs
```diff
@@ -938,6 +1092,63 @@ static IEnumerable<string> GenerateData()
                 ( "typeof(string)", "System.Type" ),
                 ( "DummyEnum.Value1", "DummyEnum" )
             ];
+
+            public static IEnumerable<ValueTupleDouble<string, string>> ConstantValuesAndTypes =>
+            [
+                ( "true", "bool" ),
+                ( "(byte)123", "byte" ),
```
timcassell: If this is testing the case of `[Params(123)] public byte x;`, I think it shouldn't have the cast (same for others). Or maybe another case to test both with and without the cast. [Edit] Also nevermind, I can see this is for testing constant reference, not constant literal.

tests/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/ArgumentsAttributeAnalyzerTests.cs
```diff
@@ -938,6 +1092,63 @@ static IEnumerable<string> GenerateData()
...
+            ];
+
+            public static IEnumerable<string> NullReferenceConstantTypes =>
+            [
+                "object",
+                "string",
+                "System.Type",
+            ];
+
+            public static IEnumerable<ValueTupleDouble<string, string>> NotConvertibleConstantValuesAndTypes =>
```
timcassell: We should also test for values that are out of range of the type, e.g. 256 -> byte, -1 -> ushort, etc. [Edit] Nevermind, I see it's already tested, I was just confused by the property name here.

timcassell:
> When it comes to category values, are these all considered the same?
> 
> * `null` value
> * `""` (empty string value)
> * No category specified
> * `"<whitespace>"` (is whitespace trimmed to `""` and thus considered equal to `null`?)

It's a good question. I checked the code, and it doesn't look like it does any special string handling. `null` likely will result in an exception, so we should not allow it. The other 3 cases appear to be valid and distinct. I don't have time to test it right now, though, so please test it yourself.

tests/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/ArgumentsAttributeAnalyzerTests.cs
silkfire: No, these were to test precise values of constants. The cast on the `byte` is because C# lacks a suffix for `byte` literals for some reason. The implicit integer to byte conversion is tested in `Providing_integer_value_types_within_target_type_range_should_not_trigger_diagnostic`.

tests/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/ArgumentsAttributeAnalyzerTests.cs
silkfire: Yep that's tested in `Providing_integer_value_types_not_within_target_type_range_should_trigger_diagnostic`. It took me a few days to build all test cases, test everything thorougly - and most importantly - get all tests to pass. The combinatorial nature of the tests results in a huge amount of tests but makes sure all possible combinations are tested.

I'm also testing nullable value types (both negative and positive tests) as well as reference types (only positive tests as reference types are always nullable).

silkfire: I still had to go with a `FormatLiteral` method because the constant value of an `enum` type is always its underlying type (default `int`). So to ensure the value fits in the underlying type of the enum type I need to cast it like this:

```csharp
DummyEnum x = (DummyEnum)42;
```

silkfire:
> > When it comes to category values, are these all considered the same?
> > 
> > * `null` value
> > * `""` (empty string value)
> > * No category specified
> > * `"<whitespace>"` (is whitespace trimmed to `""` and thus considered equal to `null`?)
> 
> It's a good question. I checked the code, and it doesn't look like it does any special string handling. `null` likely will result in an exception, so we should not allow it. The other 3 cases appear to be valid and distinct. I don't have time to test it right now, though, so please test it yourself.

Thanks for the reply. I tested with `null` and it does not throw any exceptions. It just prints an empty string (I assume?) in the result table.

timcassell:
The default `DefaultCategoryDiscoverer` uses `.Distinct(StringComparer.OrdinalIgnoreCase)`, but we support custom `ICategoryDiscoverer`, so we can't really rely on any particular behavior.

silkfire:
I'm not really sure how the combination of categories is supposed to work (or if it's intended?). This is with `BenchmarkCategory("test", "Deserialize")`: See the question mark in the `Ratio` column.

image

----
 Just using one category (`"Deserialize"`) groups them correctly:

image

timcassell:
Categories predates me. @AndreyAkinshin Can you provide more insight here?

silkfire:
I tested with passing a `null` value and get an exception; you were right - this means that the `params string[]` array will become `null`. Passing an array containing `null` values though is valid - which seems to be equivalent to not specifying a category at all, just as I guessed.

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Linq.Enumerable.SelectManySingleSelectorIterator`2.MoveNext()
   at System.Collections.Generic.HashSet`1.UnionWith(IEnumerable`1 other)
   at System.Collections.Generic.HashSet`1..ctor(IEnumerable`1 collection, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.DistinctIterator`1.ToArray()
   at BenchmarkDotNet.Running.DefaultCategoryDiscoverer.GetCategories(MethodInfo method)
```

I'll create a diagnostic for the first scenario.

silkfire:
It was as I suspected, it seems (at least in the default implementation) that categories for a method are just joined together with a comma.

So `BenchmarkCategory("Deserialize", "test")` and `BenchmarkCategory("Deserialize")` are considered distinct groups, whereas `BenchmarkCategory("Deserialize", "test")` and `BenchmarkCategory("Deserialize", "test")` are equal.

silkfire:
So the way I solved it was to split the original "Only one benchmark can be baseline" into two rules: one for when the category is empty and one for when it's not. The former has the severity `Error` and the latter `Warning`, so it won't prevent the developer from compiling their benchmark project (uses the default `ICategoryDiscoverer` implementation).

Empty categories are:

- When the benchmark method is not annotated with any `[BenchmarkCategory]` attribute
- Empty usages: `[BenchmarkCategory]` or `[BenchmarkCategory()]`
- Any variant that involves passing an empty array as an argument to `categories`:
  - `[ ]`, `new string[] {}`, ...

The algorithm is quite smart, so it won't be looking into ancestors of the class if there already is a duplicate baseline method in the current class that it's analyzing.

A separate rule has also been added that triggers if the user tries to pass a single `null` value as an argument.

timcassell:
Nice. Can you please check the build errors?

silkfire:
I'm not sure what's causing the build errors to be honest, as I haven't touched the original `BenchmarkDotNet` project. I'm trying to run the tests locally but am getting the error `Referenced assembly 'BenchmarkDotNet.IntegrationTests.FSharp, Version=0.14.1.0, Culture=neutral, PublicKeyToken=null' does not have a strong name.`

timcassell:
You can check the logs https://github.com/dotnet/BenchmarkDotNet/actions/runs/18837492926

silkfire:
I see a lot of build errors are due to `error CS1591: Missing XML comment for publicly visible type or member` which shouldn't be relevant for an analyzer. I'll disable them on the project-level.

src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.csproj
```diff
@@ -0,0 +1,41 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+
+    <!-- Avoid ID conflicts with the package project. -->
+    <PackageId>*$(MSBuildProjectFile)*</PackageId>
```
timcassell: Was it actually causing issues? I think the default package ID that uses the project name should be fine. The build is saying `*` is an illegal character. It looks like all the other projects just copy the project name explicitly.

src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.csproj
```diff
@@ -0,0 +1,41 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+
+    <!-- Avoid ID conflicts with the package project. -->
+    <PackageId>*$(MSBuildProjectFile)*</PackageId>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <NoWarn>CS1591</NoWarn>
```
timcassell:
```suggestion
    <NoWarn>$(NoWarn);CS1591</NoWarn>

```

src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.csproj
silkfire: To be honest I have no idea where I got this from. Do we need a package ID? The analyzer won't be in a separate package, will it?

timcassell: Right, it'll be packed with annotations. And you already have IsPackable set to false, so it'll probably not used.

silkfire:
All workflows are green except the **test-macos (x64)** job. Is it an environment issue or something on my part?

timcassell:
Flaky tests, unrelated to these changes.

Is there any more work you want to do here, or it's good to go?

silkfire:
I feel very confident in releasing this first iteration of the analyzers. They cover most common scenarios that were requested in the discussion and I or other contributors can always build further on top of this. Thanks for your patience and guidance during the review process!

As you mentioned among other things what is remaining is to adjust the pipeline. I noticed that the Analyzers project is built in debug mode:

`BenchmarkDotNet.Analyzers -> D:\a\BenchmarkDotNet\BenchmarkDotNet\src\BenchmarkDotNet.Analyzers\bin\Debug\netstandard2.0\BenchmarkDotNet.Analyzers.dll`

timcassell:
Alright, I'll work on that and push to your branch when I get some time.

AndreyAkinshin:
@silkfire thank you very much for working on this!

> I'm not really sure how the combination of categories is supposed to work (or if it's intended?)

The categories were introduced in the early days of BenchmarkDotNet (2016-2017). No prior design work was done: it was an experimental feature that evolved gradually based on user feedback. If some behavior feels unnatural or some corner cases don't make much sense, feel free to suggest changes.

silkfire:
Are we okay with the naming of the project, as it might be conflated with the `BenchmarkDotNet.Analysers` namespace?

timcassell:
> Are we okay with the naming of the project, as it might be conflated with the `BenchmarkDotNet.Analysers` namespace?

It's fine. It won't be visible to users anyway.

AndreyAkinshin:
> Are we okay with the naming of the project, as it might be conflated with the BenchmarkDotNet.Analysers namespace?

Fine for me too. I have thoughts on reworking the current Analysers and maybe even dropping this concept from the public API. Currently, they are used mostly for statistical post-checks (moving to [perfolizer](https://github.com/AndreyAkinshin/perfolizer) and [pragmastat](https://github.com/AndreyAkinshin/pragmastat)) and error aggregation (moving to core logic). With the new result data format, we won't need an extension point here since users will have other ways to implement their post-checks. It's a huge piece of work with no ETA yet, but long-term we can resolve this name clash.

I don't see any issues with using this name for the new package now. On the compilation level, it won't lead to any conflicts since the new package uses American spelling (Analy**z**ers) while the existing namespace uses British spelling (Analy**s**ers).

timcassell:
@silkfire I hooked up the analyzers to run against the other test projects and it produced a bunch of errors. PTAL.

Also I find this error a bit confusing.

image

I would expect the attribute to be underlined, not the class, with a message something more like "class is not generic or does not match the arity of the GenericTypeArguments".

silkfire:
> @silkfire I hooked up the analyzers to run against the other test projects and it produced a bunch of errors. PTAL.

I'll take a look at that in a couple of days when I will have more time.

> Also I find this error a bit confusing.

Regarding the specific error, there are many diagnostics in `BenchmarkDotNet.Analyzers` that come with a more detailed description when viewed in the Error List of Visual Studio (not sure how it looks in Rider). The immediate error you see is what's actually wrong. Of course, it can be adjusted if you find it to be unclear.

image

-----

Would you like to go through all of the error messages to see if they need any polish and improvement in clarity?

As far as the reasoning for underlining the class name instead of the attribute itself, it was that it's not the _attribute usage_ that's wrong, but the _class itself_ does not have type parameters (is non-generic), and therefore it's the class that should be changed.

But I'm open-minded here and can change this, too.

There is a separate diagnostic for when the arity mismatches with the number of type arguments in the attribute usage(s).

tests/BenchmarkDotNet.Analyzers.Tests/BenchmarkDotNet.Analyzers.Tests.csproj
```diff
@@ -1,51 +1,33 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
+  <Import Project="..\..\build\common.props" />
   <PropertyGroup>
-    <TargetFrameworks>net462;net6.0;net8.0</TargetFrameworks>
+    <AssemblyName>BenchmarkDotNet.Analyzers.Tests</AssemblyName>
+    <TargetFrameworks>net462;net8.0</TargetFrameworks>
```
silkfire, line 5: @timcassell What is the reasoning for removing the .NET 6 TFM?

tests/Directory.build.props
silkfire, line 1: This build settings file is what's causing the errors. Why do you want analysis to be performed on the BenchmarkDotNet test projects?

src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
```diff
@@ -16,6 +16,6 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\BenchmarkDotNet.Analyzers\BenchmarkDotNet.Analyzers.csproj" PrivateAssets="all" />
-    <None Include="$(OutputPath)\netstandard2.0\BenchmarkDotNet.Analyzers.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
+    <Content Include="..\BenchmarkDotNet.Analyzers\bin\$(Configuration)\netstandard2.0\BenchmarkDotNet.Analyzers.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
```
silkfire: Does changing the tag name to `Content` make any difference?

tests/BenchmarkDotNet.Analyzers.Tests/BenchmarkDotNet.Analyzers.Tests.csproj
timcassell: For consistency, we aren't testing that tfm in any of the other test projects.

tests/Directory.build.props
timcassell: Because there are a lot of tests with valid usages that makes a good litmus test for the analyzers. For the tests that are purposefully testing validation failure, they should disable the analyzers for that test.

It should also be added to the samples project, which I forgot to do.

src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
timcassell: It looks like it doesn't make a difference. I just copied that from another project.

timcassell:
> Would you like to go through all of the error messages to see if they need any polish and improvement in clarity?

Yes, I can do that after you fix the errors from the samples and tests.

> As far as the reasoning for underlining the class name instead of the attribute itself, it was that it's not the _attribute usage_ that's wrong, but the _class itself_ does not have type parameters (is non-generic), and therefore it's the class that should be changed.
> 
> But I'm open-minded here and can change this, too.

I would prefer the attribute to be underlined rather than the class. In my experience, in C# an attribute is easier to get wrong than a class (e.g. refactoring, everything compiles fine, but the attributes were left untouched).

src/BenchmarkDotNet.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,38 @@
...
+BDN1201  |  Usage   | Error	   | BDN1201_Attributes_GeneralParameterAttributes_MutuallyExclusiveOnProperty
+BDN1202  |  Usage   | Error	   | BDN1202_Attributes_GeneralParameterAttributes_FieldMustBePublic
+BDN1203  |  Usage   | Error	   | BDN1203_Attributes_GeneralParameterAttributes_PropertyMustBePublic
+BDN1204  |  Usage   | Error	   | BDN1204_Attributes_GeneralParameterAttributes_NotValidOnReadonlyField
+BDN1205  |  Usage   | Error	   | BDN1205_Attributes_GeneralParameterAttributes_NotValidOnConstantField
+BDN1206  |  Usage   | Error	   | BDN1206_Attributes_GeneralParameterAttributes_PropertyCannotBeInitOnly
+BDN1207  |  Usage   | Error	   | BDN1207_Attributes_GeneralParameterAttributes_PropertyMustHavePublicSetter
+BDN1300  |  Usage   | Error	   | BDN1300_Attributes_ParamsAttribute_MustHaveValues
+BDN1301  |  Usage   | Error	   | BDN1301_Attributes_ParamsAttribute_MustHaveMatchingValueType
+BDN1302  |  Usage   | Warning  | BDN1302_Attributes_ParamsAttribute_UnnecessarySingleValuePassedToAttribute
```
timcassell: Probably this should be downgraded to Info rather than Warning. WarningsAsErrors causes it to become an error, and it's not so severe (maybe a user wants the param to show up in the results table even though it's just 1 value).

src/BenchmarkDotNet.Annotations/BenchmarkDotNet.Annotations.csproj
silkfire: I copied it from another project that transitively referenced an analyzer as well as another place using `None` and confirmed it works just as intended :)

tests/Directory.build.props
silkfire: How do you disable an analyzer reference on the test level? Or do you just want me to explicitly suppress the triggered diagnostics with a pragma directive?

timcassell: Right, using pragma directives around the individual tests that need it suppressed.

tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
```diff
@@ -66,14 +66,16 @@ public ValueTask<int> SimpleValueTaskAsync(bool boolean, int number)
 
         public class WithArgumentsSource
         {
+#pragma warning disable BDN1401
```
timcassell: Looks like you removed `BDN1401`, so all these disables are redundant. (Also these are valid benchmarks, so the analyzers should not be disabled anyway.)

silkfire: Well spotted! I realized that this rule isn't just related to the `[Arguments]` attribute, hence moving it to a separate analyzer and changing its ID to `BDN1500`. It's the same non-issue that was manifested in the Samples project, which now no longer gives false positives.

silkfire:
Not sure if this is a valid issue, but locally I get lots of errors like `C:\Program Files\dotnet\sdk\9.0.306\Microsoft.Common.CurrentVersion.targets(2189,5): warning MSB9008: The referenced project ..\..\src\BenchmarkDotNet.Analyzers\BenchmarkDotNet.Analyzers.csproj does not exist. [D:\Projects\2021\BenchmarkDotNet\tests\BenchmarkDotNet.IntegrationTests\bin\Debug\net8.0\BenchmarkDotNet.IntegrationTests-3\BenchmarkDotNet.Autogenerated.csproj]`

Does the generated benchmark really need to reference the analyzer project?

tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
timcassell: I think it was a good rule, it just needed to account for both `[Arguments]` and `[ArgumentsSource]`.

timcassell:
Oh, that's probably from the Directory.build.props file. I didn't get that error until I cleaned my bin/obj dirs. It wouldn't be an issue if #2467 was done... we could just copy the project reference to each test project instead, not sure if there's a cleaner way.

tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
silkfire: Yeah exactly. I also adjusted the wording for it in the diagnostic message.

timcassell:
I pushed a fix for that.

silkfire:
What's the next step in this PR process?

timcassell:
Looks like some tests are still failing.

silkfire:
I can't really see what's wrong besides the flaky tests

timcassell:
Hm... I re-ran the tests twice and they keep failing. I'm not able to repro on my local machine. 🤔

timcassell:
It looks like updating the sdk to 9.0 caused some tests to fail even though the tests are using net8.0 tfm. The update was necessary because there were compile errors using Microsoft.CodeAnalysis.CSharp v4.14.0 otherwise. We can either revert the sdk update and downgrade the Microsoft.CodeAnalysis.CSharp version, or update all of the tests. It's probably easier to do the former. We should be updating all the tests to net10.0 soon after the GA release next week anyway.

timcassell:
@silkfire

- `BDN1100` also should underline the attribute, and update short description to something like "GenericTypeArgumentsAttribute may not be applied to an abstract class".
- `BDN1202`, `BDN1203`, `BDN1204`, `BDN1205`, `BDN1206`, `BDN1207` are not triggered when I tested them in the Samples project. (I see the tests pass, not sure if they are flawed or what the issue is.)
- `BDN1301`, `BDN1303`, `BDN1304` are not triggered.
- `BDN1502` is not triggered.

silkfire:
> @silkfire
> 
> * `BDN1202`, `BDN1203`, `BDN1204`, `BDN1205`, `BDN1206`, `BDN1207` are not triggered when I tested them in the Samples project. (I see the tests pass, not sure if they are flawed or what the issue is.)
> * `BDN1301`, `BDN1303`, `BDN1304` are not triggered.
> * `BDN1502` is not triggered.

There's been a bug in Visual Studio at least since June 2025 that prevents some diagnostics from showing. It has apparently to do with the way the IDE interprets the diagnostic `Location`. It's been reported [here](https://developercommunity.visualstudio.com/t/Roslyn-analyzer-diagnostic-not-reported/10918937) but I haven't heard any update from Microsoft unfortunately.

If you compile the project though, you should always see the reported diagnostic in the output. Have you tried that?

image

I have put a lot of work into the unit tests and gone through them all meticulously, and I have full trust in the testing framework when all tests are reported green.

Some of those diagnostics you mention do trigger though in my environment. How are you testing them?

image
image

silkfire:
I've added two new diagnostics that will trigger if you pass a single `null` value to either an `[Arguments]` or `[Params]` attribute.

timcassell:
> I've added two new diagnostics that will trigger if you pass a single `null` value to either an `[Arguments]` or `[Params]` attribute.

Apologies, I was incorrect about those. It is valid to pass a single `null` value, the attribute internally converts it to `[null]`. So it just needs to validate that the value it's being assigned to is nullable.

timcassell:
> There's been a bug in Visual Studio at least since June 2025 that prevents some diagnostics from showing. It has apparently to do with the way the IDE interprets the diagnostic `Location`. It's been reported [here](https://developercommunity.visualstudio.com/t/Roslyn-analyzer-diagnostic-not-reported/10918937) but I haven't heard any update from Microsoft unfortunately.
> 
> If you compile the project though, you should always see the reported diagnostic in the output. Have you tried that?
> 
> I have put a lot of work into the unit tests and gone through them all meticulously, and I have full trust in the testing framework when all tests are reported green.
> 
> Some of those diagnostics you mention do trigger though in my environment. How are you testing them?

Interesting. I am just testing in Visual Studio. I updated VS to test again and I get the same results. The diagnostics do not trigger even if I run a build in VS. But they do trigger if I build with the sdk on the command line. So I guess it is just an issue with my environment (probably related to that link you shared).

silkfire:
> > I've added two new diagnostics that will trigger if you pass a single `null` value to either an `[Arguments]` or `[Params]` attribute.
> 
> Apologies, I was incorrect about those. It is valid to pass a single `null` value, the attribute internally converts it to `[null]`. So it just needs to validate that the value it's being assigned to is nullable.

Ah yes I was a bit too quick there :) I'll remove them then.

timcassell (approved):
Thanks @silkfire for implementing this huge feature!

silkfire:
@timcassell Should we have a page that documents all the available analyzers to have as a reference?

timcassell:
That's a good idea.

silkfire:
We can just copy the markdown table above

AndreyAkinshin:
> We can just copy the markdown table above

If we need to sync information from the sources to the documentation, we can add this logic to the build system. `BenchmarkDotNet.Build.Runners.DocumentationRunner.Generate` appears to be a suitable entry point for this logic. In this case, the corresponding documentation page does not need to be stored in the repository; it will be autogenerated when we build the documentation website.

timcassell:
We should also make it part of the release workflow to move the updated rules from `AnalyzerReleases.Unshipped.md` to `AnalyzerReleases.Shipped.md`.

Issue #2842, opened 2025-10-17
`[ArgumentsSource]` does not support ref struct


timcassell:
This works:

```cs
[Arguments(new int[] { 1, 2, 3 })]
public void Benchmark(Span<int> args) { }
```

But this errors:

```cs
[ArgumentsSource(nameof(GetArgs))]
public void Benchmark(Span<int> args) { }

public IEnumerable<object[]> GetArgs()
{
    yield return new object[] { new int[] { 1, 2, 3 } };
}
```

> Ambiguous user defined conversions 'Span<int>.implicit operator Span<int>(ArraySegment<int>)' and 'Span<int>.implicit operator Span<int>(int[]?)' when converting from 'object' to 'Span<int>'

Looking at the generated code, it's because of this:

```cs
__argField0 = (System.Span<System.Int32>)BenchmarkDotNet.Parameters.ParameterExtractor.GetParameter(GetArgs(), 0);;
```

So the code generator should take into account the actual type of the element if the argument is a ref struct so that both attributes work the same (which it seems it already does partially, because the field type is correct, only the cast is incorrect).

dmitrenko-v:
Hi! Can I work on this one?

timcassell:
@dmitrenko-v Yes, it's marked up-for-grabs.

PR #2849, opened 2025-10-31
Added ref struct param support for [ArgumentsSource]

dmitrenko-v:
- Changed cast and type of field to arg type in code generation when param is ref struct
Related issue: #2842

### DIFF
--- a/src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
+++ b/src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
@@ -102,9 +102,13 @@ namespace BenchmarkDotNet.Parameters
             string callPostfix = source is PropertyInfo ? string.Empty : "()";
 
             MethodInfo sourceAsMethodInfo =  source as MethodInfo;
+            PropertyInfo sourceAsPropertyInfo = source as PropertyInfo;
+
+            Type indexableType = typeof(IEnumerable<object[]>);
 
             string indexPostfix;
-            if (typeof(IEnumerable<object[]>) == sourceAsMethodInfo?.ReturnType) {
+            if (sourceAsMethodInfo?.ReturnType == indexableType ||
+                sourceAsPropertyInfo?.GetMethod.ReturnType == indexableType) {
                 indexPostfix = $"[{argumentIndex}]";
             }
             else
@@ -113,7 +117,7 @@ namespace BenchmarkDotNet.Parameters
             }
 
             string methodCall;
-            if (sourceAsMethodInfo?.IsStatic ?? (source as PropertyInfo)?.GetMethod.IsStatic ?? throw new Exception($"{nameof(source)} was not {nameof(MethodInfo)} nor {nameof(PropertyInfo)}"))
+            if (sourceAsMethodInfo?.IsStatic ?? sourceAsPropertyInfo?.GetMethod.IsStatic ?? throw new Exception($"{nameof(source)} was not {nameof(MethodInfo)} nor {nameof(PropertyInfo)}"))
             {
                 // If the source member is static, we need to place the fully qualified type name before it, in case the source member is from another type that this generated type does not inherit from.
                 methodCall = $"{source.DeclaringType.GetCorrectCSharpTypeName()}.{source.Name}";


timcassell:
Can you please add tests for both `T[]` -> `Span<T>` and (in .Net Core only) `string` -> `ReadOnlySpan<char>`?

src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
```diff
@@ -91,11 +93,15 @@ public SmartArgument(ParameterDefinition[] parameterDefinitions, object value, M
...
+
+            string cast = isParamRefLike ? $"({Value.GetType().GetCorrectCSharpTypeName()})"
+                : $"({paramType.GetCorrectCSharpTypeName()})"; // it's an object so we need to cast it to the right type
 
             string callPostfix = source is PropertyInfo ? string.Empty : "()";
 
-            string indexPostfix = parameterDefinitions.Length > 1
+            string indexPostfix = parameterDefinitions.Length > 1 || isParamRefLike
                 ? $"[{argumentIndex}]" // IEnumerable<object[]>
                 : string.Empty; // IEnumerable<object>
```
timcassell: This doesn't look like a correct change. `isParamRefLike` should have no bearing on whether the value needs to be indexed.

dmitrenko-v: Yes, I found out that the issue was not only about ref struct parameter, but also if passing something that is not compilation-time constant and there is only one parameter, then it will fail because index is only applied if 
`parameterDefinitions.Length > 1` . Taking example that you provided 

```[ArgumentsSource(nameof(GetArgs))]
public void Benchmark(Span<int> args) { }

public IEnumerable<object[]> GetArgs()
{
    yield return new object[] { new int[] { 1, 2, 3 } };
}
```

In this case, function has only one parameter, but we still need index(0) to get int[] out of object[]. So I changed `parameterDefinitions.Length` check to source return type check

dmitrenko-v:
@dotnet-policy-service agree

src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
```diff
@@ -101,12 +101,19 @@ public string ToSourceCode()
 
             string callPostfix = source is PropertyInfo ? string.Empty : "()";
 
-            string indexPostfix = parameterDefinitions.Length > 1 || isParamRefLike
-                ? $"[{argumentIndex}]" // IEnumerable<object[]>
-                : string.Empty; // IEnumerable<object>
+            MethodInfo sourceAsMethodInfo =  source as MethodInfo;
+
+            string indexPostfix;
+            if (typeof(IEnumerable<object[]>) == sourceAsMethodInfo?.ReturnType) {
```
timcassell: Good fix. 👍 But this should also check for PropertyInfo.

dmitrenko-v: Thanks for hint, added this check

src/BenchmarkDotNet/Parameters/SmartParamBuilder.cs
```diff
@@ -91,16 +93,31 @@ public SmartArgument(ParameterDefinition[] parameterDefinitions, object value, M
...
-                ? $"[{argumentIndex}]" // IEnumerable<object[]>
-                : string.Empty; // IEnumerable<object>
+            MethodInfo sourceAsMethodInfo =  source as MethodInfo;
+            PropertyInfo sourceAsPropertyInfo = source as PropertyInfo;
+
+            Type indexableType = typeof(IEnumerable<object[]>);
+
+            string indexPostfix;
+            if (sourceAsMethodInfo?.ReturnType == indexableType ||
+                sourceAsPropertyInfo?.GetMethod.ReturnType == indexableType) {
```
timcassell, line 111:
```suggestion
                sourceAsPropertyInfo?.PropertyType == indexableType) {
```

timcassell (approved):
Thanks @dmitrenko-v!

PR #2865, opened 2025-11-15
Fix for issue #2864

silkfire:

PR #2866, opened 2025-11-16
Multi target analyzers

### LINKED ISSUE #2863: Upgrading from 0.15.6 to 0.15.7 causes compiler mismatch issues
We recently just tried to upgrade BenchmarkDotNet from 0.15.6 to 0.15.7.

When we try to compile however, we get this error

```
CSC : error CS9057: The analyzer assembly 'C:\Users\some.username\.nuget\packages\benchmarkdotnet.annotations\0.15.7\analyzers\dotnet\cs\BenchmarkDotNet.Analyzer
s.dll' references version '4.14.0.0' of the compiler, which is newer than the currently running version '4.11.0.0'. [C:\code\I
ntegration.Benchmark\Integration.Benchmark.csproj]
```

CSProj pretty much looks like this

```
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <IsPackable>false</IsPackable>
    <GenerateProgramFile>false</GenerateProgramFile>
    <NoWarn>$(NoWarn),CS7022</NoWarn>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="BenchmarkDotNet" />
    <PackageReference Include="BenchmarkDotNet.Diagnostics.dotMemory" />
    <PackageReference Include="BenchmarkDotNet.Diagnostics.dotTrace" />
  </ItemGroup>
  <ItemGroup>
    <Folder Include="BenchmarkDotNet.Artifacts\" />
  </ItemGroup>
  <ItemGroup>
    <RuntimeHostConfigurationOption Include="System.GC.ConserveMemory" Value="9" />
    <RuntimeHostConfigurationOption Include="System.GC.HeapHardLimit" Value="147000000" />
  </ItemGroup>
</Project>

```

and Directory.Packages.props has these specified

```
    <PackageVersion Include="BenchmarkDotNet" Version="0.15.7" />
    <PackageVersion Include="BenchmarkDotNet.TestAdapter" Version="0.15.7" />
    <PackageVersion Include="BenchmarkDotNet.Diagnostics.dotMemory" Version="0.15.7" />
    <PackageVersion Include="BenchmarkDotNet.Diagnostics.dotTrace" Version="0.15.7" />
```

The global.json file is

```
{
    "sdk": {
      "version": "8.0.415",
      "rollForward": "latestFeature"
    }
 }

```

The solution this is referenced is a dotnet 8.0.

my local SDKS look like

```
C:\code\L\some-service>dotnet --list-sdks
8.0.415 [C:\Program Files\dotnet\sdk]
9.0.307 [C:\Program Files\dotnet\sdk]
10.0.100 [C:\Program Files\dotnet\sdk]
```

Thoughts? as we are a little perplexed as to why this is failing given that the nuget versions all look correct.

timcassell:
- Fixes #2863
- Fixes Params and Arguments type conversion analyzers.

### DIFF
--- a/build/BenchmarkDotNet.Build/Runners/BuildRunner.cs
+++ b/build/BenchmarkDotNet.Build/Runners/BuildRunner.cs
@@ -67,7 +67,7 @@ public class BuildRunner
     public void BuildAnalyzers()
     {
         context.Information("BuildSystemProvider: " + context.BuildSystem().Provider);
-        string[] mccVersions = ["2.8", "3.8", "4.8", "5.0"];
+        string[] mccVersions = ["2.8", "3.8", "4.8"];
         foreach (string version in mccVersions)
         {
             context.DotNetBuild(context.AnalyzersProjectFile.FullPath, new DotNetBuildSettings
--- a/src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.csproj
+++ b/src/BenchmarkDotNet.Analyzers/BenchmarkDotNet.Analyzers.csproj
@@ -8,13 +8,11 @@
     <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <!-- Default version used by local builds. This is overridden by the build script. -->
-    <!-- Update this when tests are updated to newer TFMs. -->
     <MccVersion>4.8</MccVersion>
     <OutputPath>bin\$(Configuration)\roslyn$(MccVersion)\cs</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <DefineConstants Condition="'$(MccVersion)' >= '3.8'">$(DefineConstants);CODE_ANALYSIS_3_8</DefineConstants>
     <DefineConstants Condition="'$(MccVersion)' >= '4.8'">$(DefineConstants);CODE_ANALYSIS_4_8</DefineConstants>
-    <DefineConstants Condition="'$(MccVersion)' >= '5.0'">$(DefineConstants);CODE_ANALYSIS_5_0</DefineConstants>
     <!-- New equality comparison doesn't exist in v2.8, and we only use the default comparison anyway. -->
     <!-- Formatting in AnalyzerReleases.Unshipped.md -->
     <NoWarn>$(NoWarn);RS1024;RS2007</NoWarn>
@@ -29,8 +27,6 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.8.0" Condition="'$(MccVersion)' == '3.8'" />
     <!-- C# 12 -->
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.8.0" Condition="'$(MccVersion)' == '4.8'" />
-    <!-- C# 14 -->
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="5.0.0-2.final" Condition="'$(MccVersion)' == '5.0'" />
   </ItemGroup>
   <ItemGroup>
     <AdditionalFiles Include="AnalyzerReleases.Shipped.md" />


src/BenchmarkDotNet.Analyzers/AnalyzerHelper.cs
```diff
@@ -145,58 +146,52 @@ public static string NormalizeTypeName(INamedTypeSymbol namedTypeSymbol)
...
     {
-        var usingDirective = valueTypeContainingNamespace != null ? $"using {valueTypeContainingNamespace};" : "";
-
-        var hasNoCompilerDiagnostics = HasNoCompilerDiagnostics(string.Format(codeTemplate1, usingDirective, targetType, valueExpression), compilation, languageVersion);
-        if (hasNoCompilerDiagnostics)
+        if (valueType == "BenchmarkDotNet.IntegrationTests.ArgumentsTests.WithUndefinedEnumValue.SomeEnum")
+        {
+            System.Diagnostics.Debugger.Launch();
+        }
+        var hasCompilerDiagnostics = HasNoCompilerDiagnostics(string.Format(codeTemplate1, targetType, valueExpression), compilation);
```
silkfire: Should be `hasNoCompilerDiagnostics` :)

src/BenchmarkDotNet.Analyzers/AnalyzerHelper.cs
```diff
@@ -212,19 +207,16 @@ private static bool IsAssignableTo(string codeTemplate1, string codeTemplate2, C
...
 
-    private static bool HasNoCompilerDiagnostics(string code, Compilation compilation, LanguageVersion languageVersion)
+    private static bool HasNoCompilerDiagnostics(string code, Compilation compilation)
     {
-        var compilationTestSyntaxTree = CSharpSyntaxTree.ParseText(code, new CSharpParseOptions(languageVersion));
-
-        var syntaxTreesWithInterceptorsNamespaces = compilation.SyntaxTrees.Where(st => st.Options.Features.ContainsKey(InterceptorsNamespaces));
+        var syntaxTree = CSharpSyntaxTree.ParseText(code);
 
         var compilerDiagnostics = compilation
```
silkfire: How do you cirmcumvent InterceptorsNamespaces here?

silkfire:
Didn't quite follow on the reasoning behind removing a lot of the logic for checking assignability. I don't really want to trigger a diagnostic if the user supplies an illegal value to the attribute, e.g. multidimensional arrays `string[][]`, nullable struct arrays `int?[]` or reference types that aren't `null` or `string` or `System.Type`.

src/BenchmarkDotNet.Analyzers/Attributes/ParamsAttributeAnalyzer.cs
```diff
@@ -138,8 +138,8 @@ private static void AnalyzeFieldOrPropertyTypeSyntax(SyntaxNodeAnalysisContext c
             return;
         }
 
+#if CODE_ANALYSIS_4_8
```
silkfire: Clever! I wasn't aware you could solve it like this by multi-targeting

src/BenchmarkDotNet.Analyzers/Attributes/ParamsAttributeAnalyzer.cs
```diff
@@ -117,7 +117,7 @@ private static void AnalyzeFieldOrPropertyTypeSyntax(SyntaxNodeAnalysisContext c
 
         if (attributeSyntax.ArgumentList.Arguments.All(aas => aas.NameEquals != null))
         {
-            context.ReportDiagnostic(Diagnostic.Create(MustHaveValuesRule, Location.Create(context.FilterTree, attributeSyntax.ArgumentList.Arguments.Span)));
+            context.ReportDiagnostic(Diagnostic.Create(MustHaveValuesRule, Location.Create(context.Node.SyntaxTree, attributeSyntax.ArgumentList.Arguments.Span)));
```
silkfire: I've always used `context.FilterTree` here, is there any difference to `context.Node.SyntaxTree`?

src/BenchmarkDotNet.Analyzers/BenchmarkRunner/RunAnalyzer.cs
```diff
@@ -54,14 +54,14 @@ public class RunAnalyzer : DiagnosticAnalyzer
...
-    [
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => new DiagnosticDescriptor[]
+    {
         TypeArgumentClassMissingBenchmarkMethodsRule,
         TypeArgumentClassMustBePublicRule,
         TypeArgumentClassMustBeUnsealedRule,
         TypeArgumentClassMustBeNonAbstractRule,
         GenericTypeArgumentClassMustBeAnnotatedWithAGenericTypeArgumentsAttributeRule,
-    ];
+    }.ToImmutableArray();
```
silkfire, line 64: Is there any reason why you can't use collection expressions here?

src/BenchmarkDotNet.Analyzers/AnalyzerHelper.cs
timcassell: I reverted your changes from #2865 (I was trying to see if it was related) before I started working on the fix. This code path will likely be removed, pending testing again the issue that the PR was addressing.

src/BenchmarkDotNet.Analyzers/Attributes/ParamsAttributeAnalyzer.cs
timcassell: `FilterTree` doesn't exist in v2.8, so I had to change it to that. According to CoPilot, `FilterTree` only returns C# syntax trees, while `Node.SyntaxTree` could return VB syntax trees. Since we're only analyzing C#, it doesn't matter.

src/BenchmarkDotNet.Analyzers/BenchmarkRunner/RunAnalyzer.cs
timcassell: v2.8 uses an older version of immutable collections package that doesn't support it. I tried manually adding the latest version, but the compiler complained that it couldn't find the assembly when it ran the analyzer.

timcassell:
> Didn't quite follow on the reasoning behind removing a lot of the logic for checking assignability. I don't really want to trigger a diagnostic if the user supplies an illegal value to the attribute, e.g. multidimensional arrays `string[][]`, nullable struct arrays `int?[]` or reference types that aren't `null` or `string` or `System.Type`.

Much of the logic became unnecessary under the semantic model. The new implementation doesn't check the type of the array passed to the attribute (other than `null`), it only checks the type of each element.

src/BenchmarkDotNet.Analyzers/AnalyzerHelper.cs
timcassell: `compilation.ClassifyConversion` and `semanticModel.ClassifyConversion` just does the thing without having to manually construct a dummy syntax tree.

silkfire:
Was the issue of assigning e.g. a smaller `int` to a `byte` resolved? Otherwise looks good to me.

timcassell:
> Was the issue of assigning e.g. a smaller `int` to a `byte` resolved? Otherwise looks good to me.

Yes, it's covered by your tests. Thanks.

PR #2872, opened 2025-11-21
fix(wasm): prevent generated project from being trimmed out

FreyLuis:
When running WASM benchmarks with AOT, the autogenerated classes `Runnable_x` are trimmed out.
This PR prevents this by adding the generated assembly to the already existing `TrimmerRootDescriptor` xml file.


FreyLuis:
@dotnet-policy-service agree company="twobrains GmbH"

Issue #2871, opened 2025-11-21
Additional dependency to `System.Text.Json` NuGet package


filzrev:
After `Microsoft.Diagnostics.Tracing.TraceEvent` package dependency is updated to `v3.1.21`.

Unintended transitive package dependency seems be added.
https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/3.1.21#dependencies-body-tab

Is this additional dependency is acceptable for BenchmarkDotNet?
If not, it need to fix version to `v3.1.20` that don't using System.Text.Json

If it's acceptable,
It can use System.Text.Json for following cases.

- Replace code that using `SimpleJson` for JSON serialization
- It can use strongly typed model instead of [Dictionary<string, object> based serialization](https://github.com/dotnet/BenchmarkDotNet/blob/d2be797a77daf561630f24b235008533edb13253/src/BenchmarkDotNet/Exporters/Json/JsonExporterBase.cs#L31-L41)

timcassell:
Perhaps we should remove it (TraceEvent) as a dependency altogether, and move the diagnosers that depend on it to a separate package (`DisassemblyDiagnoser` and `EventPipeProfiler`)? cc @adamsitnik

adamsitnik:
I took at look at PerfView source code and it seems like it's a must have dependency for TraceEvent now: https://github.com/microsoft/perfview/pull/2183 (cc @brianrob)

> Perhaps we should remove it (TraceEvent) as a dependency altogether, and move the diagnosers that depend on it to a separate package (`DisassemblyDiagnoser` and `EventPipeProfiler`)? cc [@adamsitnik](https://github.com/adamsitnik)

We are providing a dedicated NuGet [package](https://www.nuget.org/packages/BenchmarkDotNet.Diagnostics.Windows) for Windows-specific diagnosers. But over the years, there was quite a lot of users who tried to use some of its features without installing the package (for example, via command line arguments that then tried to [load](https://github.com/dotnet/BenchmarkDotNet/blob/d2be797a77daf561630f24b235008533edb13253/src/BenchmarkDotNet/Diagnosers/DiagnosersLoader.cs#L76-L81) the external assembly) and ignored the error that clearly say that they should install it.

Since there are very few products that take dependency on BenchmarkDotNet, I don't mind having this extra dependency and increased size on disk. If BDN was a library that more products would directly depend on, we could do it.

> It can use System.Text.Json for following cases.

[...]

PR #2876, opened 2025-11-26
chore: Rollback `Microsoft.Diagnostics.Tracing.TraceEvent` package version to 3.1.20

filzrev:
This PR is intended to workaround of #2871

`Microsoft.Diagnostics.Tracing.TraceEvent` v3.1.21 or later has additional `System.Text.Json` dependency.
So temporary rollback version `v3.1.20`.

If it need to use latest version of `TraceEvent` package.
It need to accept dependencies to `System.Text.Json`.
or need to introduce separate package approach.

**Release notes of TraceEvent**
https://github.com/microsoft/perfview/releases


adamsitnik (commented):
I am not a big fan of this workaround, as sooner or later we are going to need to update TraceEvent to take advantage of some bug fix or new feature.

I suggest we keep it as is.

filzrev:
> I suggest we keep it as is.

Thanks for review.
I'll close this PR.

PR #2877, opened 2025-11-27
fix: Deadlock issue when invoked external command write a lot of output 

### LINKED ISSUE #2873: Running benchmarks hangs at "// Validating benchmarks:" when you have many, many SDKs installed
If you have many, many SDKs installed (in our case 89 ... this is from a localized dotnet install via the UseDotNet task in devops pipelines), and then attempt to run benchmarks, you get stuck at the

```
// Validating benchmarks:
```

Bringing up ProcessMonitor, we could see the host benchmark executable was waiting on 

```
dotnet --list-sdks
```

to finish. We assume from DotNetSdkValidator.GetInstalledDotNetSdks(). That function does:

```csharp
process.WaitForExit();
// stuff
var output = process.StandardOutput.ReadToEnd();
```

However, per the docs (https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.redirectstandardoutput?view=net-9.0#remarks) that approach will deadlock when the stdout buffer gets full. Instead you want to:

```csharp
var output = process.StandardOutput.ReadToEnd();
process.WaitForExit();
```

filzrev:
This PR intended to fix #2873 

#### 1. `DotNetSdkValidator.cs`
Modify code to call ReadToEnd before WaitForExit.

#### 2. `ProcessHelper.cs`
Remove `RedirectStandardError = true` setting because stderror is not consumed.

#### 3. `ProcessExtensions.cs`
Modify code to asynchronously  read StandardOutput to support timeout. (ReadToEnd don't support timeout)
And change code to use [KillTree](https://github.com/dotnet/BenchmarkDotNet/blob/d2be797a77daf561630f24b235008533edb13253/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs#L163-L181) when timeout happens. (It might wait 30 seconds if kill signal is not property handled though)

### DIFF
--- a/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
@@ -1,8 +1,3 @@
-using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Diagnostics;
-using System.IO;
 using BenchmarkDotNet.Characteristics;
 using BenchmarkDotNet.Detectors;
 using BenchmarkDotNet.Engines;
@@ -13,6 +8,13 @@ using BenchmarkDotNet.Portability;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.CoreRun;
 using JetBrains.Annotations;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
 
 
 namespace BenchmarkDotNet.Extensions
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
```diff
@@ -210,27 +212,51 @@ private static void GetAllChildIdsUnix(int parentId, HashSet<int> children, Time
 
         private static (int exitCode, string output) RunProcessAndReadOutput(string fileName, string arguments, TimeSpan timeout)
         {
-            var startInfo = new ProcessStartInfo
+            using var process = new Process
```
adamsitnik, line 215: You could also just use the [AsyncProcessOutputReader](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Loggers/AsyncProcessOutputReader.cs) here.

src/BenchmarkDotNet/Validators/DotNetSdkValidator.cs
```diff
@@ -84,11 +84,11 @@ private static IEnumerable<Version> GetInstalledDotNetSdks(string? customDotNetC
                         return Enumerable.Empty<Version>();
                     }
 
+                    var output = process.StandardOutput.ReadToEnd();
```
adamsitnik, line 87: Same as above, we could use the AsyncProcessOutputReader helper here.

src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
filzrev: I've modified code to use AsyncProcessOutputReader at #2878

src/BenchmarkDotNet/Validators/DotNetSdkValidator.cs
filzrev: This method is used for dedicated command.
And run without problem by using synchronous `ReadToEnd()`.

PR #2879, opened 2025-11-28
Allow filters to filter out every benchmark from a type (Fixes #2860)

### LINKED ISSUE #2860: BenchmarkRunner cannot Run(Assembly) if the filter eliminates all benchmark cases in any type
I am using [`BenchmarkRunner.Run(Assembly, config)`](https://github.com/dotnet/BenchmarkDotNet/blob/b8af587f76c980e6a16b6fbed90dd650f25b82b9/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs#L48-L53) with a custom filter.  I wanted to run all benchmarks in assembly _except_ a few filtered-out cases.  Unfortunately I am finding that this does not work if my filter happens to eliminate all benchmark cases _within any one type_.  When this happens, a validation error prevents running the benchmarks:

> // Validating benchmarks:
> //    * No [Benchmark] attribute found on 'MyTypeWithFilteredOutCases' benchmark case.

I found a workaround: discover types-with-benchmarks manually, use [BenchmarkConverter.TypeToBenchmarks](https://benchmarkdotnet.org/api/BenchmarkDotNet.Running.BenchmarkConverter.html#BenchmarkDotNet_Running_BenchmarkConverter_TypeToBenchmarks_System_Type_BenchmarkDotNet_Configs_IConfig_) to generate runInfos, and skip runInfos where config's filter removes all cases.

That works, but it seems like BenchmarkRunner would be better with this behavior by default.

User0332:
Fixes #2860 

## Previous Behavior
Running [`BenchmarkRunner.Run(Assembly, config)`](https://github.com/dotnet/BenchmarkDotNet/blob/b8af587f76c980e6a16b6fbed90dd650f25b82b9/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs#L48-L53) with a filter that eliminated all benchmarks from *any one type* within the specified assembly would error  `No [Benchmark] attribute found on 'MyTypeWithFilteredOutCases' benchmark case` and not run *any benchmarks* within the assembly.

## New Behavior

Even if all the benchmarks from one type are completely filtered out, [`BenchmarkRunner.Run(Assembly, config)`](https://github.com/dotnet/BenchmarkDotNet/blob/b8af587f76c980e6a16b6fbed90dd650f25b82b9/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs#L48-L53) will still run the selected benchmarks from all other types within the assembly.

If the filter happens to eliminate all benchmarks across all types within the assembly, `There are no benchmarks found` will be printed and zero benchmarks will run.

### DIFF
--- a/tests/BenchmarkDotNet.Tests/Running/RunningEmptyBenchmarkTests.cs
+++ b/tests/BenchmarkDotNet.Tests/Running/RunningEmptyBenchmarkTests.cs
@@ -313,11 +313,6 @@ namespace BenchmarkDotNet.Tests.Running
         [Fact]
         public void AssemblyWithBenchmarks_RunsNothing_WhenAllBenchmarksFilteredOutFromOneTypeWithBenchmarkAttributePresent()
         {
-
-            // Skip test on .NET Framework 4.6.2
-            if (RuntimeInformation.FrameworkDescription.Contains(".NET Framework 4"))
-                return;
-
             // Create a mock assembly with benchmark types
             var assemblyName = new AssemblyName("MockAssemblyWithBenchmarks");
             var assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(assemblyName, AssemblyBuilderAccess.Run);


User0332:
@dotnet-policy-service agree

timcassell:
Can you please add a test for the new behavior?

User0332:
Done!

The name is kind of long though, let me know if I should change it.

tests/BenchmarkDotNet.Tests/Running/RunningEmptyBenchmarkTests.cs
```diff
@@ -308,6 +309,47 @@ public void AssemblyWithBenchmarks_RunsSuccessfully_WhenBenchmarkAttributePresen
                 Assert.DoesNotContain(summary.ValidationErrors, validationError => validationError.Message == GetGeneralValidationError());
             }
         }
+
+        [Fact]
+        public void AssemblyWithBenchmarks_RunsNothing_WhenAllBenchmarksFilteredOutFromOneTypeWithBenchmarkAttributePresent()
+        {
+
+            // Skip test on .NET Framework 4.6.2
+            if (RuntimeInformation.FrameworkDescription.Contains(".NET Framework 4"))
```
timcassell: Why skip .Net Framework? If you really need to skip it, use `[FactEnvSpecific("reason", EnvRequirement.NonFullFramework]`.

User0332: Sorry, I had copied that over from the previous assembly test case so I just left it in. Fixed in the most recent commit

timcassell:
Thanks @User0332

Issue #2815, opened 2025-07-24
[Feature Request] Enable nullable type checks


filzrev:
Currently [BenchmarkDotNet using `<Nullable>annotations</Nullable>` setting](https://github.com/dotnet/BenchmarkDotNet/blob/master/build/common.props#L25).

Is there a plan to migrate to use `<Nullable>enable</Nullable>` in future?

**Migration path**
Currently 828 errors are raised when setting `<Nullable>enable</Nullable>` on solution-level.
Therefore, it might be better to migrate with following steps.

1. Enable nullable check with `#nullable enable` setting per-files. and fix build errors.
2. When entire code is refactored. Migrate to use `<Nullable>enable</Nullable>`. And remove `#nullable enable` settings

**Background**
When accessing [`ExecuteResult::StandardOutput` property](https://github.com/dotnet/BenchmarkDotNet/blob/daa233906cc5aba9aa94d4bc1b3dc449e42806e9/src/BenchmarkDotNet/Toolchains/Results/ExecuteResult.cs#L22).
It's expected to be non-null value is returned.
But it returns `null` in some cases (When used constructor that don't initialize `StandardOutput` property)

kapilepatel:
Can I work on this?
If anyone already started, please share here.
I welcome any suggestions you might have.

timcassell:
@kapilepatel Sure you can work on it. I'm fine if you want to enable nullable and fix all the errors at once, or do it more piece-meal like @filzrev suggested. Up to you.

PR #2881, opened 2025-11-28
chore: Add PolySharp dev dependency and cleanup #if-#endif preprocessor

filzrev:
This PR add [PolySharp](https://github.com/Sergio0694/PolySharp) dev dependency.
And cleanup #if-#endif preprocessor that are used on `SupportedOSPlatform`/`SupportedOSPlatformGuard`.

The purpose of adding PolySharp dependency if followings. 
- Using [Nullability attributes](https://github.com/Sergio0694/PolySharp?tab=readme-ov-file#documentation-) that expected to be used for #2815
- Using newer language features on `netstandard2.0` tfm (e.g. init-only properties/required field)


timcassell:
Thanks. I think PolySharp also generates DynamicDependencyAttribute, so we should be able to remove the ifdefs around it also, right?

filzrev:
I've add commit that remove `#if NET6_0_OR_GREATER` on `DynamicDependencyAttribute`

PR #2886, opened 2025-11-30
Allow parameterized arrays to be differentiated using value comparison (Fixes #2838)

### LINKED ISSUE #2838: Ratio does not get reset for new input set when using ArgumentsSource
Hello,

When using `ArgumentsSource` to drive parameterization of benchmarks, the ratio column is calculated from the first test in the suite, and not the baseline per ArgumentsSource input.

I am running `.NET 9` and BenchmarkDotNet version `0.15.4`.

Example of the problem (see ratio column)::
```
| Method                             | a               | b               | Mean                | Error             | StdDev            | Ratio |
|----------------------------------- |---------------- |---------------- |--------------------:|------------------:|------------------:|------:|
| NaiveMatrixMultiply                | Double[1048576] | Double[1048576] | 2,240,938,607.69 ns | 11,998,955.271 ns | 10,019,671.089 ns | 1.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[1048576] | Double[1048576] | 1,257,492,000.00 ns |  6,858,165.966 ns |  6,415,132.703 ns | 0.561 |
| NaiveMatrixMultiply                | Double[10]      | Double[10]      |            26.92 ns |          0.131 ns |          0.122 ns | 0.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[10]      | Double[10]      |            29.85 ns |          0.220 ns |          0.206 ns | 0.000 |
| NaiveMatrixMultiply                | Double[4096]    | Double[4096]    |       347,221.54 ns |      2,192.308 ns |      2,050.687 ns | 0.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[4096]    | Double[4096]    |       294,919.02 ns |      1,008.847 ns |        842.433 ns | 0.000 |
| NaiveMatrixMultiply                | Double[81]      | Double[81]      |           846.51 ns |          3.732 ns |          3.309 ns | 0.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[81]      | Double[81]      |           868.75 ns |          3.295 ns |          3.082 ns | 0.000 |
| NaiveMatrixMultiply                | Double[9]       | Double[9]       |            48.58 ns |          0.293 ns |          0.244 ns | 0.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[9]       | Double[9]       |            52.31 ns |          0.585 ns |          0.519 ns | 0.000 |
```

I would expect to see this (see ratio column):
```
| Method                             | a               | b               | Mean                | Error             | StdDev            | Ratio |
|----------------------------------- |---------------- |---------------- |--------------------:|------------------:|------------------:|------:|
| NaiveMatrixMultiply                | Double[1048576] | Double[1048576] | 2,240,938,607.69 ns | 11,998,955.271 ns | 10,019,671.089 ns | 1.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[1048576] | Double[1048576] | 1,257,492,000.00 ns |  6,858,165.966 ns |  6,415,132.703 ns | 0.561 |
| NaiveMatrixMultiply                | Double[10]      | Double[10]      |            26.92 ns |          0.131 ns |          0.122 ns | 1.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[10]      | Double[10]      |            29.85 ns |          0.220 ns |          0.206 ns | x.xxx |
| NaiveMatrixMultiply                | Double[4096]    | Double[4096]    |       347,221.54 ns |      2,192.308 ns |      2,050.687 ns | 1.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[4096]    | Double[4096]    |       294,919.02 ns |      1,008.847 ns |        842.433 ns | x.xxx |
| NaiveMatrixMultiply                | Double[81]      | Double[81]      |           846.51 ns |          3.732 ns |          3.309 ns | 1.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[81]      | Double[81]      |           868.75 ns |          3.295 ns |          3.082 ns | x.xxx |
| NaiveMatrixMultiply                | Double[9]       | Double[9]       |            48.58 ns |          0.293 ns |          0.244 ns | 1.000 |
| CacheBlockedMultiDimMatrixMultiply | Double[9]       | Double[9]       |            52.31 ns |          0.585 ns |          0.519 ns | x.xxx |
```

Here is my benchmark.

```
public class MatrixBenchmarks
{
    private static double[,] GenerateRandomMatrix(Random rng, int m, int n)
    {
        double scale = 10;
        double[,] matrix = new double[m, n];

        for (int i = 0; i < m; i++)
        {
            for (int j = 0; j < n; j++)
            {
                matrix[i, j] = scale * rng.NextDouble();
            }
        }

        return matrix;
    }

    public Random rng = new();

    [Benchmark(Baseline = true)]
    [ArgumentsSource(nameof(MultiDimMatrices))]
    public double[,] NaiveMatrixMultiply(double[,] a, double[,] b)
    {
        return Matrix.MultiplyMultiDimNaive(a, b);
    }

    [Benchmark]
    [ArgumentsSource(nameof(MultiDimMatrices))]
    public double[,] CacheBlockedMultiDimMatrixMultiply(double[,] a, double[,] b)
    {
        return Matrix.MultiplyMultiDimCacheBlocked(a, b);
    }

    public IEnumerable<object[]> MultiDimMatrices()
    {
        yield return new object[] { GenerateRandomMatrix(rng, 3, 3), GenerateRandomMatrix(rng, 3, 3) };
        yield return new object[] { GenerateRandomMatrix(rng, 9, 9), GenerateRandomMatrix(rng, 9, 9) };
        yield return new object[] { GenerateRandomMatrix(rng, 1, 10), GenerateRandomMatrix(rng, 10, 1) };
        yield return new object[] { GenerateRandomMatrix(rng, 64, 64), GenerateRandomMatrix(rng, 64, 64) };
        yield return new object[] { GenerateRandomMatrix(rng, 1024, 1024), GenerateRandomMatrix(rng, 1024, 1024) };
    }
}
```

Thank you.

User0332:
Solves a larger problem dealing with array comparison that fixes #2838

## Previous Behavior

In the following example,

```cs
public class ArrayOnDifferentMethods
{
	public IEnumerable<int[]> GetArrays()
	{
		yield return new int[] { 1, 2, 3 };
		yield return new int[] { 2, 3, 4 };
	}

	[Benchmark(Baseline = true)]
	[ArgumentsSource(nameof(GetArrays))]
	public void AcceptsArrays(int[] arr)
	{
		int[] x;

		if (arr[0] == 1) x = [.. arr.Order()];
		else x = [];

		if (x.Length != 3 && x.Length != 0)
			throw new ArgumentException("Incorrect length");
	}

	[Benchmark]
	[ArgumentsSource(nameof(GetArrays))]
	public void AcceptsArrays2(int[] arr)
	{
		if (arr.Length != 3)
			throw new ArgumentException("Incorrect length");
	}
}
```
the arguments `int[] { 1, 2, 3 }` and `int[] { 2, 3, 4 }` would not be recognized as different because their [`ParameterInstances.ValueInfo`](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Parameters/ParameterInstances.cs#L34) property--which [`DefaultOrderer`](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Order/DefaultOrderer.cs#L128) uses to generate logical groups--depended on `object.ToString()`, which for an array just provides the type and length of the array (in this case, they were equal, so the library interpreted them as the same argument). Because they were seen as the same, the cases where `arr = { 1, 2, 3 }` and `arr = { 2, 3, 4 }` were not separated into different logical groups and the runtime ratio for *all cases* was calculated with respect to the first case `AcceptsArrays({ 1, 2, 3 })`. The desired behavior would be for the cases with corresponding arguments to be grouped properly and have the baseline runtime calculated once for `arr = { 1, 2, 3 }` and another time for `arr = { 2, 3, 4 }`.

## Solution

Have [`ParameterInstances.ValueInfo`](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Parameters/ParameterInstances.cs#L34) depend on a customizable [`ToValueText`](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Parameters/ParameterInstance.cs#L65) (which eventually propagates down to a new property [`IParam.ValueText`](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Code/IParam.cs#L19)) rather than `ToString()` . For most param types, [`IParam.ValueText`](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Code/IParam.cs#L19) will point to [`IParam.ValueText`](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Code/IParam.cs#L14), but for arrays, it will contain [extra information like `Array.Rank` as well as a hash computed based off of the array's *elements*](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/src/BenchmarkDotNet/Code/ArrayParam.cs#L15) to properly distinguish between different-valued arrays of the same size.

Downsides of this approach
- As the array size grows, this approach to differentiating different arrays might grow expensive.
- This approach is prone to hash collisions; it uses hashes because it is unreasonable to add every array element to a string representation (especially if the array is large). This issue is mitigated by the fact that the array length, rank, *and* value hash must all be equivalent for two arrays to be seen as equivalent. 

## New Behavior

Benchmark cases accepting an array parameter or argument will be grouped by array values rather than just length. 

The new test case [ArrayArgumentsCanBeGrouped](https://github.com/User0332/BenchmarkDotNet/blob/975a572264664b8ac62dd96eb9f0815aea091f5f/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs#L548C21-L548C47) tests for this behavior.

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ArgumentsTests.cs
@@ -1,4 +1,5 @@
 ﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Tests.XUnit;
 using BenchmarkDotNet.Toolchains;
@@ -311,9 +312,9 @@ namespace BenchmarkDotNet.IntegrationTests
                     throw new ArgumentNullException(nameof(array));
 
                 for (int i = 0; i < 10; i++)
-                for (int j = 0; j < i; j++)
-                    if (array[i][j] != i)
-                        throw new ArgumentException("Invalid value");
+                    for (int j = 0; j < i; j++)
+                        if (array[i][j] != i)
+                            throw new ArgumentException("Invalid value");
             }
 
             public IEnumerable<object> CreateMatrix()
@@ -540,6 +541,46 @@ namespace BenchmarkDotNet.IntegrationTests
 
                 if (!notEven.All(n => n % 2 != 0))
                     throw new ArgumentException("Even");
+            }
+        }
+
+        [Theory, MemberData(nameof(GetToolchains), DisableDiscoveryEnumeration = true)]
+        public void ArrayArgumentsCanBeGrouped(IToolchain toolchain)
+        {
+            var summary = CanExecute<ArrayOnDifferentMethods>(toolchain, (config) => config.AddLogicalGroupRules(BenchmarkLogicalGroupRule.ByParams)); // We must group by params to test array argument grouping
+
+            // There should be two logical groups, one for the first argument and one for the second argument
+            // (thus there should be two pairs per descriptor, and each pair should be distinct because it belongs to a different group)
+
+            var descriptorGroupPairs = summary.BenchmarksCases.Select(benchmarkCase => (benchmarkCase.Descriptor, summary.GetLogicalGroupKey(benchmarkCase))).GroupBy(benchmarkCase => benchmarkCase.Descriptor);
+
+            Assert.True(
+                descriptorGroupPairs.All(group => group.Select(pair => pair.Item2).Distinct().Count() == 2)
+            );
+        }
+
+        public class ArrayOnDifferentMethods
+        {
+            public IEnumerable<int[]> GetArrays()
+            {
+                yield return new int[] { 1, 2, 3 };
+                yield return new int[] { 2, 3, 4 };
+            }
+
+            [Benchmark(Baseline = true)]
+            [ArgumentsSource(nameof(GetArrays))]
+            public void AcceptsArrays(int[] arr)
+            {
+                if (arr.Length != 3)
+                    throw new ArgumentException("Incorrect length");
+            }
+
+            [Benchmark]
+            [ArgumentsSource(nameof(GetArrays))]
+            public void AcceptsArrays2(int[] arr)
+            {
+                if (arr.Length != 3)
+                    throw new ArgumentException("Incorrect length");
             }
         }
 
@@ -1026,6 +1067,16 @@ namespace BenchmarkDotNet.IntegrationTests
             }
         }
 
-        private void CanExecute<T>(IToolchain toolchain) => CanExecute<T>(CreateSimpleConfig(job: Job.Dry.WithToolchain(toolchain)));
+        private Reports.Summary CanExecute<T>(IToolchain toolchain, Func<IConfig, IConfig> furtherConfigure = null)
+        {
+            var config = CreateSimpleConfig(job: Job.Dry.WithToolchain(toolchain));
+
+            if (furtherConfigure is not null)
+            {
+                config = furtherConfigure(config);
+            }
+
+            return CanExecute<T>(config);
+        }
     }
 }


timcassell:
2 issues I have with this.

Hashes should not be used to determine equality, as you mentioned they can collide. If we really want to go down this road, we should probably refactor the system to get it correct.

Users may be surprised to see it work for arrays, but not lists and other collection types.

I wonder if we should just leave it up to the user somehow (I'm not sure if the current system lets users override it already).

User0332:
I agree with the hashing equality thing and that we probably refactor to not use ValueText

On the docs there is a snippet on wrapping things like arrays and other collections which don't have signatures built in to their ToString in a custom class that provides a unique ToString, so maybe this is a wontfix?

https://benchmarkdotnet.org/articles/features/parameterization.html#another-example:~:text=ns%20%7C%200.3063%20ns%20%7C-,Another%20example,-If%20the%20values

User0332:
On the other hand I could bring back my old ParameterComparer code that worked for arrays via StructuralComparer (so we could probably get it to work with most collections through LINQ conversions) and refactor the way logical groups are organized (to go for object comparison rather than ValueInfo comparsion)

timcassell:
Well, looking at the OP again, it looks like the display value of the text is not being taken into account for the array comparisons, so that should be fixed at least (so it will at least take length into account). I also see that it shows multi dimensional arrays as single dimension, so that should also be fixed. But probably yeah we leave the true equality comparisons as won't fix, and the user can wrap it in a struct that implements `IComparable` and do it themself.

timcassell:
> On the other hand I could bring back my old ParameterComparer code that worked for arrays via StructuralComparer (so we could probably get it to work with most collections through LINQ conversions) and refactor the way logical groups are organized (to go for object comparison rather than ValueInfo comparsion)

I'm open to trying that out.

User0332:
> > On the other hand I could bring back my old ParameterComparer code that worked for arrays via StructuralComparer (so we could probably get it to work with most collections through LINQ conversions) and refactor the way logical groups are organized (to go for object comparison rather than ValueInfo comparsion)
> 
> I'm open to trying that out.

I'll try this first and otherwise will just go for fixing #2838 

Closing this PR for now since both changes would go in a different direction.

PR #2887, opened 2025-12-01
Compare parameters by object value rather than string representation (Related to #2838)

User0332:
This is related to issue #2838 and PR #2886

## Previous Behavior

Parameters and arguments passed to benchmarks were grouped into logical groups based on `ParameterInstances.ValueInfo` (a string representation), which failed for types like arrays (whose string representations just contain length and element type), (partially) causing #2838. Any type where objects are not equal but can have the same `ToString()` was affected by this behavior.

## New Behavior

When using `DefaultOrderer` to group benchmark cases, parameters are compared by their actual value as opposed to their string representations, leading to correct benchmark case grouping. This is done through `ParameterEqualityComparer`, so any parameter or custom wrapping struct for a parameter overriding `object.Equals` will now be properly distinguished regardless of its `ToString` override. For any `IEnumerable`s (both non-generic and generic), there is now built-in functionality in `ParameterComparer` which compares enumerables element-by-element for value-based equality.

### DIFF
--- a/src/BenchmarkDotNet/Parameters/ParameterComparer.cs
+++ b/src/BenchmarkDotNet/Parameters/ParameterComparer.cs
@@ -1,215 +1,287 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
 using BenchmarkDotNet.Portability;
+
+#if NETSTANDARD2_0
+using System.Collections.Concurrent;
+using System.Linq;
+#endif
 
 namespace BenchmarkDotNet.Parameters
 {
     internal class ParameterComparer : IComparer<ParameterInstances>
     {
+#if NETSTANDARD2_0
+        private static readonly ConcurrentDictionary<Type, MemberInfo[]> s_tupleMembersCache = new();
+
+        private static MemberInfo[] GetTupleMembers(Type type)
+            => s_tupleMembersCache.GetOrAdd(type, t =>
+            {
+                var members = type.FullName.StartsWith("System.Tuple`")
+                    ? type.GetProperties(BindingFlags.Public | BindingFlags.Instance).Cast<MemberInfo>()
+                    : type.GetFields(BindingFlags.Public | BindingFlags.Instance).Cast<MemberInfo>();
+                return members
+                    .Where(p => p.Name.StartsWith("Item"))
+                    .OrderBy(p => p.Name)
+                    .ToArray();
+            });
+#endif
+
         public static readonly ParameterComparer Instance = new ParameterComparer();
 
         public int Compare(ParameterInstances x, ParameterInstances y)
         {
             if (x == null && y == null) return 0;
-            if (x != null && y == null) return 1;
             if (x == null) return -1;
+            if (y == null) return 1;
+
             for (int i = 0; i < Math.Min(x.Count, y.Count); i++)
             {
-                var compareTo = CompareValues(x[i]?.Value, y[i]?.Value);
-                if (compareTo != 0)
-                    return compareTo;
+                var comparison = CompareValues(x[i]?.Value, y[i]?.Value);
+                if (comparison != 0)
+                {
+                    return comparison;
+                }
             }
 
             return string.CompareOrdinal(x.DisplayInfo, y.DisplayInfo);
         }
 
-        private int CompareValues<T1, T2>(T1 x, T2 y)
+        private static int CompareValues<T1, T2>(T1 x, T2 y)
         {
-            // Detect IComparable implementations.
-            // This works for all primitive types in addition to user types that implement IComparable.
-            if (x != null && y != null && x.GetType() == y.GetType())
+            if (x == null || y == null || x.GetType() != y.GetType())
             {
-                if (x is IStructuralComparable xStructuralComparable)
+                return string.CompareOrdinal(x?.ToString(), y?.ToString());
+            }
+
+            if (x is IStructuralComparable xStructuralComparable)
+            {
+                try
                 {
-                    if (x is Array xArr && y is Array yArr)
-                    {
-                        if (xArr.Rank != yArr.Rank) return xArr.Rank.CompareTo(yArr.Rank);
-
-                        for (int dim = 0; dim < xArr.Rank; dim++)
-                        {
-                            if (xArr.GetLength(dim) != yArr.GetLength(dim))
-                                return xArr.GetLength(dim).CompareTo(yArr.GetLength(dim));
-                        }
-
-                        //  1D, 2D, and 3D array comparison is optimized with dedicated methods
-                        if (xArr.Rank == 1) return StructuralComparisonWithFallback(xArr, yArr);
-
-                        if (xArr.Rank == 2 && !RuntimeInformation.IsAot)
-                        {
-                            return (int) GetType()
-                                .GetMethod(nameof(CompareTwoDimensionalArray), BindingFlags.NonPublic | BindingFlags.Instance)
-                                .MakeGenericMethod(xArr.GetType().GetElementType(), yArr.GetType().GetElementType())
-                                .Invoke(this, [xArr, yArr]);
-                        }
-
-                        if (xArr.Rank == 3 && !RuntimeInformation.IsAot)
-                        {
-                            return (int) GetType()
-                                .GetMethod(nameof(CompareThreeDimensionalArray), BindingFlags.NonPublic | BindingFlags.Instance)
-                                .MakeGenericMethod(xArr.GetType().GetElementType(), yArr.GetType().GetElementType())
-                                .Invoke(this, [xArr, yArr]);
-                        }
-
-                        return CompareEnumerables(xArr, yArr);
-                    }
-                    else // Probably a user-defined IStructuralComparable or tuple
-                    {
-                        return StructuralComparisonWithFallback(xStructuralComparable, (IStructuralComparable) y);
-                    }
+                    return StructuralComparisons.StructuralComparer.Compare(x, y);
                 }
-                else if (x is IComparable xComparable)
+                // https://github.com/dotnet/BenchmarkDotNet/issues/2346
+                // https://github.com/dotnet/runtime/issues/66472
+                // Unfortunately we can't rely on checking the exception message because it may change per current culture.
+                catch (ArgumentException)
                 {
-                    try
+                    if (TryFallbackStructuralCompareTo(x, y, out int comparison))
                     {
-                        return xComparable.CompareTo(y);
+                        return comparison;
                     }
-                    // Some types, such as Tuple and ValueTuple, have a fallible CompareTo implementation which can throw if the inner items don't implement IComparable.
-                    // See: https://github.com/dotnet/BenchmarkDotNet/issues/2346
-                    // For now, catch and ignore the exception, and fallback to string comparison below.
-                    catch (ArgumentException ex) when (ex.Message.Contains("At least one object must implement IComparable."))
-                    {
-                    }
+                    // A complex user type did not handle a multi-dimensional array or tuple, just re-throw.
+                    throw;
                 }
-                else if (x is IEnumerable xEnumerable  && y is IEnumerable yEnumerable) // General collection comparison support
-                {
-                    return CompareEnumerables(xEnumerable, yEnumerable);
-                }
+            }
+
+            if (x is IComparable xComparable)
+            {
+                // Tuples are already handled by IStructuralComparable case, if this throws, it's the user's own fault.
+                return xComparable.CompareTo(y);
+            }
+
+            if (x is IEnumerable xEnumerable) // General collection comparison support
+            {
+                return CompareEnumerables(xEnumerable, (IEnumerable) y);
             }
 
             // Anything else to differentiate between objects.
             return string.CompareOrdinal(x?.ToString(), y?.ToString());
         }
 
-        private int StructuralComparisonWithFallback(IStructuralComparable x, IStructuralComparable y)
+        private static bool TryFallbackStructuralCompareTo(object x, object y, out int comparison)
         {
-            try
+            // Check for multi-dimensional array and ITuple and re-try for each element recursively.
+            if (x is Array xArr)
             {
-                return StructuralComparisons.StructuralComparer.Compare(x, y);
-            }
-            // Some types, such as Tuple and ValueTuple, have a fallible CompareTo implementation which can throw if the inner items don't implement IComparable.
-            // See: https://github.com/dotnet/BenchmarkDotNet/issues/2346
-            // For now, catch and ignore the exception, and fallback to string comparison below.
-            catch (ArgumentException ex) when (ex.Message.Contains("At least one object must implement IComparable."))
-            {
-                var ITuple = Type.GetType("System.Runtime.CompilerServices.ITuple");
-
-                if (ITuple.IsAssignableFrom(x.GetType()))
+                Array yArr = (Array) y;
+                if (xArr.Rank != yArr.Rank)
                 {
-                    var lengthProperty = ITuple.GetProperty("Length");
-
-                    var xLength = (int) lengthProperty.GetValue(x);
-                    var yLength = (int) lengthProperty.GetValue(y);
-
-                    if (xLength != yLength) return xLength.CompareTo(yLength);
-
-                    var indexerProperty = ITuple.GetProperty("Item");
-
-                    object ItemGetter(IStructuralComparable tup, int i) => indexerProperty.GetValue(tup, [i]);
-
-                    var xFlatTransformed = Enumerable.Range(0, xLength).Select(i => ItemGetter(x, i).ToString());
-                    var yFlatTransformed = Enumerable.Range(0, xLength).Select(i => ItemGetter(y, i).ToString());
-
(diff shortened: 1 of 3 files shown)


User0332:
The failing exporter checks seem to be because I changed the schema of the logical group keys, is there a way I could update what verify is expecting?

E.g. verify expects `'Only 1 job in a group can have "Baseline = true" applied to it, group Invalid_TwoJobBaselines.Foo in class Invalid_TwoJobBaselines has 2'` where the logical group key is `"Invalid_TwoJobBaselines.Foo"` but with my new parameter-grouping keys it outputs `'Only 1 job in a group can have "Baseline = true" applied to it, group Distinct Param Set 0-Invalid_TwoJobBaselines.Foo in class Invalid_TwoJobBaselines has 2'` where the logical group key is `"Distinct Param Set 0-Invalid_TwoJobBaselines.Foo"`

User0332:
I guess I could have logical groups have a key for grouping and a display name for errors, but I feel this would just add unnecessary nonfunctional code

timcassell:
You can just update the verified files.

src/BenchmarkDotNet/Parameters/ParameterEqualityComparer.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace BenchmarkDotNet.Parameters
+{
+    internal class ParameterEqualityComparer : IEqualityComparer<ParameterInstances>
+    {
+        public static readonly ParameterEqualityComparer Instance = new ParameterEqualityComparer();
+
+        public bool Equals(ParameterInstances? x, ParameterInstances? y)
+        {
+            return ParameterComparer.Instance.Compare(x, y) == 0;
```
timcassell: 2 values who have equivalent comparisons are not necessarily equal. You should rather implement the equality comparer properly.

User0332:
> You can just update the verified files.

Won't this not affect the CI though?
Or should I not worry about that

timcassell:
> > You can just update the verified files.
> 
> Won't this not affect the CI though? Or should I not worry about that

I don't understand the question. The test compares the output of the run to the verified file.

User0332:
> > > You can just update the verified files.
> > 
> > 
> > Won't this not affect the CI though? Or should I not worry about that
> 
> I don't understand the question. The test compares the output of the run to the verified file.

Sorry, didn't realize that the GitHub actions pulled tests from my PR and not master.

Will implement these changes.

User0332:
This effectively closes #2838 because it correctly groups parameters into different groups based on value. However, since the example in #2838 passes randomly generated parameters will improper seeding, the ratios won't come out as desired. If the parameters are properly seeded to be deterministic and effectively constant across calls of the ArgumentsSource, then the desired behavior will be exhibited.

With super large arrays like the ones in #2838, though, the report generation takes a few minutes (I think its because of a lot of repeated value comparison work when sorting for the summary table)

timcassell:
> With super large arrays like the ones in #2838, though, the report generation takes a few minutes (I think its because of a lot of repeated value comparison work when sorting for the summary table)

That's likely due to the boxing of every element. Perhaps it can be accelerated by using `IStructuralComparable.CompareTo` similar to the `IComparable` path, then special-case the common 2d and 3d arrays, and use the slow path for 4+d arrays. (And same thing for `IStructuralEquatable`.)

Btw, apparently md arrays throw from `IStructural...` (https://github.com/dotnet/runtime/issues/66472), which we will need to handle.

User0332:
I already flatten all enumerators (including md arrays) into an `object[]` before using `StructuralComparer`, but unfortunately since I just check for belonging `IEnumerable` and then perform a linq `OfType` query on them to convert it to `object[]`, I'm not sure if there is a way to avoid the boxing (let me know if im not understanding your comment correctly). 

Another solution I was thinking of was to id parameters with indices of the order they were returned from ArgumentsSource/Arguments attribute with; this would avoid having to do object comparison entirely as we could use the indices to track params

timcassell:
> I already flatten all enumerators (including md arrays) into an `object[]` before using `StructuralComparer`, but unfortunately since I just check for belonging `IEnumerable` and then perform a linq `OfType` query on them to convert it to `object[]`, I'm not sure if there is a way to avoid the boxing (let me know if im not understanding your comment correctly).

Right, and that flattening is the expensive part, because it boxes every element. My idea is first check for `IStructuralComparable`, then simply run `CompareTo`. If the check fails, or if it throws the exception because md arrays don't support it, then check for md array and special-case 2d and 3d, then finally fallback to the code you already have. For the special-case, you can use reflection to create generic functions to avoid the boxing (`MethodInfo.MakeGenericMethod`).

> Another solution I was thinking of was to id parameters with indices of the order they were returned from ArgumentsSource/Arguments attribute with; this would avoid having to do object comparison entirely as we could use the indices to track params

It's clever, but it won't work because 2 methods can have `[Arguments(...)]` with different order (or number of attrs).

User0332:
Thanks for clarifying, will do!

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,25 +20,54 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
+                else if (x is IEnumerable xEnumerable  && y is IEnumerable yEnumerable) // collection equality support
                 {
+                    if (x is Array xArr && y is Array yArr) // check rank here for arrays because their values will get compared when flattened
+                    {
+                        if (xArr.Rank != yArr.Rank)
+                            return xArr.Rank.CompareTo(yArr.Rank);
+                    }
+
+                    var xFlat = xEnumerable.OfType<object>().ToArray();
+                    var yFlat = yEnumerable.OfType<object>().ToArray();
```
timcassell: Should probably manually iterate the enumerables to avoid materializing the entirety of the boxed elements at once.

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,25 +20,54 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
-                // Some types, such as Tuple and ValueTuple, have a fallible CompareTo implementation which can throw if the inner items don't implement IComparable.
-                // See: https://github.com/dotnet/BenchmarkDotNet/issues/2346
-                // For now, catch and ignore the exception, and fallback to string comparison below.
-                catch (ArgumentException ex) when (ex.Message.Contains("At least one object must implement IComparable."))
+                else if (x is IEnumerable xEnumerable  && y is IEnumerable yEnumerable) // collection equality support
                 {
+                    if (x is Array xArr && y is Array yArr) // check rank here for arrays because their values will get compared when flattened
+                    {
+                        if (xArr.Rank != yArr.Rank)
+                            return xArr.Rank.CompareTo(yArr.Rank);
```
timcassell: You could also short-circuit for `LongLength`.

User0332:
I implemented the changes for ParameterComparer but realized that it alone is not actually the culprit, so I reverted a few of the things (like the special case methods) to maintain simplicity.

I'm currently investigating to see what else might be causing the slowdown

User0332:
Ok, so it *was* a combination of ParameterComparer and ParameterEqualityComparer. It's sped up by magnitudes now. Thanks for the suggestion!

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,30 +21,154 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
+                        return xComparable.CompareTo(y);
+                    }
+                    // Some types, such as Tuple and ValueTuple, have a fallible CompareTo implementation which can throw if the inner items don't implement IComparable.
+                    // See: https://github.com/dotnet/BenchmarkDotNet/issues/2346
+                    // For now, catch and ignore the exception, and fallback to string comparison below.
+                    catch (ArgumentException ex) when (ex.Message.Contains("At least one object must implement IComparable."))
+                    {
+                    }
+                }
+                else if (x is IStructuralComparable)
```
timcassell: I would check this before `IComparable` (tuples implement it deeply).

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,30 +21,154 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
+                        if (xArr.Rank != yArr.Rank) return xArr.Rank.CompareTo(yArr.Rank);
+
+                        for (int dim = 0; dim < xArr.Rank; dim++)
+                        {
+                            if (xArr.GetLength(dim) != yArr.GetLength(dim))
+                                return xArr.GetLength(dim).CompareTo(yArr.GetLength(dim));
+                        }
+
+                        //  1D, 2D, and 3D array comparison is optimized with dedicated methods
+                        if (xArr.Rank == 1) return StructuralComparisonWithFallback(xArr, yArr);
```
timcassell: We need the fallback case for all structural comparisons, not just arrays (md arrays can be nested in the type).

In fact, we can simplify it by just wrapping the normal case in a try/catch, and do the specialized array checks in the catch clause.

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,30 +21,154 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
+                return CompareEnumerables(xFlatHashed, yFlatHashed);
+            }
+        }
+
+        private int CompareEnumerables(IEnumerable nonGenericX, IEnumerable nonGenericY)
+        {
+            // Use this instead of StructuralComparisons.StructuralComparer to avoid resolving the whole enumerable to object[]
+
+            var x = nonGenericX.OfType<object>();
+            var y = nonGenericY.OfType<object>();
```
timcassell: Better to simply enumerate directly (`GetEnumerator`), like how the BCL implements SequenceEquals.

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,30 +21,154 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
+                else if (x is IStructuralComparable)
+                {
+                    if (x is Array xArr && y is Array yArr)
+                    {
+                        if (xArr.Rank != yArr.Rank) return xArr.Rank.CompareTo(yArr.Rank);
+
+                        for (int dim = 0; dim < xArr.Rank; dim++)
+                        {
+                            if (xArr.GetLength(dim) != yArr.GetLength(dim))
+                                return xArr.GetLength(dim).CompareTo(yArr.GetLength(dim));
```
timcassell: Just use `LongLength`. It's O(1) complexity and is equivalent since `Rank` is checked first.

timcassell (changes requested):

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
User0332: I did this because an array `int[3, 2, 3] != int[3, 3, 2]` even though `Rank` and `Length` are both equal in this case

timcassell: Nevermind, just realized a 2x3x4 != 2x4x3.

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
User0332: The other three cases (Rank 2, Rank 3, and Rank 4+) never fail because they recurse back into CompareValues, and the comparability stuff is handled there, so I only used the fallback for 1d arrays.

As for other structural comparisons (arrays), I am unable to add a similar fallback because the element structure of the type would be unknown. In my new commit, it would just catch the error and fall to the default case at the bottom of the method

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,30 +21,154 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
+
+                        //  1D, 2D, and 3D array comparison is optimized with dedicated methods
+                        if (xArr.Rank == 1) return StructuralComparisonWithFallback(xArr, yArr);
+
+                        if (xArr.Rank == 2)
+                        {
+                            return (int) GetType()
+                                .GetMethod(nameof(CompareTwoDimensionalArray), BindingFlags.NonPublic | BindingFlags.Instance)
+                                .MakeGenericMethod(xArr.GetType().GetElementType(), yArr.GetType().GetElementType())
+                                .Invoke(this, [xArr, yArr]);
```
timcassell: We should guard these cases with `RuntimeInformation.IsAot` to prevent runtime exceptions from `MakeGenericMethod` in AOT runtimes.

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,30 +21,154 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
+                    return CompareEnumerables(xEnumerable, yEnumerable);
+                }
+            }
+
+            // Anything else to differentiate between objects.
+            var stringComp = string.CompareOrdinal(x?.ToString(), y?.ToString());
+
+            if (stringComp != 0) return stringComp;
+
+            return x?.GetHashCode().CompareTo(y?.GetHashCode() ?? 0) ?? 0;
```
timcassell: I don't think we want to use hash codes for this (they are not guaranteed to be stable).

User0332: Should we just stick with the string comparison then?

timcassell: Yes

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
timcassell: Right I understand those don't fail in the current code. But the else clause could fail for the same reason.

```cs
else // Probably a user-defined IStructuralComparable, as tuples would be handled by the IComparable case
```

e.g. `(object, int[,])` tuple (after you move the structural check first per other comment).

timcassell: As for structural fallbacks, you can check if the type is array (which you already do), and also check for `ITuple` (will require reflection in netstandard2.0).

src/BenchmarkDotNet/Parameters/ParameterComparer.cs
```diff
@@ -18,30 +21,154 @@ public int Compare(ParameterInstances x, ParameterInstances y)
...
-        private int CompareValues(object x, object y)
+        private int CompareValues<T1, T2>(T1 x, T2 y)
         {
             // Detect IComparable implementations.
             // This works for all primitive types in addition to user types that implement IComparable.
-            if (x != null && y != null && x.GetType() == y.GetType() &&
-                x is IComparable xComparable)
+            if (x != null && y != null && x.GetType() == y.GetType())
+            {
+                if (x is IComparable xComparable)
```
timcassell: Since you made the method generic, you can also check for generic comparable before non-generic to potentially reduce boxing.

```cs
if (x is IComparable<T2> xComparableGeneric)
```

timcassell: Actually, nevermind. This would still box. We'd need generic bridges for it to be worth it.

User0332:
Thanks for restructuring the fallbacks

User0332:
Is it guaranteed that the Orderer will always receive the same `ImmutableArray` of benchmark cases? Because then we could easily save the result of `allBenchmarksCases.Select(benchmarkCase => benchmarkCase.Parameters).Distinct(ParameterEqualityComparer.Instance).ToArray()` to avoid computing it multiple times.

timcassell:
> Is it guaranteed that the Orderer will always receive the same `ImmutableArray` of benchmark cases? Because then we could easily save the result of `allBenchmarksCases.Select(benchmarkCase => benchmarkCase.Parameters).Distinct(ParameterEqualityComparer.Instance).ToArray()` to avoid computing it multiple times.

No, the orderer is a singleton, unfortunately. It will receive the same cases per summary, but users can run multiple times to get multiple summaries.

User0332:
Or we could overload `GetLogicalGroupKey` to accept a summary instead of `ImmutableArray<BenchmarkCase> allBenchmarksCases` and then use an internal mapping to map summaries to the distinct `paramSets` based on reference equality

It was a quick thing so I tried it out already and I can push it if you'd like

timcassell:
> Or we could overload `GetLogicalGroupKey` to accept a summary instead of `ImmutableArray<BenchmarkCase> allBenchmarksCases` and then use an internal mapping to map summaries to the distinct `paramSets` based on reference equality
> 
> It was a quick thing so I tried it out already and I can push it if you'd like

How? It's called from more than just Summary.

User0332:
I left the old method signsture in place which is non-caching, and there is another one for summaries that is caching

timcassell:
Let's just leave it for now. We can refactor the interface separately. Thanks.

timcassell:
I optimized it in https://github.com/dotnet/BenchmarkDotNet/commit/e0adbb9ffe378a72c1a7fd9dd6036460ed1b03b6 to just scan the array once instead of building a set and then scanning the set.

PR #2888, opened 2025-12-02
chore: Enable nullable setting per types (Part.1)

filzrev:
This PR intended to fix a part of #2815

By adding `#nullable enable` line per files and fix build errors.
And create separate commit for each files.

Almost of the code has been modified while retaining existing behaviors.
But following files code are manually modified.

#### 1. `SourceCodeProvider.cs`
Remove internal code that seems not be used.

#### 2. `StringAndTextExtensions.cs`

Remove null check because it's internal and `null` seems not passed.

#### 3. `SummaryDto.cs`
Modify some properties default value from `null` to `EmptyString`.
These properties are used for XML serialization/deserialization.
So it might make some differences on generated XMLs.

#### 4. `ThreadingStats.cs`
Add `#if NETSTANDARD2_0` directives and use `Monitor.LockContentionCount`/`ThreadPool.CompletedWorkItemCount` directly when targeted to .NET 6 or later.

### DIFF
--- a/src/BenchmarkDotNet/Disassemblers/SourceCodeProvider.cs
+++ b/src/BenchmarkDotNet/Disassemblers/SourceCodeProvider.cs
@@ -5,13 +5,15 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 
+#nullable enable
+
 namespace BenchmarkDotNet.Disassemblers
 {
     internal class SourceCodeProvider : IDisposable
     {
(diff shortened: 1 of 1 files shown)


timcassell:
Please fix merge conflicts.

filzrev:
> Please fix merge conflicts.

I've removed conflicted file commit (`WindowsDisassembler.cs`) and rebased to latest master.

timcassell (approved):
Thanks

PR #2897, opened 2025-12-03
Allow benchmark methods to use names that are used by code generation templates (fixes #2821)

### LINKED ISSUE #2821: Build error when using certain method names
Example benchmark method:

```cs
[Benchmark] public void System() { }
```

Fails with

```
error CS0119: 'Benchmarks.System()' is a method, which is not valid in the given context
```

We should use `base.MethodName` for instance methods, or `<FullyQualifiedTypeName>.MethodName` for static methods, rather than the simple `MethodName`. And qualify all types with `global::`

User0332:
Fixes #2821.

## Solution

Uses the suggested solution from #2821, using `base.MethodName` for instance methods, `FQTypeName.MethodName` for static methods, and prefixing all type/namespace names with the `global::` namespace prefix to avoid accidentally referencing methods with names like `System` or `BenchmarkDotNet`.

Adds a new test that ensures the method names `System` and `BenchmarkDotNet` can be used as benchmark method names.


timcassell:
It may not be strictly necessary, but we should probably also use `this.` to reference members of the generated class within itself. e.g. `this.__Overhead`, etc.

timcassell:
Need to fix merge conflicts

User0332:
I might not be able to get to this for a bit so if anyone who is seeing this wants to build off this PR & finish it up I'd be open to that

Issue #2779, opened 2025-06-15
`AllocationQuantumIsNotAnIssueForNetCore21Plus` tests randomly failed on `macos (arm64)`


filzrev:
`AllocationQuantumIsNotAnIssueForNetCore21Plus` test is recently enabled by PR #2562

But it's randomly failed on CI.
Because extra memory allocation is reported.
Currently it's confirmed only on `macos(arm64)` environment . (But it might be occurred on another arm64 images on windows/linux)

**Error Log**

```
2025-06-15T05:07:55.9366620Z [xUnit.net 00:09:48.63]     AllocationQuantumIsNotAnIssueForNetCore21Plus(toolchain: .NET 8.0) [FAIL]
2025-06-15T05:07:55.9368370Z [xUnit.net 00:09:48.63]       Assert.Equal() Failure: Values differ
2025-06-15T05:07:55.9368720Z [xUnit.net 00:09:48.63]       Expected: 88
2025-06-15T05:07:55.9368990Z [xUnit.net 00:09:48.63]       Actual:   760
...
2025-06-15T05:07:55.9500980Z [xUnit.net 00:09:48.63]         | Method              | Mean     | Error | Allocated |
2025-06-15T05:07:55.9501490Z [xUnit.net 00:09:48.63]         |-------------------- |---------:|------:|----------:|
2025-06-15T05:07:55.9502130Z [xUnit.net 00:09:48.63]         | SixtyFourBytesArray | 551.2 ms |    NA |     760 B |
```

If it can't determine root cause of this memory allocation issue.
This test should be temporary skipped for ARM64 environment.

timcassell:
Hm... My guess is because it's a long-running benchmark, it's not able to get up to tier1 JIT before the measurement happens. It's not an uncommon thing to happen in real-world benchmarks. Maybe could try adding `[MethodImpl(CodeGenHelper.AggressiveOptimization)]` to it? (I know most real benchmarks won't do that, but I'm not sure what else we can do.)

timcassell:
Actually I realized we specifically disable tiered JIT for these tests for exactly that reason. Then I'm really not sure what the cause is...

filzrev:
I've confirmed it can be reproduced on `macos-13 (x64)` runner image also  (It's occurred 3 times on 100 runs )
And it's not occurred when using windows(x64/arm64) and ubuntu(x64/arm64) images. (It's also testes 100 times)

So it seems to be a `macos` specific issues.

[...]

PR #2898, opened 2025-12-04
chore: Update BenchmarkDotNet.Samples project

filzrev:

#### 1. Add code to return exit code
Currently `void Main` is used, and it return ExitCode: `0` regardless validation/build/benchmark failure exists.
This PR add code to return error code when failed. (Except for non-critical validation error)

#### 2. Add `Properties/launchSettings.json`
Add profiles to switch settings (VisualStudio or Use `--launch-profile` argument).
By default, `Default` profile is loaded. (It can be suppressed when pass `--no-launch-profile`)

~~`"workingDirectory": ".",` setting is added to output logs based on current directory.~~   
~~(It's `project directory` when running on VS)~~
~~Previous behavior output logs to `bin/$(Configuration)/$(TargetFramework)`.~~

#### 3. Add code for debugging
Add dedicated config that is used when build with `Debug` configuration.
It's expected to be used for Debugging with "Run IntroBasic Benchmarks" profile.

Note:  
When target benchmark define local config with `[Config(typeof(Config))]`.   
Job is merged with `ConfigUnionRule.Union` rule. And it's not works as expected.  
(Because both Job is executed that included GlobalConfig/LocalConfig)

### DIFF
--- a/samples/BenchmarkDotNet.Samples/Program.cs
+++ b/samples/BenchmarkDotNet.Samples/Program.cs
@@ -1,13 +1,10 @@
-﻿using BenchmarkDotNet.Attributes;
-using BenchmarkDotNet.Configs;
-using BenchmarkDotNet.Diagnosers;
+﻿using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Engines;
 using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.Extensions;
 using BenchmarkDotNet.Loggers;
-using BenchmarkDotNet.Reports;
 using BenchmarkDotNet.Running;
 using System;
-using System.Collections.Generic;
 using System.Linq;
 
 namespace BenchmarkDotNet.Samples;
@@ -49,37 +46,13 @@ public class Program
             args = args.Where(x => x != "--inProcess").ToArray();
 
         DebugConfig config = isInProcess
-                            ? new DebugInProcessConfig()
-                            : new DebugBuildConfig();
+            ? new DebugInProcessConfig()
+            : new DebugBuildConfig();
 
         return config.AddAnalyser(DefaultConfig.Instance.GetAnalysers().ToArray())
-                     .AddDiagnoser(
-                         MemoryDiagnoser.Default,
-#if NETCOREAPP3_0_OR_GREATER
-                         new ThreadingDiagnoser(new ThreadingDiagnoserConfig(displayCompletedWorkItemCountWhenZero: false, displayLockContentionWhenZero: false)),
+                     .AddExporter(MarkdownExporter.Default)
+                     .AddValidator(DefaultConfig.Instance.GetValidators().ToArray())
+                     .WithArtifactsPath(DefaultConfig.Instance.ArtifactsPath);
 #endif
-                         new ExceptionDiagnoser(new ExceptionDiagnoserConfig(displayExceptionsIfZeroValue: false))
-                      )
-                      .AddExporter(MarkdownExporter.Default)
-                      .AddValidator(DefaultConfig.Instance.GetValidators().ToArray())
-                      .WithArtifactsPath(DefaultConfig.Instance.ArtifactsPath);
-#endif
-    }
-}
-
-file static class ExtensionMethods
-{
-    public static bool HasError(this Summary[] summaries)
-    {
-        if (summaries.Length == 0)
-        {
-            var hashSet = new HashSet<string>(["--help", "--list", "--info", "--version"]);
-            return !Environment.GetCommandLineArgs().Any(hashSet.Contains);
-        }
-
-        if (summaries.Any(x => x.HasCriticalValidationErrors))
-            return true;
-
-        return summaries.Any(x => x.Reports.Any(r => !r.Success));
     }
 }
--- a/samples/BenchmarkDotNet.Samples/Properties/launchSettings.json
+++ b/samples/BenchmarkDotNet.Samples/Properties/launchSettings.json
@@ -3,15 +3,18 @@
     "Default": {
       "commandName": "Project",
       "commandLineArgs": "",
-      "workingDirectory": ".",
       "environmentVariables": {
       }
     },
     "Run IntroBasic Benchmarks": {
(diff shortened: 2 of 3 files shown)


filzrev:
CI tests(`test-macos(arm64)`) failed on `AllocationQuantumIsNotAnIssueForNetCore21Plus`
It's expected to caused by #2779 and  fixed  by rerun.

samples/BenchmarkDotNet.Samples/Program.cs
```diff
@@ -1,9 +1,75 @@
...
+
+        if (summaries.HasError())
+            return 1;
+
+        return 0;
+    }
+
+    private static ManualConfig GetDebugConfig()
+    {
+        return DefaultConfig.Instance
```
timcassell: I sometimes want to debug the separate process, this should take an option and choose `DebugBuildConfig` or `DebugInProcessConfig`.

timcassell: Or you could just check if the debugger is attached and choose accordingly.

filzrev:
I've modified code to use `DebugBuildConfig`/`DebugInProcessConfig`.

By default, `DebugBuildConfig` is used on `Debug` build. (Target benchmarks are also build with `Debug` configuration).
When `--inProcess` argument is passed. It use `DebugInProcessConfig` instead.

**Other changes**
- Add some settings that are used on `DefaultConfig`.
- Add `MemoryDiagnoser`/`ExceptionDiagnoser`/`ThreadingDiagnoser`(Requires .NET 3.0+)
- `RunStrategy.Monitoring` setting is removed from config. and need to specified as command line argument.

samples/BenchmarkDotNet.Samples/Program.cs
filzrev: I've changed code to use config that based on `DebugInProcessConfig` when `--inProcess` option is specified.
In this case, `--inProcess` option need to be removed from argument list.
Because when passed this argument, it add another in-process `Job`.

samples/BenchmarkDotNet.Samples/Program.cs
```diff
@@ -1,9 +1,85 @@
...
+                            ? new DebugInProcessConfig()
+                            : new DebugBuildConfig();
+
+        return config.AddAnalyser(DefaultConfig.Instance.GetAnalysers().ToArray())
+                     .AddDiagnoser(
+                         MemoryDiagnoser.Default,
+#if NETCOREAPP3_0_OR_GREATER
+                         new ThreadingDiagnoser(new ThreadingDiagnoserConfig(displayCompletedWorkItemCountWhenZero: false, displayLockContentionWhenZero: false)),
+#endif
+                         new ExceptionDiagnoser(new ExceptionDiagnoserConfig(displayExceptionsIfZeroValue: false))
```
timcassell: We have samples for these diagnosers, I don't think they should be added always.

samples/BenchmarkDotNet.Samples/Properties/launchSettings.json
```diff
@@ -0,0 +1,35 @@
...
+      "commandName": "Project",
+      "commandLineArgs": "",
+      "workingDirectory": ".",
+      "environmentVariables": {
+      }
+    },
+    "Run IntroBasic Benchmarks": {
+      "commandName": "Project",
+      "commandLineArgs": "--filter *IntroBasic*",
+      // "commandLineArgs": "--filter *IntroBasic* --inProcess --strategy Monitoring", // Use this setting for in-process debug
```
timcassell: It seems that VS doesn't mind it, but I don't think json usually supports comments (it's red in github). You could just add another profile for it I think.

samples/BenchmarkDotNet.Samples/Program.cs
```diff
@@ -1,9 +1,85 @@
...
+#if !DEBUG
+        return null; // `DefaultConfig.Instance` is used.
+#else
+        bool isInProcess = args.Contains("--inProcess");
+        if (isInProcess)
+            args = args.Where(x => x != "--inProcess").ToArray();
+
+        DebugConfig config = isInProcess
+                            ? new DebugInProcessConfig()
+                            : new DebugBuildConfig();
```
timcassell: Please match indentation
```suggestion
        DebugConfig config = isInProcess
            ? new DebugInProcessConfig()
            : new DebugBuildConfig();
```

samples/BenchmarkDotNet.Samples/Program.cs
filzrev: > I don't think they should be added always.

I'll remove these settings from Samples project later.

---

I'm using `Samples` projects for debugging BenchmarkDotNet and running custom benchmarks with latest code base.
So I'm want to use some settings by default.
 
Is it acceptable to add another samples project? 
Example
- `samples/BenchmarkDotNet.Samples.Sandbox`
- `sandbox/BenchmarkDotNet.Sandbox`

It's expected to be used with following purpose.
- Debugging `BenchmarkDotNet.Analyzers`  (By specifying `BenchmarkDotNet.Analyzers` debug target project)
- Define pre-configured benchmark config and switch config by `launchSettings.json` without rebuild.
- Compare BenchmarkDotNet local build version with NuGet published versions

samples/BenchmarkDotNet.Samples/Properties/launchSettings.json
filzrev: I'll separate these profiles.

samples/BenchmarkDotNet.Samples/Program.cs
timcassell: It's fine to add a new project.

filzrev:
I've fixed code that pointed by review comments.

Additionally, I've added following changes.

1.  Move `HasError` extension method to BenchmakrDotNet` project. (It'll be expected be used on another proj) 
2. Remove `workingDirectory` setting form `launchSettings.json`.  
    Because it affects search inside VisualStudio.  
     (I'll add PR that prints artifactsDirectory/LogFile path instead)

Issue #2896, opened 2025-12-03
DotNetSdkValidator failed when using CsProjClassicNetToolchain


filzrev:
## Steps to reproduce problem

1. Create simple benchmark dotnet project
2. Modify project to use `<TargetFrameworks>net48;net8.0</TargetFrameworks>`
3. Modify Program.cs with following content
    ```
    internal class Program
    {
        static void Main(string[] args)
        {
            var config = DefaultConfig.Instance
                .AddJob(Job.Dry.WithToolchain(CsProjClassicNetToolchain.Net48));
                // .AddJob(Job.Dry.WithToolchain(CsProjCoreToolchain.NetCoreApp80));

            BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args, config);
        }
    }

    public class Benchmarks
    {
        [Benchmark]
        public void Benchmark1() { };
    }
    ```
4. Run program by using `net8.0` framework.
5. Following error is shown.
    ```
    // Validating benchmarks:
    // Benchmark Benchmark.IRunEveryWhere: Dry(Toolchain=.NET Framework 4.8, IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=1, WarmupCount=1)
    //    * The required .NET Framework SDK version 8.0 or higher is not installed.
    ```

---
This problem is not occurred when running benchmark host project with `net48`.
And it seems not occurred when using BenchmarkDotNet v0.14.0

timcassell:
I cannot repro. I suspect it's a machine-specific error. If you can find the cause and send a fix, that'd be great!

filzrev:
I've confirmed it can be reproduced on GitHub actions environment also. 

**Logs**
https://github.com/filzrev/BenchmarkDotNet/actions/runs/19913622689/job/57087688449

Note: Returned summary is not validated and process exit with `0`.

As far as I've confirmed related source code.
`DotNetSdkValidator` is executed when using `CsProjClassicNetToolchain` and It's expected always failed when running on .NET 8.

timcassell:
I don't see failures in that link.

`CsProjClassicNetToolchain` runs a different validation method than `CsProjCoreToolchain` that checks only for .Net Framework sdks.

timcassell:
Sorry I misspoke. `CsProjClassicNetToolchain` does check if the cli is installed because it's required, but it doesn't care what version it is.

filzrev:
> I don't see failures in that link.

[...]

PR #2899, opened 2025-12-05
fix: DotNetSdkValidator failed when using CsProjClassicNetToolchain

filzrev:
This PR intended to fix issue #2896

When using `CsProjClassicNetToolchain` and running benchmarks on `.NET 8` host.

Following code isn't works as expected and fallback to `GetCurrentRuntime()` (It returns .NET 8)
https://github.com/dotnet/BenchmarkDotNet/blob/a62d90ad9e444fde6bbfd99df21db9ea21b2d052/src/BenchmarkDotNet/Portability/RuntimeInformation.cs#L173-L176

Modify `ValidateFrameworkSdks` code with following logics.
1. When RuntimeCharacteristic is specified `WithRuntime(runtime)`. Use specified runtime. (It's existing behavior)
2. Otherwise, resolve runtime with `ClrRuntime.GetTargetOrCurrentVersion`

### DIFF
--- a/src/BenchmarkDotNet/Validators/DotNetSdkValidator.cs
+++ b/src/BenchmarkDotNet/Validators/DotNetSdkValidator.cs
@@ -32,7 +32,11 @@ namespace BenchmarkDotNet.Validators
 
         public static IEnumerable<ValidationError> ValidateFrameworkSdks(BenchmarkCase benchmark)
         {
-            var requiredSdkVersion = benchmark.GetRuntime().RuntimeMoniker.GetRuntimeVersion();
+            var targetRuntime = benchmark.Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic)
+                ? benchmark.Job.Environment.Runtime
+                : ClrRuntime.GetTargetOrCurrentVersion(benchmark.Descriptor.WorkloadMethod.DeclaringType.Assembly);
+            var requiredSdkVersion = targetRuntime.RuntimeMoniker.GetRuntimeVersion();
+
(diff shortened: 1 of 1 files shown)


src/BenchmarkDotNet/Validators/DotNetSdkValidator.cs
```diff
@@ -32,7 +32,11 @@ public static IEnumerable<ValidationError> ValidateCoreSdks(string? customDotNet
 
         public static IEnumerable<ValidationError> ValidateFrameworkSdks(BenchmarkCase benchmark)
         {
-            var requiredSdkVersion = benchmark.GetRuntime().RuntimeMoniker.GetRuntimeVersion();
+            var targetRuntime = benchmark.Job.Environment.HasValue(EnvironmentMode.RuntimeCharacteristic)
+                ? benchmark.Job.Environment.Runtime
+                : ClrRuntime.GetTargetOrCurrentVersion(benchmark.Descriptor.WorkloadMethod.DeclaringType.Assembly);
```
timcassell, line 37: What's the difference? It looks like `benchmark.GetRuntime()` does the exact same thing.

timcassell: Oh I see, the other code is calling `RuntimeInformation.GetTargetOrCurrentRuntime` and this changes it to `ClrRuntime.GetTargetOrCurrentVersion`.

timcassell:
> This PR intended to fix issue #2896

FYI if you just put `Fixes #XXXX`, GitHub will automatically link the issue.

PR #2900, opened 2025-12-06
chore: Enable nullable for job/characteristics relating code

filzrev:
This PR intended to fix a part issue of #2815

**What's changed in this PR**
Add `#nullable enable` setting and fix nullable annotations.
And fix nullable errors.

**Other changes**
Remove `nullable` annotation from `CharacteristicObject` constructor's `id` parameter.
when `null` or `""` is passed to CharacteristicObject constructor.
It skip setting `IdCharacteristic`. and `Default` is used when getting Id as fallback value.

### DIFF
--- a/src/BenchmarkDotNet/Characteristics/CharacteristicObject.cs
+++ b/src/BenchmarkDotNet/Characteristics/CharacteristicObject.cs
@@ -5,6 +5,8 @@ using System.Linq;
 using System.Reflection;
 using JetBrains.Annotations;
 
+#nullable enable
+
 namespace BenchmarkDotNet.Characteristics
 {
     // TODO: better naming.
@@ -17,10 +19,10 @@ namespace BenchmarkDotNet.Characteristics
             DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties
             | DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields;
 
-        protected static string ResolveId(CharacteristicObject obj, string actual)
+        protected static string ResolveId(CharacteristicObject obj, string? actual)
         {
             if (!string.IsNullOrEmpty(actual) && actual != IdCharacteristic.FallbackValue)
-                return actual;
+                return actual!;
 
             string result = CharacteristicSetPresenter.Display.ToPresentation(obj);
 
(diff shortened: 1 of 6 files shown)


src/BenchmarkDotNet/Characteristics/Characteristic.cs
```diff
@@ -24,7 +26,7 @@ public abstract class Characteristic
                 null, fallbackValue,
                 false);
 
-        public static Characteristic<T> Create<TOwner, [DynamicallyAccessedMembers(CharacteristicObject.CharacteristicMemberTypes)] T>(string memberName, Func<CharacteristicObject, T, T> resolver, T fallbackValue, bool ignoreOnApply)
+        public static Characteristic<T> Create<TOwner, [DynamicallyAccessedMembers(CharacteristicObject.CharacteristicMemberTypes)] T>(string memberName, Func<CharacteristicObject, T?, T> resolver, T fallbackValue, bool ignoreOnApply)
```
filzrev, line 29: It need to allow `null` because some characteristics expected to return `null`(e.g. RuntimeCharacteristic/EnvironmentVariablesCharacteristic)

src/BenchmarkDotNet/Characteristics/CharacteristicObject.cs
```diff
@@ -25,7 +25,7 @@ protected static string ResolveId(CharacteristicObject obj, string actual)
             string result = CharacteristicSetPresenter.Display.ToPresentation(obj);
 
             if (result.Length == 0)
-                result = IdCharacteristic.FallbackValue;
+                result = IdCharacteristic.FallbackValue!;
```
filzrev, line 30: `IdCharacteristic` specify `Default` as fallback value. so it must not be null.

src/BenchmarkDotNet/Jobs/Job.cs
```diff
@@ -41,29 +41,29 @@ public Job(string? id) : base(id)
...
         {
             Apply(others);
         }
 
-        public EnvironmentMode Environment => EnvironmentCharacteristic[this];
-        public RunMode Run => RunCharacteristic[this];
-        public InfrastructureMode Infrastructure => InfrastructureCharacteristic[this];
-        public AccuracyMode Accuracy => AccuracyCharacteristic[this];
-        public MetaMode Meta => MetaCharacteristic[this];
+        public EnvironmentMode Environment => EnvironmentCharacteristic[this]!;
```
filzrev, line 62: Following getter field are set by constructor and should not be null.
- Environment
- Run
- Accuracy
- Meta

PR #2902, opened 2025-12-07
chore: Enable nullable annotation for validators

filzrev:
This PR intended to fix a part issue #2815

### DIFF
--- a/src/BenchmarkDotNet/Validators/ValidationParameters.cs
+++ b/src/BenchmarkDotNet/Validators/ValidationParameters.cs
@@ -19,7 +19,7 @@ namespace BenchmarkDotNet.Validators
         }
 
         // Note: Following implicit operators are expected to be used for test projects.
-        public static implicit operator ValidationParameters(BenchmarkCase[] benchmarksCase) => new ValidationParameters(benchmarksCase, config: null!);
-        public static implicit operator ValidationParameters(BenchmarkRunInfo benchmarkRunInfo) => new ValidationParameters(benchmarkRunInfo.BenchmarksCases, config: null!);
+        public static implicit operator ValidationParameters(BenchmarkCase[] benchmarksCase) => new ValidationParameters(benchmarksCase, DefaultConfig.Instance.CreateImmutableConfig());
+        public static implicit operator ValidationParameters(BenchmarkRunInfo benchmarkRunInfo) => new ValidationParameters(benchmarkRunInfo.BenchmarksCases, benchmarkRunInfo.Config);
     }
 }


src/BenchmarkDotNet/Validators/ValidationParameters.cs
```diff
@@ -10,16 +10,16 @@ public class ValidationParameters
...
+        public ValidationParameters(IReadOnlyList<BenchmarkCase> benchmarks, ImmutableConfig config)
         {
             Benchmarks = benchmarks;
             Config = config;
         }
 
-        // to have backward compatibility for people who implemented IValidator(Benchmark[] benchmarks)
-        public static implicit operator ValidationParameters(BenchmarkCase[] benchmarksCase) => new ValidationParameters(benchmarksCase, null);
-        public static implicit operator ValidationParameters(BenchmarkRunInfo benchmarkRunInfo) => new ValidationParameters(benchmarkRunInfo.BenchmarksCases, null);
+        // Note: Following implicit operators are expected to be used for test projects.
```
filzrev, line 21: Remove backward compatibility comments.
Because when `null` is passed to config parameter.
It'll throw NullReferenceException when accessing config property.

These implicit operators seems to be used for test projects.

filzrev:
Please re-run `test-macos (arm64)` workflow.
It's failed by flaky tests.

src/BenchmarkDotNet/Validators/ParamsAllValuesValidator.cs
```diff
@@ -24,7 +27,7 @@ public IEnumerable<ValidationError> Validate(ValidationParameters input) =>
                 .SelectMany(type => type.GetTypeMembersWithGivenAttribute<ParamsAllValuesAttribute>(ReflectionFlags))
                 .Distinct()
                 .Select(member => GetErrorOrDefault(member.ParameterType))
-                .Where(error => error != null);
+                .OfType<ValidationError>();
```
timcassell: We have a `WhereNotNull` extension.

src/BenchmarkDotNet/Validators/ValidationParameters.cs
```diff
@@ -10,16 +10,16 @@ public class ValidationParameters
     {
         public IReadOnlyList<BenchmarkCase> Benchmarks { get; }
 
-        public ImmutableConfig? Config { get; }
+        public ImmutableConfig Config { get; }
```
timcassell, line 13: I don't quite understand why this is changed from nullable when null is explicitly passed in the implicit operators. We should just remove those operators if they are no longer useful.

filzrev: > We should just remove those operators if they are no longer useful.

I think that's the correct way to fix it.

Though when removing these implicit operators. It cause about 39 build errors on test projects. and without implicit operators.
It need to add a lot of additional code. (e.g. Create `ToValidationParameters` extension method and call this method  on every location that using implicit operators)

timcassell: What about changing `null` to `DefaultConfig.Instance.CreateImmutableConfig()` for now?

filzrev: I've modified code to use `DefaultConfig.Instance.CreateImmutableConfig()` on implicit operators that accept `BenchmarkCase[] benchmarksCase` as parameter.
And use actual config when accept `(BenchmarkRunInfo benchmarkRunInfo)` as parameter.

In anyway, this config is not be referenced by existing unit tests/integration tests.

PR #2908, opened 2025-12-10
chore: Replace `string.IsNullOrEmpty` method to IsBlank/IsNotBlank extension methods

filzrev:
This PR replace code that using `string.IsNullOrEmpty` method to use `IsBlank`/`IsNotBlank` extension methods.

#### Background
It's required to simplify code when enabled `nullable` features.

When targeting `netstandard2.0` tfm.
[Nullable annotations are not set on `string.IsNullOrEmpty`](https://github.com/dotnet/standard/issues/1118)

So following code needs extra `!` annotation when targeting `netstandard2.0`.

```csharp
string? value = null;
if (!string.IsNullOrEmpty(value))
{
    var length = value.Length!; // `!` is required
}
```
It can't change existing API's behaviors.
So replace to existing extension methods (`IsBlank`/`IsNotBlank`)

Note: These extension methods using `IsNullOrWhiteSpace` instead of `IsNullOrEmpty`.

1. `StringAndTextExtensions.cs` / `CommonExtensions.cs`
Add `NotNullWhen` attributes to existing IsBlank/IsNotBlank/IsNullOrEmpty extension methods.

2. Other files
Replace code that using `string.IsNullOrEmpty` to use `IsBlank`/`IsNotBlank`

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -232,7 +232,7 @@ namespace BenchmarkDotNet.ConsoleArguments
 
         private static bool Validate(CommandLineOptions options, ILogger logger)
         {
-            if (options.BaseJob.IsNotBlank() && !AvailableJobs.ContainsKey(options.BaseJob))
+            if (options.BaseJob.IsBlank() || !AvailableJobs.ContainsKey(options.BaseJob))
             {
                 logger.WriteLineError($"The provided base job \"{options.BaseJob}\" is invalid. Available options are: {string.Join(", ", AvailableJobs.Keys)}.");
                 return false;
@@ -548,7 +548,7 @@ namespace BenchmarkDotNet.ConsoleArguments
 
                 case RuntimeMoniker.Mono:
                     {
-                        var runtime = new MonoRuntime("Mono", options.MonoPath?.FullName!);
+                        var runtime = new MonoRuntime("Mono", options.MonoPath?.FullName);
                         return baseJob.WithRuntime(runtime).WithId(runtime.Name);
                     }
 


src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -232,7 +232,7 @@ private static Parser CreateParser(ILogger logger)
 
         private static bool Validate(CommandLineOptions options, ILogger logger)
         {
-            if (!AvailableJobs.ContainsKey(options.BaseJob))
+            if (options.BaseJob.IsNotBlank() && !AvailableJobs.ContainsKey(options.BaseJob))
```
timcassell:
```suggestion
            if (options.BaseJob.IsBlank() || !AvailableJobs.ContainsKey(options.BaseJob))
```

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -548,7 +548,7 @@ private static Job CreateJobForGivenRuntime(Job baseJob, string runtimeId, Comma
 
                 case RuntimeMoniker.Mono:
                     {
-                        var runtime = new MonoRuntime("Mono", options.MonoPath?.FullName);
+                        var runtime = new MonoRuntime("Mono", options.MonoPath?.FullName!);
```
timcassell: Code smell, null-propagation followed by null-forgiving.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
filzrev: Thank you for pointing that out.
This is definitely an wrong code on empty text is passed.

PR #2911, opened 2025-12-10
Add net11 support

LoopedBard3:
Add net11.0 support. Based on the update for net10.0: https://github.com/dotnet/BenchmarkDotNet/pull/2642.

(Answered) ~Open question: Is the update to src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs and tests/BenchmarkDotNet.Tests/Helpers/DotNetRuntimeHelper.cs correct~?

Once merged we plan on updating the dotnet/performance repo to latest BDN, in part for the WasmNet11_0 runtime moniker support.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
+++ b/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
@@ -269,7 +269,7 @@ $@"<?xml version=""1.0"" encoding=""utf-8""?>
                     if (HardwareIntrinsics.IsX86BaseSupported) yield return "base";
                     if (HardwareIntrinsics.IsX86Sse42Supported)
                     {
-                        if (runtimeMoniker <= RuntimeMoniker.NativeAot11_0) yield return "sse4.2";
+                        if (runtimeMoniker <= RuntimeMoniker.NativeAot10_0) yield return "sse4.2";
                         if (runtimeMoniker <= RuntimeMoniker.NativeAot90) yield return "popcnt";
                     }
                     if (HardwareIntrinsics.IsX86AvxSupported) yield return "avx";


src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -262,7 +262,7 @@ private IEnumerable<string> GetCurrentProcessInstructionSets(Platform platform)
                     if (HardwareIntrinsics.IsX86BaseSupported) yield return "base";
                     if (HardwareIntrinsics.IsX86Sse42Supported)
                     {
-                        if (runtimeMoniker <= RuntimeMoniker.NativeAot10_0) yield return "sse4.2";
+                        if (runtimeMoniker <= RuntimeMoniker.NativeAot11_0) yield return "sse4.2";
```
timcassell: Not sure if this needs to change. cc @tannergooding

tests/BenchmarkDotNet.Tests/Helpers/DotNetRuntimeHelper.cs
```diff
@@ -8,7 +8,7 @@ public static class DotNetRuntimeHelper
         public static string GetExpectedDotNetCoreRuntimeName()
         {
             return Environment.GetEnvironmentVariable("DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX") == "2"
-                ? ".NET 10.0"
+                ? ".NET 11.0"
```
timcassell: This should not be changed.

LoopedBard3:
@timcassell Do you have any idea why the CurrentRuntimeIsProperlyRecognized is still failing despite reverting it back to be teh same as master? Running locally I didn't hit the issue, but I think it was using net8.0 instead of net10.0.

timcassell:
> @timcassell Do you have any idea why the CurrentRuntimeIsProperlyRecognized is still failing despite reverting it back to be teh same as master? Running locally I didn't hit the issue, but I think it was using net8.0 instead of net10.0.

Because of your change in DotNetRuntimeHelper.cs.

tests/BenchmarkDotNet.Tests/Helpers/DotNetRuntimeHelper.cs
LoopedBard3: Thanks for this catch! Went ahead and reverted this files change.

LoopedBard3:
Thank you for catching this, went ahead and undid that change.

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
timcassell: Maybe that check shouldn't even be there and it just always yields sse4.2 like the code previous to #2828.

tannergooding: This is a case where `sse42` is now `base` in .NET 11+, but still needs to be specified for older versions.

It was intentional this was set to `<= NativeAot10_0` accordingly, because it no longer exists in .NET 11. Much as `popcnt` is `<= NativeAot90` because it doesn't exist in .NET 10 and is implied by `sse4.2`

LoopedBard3: Cool, I went ahead and reverted this in the latest commit so there are no changes to the NativeAot/Generator.cs file.

timcassell:
Thanks @LoopedBard3

PR #2913, opened 2025-12-11
chore: Update roslyn package versions and analyzer's target

filzrev:
This PR update roslyn version `5.0.0`. And add `BenchmarkDotNet.Analyzers` project target.

**What's changed in this PR**

1. Update "Microsoft.CodeAnalysis.CSharp" to `5.0.0`
2. Update `System.Numerics.Vectors` to latest version and add `System.Buffers` to suppress warnings. 
~~3. Add `BenchmarkDotNet.Analyzers` project target for roslyn `5.0`  ~~
~~    (Currently no C# 14 specific rules exists. So it might not need to add?)~~


timcassell:
> 3. Add `BenchmarkDotNet.Analyzers` project target for roslyn `5.0`
>    (Currently no C# 14 specific rules exists. So it might not need to add?)

Right, it's not necessary. We can add newer version when we need to analyze newer language features.

filzrev:
> Right, it's not necessary. We can add newer version when we need to analyze newer language features.

I've removed code for analyzers. and force pushed.

PR #2933, opened 2025-12-17
feat: Change default job order to numeric order

### LINKED ISSUE #2919: Job summary does not sort naturally by runtime version
I ran a benchmark targeting .NET 8-10 + .NET Framework earlier with v0.15.8, and noticed the summary looked similar to the below:

```md
| Method           | Job                  | Runtime              |
|----------------- |--------------------- |--------------------- |
| ABenchmarkMethod | .NET 10.0            | .NET 10.0            |
| ABenchmarkMethod | .NET 8.0             | .NET 8.0             |
| ABenchmarkMethod | .NET 9.0             | .NET 9.0             |
| ABenchmarkMethod | .NET Framework 4.6.2 | .NET Framework 4.6.2 |
```

I would expect .NET 10 to be the third row, not the first.

There appears to be a numeric comparer built-in for jobs, but it doesn't appear to be used in the case above:

https://github.com/dotnet/BenchmarkDotNet/blob/46417907f0442c2aff4da5168cc0b96ea7eb1b9b/src/BenchmarkDotNet/Jobs/JobComparer.cs#L15

The command I used to run the benchmarks was:

`dotnet run --configuration Release --framework net9.0 --filter "*ABenchmarkMethod*" --runtimes net462 net8.0 net9.0 net10.0`

I had a quick look at the code to see if I could submit a PR to fix this, but it wasn't immediately clear to me where the right place for the fix would be.

filzrev:
This PR intended to fix #2919 

**What's changed in this PR**
- Change `JobOrderPolicy` default to numeric order.
- Add `JobOrderPolicy.Ordinal` enum value to.

Note: `JobOrderPolicy.Numeric` enum value is retained for compatibility purpose.

**Other Changes**
- Rename `JobComparer.Instance` to `JobComparer.Default`
- Add code to use `CompareOptions.NumericOrdering` on .NET 10 (It'll be used when `net10.0` tfm is added in future)
- Modify test code of `JobOrderTests.cs` (It's confirmed that these tests on `net10.0` tfm also)

### DIFF
--- a/src/BenchmarkDotNet/BenchmarkDotNet.csproj
+++ b/src/BenchmarkDotNet/BenchmarkDotNet.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <AssemblyTitle>BenchmarkDotNet</AssemblyTitle>
-    <TargetFrameworks>netstandard2.0;net6.0;net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0;net8.0;net9.0;net10.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <NoWarn>$(NoWarn);1701;1702;1705;1591;3005;NU1702;CS3001;CS3003</NoWarn>
     <AssemblyName>BenchmarkDotNet</AssemblyName>
--- a/tests/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj
+++ b/tests/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <AssemblyTitle>BenchmarkDotNet.Tests</AssemblyTitle>
-    <TargetFrameworks>net8.0;net462</TargetFrameworks>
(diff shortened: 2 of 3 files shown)


src/BenchmarkDotNet/Order/JobOrderPolicy.cs
```diff
@@ -3,12 +3,17 @@
 public enum JobOrderPolicy
 {
     /// <summary>
-    /// Compare job characteristics in ordinal order.
+    /// Compare job characteristics in default order.
     /// </summary>
     Default,
```
timcassell: We can remove `Default` value (breaking change is ok for next version).

src/BenchmarkDotNet/Jobs/JobComparer.cs
```diff
@@ -10,15 +10,28 @@ namespace BenchmarkDotNet.Jobs
     internal class JobComparer : IComparer<Job>, IEqualityComparer<Job>
     {
         private readonly IComparer<string> Comparer;
+        private static IComparer<string> NumericComparer =>
+#if NET10_0_OR_GREATER
```
timcassell, line 14: You can add `net10.0` tfm to the csproj to make this light up.

filzrev: I've added `net10.0` tfm. and add enable unit tests under `net10.0`.

PR #2943, opened 2025-12-26
Docs: add Linux thread priority guidance

### LINKED ISSUE #2024: Add information to the docs about thread priority (Linux)
Some may wonder how to solve for the thread priority exception when optimizations validator is enabled. If you run with `sudo` or as a privileged user, it obviously solves this, but then all the compiled files are saved as root. It took me a bit, but the best solution I came up with to get around this is to put this into the Benchmark project's .csproj:

- within `.csproj` add a target:
	- For non-interactive, but the user's password must be provided on the command line, eg `export SUDOPASS=user_password_here && dotnet run` or to keep your password from being in the global environment: `(export SUDOPASS=user_password_here && dotnet run)`
	```xml
	<Target Name="PostBuild" AfterTargets="PostBuildEvent">
		<Exec Command="${SUDOPASS} | sudo --stdin setcap 'cap_sys_nice=eip' $(TargetDir)$(AssemblyName)" />
	</Target>
	```
	- for a version which prompts the user for their password and can be run with a simple `dotnet run`:
	```xml
	<Target Name="PostBuild" AfterTargets="PostBuildEvent">
		<Exec Command="sudo setcap 'cap_sys_nice=eip' $(TargetDir)$(AssemblyName)"/>
	</Target>
	```
        
This would have helped me, maybe it would help others. I'm happy to submit a PR, but I was unsure where this should be placed or if you wanted this information added at all.

meiranzheng:
Fixes #2024.

Adds a short FAQ entry explaining the "Failed to set up priority Highest for thread" message on Linux (CAP_SYS_NICE / setcap), and suggests using setcap instead of running the whole benchmark as root.

### DIFF
--- a/docs/articles/faq.md
+++ b/docs/articles/faq.md
@@ -83,6 +83,16 @@ For example, you can use the `SimpleJob` or `ShortRunJob` attributes:
     }
     ```
 
+* **Q** I'm running benchmarks on Linux and I see: `// ! Failed to set up priority Highest for thread ... Make sure you have the right permissions.` What can I do?
+
+    **A** This message can appear when using in-process toolchains (for example via `[InProcess]`) because BenchmarkDotNet tries to set `ThreadPriority.Highest`. On Linux, raising scheduling priority requires additional permissions. Instead of running the whole benchmark as root (which can create root-owned build artifacts), you can grant the benchmark executable the `CAP_SYS_NICE` capability:
+
+    ```bash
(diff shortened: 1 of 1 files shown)


timcassell:
It was disabled in #2924, cc @filzrev

filzrev:
`#2924` suppress setting **Process** priority on non-Windows environment.
Setting **Thread** priority behavior is not changed by this PR.

And reported behavior seems not occurred on CI tests.

PR #2949, opened 2025-12-31
Revert "chore: Skip setting `process priority` when running on non windows environment"

adamsitnik:
Reverts dotnet/BenchmarkDotNet#2924 for reasons described here: https://github.com/dotnet/BenchmarkDotNet/pull/2924#pullrequestreview-3621119800

### DIFF
--- a/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ProcessExtensions.cs
@@ -27,10 +27,6 @@ namespace BenchmarkDotNet.Extensions
 
         public static void EnsureHighPriority(this Process process, ILogger logger)
         {
-            // Process.set_PriorityClass requires root on Unix
-            if (!OsDetector.IsWindows())
-                return;
-
             try
             {
                 process.PriorityClass = ProcessPriorityClass.High;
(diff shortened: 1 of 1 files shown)


filzrev:
@adamsitnik 
Sorry, I don't noticed usage scenarios that run benchmarks as **root** users.

Is it acceptable to add logics to skip setting process priority **when running on running GitHub Actions Hosted runner**
by using following code?

**Additional code to `Portablity/RuntimeInformation.cs`**

```csharp
// GitHub Actions environment variables https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables
public static readonly bool IsRunningOnGitHubActions = Environment.GetEnvironmentVariable("GITHUB_ACTIONS") == "true";

public static readonly bool IsRunningOnGitHubHostedRunner = IsRunningOnGitHubActions
                                                                 && Environment.GetEnvironmentVariable("RUNNER_ENVIRONMENT") == "github-hosted";
```
```

adamsitnik:
> @adamsitnik Sorry, I don't noticed usage scenarios that run benchmarks as **root** users.

No problem. I have developed this skill after years of working on the .NET Team and being constantly warned about very subtle breaking changes.

> Is it acceptable to add logics to skip setting process priority **when running on running GitHub Actions Hosted runner** by using following code?
> 
> **Additional code to `Portablity/RuntimeInformation.cs`**
> 

I would prefer us to modify the code that creates configs for tests and disable it there in explicit way. Because our users could in theory run as sudo using GH actions.

PR #2952, opened 2026-01-02
Update test project to a bit newer TFM

Youssef1313:


tests/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks.csproj
```diff
@@ -41,10 +40,9 @@
     <ProjectReference Include="..\..\src\BenchmarkDotNet\BenchmarkDotNet.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <!-- Microsoft.NET.Test.Sdk breaks netcoreapp2.0 -->
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.14.1" Condition=" '$(TargetFramework)' != 'netcoreapp2.0' " />
```
Youssef1313, line 45: Excluding Microsoft.NET.Test.Sdk on netcoreapp2.0 likely meant that the tests never executed for netcoreapp2.0.

tests/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks.csproj
```diff
@@ -41,10 +40,9 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageReference Include="xunit" Version="2.9.3" />
-    <!-- We use older version 2.4.1 for combatibility with netcoreapp2.0. -->
+    <!-- We use older version 2.4.1 for combatibility with net461 -->
```
Youssef1313, line 45: Probably you can consider updating the tests from net461 to net462.

tests/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks.csproj
timcassell: I never had any problems running the tests in VS (we don't run these in CI so it doesn't matter there).

timcassell:
I still don't understand the need for this. I can run the tests without issue in VS 2026 and 2022.

tests/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks.csproj
Youssef1313: And do the netcoreapp2.0 tests specifically run correctly?
I haven't tried but I doubt. Without this package reference, there is no testhost to run the tests.

timcassell: Why do you doubt?

image

Youssef1313: The testhost.exe is supposed to come through Microsoft.NET.Test.Sdk. So, without this package, there is no testhost.exe.

However, for this specific case, it looks like Microsoft.NET.Test.Sdk is still brought via xunit.runner.visualstudio 2.4.1. But that's no longer the case for newer xunit.runner.visualstudio packages that don't declare Microsoft.NET.Test.Sdk as a dependency.

In addition, version 15.0.0 of Microsoft.NET.Test.Sdk which is brought by xunit.runner.visualstudio 2.4.1 is also extremely old.

I personally think it's a lot of hassle with low value. A tfm is out of support for a reason. It's not supported, hence it's not worth testing against, especially when it comes with many drawbacks like forcing old dependencies, forcing suppressing warnings (including NuGet security warnings), and preventing moving forward (e.g, migrating from VSTest to Microsoft.Testing.Platform). IMHO, these are big enough drawbacks compared to the value of "manually testing out of support TFMs".

timcassell: I think if we want to update our testing framework, we can update the TFMs if it's necessary for the update. I don't see the point of updating the TFMs by themselves. The `ManualRunning.MultipleFrameworks` project is small enough for someone to experiment with newer XUnit or TUnit before doing it for the entire repo.

Issue #1808, opened 2021-09-13
Add IAsyncEnumerable support for Consumer


rlgordey:
Please, add IAsyncEnumerable to list of Consume targets.

timcassell:
I'm guessing you mean when returning `IAsyncEnumerable(<T>)` from the benchmark method, it should `await foreach` on the result and measure all the awaits and yields, and pass the `T` yielded values to `Consumer.Consume()`? Otherwise, this ask wouldn't make too much sense for just `Consumer.Consume(IAsyncEnumerable)`.

rlgordey:
> 
> I'm guessing you mean when returning `IAsyncEnumerable(<T>)` from the benchmark method, it should `await foreach` on the result and measure all the awaits and yields, and pass the `T` yielded values to `Consumer.Consume()`? 
>

Yes @timcassell you are correct.

timcassell:
In that case, this looks like a great candidate for a follow-up to my suggestion for fixing ValueTasks. https://github.com/dotnet/BenchmarkDotNet/issues/1595#issuecomment-778923271 Which is refactoring the engine to use an `IMeasurer` interface instead of `System.Action WorkloadAction` delegates.

timcassell:
For completeness, this should also include `IAsyncEnumerator<T>`.

timcassell:
I can take this if #1968 gets merged.

sramekpete:
Any progress on this one? I am actually trying to compare IEnumerable<T> and IAsyncEnumerable<T> performance, but having hard times to understand how to do it properly as `Consume` for `IAsyncEnumerable` doesn't exist.

I am happy to give a hand with implementation if someone would guide me.

If this is not something of an interest anymore, can someone just clarify how to properly consume `IAsyncEnumerable` in benchmark method, please?

I am now consuming IAsyncEnumerable by having one line at the end of the method.

`await foreach (var _ in result) { }`

Would that do the measurement somehow relevant to `Consumer.Consume()` method?

Thanks for a response and great work on improvements!

timcassell:
Unfortunately, this isn't possible to implement correctly without the async refactor. `await foreach (var _ in result) { }` is the best you can do for now (it still includes the overhead of the `async Task`).

`IEnumerable<T>` and `IAsyncEnumerable<T>` aren't really comparable, as they serve different purposes. What are you trying to achieve?

sramekpete:
@timcassell Thanks for a quick reply!

[...]

Issue #1045, opened 2019-02-03
Dry jobs can eat iteration failures


danmosemsft:
## BDN Version
`Benchmarks 0.1.1-preview2-190203-1`

## Repro 
Consider a benchmark where the method under test sets some result which the `[GlobalCleanup]` examines. If the result is not set, `[GlobalCleanup]` will throw. Assume the test method itself has a bug, and throws an exception:
```c#
using System;
using System.IO;
using System.Text;
using System.Linq;
using System.Reflection;
using System.Collections.Generic;
using System.Threading.Tasks;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

namespace Foo
{
    [DryJob] // Important!
    public class BarBaz
    {
        string _result;

        public static void Main(string[] args)
            => BenchmarkRunner.Run(typeof(BarBaz));

        [GlobalCleanup]
        public void CheckResults()
        {
            Console.WriteLine("global cleanup");
            Console.WriteLine("result " + _result.Length);
        }

        [Benchmark]
        public void BizBoz()
        {`
            Console.WriteLine("starting iteration");
            if ((new Random()).Next(2) < 100) throw new Exception("bad iteration");
            _result = "foo";
            Console.WriteLine("finishing iteration");
        }
    }
}
```
Run with
`dotnet run -c Release`

## Expected
The callstack for the test method exception, possibly followed by the callstack for the cleanup exception. BDN should not hide a test method failure.

## Actual
The callstack for the cleanup exception.
```
...
// BeforeActualRun
starting iteration
global cleanup

System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Foo.BarBaz.CheckResults() in /home/dan/git/corefxlab/tfoo.cs:line 29
   at BenchmarkDotNet.Engines.Engine.Dispose()
   at BenchmarkDotNet.Autogenerated.Runnable_0.Run(IHost host, String benchmarkName) in /home/dan/git/corefxlab/tests/Benchmarks/bin/Release/netcoreapp3.0/41182a39-ec98-4e84-a2ad-bc9ea467e163/41182a39-ec98-4e84-a2ad-bc9ea467e163.notcs:line 157
   --- End of inner exception stack trace ---
[...]

PR #2958, opened 2026-01-05
Async Refactor

### LINKED ISSUE #2159: ctrl+c causes crash
It doesn't happen 100% of the time, but it happens often enough. I believe it's caused by the child process not killed gracefully, but it's hard to say. We can hook into `Console.CancelKeyPress` event, so we should be able to gracefully kill the child process instead of it crashing.

### LINKED ISSUE #2162: [Feature][Breaking change] Remove ExportToFiles from IExporter
https://github.com/dotnet/BenchmarkDotNet/blob/1fb1015556039664451dd101b4e2ef9501e03805/src/BenchmarkDotNet/Exporters/IExporter.cs#L12

I would like to suggest to remove ExportToFiles from IExporter, because it is not said that the output is a file but it could be, a database, a rest endpoint, a websock, named pipe, etc ...

### LINKED ISSUE #2299: Unable to benchmark with Mono runtime on Linux
Benchmarking using Mono as a runtime on Linux seems to be broken since 0.13.3, with an error message saying "The Roslyn toolchain is only supported on .NET Framework". This leads me to believe that #2120 is the cause for the issue. Simply downgrading to 0.13.2 allows the benchmark to run as expected, and is correctly reported as using the Mono runtime.

Minimal repro (just needs the .NET 7 SDK to build and Mono to run): [BDNMinimalRepro.zip](https://github.com/dotnet/BenchmarkDotNet/files/11358729/BDNMinimalRepro.zip)

### LINKED ISSUE #2442: Please, add async support for iteration setup and cleanup
It does not seem that `[IterationSetup]` and `[IterationCleanup]` may mark `async` methods for now. The code is failing with `…the wrong return type` error. Is there any reason why global counterparts are allowed to be async, while iteration bound methods are not?

timcassell:
## Core changes:
1. Added `BenchmarkRunner.RunAsync`/`BenchmarkSwitcher.Run*Async` APIs.
2. Added cooperative cancellation support.
3. Async benchmarks are now properly awaited instead of synchronously blocked
    - This follows the suggested implementation in https://github.com/dotnet/BenchmarkDotNet/issues/1595#issuecomment-731208863
    - Allows async benchmarks to work properly in-process with UI frameworks (currently it deadlocks).
4. Interfaces changed to support async.
5. `IterationSetup/Cleanup` now support async methods.
6. `IClock` is passed to the `WorkloadActionNoUnroll` etc. benchmark methods and they pass back `ClockSpan`, instead of the `Engine` starting and stopping the clock.
    - This is to avoid measuring workload setup before the core loop and async continuations for more accurate measurements. (It also helps to simplify the jit stage.)
 
## Behavior changes:
1. `await Task.Yield()` continues on the current synchronization context if it exists, or the current task scheduler. Previously this meant it continued on a ThreadPool thread, now it means it will likely run on the same thread via the new `BenchmarkSynchronizationContext`. This should be benign.

## Breaking changes:
1. `IHost`
    - Removed unused `void Write(string message)`
    - Added `CancellationToken CancellationToken { get; }` and `ValueTask Yield()`
    - Changed `void SendSignal(HostSignal hostSignal)` to `ValueTask SendSignalAsync(HostSignal hostSignal)`
2. `IEngine`
    - `RunResults Run()` changed to `ValueTask<RunResults> RunAsync()`
3. `EngineParameters`
    - `WorkloadActionNoUnroll` etc. changed from `Action<long>` to `Func<long, IClock, ValueTask<ClockSpan>>`
    - `GlobalSetupAction` etc. changed from `Action` to `Func<ValueTask>`
4. `IExecutor`
    - `ExecuteResult Execute(ExecuteParameters executeParameters)` changed to `ValueTask<ExecuteResult> ExecuteAsync(ExecuteParameters executeParameters, CancellationToken cancellationToken)`
5. `IGenerator`
    - Changed `GenerateResult GenerateProject(BuildPartition buildPartition, ILogger logger, string rootArtifactsFolderPath)` to `ValueTask<GenerateResult> GenerateProjectAsync(BuildPartition buildPartition, ILogger logger, string rootArtifactsFolderPath, CancellationToken cancellationToken)`
7. `IBuilder`
    - Changed `BuildResult Build(GenerateResult generateResult, BuildPartition buildPartition, ILogger logger)` to `ValueTask<BuildResult> BuildAsync(GenerateResult generateResult, BuildPartition buildPartition, ILogger logger, CancellationToken cancellationToken)`
8. `IDiagnoser`
    - Changed `void Handle(HostSignal signal, DiagnoserActionParameters parameters)` to `ValueTask HandleAsync(HostSignal signal, DiagnoserActionParameters parameters, CancellationToken cancellationToken)`
9. `IDiagnoser`, `IToolchain`, `IValidator`
    - Changed `IEnumerable<ValidationError> Validate(ValidationParameters validationParameters)` to `IAsyncEnumerable<ValidationError> ValidateAsync(ValidationParameters validationParameters);`
10. `IExporter`
    - Removed `ExportToLog` and `ExportToFiles`
    - Added `ValueTask ExportAsync(Summary summary, ILogger logger, CancellationToken cancellationToken)`

## Other Changes:
1. Added support for custom awaitable returns
    - Not supported in `InProcessNoEmitToolchain`
    - Does not include extension await
2. New attributes
    - `[AsyncCallerType]` allows users to override the type used in the async method that calls their benchmark method.
        - Not supported in `InProcessNoEmitToolchain`
    - `[BenchmarkCancellation]` allows benchmarks to respond to cancellation requests from the user.
    - `[AggressivelyOptimizeMethods]` instructs the assembly weaver to apply `AggressiveOptimization` to all methods in the annotated type and nested types.
        - This is intended for internal use, but users can also use it. It's needed because C# doesn't allow to apply attributes to compiler-generated async state machines.
3. Simplified `EngineJitStage`
4. `ReturnValueValidator` is now able to validate async results.
5. Anonymous pipe IPC changed to TCP loopback.
6. Wasm toolchain now supports IPC for all JS engines (WebSocket or StdOut+File depending on the engine; configurable).
7. Added new analyzers and a code fixer.

Fixes #2442
Fixes #2159
Fixes #2299
Fixes #2162
Unblocks #1808

### DIFF
--- a/src/BenchmarkDotNet/Environments/Runtimes/WasmRuntime.cs
+++ b/src/BenchmarkDotNet/Environments/Runtimes/WasmRuntime.cs
@@ -37,8 +37,7 @@ namespace BenchmarkDotNet.Environments
         public int ProcessTimeoutMinutes { get; }
 
         /// <summary>
-        /// Specifies the IPC mechanism to use. Default is <see cref="WasmIpcType.Auto"/> which automatically detects
-        /// based on the JavaScript engine name.
+        /// Specifies the IPC mechanism to use. Default is <see cref="WasmIpcType.Auto"/> which automatically detects the JavaScript engine capabilities.
         /// </summary>
         public WasmIpcType IpcType { get; }
 


src/BenchmarkDotNet/Toolchains/InProcess/Emit/Implementation/Emitters/AsyncCoreEmitter.cs
```diff
@@ -0,0 +1,903 @@
...
+using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using static BenchmarkDotNet.Toolchains.InProcess.Emit.Implementation.RunnableConstants;
+
+namespace BenchmarkDotNet.Toolchains.InProcess.Emit.Implementation;
+
+partial class RunnableEmitter
+{
+    // TODO: update this to support runtime-async.
```
timcassell, line 18: @VSadov Should we emit runtime-async methods for this? Is it supported already? If so, what's the pattern?

timcassell:
@copilot review

timcassell:
I did the minimum changes required for proper async benchmarks, but I'm not sure if we want to take it further. We could make all interface methods async (`IGenerator.GenerateProjectAsync`, `IBuilder.BuildAsync`, etc), and we could maybe add `CancellationToken` support (I'm not sure the best way to signal cancellation to the child process, though).

@stephentoub Can we get your input here?

stephentoub:
> @stephentoub Can we get your input here?

Input on what specifically?

timcassell:
> > @stephentoub Can we get your input here?
> 
> Input on what specifically?

On the whole async API design. More specifically my previous comment about how much of the surface we should make async and whether it even makes sense to add cancelation token support.

timcassell:
@pavelsavara Can you sanity check the wasm changes here? Claude had the genius idea to use file-based IPC for JS shells that don't support WebSocket, so I was able to use that and still keep v8 as the default engine. (I don't expect a full review, this PR is too massive.)

timcassell:
🤖 *This review was generated by Claude.*

## PR #2958 Review: Async Refactoring

### Overview

This is a massive PR (300+ files, 18K+ additions) that converts BenchmarkDotNet's entire pipeline to async. The core goals are well-motivated: proper async benchmark awaiting (fixing deadlocks with UI frameworks), cooperative cancellation, and async `RunAsync`/`RunAllAsync` APIs.

The architecture is sound. The key design decisions — `BenchmarkSynchronizationContext` for thread affinity, `WorkloadContinuerAndValueTaskSource` for zero-alloc async loops, TCP loopback replacing anonymous pipes, `IClock` passed into workload methods — are all well-reasoned.

### Engine & Core Pipeline

- The async flow through `BenchmarkRunnerClean` → `Engine.RunAsync()` → workload actions is clean and correct.
- `ConfigureAwait(true)` usage in `Engine.cs` is intentional and appropriate — it ensures continuations run on the `BenchmarkSynchronizationContext` thread for consistent measurements.
- Exception handling in `RunAsync()` properly preserves the original exception when cleanup also throws (#1045).

### IPC Changes (Anonymous Pipes → TCP Loopback)

- Binding to `IPAddress.Loopback` with port 0 and backlog of 1 is appropriate for single-process IPC.
- The signal/acknowledgment protocol between `TcpHost` and `Broker` is well-designed, with proper handling of the `Cancel` signal allowing `AfterAll` to still be received.
- `Broker.ProcessDataCore` correctly uses a linked `CancellationTokenSource` for `AcceptConnection` and properly scopes the cancellation registration.
- Socket error handling (`ConnectionReset`, `Shutdown`, `OperationAborted`) covers the expected failure modes.

### InProcess Emit (IL Emission)

This is the most impressive part of the PR. The async state machine emission is correct and matches Roslyn-generated patterns:

- State initialized to -1, set to -2 on completion/exception — correct per C# compiler convention.
- `AsyncTaskMethodBuilder` lifecycle (`Create` → `Start` → `SetResult`/`SetException`) is properly implemented.
- Value type vs reference type awaiter handling (`Call` vs `Callvirt`, `ldloca` vs `ldloc`) is correct.
- The `WorkloadContinuerAndValueTaskSource` integration for reusable async state machines in the benchmark loop is sophisticated and well-executed.
- Setup/cleanup emission correctly wraps synchronous methods in completed `ValueTask` for uniform API.

### Validators, Diagnosers, Exporters

- `IValidator.ValidateAsync` returning `IAsyncEnumerable<ValidationError>` is a clean async streaming interface.
- `BenchmarkCancellationValidator` is thorough — correctly inspects fields and properties with comprehensive error messages.
- Exporters make good use of async file I/O and `await using` with `ProcessCleanupHelper` and async channels for streaming process output.

### Analyzers & Code Fixers

- `AsyncBenchmarkAnalyzer` and `BenchmarkCancellationAttributeAnalyzer` provide good compile-time guidance for users adopting the new async APIs.
- Comprehensive validation of `[BenchmarkCancellation]` attribute usage (type checks, accessibility, mutability).

### Testing

Good coverage with `RunAsyncTests`, `CancellationTokenTests`, `CustomTaskAndAwaitableTests`, and `CallerThreadTests` covering the key new functionality, plus analyzer test coverage.

### Minor Observations

None of these affect the overall design or implementation:

- `Broker` doesn't dispose `IpcListener` itself — ownership is managed by the outer scope in `Executor.cs`, but the implicit ownership model could be documented.
- `TcpClient.Connect` in `IpcHelper` has no explicit timeout, relying on the OS default.
- `IHost.Yield()` purpose (yielding back to JS in WASM) could be documented on the interface.

### Verdict

The async refactoring is well-architected and the implementation is solid for a change of this scope. The IL emission work is particularly impressive. No issues found.

PR #2964, opened 2026-01-11
feat: Show log file path on benchmark finished

filzrev:
Currently ArtifactsDirectory and generated log file path is not shown on console.
And it need to find log files from `bin\Debug\net8.0\BenchmarkDotNet.Artifacts`.

This PR add feature to show following paths as clickable link when benchmark finished.
- `ArtifactsDirectory`
- `LogFilePath`

This behavior is skipped when `--disableLogFile` is passed.

**Sample Image**
image

**What's Changed in this PR**
1. Add `Helpers/ConsoleHelper.cs` to output clickable link by using ANSI Escape Sequence OSC8.  
    When stdout is redirected or terminal don't support OSC8. It write URL as plain text. 
3. Add `Helpers/PathHelper.cs` to gets relative path. (It's required because `netstandard2.0` don't support `Path.GetRelativePath`)
4. Modify `CompositeLogger.cs` and add code to get ConsoleLogger-like logger.
5. Modify `BenchmarkRunnerClean.cs` and add code to output clickable links.
6. Add `Helpers/PathHelperTests.cs` unit tests for `PathHelper.GetRelativePath`.

**Other Changes**
- Modify `ConfigParser.cs` to suppress exception thrown when running on VS TestExplorer.
- Modify `BenchmarkRunnerClean.cs`  and insert empty line after `Global total time`.

### DIFF
--- a/src/BenchmarkDotNet/Helpers/ConsoleHelper.cs
+++ b/src/BenchmarkDotNet/Helpers/ConsoleHelper.cs
@@ -1,0 +1,135 @@
+using BenchmarkDotNet.Detectors;
+using BenchmarkDotNet.Extensions;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+using System.Text.RegularExpressions;
+
+namespace BenchmarkDotNet.Helpers;
+
+internal static class ConsoleHelper
+{
+    private const string ESC = "\e";          // Escape sequence.
+    private const string OSC8 = $"{ESC}]8;;"; // Operating System Command 8
+    private const string ST = ESC + @"\";     // String Terminator
+
+    /// <summary>
+    /// Try to gets clickable link text for console.
+    /// If console doesn't support clickable link, it returns false.
+    /// </summary>
+    public static bool TryGetClickableLink(string link, string? linkCaption, out string result)
+    {
+        if (!IsClickableLinkSupported)
+        {
+            result = "";
+            return false;
+        }
+
+        result = @$"{OSC8}{link}{ST}{linkCaption ?? link}{OSC8}{ST}";
+        return true;
+    }
+
+    public static bool IsWindowsTerminal => _isWindowsTerminal.Value;
+
+    public static bool IsClickableLinkSupported => _isClickableLinkSupported.Value;
+
+    private static readonly Lazy<bool> _isWindowsTerminal = new(()
+        => Environment.GetEnvironmentVariable("WT_SESSION") != null);
+
+    private static readonly Lazy<bool> _isClickableLinkSupported = new(() =>
+    {
+        if (Console.IsOutputRedirected)
+            return false;
+
+        // The current console doesn't have a valid buffer size, which means it is not a real console.
+        if (Console.BufferHeight == 0 || Console.BufferWidth == 0)
+            return false;
+
+        // Disable clickable link on CI environment.
+        if (Environment.GetEnvironmentVariable("CI").IsNotBlank())
+            return false;
+
+        // dumb terminal don't support ANSI escape sequence.
+        var term = Environment.GetEnvironmentVariable("TERM") ?? "";
+        if (term == "dumb")
+            return false;
+
+        if (OsDetector.IsWindows())
+        {
+            try
+            {
+                // conhost.exe don't support clickable link with OSC8.
+                if (IsRunningOnConhost())
+                    return false;
+
+                // ConEmu and don't support OSC8.
+                var conEmu = Environment.GetEnvironmentVariable("ConEmuANSI");
+                if (conEmu != null)
+                    return false;
+
+                // Return true if Virtual Terminal Processing mode is enabled.
+                return IsVirtualTerminalProcessingEnabled();
+            }
+            catch
+            {
+                return false; // Ignore unexpected exception.
+            }
+        }
+        else
+        {
+            // screen don't support OSC8 clickable link.
+            if (Regex.IsMatch(term, "^screen"))
+                return false;
+
+            // Other major terminal supports OSC8 by default. https://github.com/Alhadis/OSC8-Adoption
+            return true;
+        }
+    });
+
+    [SupportedOSPlatform("windows")]
+    private static bool IsVirtualTerminalProcessingEnabled()
+    {
+        const uint STD_OUTPUT_HANDLE = unchecked((uint)-11);
+        IntPtr handle = NativeMethods.GetStdHandle(STD_OUTPUT_HANDLE);
+        if (handle == IntPtr.Zero)
+            return false;
+
+        if (NativeMethods.GetConsoleMode(handle, out uint consoleMode))
+        {
+            const uint ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004;
+            if ((consoleMode & ENABLE_VIRTUAL_TERMINAL_PROCESSING) > 0)
+            {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    [SupportedOSPlatform("windows")]
+    private static bool IsRunningOnConhost()
+    {
+        IntPtr hwnd = NativeMethods.GetConsoleWindow();
+        if (hwnd == IntPtr.Zero)
+            return false;
+
+        NativeMethods.GetWindowThreadProcessId(hwnd, out uint pid);
+        using var process = Process.GetProcessById((int)pid);
+        return process.ProcessName == "conhost";
+    }
+
+    [SupportedOSPlatform("windows")]
+    private static class NativeMethods
+    {
+        [DllImport("kernel32.dll", SetLastError = true)]
+        public static extern IntPtr GetStdHandle(uint nStdHandle);
+
+        [DllImport("kernel32.dll", SetLastError = true)]
+        public static extern bool GetConsoleMode(IntPtr hConsoleHandle, out uint lpMode);
+
+        [DllImport("kernel32.dll", SetLastError = true)]
+        public static extern IntPtr GetConsoleWindow();
+
+        [DllImport("user32.dll", SetLastError = true)]
+        public static extern uint GetWindowThreadProcessId(IntPtr hWnd, out uint lpdwProcessId);
+    }
+}
--- a/src/BenchmarkDotNet/Loggers/CompositeLogger.cs
+++ b/src/BenchmarkDotNet/Loggers/CompositeLogger.cs
@@ -2,7 +2,7 @@ using System.Collections.Immutable;
 
 namespace BenchmarkDotNet.Loggers
 {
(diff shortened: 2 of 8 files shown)


timcassell:
There appears to be a bit of overlap with `TaskbarProgress`. Can you move common functionality to a separate helper (or to `ConsoleHelper`)?

src/BenchmarkDotNet/Loggers/CompositeLogger.cs
```diff
@@ -57,5 +61,14 @@ public void Flush()
...
         }
+
+        /// <summary>
+        /// Try to gets logger that has id of ConsoleLogger.
+        /// </summary>
+        public bool TryGetConsoleLogger([NotNullWhen(true)] out ILogger? consoleLogger)
+        {
+            consoleLogger = loggers.FirstOrDefault(x => x.Id == nameof(ConsoleLogger));
+            return consoleLogger != null;
+        }
```
timcassell:
```suggestion
        public bool TryGetConsoleLogger([NotNullWhen(true)] out ConsoleLogger? consoleLogger)
        {
            consoleLogger = loggers.OfType<ConsoleLogger>().FirstOrDefault();
            return consoleLogger != null;
        }
```

src/BenchmarkDotNet/Helpers/ConsoleHelper.cs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// <summary>
+    /// Write clickable link to console.
+    /// If console doesn't support OSC 8 hyperlinks. It writes plain link with markdown syntax.
+    /// </summary>
+    public static void WriteAsClickableLink(ILogger consoleLogger, string link, string? linkCaption = null, LogKind logKind = LogKind.Info)
+    {
+        if (consoleLogger.Id != nameof(ConsoleLogger))
+            throw new NotSupportedException("This method is expected logger that has ConsoleLogger id.");
+
```
timcassell:
```suggestion
    public static void WriteAsClickableLink(ConsoleLogger consoleLogger, string link, string? linkCaption = null, LogKind logKind = LogKind.Info)
    {
```

src/BenchmarkDotNet/Helpers/ConsoleHelper.cs
```diff
@@ -0,0 +1,168 @@
...
+{
+    private const string ESC = "\e";          // Escape sequence.
+    private const string OSC8 = $"{ESC}]8;;"; // Operating System Command 8
+    private const string ST = ESC + @"\";     // String Terminator
+
+    /// <summary>
+    /// Write clickable link to console.
+    /// If console doesn't support OSC 8 hyperlinks. It writes plain link with markdown syntax.
+    /// </summary>
+    public static void WriteLineAsClickableLink(ILogger consoleLogger, string link, string? linkCaption = null, LogKind logKind = LogKind.Info, string prefixText = "", string suffixText = "")
```
timcassell:
```suggestion
    public static void WriteLineAsClickableLink(ConsoleLogger consoleLogger, string link, string? linkCaption = null, LogKind logKind = LogKind.Info, string prefixText = "", string suffixText = "")
```

timcassell:
I wonder if we should add a method to `ILogger` interface so custom loggers can also participate in links (it's a shame netstandard2.0 doesn't support DIMs, so it would be a breaking change).

timcassell:
It would also be good to use this for `IDiagnoser.DisplayResults` implementations (e.g. `DisassemblyDiagnoser`, etc.) (more reason to add it to `ILogger` interface I guess).

src/BenchmarkDotNet/Loggers/CompositeLogger.cs
filzrev: This code intended to identify logger that has Id to support ConsoleLogger-like logger.

It's required to support replace existing console logger to custom console logger.
(e.g. https://github.com/dotnet/BenchmarkDotNet/issues/190)

In this case, benchmark progress log is not outputted on console.
So this feature is useful to confirm logs when benchmark completed.

src/BenchmarkDotNet/Helpers/ConsoleHelper.cs
filzrev: It need to support custom logger that have ConsoleLogger Id also.

So using ILogger interface as argument.
And block non ConsoleLogger-like loggers.

filzrev:
> I wonder if we should add a method to ILogger interface so custom loggers can also participate in links (it's a shame netstandard2.0 doesn't support DIMs, so it would be a breaking change).

ANSI escape sequence is not correctly shown when written to non-console logger.
And clickable link is useful for TerminalLogger (FileLogger).
So I've output clickable links for ConsoleLogger-like logger only.

> It would also be good to use this for IDiagnoser.DisplayResults implementations (e.g. DisassemblyDiagnoser, etc.) (more reason to add it to ILogger interface I guess).

As described above. I want to use this feature with custom TeraminalLogger/QuietLogger.
So it is desirable that this be executed at the end of processing.

> There appears to be a bit of overlap with TaskbarProgress. Can you move common functionality to a separate helper (or to ConsoleHelper)?
I'll check TaskbarProgress implementation later.

timcassell:
> ANSI escape sequence is not correctly shown when written to non-console logger.

Non-console loggers would implement the link as they see fit. You'd just pass the raw link, not the escape sequence.

filzrev:
I've tried to extend ILogger interface.
But without `Default Interface Methods` feature. 
it require a lot of additional code to support `WriteLink` API. and it affects existing thirdparty custom loggers.

So I wish to keep existing implementation direction.

> There appears to be a bit of overlap with TaskbarProgress.

I've confirmed TaskbarProgress API.
But it seems there aren't many duplicate parts. (Only `GetConsoleMode` API and related constants. and )
It might be better to merge ANSI Escape Sequence handling code and Console related APIs in future.
But for now, I'd like to leave it as is.

Additionally, I've removed code that output link with `markdown link syntax`(82e72e0).
Because it's hard to read when caption and link path is long text.

timcassell:
What about add a new `interface ILinkLogger : ILogger` that we can type-check for? Then we can add an `ILogger` extension that handles it.

filzrev:
> What about add a new interface ILinkLogger : ILogger that we can type-check for? Then we can add an ILogger extension that handles it.

I've modified following codes at 956d44c
- Add `IConsoleLogger: ILogger` interface and define related extension methods.
- Separate logics that relating to `ANSI Escape Sequences` to ConsoleHelper.
- Use StringBuilder to output clickable link line. Because it may be called from multiple thread in future.
- Change code to resolve IConsoleLogger from `FirstOrDefault` to `SingleOrDefault`.   
   Because logger that write to stdout should not be multiple instance.

timcassell:
I was thinking more like:

```cs
public interface ILinkLogger : ILogger
{
    void WriteLink(string link, string? linkCaption = null, LogKind logKind = LogKind.Info);
}

public static class ILoggerExtensions
{
    public void WriteLink(this ILogger logger, string link, string? linkCaption = null, LogKind logKind = LogKind.Info)
        => logger is ILinkLogger linkLogger
            ? linkLogger.WriteLink(link, linkCaption, logKind)
            : logger.Write(logKind, link);
}
```

And `ConsoleLogger` implements it with the escape sequences. Because the console is just a UI, a different UI may display links differently.

We don't need prefix and suffix because you can just use `logger.Write(prefix); logger.WriteLink(...); logger.Write(suffix);`.

`CompositeLogger` can implement `ILinkLogger` and just call the extension on each logger.

filzrev:
I've updated code based on review comment.

 > We don't need prefix and suffix because you can just use logger.Write(prefix); logger.WriteLink(...); logger.Write(suffix);

It's required for temporary workaround for Windows Terminal inside helper method.
And when using multiple `Write` to console.  Ot might be mixed when another thread call directly `Console.Write/WriteLine`.

---
`ILoggerExtensions` is exposed with `public` visibility.
This is because it can be used for generic purposes.

timcassell:
I'm not concerned about race conditions. Loggers are already assumed to not be thread-safe. In the one place we do parallel processing (builds), we use `NullLogger` to avoid the issue.

I don't like the new interface just being a marker. I feel like you ignored my comment

> Because the console is just a UI, a different UI may display links differently.

filzrev:
> I don't like the new interface just being a marker. I feel like you ignored my comment

I've removed related doc comment on latest commit.

> Because the console is just a UI, a different UI may display links differently.

`WriteLink` method should be implemented as `ILogger` interface method.
But currently it's implemented as `ILogger`'s extension method to support `netstandard 2.0` .
(It require a lot of tasks without `default interface method` method)

And currently ILinkLogger is just be used as marker interface to distinguish logger that support ANSI escape sequence and OSC8.
To support other type of UI. It need to add another interface. (e.g. To support clickable link for VSTestLogger. It need to add `file:///` prefix)

timcassell:
That doesn't make sense, why would you use separate marker interfaces for every implementation? If you put the method on the interface only 1 interface is needed; the console logger can implement it with escape sequence, and VS test logger can implement it with the file prefix. The extension exists simply to simulate DIM for old runtimes.

filzrev:
> What doesn't make sense, why would you use separate marker interfaces for every implementation?

Ignore above my comment.
I re-read my own comments, and there is no need to implement a different marker interface.

My original intent was that if new type of `ILinkLogger` type is added.
It need to conditionaly switch by type inside extension method.
So `ILoggerExtension need to know all `ILinkLogger` derived classes.

It can be avoided when using `Default Interface Implementation` or changing logger base class something like `LoggerBase` class.
But it's over engineering because additional `ILinkLogger` is not expected to be implemented.

timcassell:
> It can be avoided when using `Default Interface Implementation` or changing logger base class something like `LoggerBase` class.

That's what the extension is supposed to do, simulate DIM (if the logger doesn't implement `ILinkLogger`, it uses a default implementation, otherwise it calls the interface method).

> But it's over engineering because additional `ILinkLogger` is not expected to be implemented.

It is expected to be implemented, though? You just gave an example of console logger and VS test logger needing different implementations. Custom user loggers could also have different implementations that we don't know about.

filzrev:
> It is expected to be implemented, though?

Adding `file:///` prefix for link reduce log readability and it's now used frequently for normal use cases.
So it'll not expected to be added.

> Custom user loggers could also have different implementations that we don't know about.

Currently `ILinkLogger` is defined as `internal interface` to block thirdparty logger that support link.
(It's required because `ILoggerExtensions` need to know all logger that implement `ILinkLogger`)

timcassell:
> to block thirdparty logger that support link.

Why?

> It's required because `ILoggerExtensions` need to know all logger that implement `ILinkLogger`

Also why? With my suggested extension above, it doesn't need to know the implementation details.

PR #2966, opened 2026-01-11
chore: Suppress exception thrown internally on mono is not installed environment

filzrev:
Currently, when accessing `HostEnvironmentInfo` instance in the first time.
It  **internally**  throw exception on mono is not installed environment.
and interrupt debugger when using break on exception thrown setting.

This PR suppress exception thrown when running on `mono` is not installed environment.

**1. `Helpers/ProcessHelper.cs`**
Add `TestCommandExists` helper method to check command existence with where/which command.
On `which` command is not installed environment. It fallback to call actual command with `--version` argument.

**2. `HostEnvironmentInfo.cs`**
Replace code to use `TestCommandExists` helper method.

### DIFF
--- a/src/BenchmarkDotNet/Helpers/ProcessHelper.cs
+++ b/src/BenchmarkDotNet/Helpers/ProcessHelper.cs
@@ -90,11 +90,9 @@ namespace BenchmarkDotNet.Helpers
                 {
                     FileName = OsDetector.IsWindows() ? "where" : "which",
                     Arguments = commandName,
-                    RedirectStandardOutput = true,
                     UseShellExecute = false,
                     CreateNoWindow = true
                 })!;
-                process.StandardOutput.ReadToEnd();
                 process.WaitForExit();
                 return process.ExitCode == 0;
             }
@@ -110,11 +108,9 @@ namespace BenchmarkDotNet.Helpers
                 {
                     FileName = commandName,
                     Arguments = arguments,
-                    RedirectStandardOutput = true,
                     UseShellExecute = false,
                     CreateNoWindow = true
                 })!;
-                process.StandardOutput.ReadToEnd();
                 process.WaitForExit();
                 return process.ExitCode == 0;
             }


src/BenchmarkDotNet/Helpers/ProcessHelper.cs
```diff
@@ -79,5 +80,48 @@ internal static (int exitCode, ImmutableArray<string> output) RunAndReadOutputLi
...
+                // On some environment. which command is not installed. (e.g. Alpine Linux)
+            }
+
+            // Check command existence by executing actual command with --version argument.
+            try
+            {
+                using var process = Process.Start(new ProcessStartInfo
+                {
+                    FileName = commandName,
+                    Arguments = "--version",
```
timcassell: Should pass this as an argument rather than hard-code it. I imagine this helper would be used by other things to check for existence of a command also.

src/BenchmarkDotNet/Helpers/ProcessHelper.cs
```diff
@@ -79,5 +80,48 @@ internal static (int exitCode, ImmutableArray<string> output) RunAndReadOutputLi
...
+                    RedirectStandardOutput = true,
```
timcassell: Why redirect it if we're not using the result?

filzrev: It's added to avoid deadlock.
When it's resolved to unexpected command. and `--version` is not supported and write a lot of output to stdout.

Since it's an extremely rare case, it's fine to delete it.

timcassell: I don't think there's a deadlock issue if output is not redirected.

filzrev: I've removed `RedirectStandardOutput = true` and ` process.StandardOutput.ReadToEnd();` statements.

I've added these code because review by LLM suggest to add these lines.
- To avoid deadlock issue that described above.
- On some environment stdout is not exists and exception thrown when try to stdout.

But as far as I've tested. It seems works without problem.
So I removed these code.

PR #2935, opened 2025-12-18
Fix NETSDK1152 in auto-generated projects

timcassell:
https://github.com/dotnet/BenchmarkDotNet/pull/2931#issuecomment-3671856363

PR #2931, opened 2025-12-16
[msbuild] refactor `BenchmarkDotNet.Weaver.targets` to support mobile platforms

jonathanpeppers:
Context: https://github.com/dotnet/BenchmarkDotNet/pull/2929
Context: https://github.com/dotnet/android/blob/4aa9af89102af2e745a8507992187d3c5993d638/Documentation/guides/MSBuildBestPractices.md

In initially testing BenchmarkDotNet with .NET MAUI, I found that the weaver was not being executed because the `Publish` target is not called during the build process for Android and iOS projects. I think the target could actually run much sooner during builds and achieve better results.

To address this, I refactored the `BenchmarkDotNet.Weaver.targets` file to run after `CoreCompile` on the `@(IntermediateAssembly)` in the `obj` directory. This is similar to what other targets do, such as the XamlC compiler:

https://github.com/dotnet/maui/blob/8224becbb3a8a6bb1caaca4bbe70c56e88875506/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.targets#L213-L255

Other general MSBuild improvements:

* Defined an MSBuild property for everything that seems useful. This allows consuming projects to configure the behavior (or workarounds!) as needed.

* Made the MSBuild target incremental by using inputs and outputs. If the `.dll` file is an input, we can write a `.stamp` file as an output to indicate that the weaver has already been run for that assembly. If the `.dll` file changes, the weaver will run again.

I tested this change with a .NET MAUI on 4 platforms and also the existing `BenchmarkDotNet.Samples` console app.

PR #2967, opened 2026-01-12
Add support for composite r2r runs

BrzVlad:
This is achieved by using already existing properties in the sdk: PublishReadyToRun and PublishReadyToRunComposite. The compositer2r toolchain adds these properties to a standard project template. In order to test with custom built runtime, the toolchain reuses the already existing `customruntimepack` and `aotcompilerpath` bdn args.

Example command used locally from microbenchmarks project:

`
dotnet run -c Release -f net10.0 --runtimes r2r10_0 --filter "System.Tests.Perf_Int32.TryFormat" --customruntimepack /home/vbrezae/runtime3/artifacts/bin/microsoft.netcore.app.runtime.linux-arm64/Release/ --aotcompilerpath /home/vbrezae/runtime3/artifacts/packages/Release/Shipping/crossgen2pack/
`

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/R2RTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/R2RTests.cs
@@ -28,7 +28,8 @@ namespace BenchmarkDotNet.IntegrationTests
             return ManualConfig.CreateEmpty()
                 .AddJob(Job.Dry
                     .WithRuntime(GetCurrentR2RRuntime())
-                    .WithToolchain(toolchain));
+                    .WithToolchain(toolchain))
+                .WithBuildTimeout(TimeSpan.FromSeconds(240));
         }
 
         private static string GetTargetFrameworkMoniker()


src/BenchmarkDotNet/Toolchains/CompositeR2R/CompositeR2RGenerator.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        private readonly string CustomRuntimePack;
+        private readonly string Crossgen2Pack;
+
+        public CompositeR2RGenerator(string targetFrameworkMoniker, string cliPath, string packagesPath, string customRuntimePack, string crossgen2Pack)
+            : base(targetFrameworkMoniker, cliPath, packagesPath, runtimeFrameworkVersion: null)
+        {
+            CustomRuntimePack = customRuntimePack;
+            Crossgen2Pack = crossgen2Pack;
+        }
```
timcassell, line 25: I believe R2R doesn't apply to dynamically invoked methods, correct? If so, you should set the call type to direct:

```suggestion
            Crossgen2Pack = crossgen2Pack;
            BenchmarkRunCallType = Code.CodeGenBenchmarkRunCallType.Direct;
        }
```

src/BenchmarkDotNet.Diagnostics.dotTrace/.DotTraceDiagnoser.cs.swp
timcassell, line 1: Remove this file.

timcassell (commented):
There has been user interest in a R2R toolchain (#2514). I think it would be good to make it work for general use, not just custom runtime. Could also make "Composite" an optional setting.

BrzVlad:
> There has been user interest in a R2R toolchain (#2514). I think it would be good to make it work for general use, not just custom runtime. Could also make "Composite" an optional setting.

@timcassell I didn't really make it clear, but it actually works already with a normal sdk. If there is no passed custom runtime pack and crossgen2 pack then the `TrickRuntimePackLocation` target has no effect.

On a separate note, after rebasing to latest main, I tried using the locally built bdn with a sample project. The autogenerated project failed to build with the following error:

`/home/vbrezae/.dotnet/sdk/10.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.CrossGen.targets(467,5): error : Unhandled exception. System.CommandLine.CommandLineException: Multiple input files matching same simple name /home/vbrezae/Xamarin/tests/bdn-test/bdn-proj/bin/Release/net10.0/bdn-test-CompositeR2R-1/obj/BenchmarkDotNet.Autogenerated/release_net10.0_linux-arm64/bdn-test-CompositeR2R-1.dll /home/vbrezae/Xamarin/tests/bdn-test/bdn-proj/bin/Release/net10.0/bdn-test-CompositeR2R-1/bin/Release/net10.0/linux-arm64/publish/bdn-test-CompositeR2R-1.dll [/home/vbrezae/Xamarin/tests/bdn-test/bdn-proj/bin/Release/net10.0/bdn-test-CompositeR2R-1/BenchmarkDotNet.Autogenerated.csproj::TargetFramework=net10.0]`

This error is triggered by `crossgen2` when it receives the same assembly twice as an input. As you can see in the error message, it receives the `bdn-test-CompositeR2R-1.dll` twice. One usage comes from the `obj` folder which is suspicious. After a bit of bisecting, turns out that this started happening after https://github.com/dotnet/BenchmarkDotNet/commit/9368a7d28be0153a0d4a1122979927f190f2f97b. Would you be able to take a look at this ?

timcassell:
That sounds like the same issue as https://github.com/dotnet/BenchmarkDotNet/pull/2931#issuecomment-3671856363. Thanks for finding the problematic commit. I'll take a look when I get some time.

BrzVlad:
@timcassell Thanks!! https://github.com/dotnet/BenchmarkDotNet/pull/2935 did indeed fix the issue.

Regarding the simple R2R support, does it sound good to rename the toolchain/moniker to ex. `R2R10_0` and have an additional --composite bdn arg that enables composite r2r ?

cc @adamsitnik

timcassell:
I do think it should be renamed, and add `public static readonly IToolchain Net10_0` etc fields like other toolchains, and add versions at least as far back as current supported runtimes (net8.0), optionally as far back as R2R was supported (netcoreapp3.0 according to CoPilot, but optional since users could manually construct the toolchain for the old runtimes if they need it).

I think we don't actually need a composite option in the toolchain, just always have it enabled. Users can use `MsBuildArgument` to override it if they want.

timcassell:
Also please add integration tests.

timcassell:
Also I think you should add `public class R2RRuntime : CoreRuntime` so it will have the correct `RuntimeMoniker`. And update `ToolchainExtensions.GetToolchain` to recognize it.

BrzVlad:
@timcassell I believe I've addressed all your comments. The newly added test is passing now, it is only testing the build step, there is no check at runtime. We could probably check the existence of some r2r binaries but I don't think it is really worth it and it would just make it susceptible to false positive failures in the future.

src/BenchmarkDotNet/Toolchains/R2R/R2RGenerator.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            File.WriteAllText(artifactsPaths.ProjectFilePath, content);
+
+            GatherReferences(buildPartition, artifactsPaths, logger);
+        }
+
+        protected override string GetExecutablePath(string binariesDirectoryPath, string programName)
+            => OsDetector.IsWindows()
+                ? Path.Combine(binariesDirectoryPath, $"{programName}.exe")
+                : Path.Combine(binariesDirectoryPath, programName);
```
timcassell: I'm pretty sure the base method does this already, so this is redundant.

src/BenchmarkDotNet/Toolchains/R2R/R2RToolchain.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace BenchmarkDotNet.Toolchains.R2R
+{
+    [PublicAPI]
+    public class R2RToolchain : CsProjCoreToolchain, IEquatable<R2RToolchain>
+    {
+        [PublicAPI] public static readonly IToolchain R2R80 = From(new NetCoreAppSettings("net8.0", null, "R2R80"));
+        [PublicAPI] public static readonly IToolchain R2R90 = From(new NetCoreAppSettings("net9.0", null, "R2R90"));
+        [PublicAPI] public static readonly IToolchain R2R10_0 = From(new NetCoreAppSettings("net10.0", null, "R2R10_0"));
+        [PublicAPI] public static readonly IToolchain R2R11_0 = From(new NetCoreAppSettings("net11.0", null, "R2R11_0"));
```
timcassell:
```suggestion
        [PublicAPI] public static readonly IToolchain R2R80 = From(new NetCoreAppSettings("net8.0", null, "R2R 8.0"));
        [PublicAPI] public static readonly IToolchain R2R90 = From(new NetCoreAppSettings("net9.0", null, "R2R 9.0"));
        [PublicAPI] public static readonly IToolchain R2R10_0 = From(new NetCoreAppSettings("net10.0", null, "R2R 10.0"));
        [PublicAPI] public static readonly IToolchain R2R11_0 = From(new NetCoreAppSettings("net11.0", null, "R2R 11.0"));
```

tests/BenchmarkDotNet.IntegrationTests/R2RTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+    }
+
+    public class R2RBenchmark
+    {
+        [Benchmark]
+        public void SimpleMethod()
+        {
+            // I don't believe there is a way to verify at runtime that we are actrually running under r2r
```
timcassell:
```suggestion
            // I don't believe there is a way to verify at runtime that we are actually running under r2r
```

timcassell: I asked CoPilot and it said

> Every R2R assembly contains a special data directory entry called:
> `READYTORUN_HEADER`
> You can detect it by reading the PE headers.

But it's probably not worth adding that.

src/BenchmarkDotNet/Toolchains/R2R/R2RGenerator.cs
BrzVlad: The parent method was returning `.exe` or `.dll` which wasn't correct for the publish result. Reduced duplication by overriding `GetExecutableExtension` instead.

PR #2962, opened 2026-01-07
chore: Migrate some part of SimpleJson dependencies to System.Text.Json

filzrev:
This PR intended to reduce dependency to `SimpleJson`.
By migrate  Disassembler related serialize/deserialize code to use `System.Text.Json`.

**What's changed in this PR**
1. Add **explicit** dependency to `System.Text.Json`
2. Add `BdnJsonSerializer` class that use source-generator based serialization/deserialization
4. Migrate JSON serialization code to use BdnJsonSerializer. (Except for JsonExporter related code)
5. Add `AwesomeAssertions` dependency to test object equivalency with custom condition.
6. Add `DisassemblerModelSerializationTests` unit tests.

PR #2968, opened 2026-01-12
Simplify type reflection

filzrev:
This PR is follow up of #2962. and intended to remove SimpleJson dependencies from following codes.
1. `FolderNameHelper.cs`
2. `SourceCodeHelper.cs`


timcassell (approved):
Is unrelated to SimpleJson, it's just simplifying type reflection. Thanks @filzrev.

Issue #2494, opened 2024-01-05
Rider and R# do not distinguish parametrized TestAdapter cases


AndreyAkinshin:
The problem is caused by [using equal FQNs](https://github.com/microsoft/vstest/blob/main/docs/RFCs/0033-Hierarchy-TestCase-Property.md#overview) for parameterized benchmarks due to [RSRP-495522](https://youtrack.jetbrains.com/issue/RSRP-495522). Until this issue is resolved, we should use different FQNs for all the cases.

PR #2978, opened 2026-01-24
Fix #2793: Use proper FQN in VS Test Explorer to avoid hierarchy split

### LINKED ISSUE #2793: Wrong benchmark hierarchy shown on VS TestExplorer
Currently following hierarchy are shown for `BenchmarkDotNet.Samples` when group by `Project, Class`.

Image

It's caused by following workaround code for Rider/R#. 
https://github.com/dotnet/BenchmarkDotNet/blob/799ecfc514f310efcd93d274199795f2ae4e276e/src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs#L34-L36

It seems FullyQualfiedName's dot(`.`) char is used for hierarchy separator.
So when job display name contains `.` char. It cause wrong grouping.

```

Shrinidhi203:
- Fix incorrect VS Test Explorer hierarchy when job display name contains `.` (e.g., ".NET 8.0.6")
  - Detect Visual Studio environment via `VSAPPIDNAME` environment variable
  - Use benchmark method FQN instead of displayName for test case in VS, preventing dots in job names from being
  misinterpreted as namespace separators
  - Preserve existing Rider/R# workaround (#2494) for non-VS environments

  ## Test plan
  - [ ] Manual testing: Verify in VS Test Explorer that benchmarks with multi-version jobs (containing dots in names
  like ".NET 8.0.6") display correct hierarchy without unintended grouping

  ### Why no automated tests
  The `ToVsTestCase` extension method and `BenchmarkCaseExtensions` class are `internal` to the
  `BenchmarkDotNet.TestAdapter` assembly. Adding unit tests would require exposing internals via `InternalsVisibleTo`,
  which introduces cross-project dependencies and framework compatibility complexity between TestAdapter
  (`netstandard2.0`) and the test project. Since the fix is straightforward and follows the exact pattern suggested in
  the issue, manual verification in VS Test Explorer is sufficient.

### DIFF
--- a/src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs
+++ b/src/BenchmarkDotNet.TestAdapter/BenchmarkCaseExtensions.cs
@@ -35,6 +35,15 @@ namespace BenchmarkDotNet.TestAdapter
             // See: https://github.com/dotnet/BenchmarkDotNet/issues/2494
             var fullyQualifiedName = displayName;
 
+            // Use benchmark method FQN on Visual Studio environment to avoid TestExplorer hierarchy split
+            // when job display name contains '.' which is interpreted as a namespace separator by VS.
+            // See: https://github.com/dotnet/BenchmarkDotNet/issues/2793
+            if (Environment.GetEnvironmentVariable("VSAPPIDNAME") != null)
+            {
+                var benchmarkMethodName = benchmarkMethod.Name;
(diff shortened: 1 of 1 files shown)


timcassell:
@filzrev Can you verify this?

filzrev:
I've tested on local VS environment.
And #2793 issue seems be resolved by this PR.

I've not tested on Rider and Resharper environment.
Because these tests seems not discover benchmarks on test explorer.

Issue #2980, opened 2026-01-24
NullReferenceException when ParamsSource references a write-only property


Shrinidhi203:
## Description
  `BenchmarkConverter.GetValidValuesForParamsSource()` throws `NullReferenceException` when a user mistakenly references
   a write-only property (no getter) in `[ParamsSource]` attribute.

  ## Location
  File: `src/BenchmarkDotNet/Running/BenchmarkConverter.cs`, line 324

  ## Problem Code
  ```csharp
  var paramsSourceProperty = sourceType.GetAllProperties()
      .FirstOrDefault(property => property.Name == sourceName && property.GetMethod.IsPublic);

  The predicate property.GetMethod.IsPublic does not check if GetMethod is null. Write-only properties have GetMethod ==
   null, causing NullReferenceException.

y1hao:
Could I take this _good first issue_?

PR #2984, opened 2026-01-29
fix: Skip properties that have no getter method during scanning

y1hao:
Fixes https://github.com/dotnet/BenchmarkDotNet/issues/2980.

When a property that's used as `ParamSource` has no getter method (a write-only property), `BenchmarkConverter.GetValidValuesForParamsSource()` throws `NullReferenceException`. This PR makes it skip such write-only property since they can't be used as the param source, this will result in an `InvalidBenchmarkDeclarationException` which is clearer.

### DIFF
--- a/tests/BenchmarkDotNet.Tests/ParamsSourceTests.cs
+++ b/tests/BenchmarkDotNet.Tests/ParamsSourceTests.cs
@@ -52,8 +52,10 @@ namespace BenchmarkDotNet.Tests
                 set { _writeOnlyValue = value; }
             }
 
+#pragma warning disable BDN1305 // Test intentionally uses write-only property
             [ParamsSource(nameof(WriteOnlyValues))]
             public int MyParam { get; set; }
+#pragma warning restore BDN1305
 
             [Benchmark]
             public void Run() { }


y1hao:
@dotnet-policy-service agree

timcassell:
Can you also add an analyzer rule for it?

y1hao:
Sure. I haven't written analyzer rules before though, will take a look this weekend.

y1hao:
Hi @timcassell, I just added the analyzer to capture this scenario. Sorry it took a bit of time.

tests/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/GeneralParameterAttributesAnalyzerTests.cs
```diff
@@ -1301,6 +1302,150 @@ public static TheoryData<string> NonPublicPropertySettersTheoryData()
...
+        [Fact]
+        public async Task UsingNameofWithReadWriteProperty_ShouldNotReportError()
+        {
+            var testCode = /* lang=c#-test */ """
+                using System.Collections.Generic;
+                using BenchmarkDotNet.Attributes;
+
+                public class BenchmarkClass
+                {
+                    public static IEnumerable<int> ValidValues => new[] { 1, 2, 3 };
```
timcassell: Copy/paste bug? The test is the same as the following test, it doesn't match the test name.

timcassell:
Can you also add the new rule to the `AnalyzerReleases.Unshipped.md` file? See `AnalyzerReleases.Shipped.md` for the format (don't include the version).

tests/BenchmarkDotNet.Analyzers.Tests/AnalyzerTests/Attributes/GeneralParameterAttributesAnalyzerTests.cs
y1hao: Ah, yeah, sorry about that. Updated.

y1hao:
> Can you also add the new rule to the `AnalyzerReleases.Unshipped.md` file? See `AnalyzerReleases.Shipped.md` for the format (don't include the version).

Done. btw, I checked the failures in CI - that's our new analyzer rule catching our new unit test code for that scenario. I'm not sure what needs to be done for these errors?

timcassell:
> Done. btw, I checked the failures in CI - that's our new analyzer rule catching our new unit test code for that scenario. I'm not sure what needs to be done for these errors?

Suppress it in source (`#pragma warning disable`).

Issue #2648, opened 2024-09-25
Ratio value is not 1.00 for baseline


jjanuszkiewicz:
For starters, please excuse me if this is not a bug, but my misunderstanding of what the Ratio value represents.

I've created 2 benchmarks, one of them is marked as baseline. I'm getting results such as:

    | Method        | Mean       | Error     | StdDev    | Median   | Min      | Max        | Ratio | RatioSD | Baseline | Allocated | Alloc Ratio |
    |-------------- |-----------:|----------:|----------:|---------:|---------:|-----------:|------:|--------:|--------- |----------:|------------:|
    | SomeMethod_V1 | 1,161.1 ms | 319.86 ms | 943.10 ms | 739.0 ms | 524.7 ms | 5,106.1 ms |  1.40 |    1.35 | Yes      |  12.68 MB |        1.00 |
    | SomeMethod_V2 |   498.8 ms |  18.31 ms |  53.99 ms | 469.1 ms | 442.1 ms |   684.2 ms |  0.60 |    0.25 | No       |  10.04 MB |        0.79 |

I expected `Ratio` to be `1.00` for `SomeMethod_V1`, which is the baseline (it is `1.00` for `Alloc Ratio`).

My benchmark code:

	[MemoryDiagnoser]
	[SimpleJob(RunStrategy.Monitoring, warmupCount: 1, iterationCount: 100)]
	[MinColumn, MaxColumn, MeanColumn, MedianColumn, BaselineColumn]
	public class SomeMethodBenchmarks
	{
		[GlobalSetup]
		public async Task GlobalSetup() { /* ... */ }

		[GlobalCleanup]
		public void GlobalCleanup() { /* ... */ }

		[Benchmark(Baseline = true)]
		public async Task SomeMethod_V1() { /* ... */ }

		[Benchmark]
		public async Task SomeMethod_V2() { /* ... */ }
	}

I'm using `BenchmarkDotNet 0.14.0` on .NET 8.0.8.

Am I doing something wrong, or is this a bug?

eNeRGy164:
I got one as well that surprised me a bit, not a big difference though.
Latest nightly, ratio `1.001`

![Screenshot 2024-09-26 213336](https://github.com/user-attachments/assets/77ffba62-f96f-496f-9eb4-34a840e479ce)

wild0ne:
[...]

PR #2988, opened 2026-01-30
fix summary ratio and ratioSD values for baseline #2648

wild0ne:
Ratio and RatioSD summary values are fixed for baseline (1.0 and 0.0, respectively)

### DIFF
--- a/tests/BenchmarkDotNet.Tests/Exporters/VerifiedFiles/MarkdownExporterVerifyTests.GroupExporterTest_Invalid_TwoJobBaselines.verified.txt
+++ b/tests/BenchmarkDotNet.Tests/Exporters/VerifiedFiles/MarkdownExporterVerifyTests.GroupExporterTest_Invalid_TwoJobBaselines.verified.txt
@@ -10,10 +10,10 @@ Frequency: 2531248 Hz, Resolution: 395.062 ns, Timer: TSC
 
  Method | Job  | Mean     | Error   | StdDev  | Ratio | RatioSD | Rank | LogicalGroup                                  | Baseline |
 ------- |----- |---------:|--------:|--------:|------:|--------:|-----:|---------------------------------------------- |--------- |
- Foo    | Job1 | 114.5 ns | 5.88 ns | 8.80 ns |  1.01 |    0.11 |    1 | DistinctParamSet0-Invalid_TwoJobBaselines.Foo | Yes      |
+ Foo    | Job1 | 114.5 ns | 5.88 ns | 8.80 ns |  1.00 |    0.00 |    1 | DistinctParamSet0-Invalid_TwoJobBaselines.Foo | Yes      |
  Foo    | Job2 | 314.5 ns | 5.88 ns | 8.80 ns |  2.76 |    0.22 |    2 | DistinctParamSet0-Invalid_TwoJobBaselines.Foo | Yes      |
         |      |          |         |         |       |         |      |                                               |          |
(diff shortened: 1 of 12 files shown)


wild0ne:
> @wild0ne please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

src/BenchmarkDotNet/Mathematics/RatioStatistics.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+            else
+            {
+                var divided = Divide(x, y);
+                Mean = divided.Mean;
+                StandardDeviation = divided.StandardDeviation;
+            }
+        }
+
+        private static Statistics Divide(Statistics x, Statistics y)
```
timcassell, line 35: You can just call `Statistics.Divide` rather than duplicate it.

wild0ne: There's no use of `Statistics.Divide `other than the ratio calculation

So I moved the whole implementation into another class

Yes, the code we are commenting on has been moved, not duplicated

wild0ne:
@AndreyAkinshin Hey Andrey! It'd be great to have your thoughts on this PR. Thanks!

AndreyAkinshin (approved):
This PR addresses a real user-facing problem — the baseline ratio deviating from 1.00 due to the all-pairs division approach — and doesn't introduce any regressions for non-baseline ratios. The fix is scoped and safe, so let's merge it.

**Status update on the broader problem:**

I'm working on a new statistical engine called [Pragmastat](https://pragmastat.dev). The latest version already includes `Ratio` and `RatioBounds` functions designed to replace the current BenchmarkDotNet ratio implementation. That implementation is quite flawed — when I originally wrote it, I wasn't aware of better solutions and picked an approach that felt reasonable at the time. After several years of research, I've arrived at a much better way to estimate ratios between benchmarks, which you can explore in the current Pragmastat.

The underlying research work is finished. What remains is implementing a few additional estimators and reworking the summary tables before it can be merged into BenchmarkDotNet. After that, we can completely drop the current ratio and statistics implementation, since it has multiple flaws and can be misleading in certain cases.

No ETA, but I hope to finish in the upcoming months. I've given similar estimates before, but this time I'm powered by an LLM-based coding agent, so progress has gotten a huge acceleration boost.

**Bottom line:** this PR is good to merge as-is. Further polishing of the current ratio logic isn't worth the investment — it will all be replaced by the Pragmastat-based engine.

AndreyAkinshin (approved):

Issue #2824, opened 2025-08-21
BenchmarkDotNet freezes when running multiple benchmark classes on .NET 10 (works fine on .NET 8)


RealDotNetDave:
# BenchmarkDotNet freezes when running multiple benchmark classes on .NET 10 (works fine on .NET 8)

## Description
After migrating my benchmark test suite to **.NET 10**, the benchmark run intermittently **freezes** mid-execution. The exact same suite works fine on **.NET 8**. This has been happening for months and I haven’t identified the cause.

## Observed Behavior
1. Usually occurs when running **multiple benchmark classes**, especially ones testing **collections**.
2. **No exception** is thrown — the run just **freezes** and nothing new is logged.
3. The **.NET host process** does not terminate after closing the console window.
   - CPU usage continues to fluctuate.
   - The benchmark output folder can’t be deleted until I kill the process (or reboot).
   - Occasionally, I need a full reboot to clear it.
4. Running **one benchmark class at a time** typically succeeds.

## Expected Behavior
All benchmark classes should run sequentially and the process should exit normally with a summary, logs, and unlocked files.

## Actual Behavior
When multiple benchmark classes run back-to-back, the process sometimes freezes indefinitely with no error output, leaving a lingering .NET host process and locked files.

## Suspected Cause
Something may not be released/cleaned up when **BenchmarkDotNet starts a new benchmark class**, particularly in collection-focused benchmarks.

## Reproduction Steps
1. Configure a BenchmarkDotNet job targeting .NET 10.
2. Run **two or more** benchmark classes **sequentially** in the same process.
3. Observe that execution sometimes **freezes** after one class completes.

## Configuration Snippet
```csharp

        var config = DefaultConfig.Instance
            .AddJob(Job.Default.WithRuntime(CoreRuntime.Core10_0))
            .WithSummaryStyle(SummaryStyle.Default.WithTimeUnit(BenchmarkDotNet.Horology.TimeUnit.Nanosecond));

        _ = config.WithOption(ConfigOptions.DisableOptimizationsValidator, true);

        // Running multiple classes back-to-back
        _ = BenchmarkRunner.Run<CollectionCreatingBenchmark>(config);
        _ = BenchmarkRunner.Run<CollectionCreatingLoopBenchmark>(config);
```

## Environment
- **BenchmarkDotNet version**: 0.15.2
- **.NET SDK version(s)**: 10.0.100-preview.7.25380.108
- **.NET runtimes tested**:
[...]

PR #2992, opened 2026-02-06
chore: modify broker related code

filzrev:
This PR intended to fix part of issue of #2824

#### 1. Add IDisposable interface ~~and remove OnProcessExited event handler~~

To ensure dispose operation is called synchronously.
Add IDisposable interface.

And modify OnProcessExited callback to dispose DisposeLocalCopyOfClientHandle.
It's required  when child process is spawned and exits quickly without `BeforeAnythingElse` signal.
Without this code,  `reader.ReadLine();` infinitely waiting, because anonymous pipe is not closed.

#### 2. Add Result enum
Add `Result` enum to return `ProcessDataBlocking` result for error logging.

#### 3. Modify Task.Run` code

On current implementation.
Exception that thrown inside ProcessDataBlocking  is silently ignored.

Because task is not awaited and `TaskScheduler.UnobservedTaskException` is not registerd.
So when exception thrown, it cause benchmark freeze at `finished.WaitOne` line. 

This PR move `finished.Set` code to try-finally block.
To ensure `finished.WaitOne` is signaled on unexpected exception thrown.

And add `GetAwaiter().GetResult()` code to wait task completion to throw exception.

#### 4. ProcessDataBlocking code changes

- Modify code to return `Result` enum.
- Replace `else if` blocks to if-continue to improve code readability.
- Add null check logic of ReadLine() results when parsing InProcessDiagnoserResults line.
- Add line.StartsWith check to avoid ArgumentOutOfRangeException when getting substring.
  (As far as I've confirmed on original issue's logs. It seems there is cases that incomplete log line is returned from ReadLine())

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliExecutor.cs
+++ b/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliExecutor.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Pipes;
@@ -79,12 +80,14 @@ namespace BenchmarkDotNet.Toolchains.DotNetCli
 
             startInfo.SetEnvironmentVariables(benchmarkCase, resolver);
 
-            using (Process process = new() { StartInfo = startInfo })
-            using (ConsoleExitHandler consoleExitHandler = new(process, logger))
-            using (AsyncProcessOutputReader processOutputReader = new(process, logOutput: true, logger, readStandardError: false))
-            {
-                using Broker broker = new(logger, process, diagnoser, compositeInProcessDiagnoser, benchmarkCase, benchmarkId, inputFromBenchmark, acknowledgments);
+            using Process process = new() { StartInfo = startInfo };
+            using ConsoleExitHandler consoleExitHandler = new(process, logger);
+            using AsyncProcessOutputReader processOutputReader = new(process, logOutput: true, logger, readStandardError: false);
 
+            List<string> results;
+            List<string> prefixedOutput;
+            using (Broker broker = new(logger, process, diagnoser, compositeInProcessDiagnoser, benchmarkCase, benchmarkId, inputFromBenchmark, acknowledgments))
+            {
                 logger.WriteLineInfo($"// Execute: {process.StartInfo.FileName} {process.StartInfo.Arguments} in {process.StartInfo.WorkingDirectory}");
 
                 diagnoser?.Handle(HostSignal.BeforeProcessStart, broker.DiagnoserActionParameters);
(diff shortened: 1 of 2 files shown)


filzrev:
I've removed redundant `Task.Run` and `ManualResetEvent`.

It’s remained to preserve the original code behavior. (Run ProcessDataBlocking on worker thread)
But it seems redundant because it block caller thread (Main Thread) in any case.

> big thanks for taking care of this edge case bug!

[Original code](https://github.com/dotnet/BenchmarkDotNet/blob/03147e982dd924715ca5f44e6af42e6b5f48aca4/src/BenchmarkDotNet/Loggers/Broker.cs#L70-L74) contains OnProcessExited callback already and disposing client handles inside OnProcessExited .

So expected edge case is following
- OnProcessExited **asynchronous** callback is not called because `Process.Dispose` is invoked before callback.

src/BenchmarkDotNet/Loggers/Broker.cs
```diff
@@ -102,17 +113,27 @@ private void ProcessDataBlocking()
                     var resultsStringBuilder = new StringBuilder();
                     for (int i = 0; i < resultsLinesCount;)
                     {
+                        line = reader.ReadLine();
+                        if (line == null)
+                            return Result.EndOfStream;
+
+                        if (!line.StartsWith("// InProcessDiagnoserResults "))
```
timcassell:
```suggestion
                        if (!line.StartsWith($"{CompositeInProcessDiagnoser.ResultsKey} "))
```

src/BenchmarkDotNet/Loggers/Broker.cs
```diff
@@ -46,6 +49,20 @@ public Broker(ILogger logger, Process process, IDiagnoser? diagnoser, CompositeI
 
         internal List<string> PrefixedOutput { get; } = [];
 
+        public void Dispose()
+        {
+            // Dispose all the pipes to let reading from pipe finish with EOF and avoid a reasource leak.
```
timcassell:
```suggestion
            // Dispose all the pipes to let reading from pipe finish with EOF and avoid a resource leak.
```

src/BenchmarkDotNet/Loggers/Broker.cs
```diff
@@ -32,10 +36,9 @@ public Broker(ILogger logger, Process process, IDiagnoser? diagnoser, CompositeI
             this.inputFromBenchmark = inputFromBenchmark;
             this.acknowledgments = acknowledgments;
             DiagnoserActionParameters = new DiagnoserActionParameters(process, benchmarkCase, benchmarkId);
-            finished = new ManualResetEvent(false);
 
             process.EnableRaisingEvents = true;
-            process.Exited += OnProcessExited;
+            process.Exited += (_, _) => DisposeLocalCopyOfClientHandles();
```
timcassell: The exited handler should probably be removed when the broker is disposed.

(Should the broker also be disposed before the wait for process exit?)

filzrev:
It seems PR #2958 including `AnonymousPipe` to `NamedPipe` migration.
NamedPipe supporting async operations and more reliable than AnonymousPipe.

So it's not needed if #2958 is expected to be merged  in the near future.
In this case following changes are expected o be merged to PR. 

1. Run dispose operation synchronously by IDisposable with following reasons  
  OnProcessExited handler is called asynchronously and it might cause race conditions.   
  OnProcessExited handler is not called when process instance is disposed.
 2. null/length check when parsing `InProcessDiagnoserResults` line (It's expected to be happens pipe is closed during processing)

timcassell:
It's fine to merge this first. I still have some work to do in that PR, and I can fix the merge conflicts with this.

timcassell:
The change to named pipe in #2958 is showing freezes on Windows CI just like the issue this is trying to fix. I don't think it's a cause, but it does make it more repro-able. I'm going to go ahead and merge this then update that branch with these changes and see if it helps.

Issue #2947, opened 2025-12-31
Drop extremely outdated TFMs


Youssef1313:
As of today, it looks like BDN supports back to netcoreapp2.0 which is out of support since October 2018, that's **7 years**, and can be very limiting. See https://github.com/Youssef1313/YTest.MTP.XUnit2/issues/12 for example.

timcassell:
BDN targets netstandard2.0, netcoreapp2.0 support is simply a by-product of that because it supports the tfm. We don't explicitly target core tfms older than net6.0. So I'm not really sure what the ask is here.

Youssef1313:
@timcassell There are test projects that explicitly target old TFMs like netcoreapp2.0.

https://github.com/dotnet/BenchmarkDotNet/blob/7bd81aa510439aee24888ce087f3a834094c31f4/tests/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks/BenchmarkDotNet.IntegrationTests.ManualRunning.MultipleFrameworks.csproj#L6

Plus some production projects also target old TFMs like netcoreapp3.1:

https://github.com/dotnet/BenchmarkDotNet/blob/7bd81aa510439aee24888ce087f3a834094c31f4/src/BenchmarkDotNet.Diagnostics.dotMemory/BenchmarkDotNet.Diagnostics.dotMemory.csproj#L4

https://github.com/dotnet/BenchmarkDotNet/blob/7bd81aa510439aee24888ce087f3a834094c31f4/src/BenchmarkDotNet.Diagnostics.dotTrace/BenchmarkDotNet.Diagnostics.dotTrace.csproj#L4

timcassell:
The test project is inconsequential to release packages, it's there just to make sure we don't accidentally break any runtimes that actually support netstandard2.0 (and we don't even run it in CI, it's a manual test).

As for the dotMemory and dotTrace packages, I'm not sure, but in general the philosophy here is we don't explicitly remove support for runtimes unless there's a very good reason to. cc @AndreyAkinshin

@filzrev previously mentioned migrating from XUnit to TUnit that actually supports netstandard2.0, which I think is reasonable and should negate the need for the linked issue.

Youssef1313:
> The test project is inconsequential to release packages, it's there just to make sure we don't accidentally break any runtimes that actually support netstandard2.0 (and we don't even run it in CI, it's a manual test).

[...]

PR #2996, opened 2026-02-10
Update ManualRunning.MultipleFrameworks TFMs

Youssef1313:
Part of https://github.com/dotnet/BenchmarkDotNet/issues/2947#issuecomment-3874812895

> - net461 has known issues with netstandard2.0 consumption — Microsoft themselves recommend net462 as the practical minimum (and ideally net472+).
> - netcoreapp2.0 has been EOL for 7+ years. Since this test project doesn't run in CI, keeping it there creates an illusion of coverage rather than actual validation.


timcassell:
With netcoreapp2.0 removed, you should add net10.0 to replace it, and update the `MultipleFrameworksTest`.

PR #2998, opened 2026-02-11
Wasm tests use `Microsoft.NET.Sdk.WebAssembly` and support CoreClr runtime

ilonatommy:
## Problem

The autogenerated benchmark project uses Microsoft.NET.Sdk, which does not set UseMonoRuntime=true. This causes ProcessFrameworkReferences to resolve the CoreCLR runtime pack (Microsoft.NETCore.App.Runtime.browser-wasm) instead of the Mono one (Microsoft.NETCore.App.Runtime.Mono.browser-wasm). The CoreCLR pack does not exist for browser-wasm, so the build fails.

PR #2994 worked around this by adding -r browser-wasm to the restore command, but this only defers the error — the runtime pack is still not resolved correctly during build.

## Fix

Switch the default SDK to `Microsoft.NET.Sdk.WebAssembly`, which auto-defaults `UseMonoRuntime=true` and resolves the correct Mono runtime pack. Add `IsMonoRuntime` flag to WasmRuntime so users can opt into the CoreCLR runtime pack (`--wasmCoreCLR` CLI flag), which falls back to `Microsoft.NET.Sdk`.

## Changes
- SDK switch: Default to Microsoft.NET.Sdk.WebAssembly (Mono). When IsMonoRuntime=false, use Microsoft.NET.Sdk (CoreCLR).
- --wasmCoreCLR CLI flag: New option to select the CoreCLR runtime pack for WASM benchmarks.
- DotNetCliPublisher: WASM toolchain now uses DotNetCliPublisher instead of DotNetCliBuilder, matching the template's DefaultTargets="publish" and eliminating the ForcedNoDependencies workaround.
- ArtifactsPath fix: Point ArtifactsPath to a .artifacts/ subdirectory so DefaultItemExcludes ($(ArtifactsPath)/**) doesn't exclude wwwroot/ from Content globs.
- Embed benchmark-main.mjs: Entry point is now an embedded template resource placed into wwwroot/ during generation, replacing the AppBundle/test-main.js test asset.
- MainJS constant: Remove dead net5.0/net6.0 code path.
- Remove: -r browser-wasm from restore, WasmGenerateRunV8Script, WasmMainJSPath dependency.

## Testing 
Integration tests: 4/4 WasmTests pass (net8.0, AOT + Interpreter × V8 + JSC)
benchmarks_ci.py: 5/5 Perf_Boolean.Parse benchmarks pass on Wasm AOT


maraf:
> The failure is caused by the fact that autogenerated project uses Microsoft.NET.Sdk and thus, keeps UseMonoRuntime=false by default. As a result, restore tries to download clr runtime, not mono runtime.

Instead of explicitly passing UseMonoRuntime, we should disable runtime/targeting pack download, similar to https://github.com/dotnet/runtime/pull/124288

ilonatommy:
> Instead of explicitly passing UseMonoRuntime, we should disable runtime/targeting pack download, similar to [dotnet/runtime#124288](https://github.com/dotnet/runtime/pull/124288)

We don't pass it explicitly. We decided to change the sdk type instead, to change the value UseMonoRuntime defaults to.

src/BenchmarkDotNet/Templates/WasmCsProj.txt
```diff
@@ -32,8 +30,10 @@
 
   <ItemGroup>
     <Compile Include="$CODEFILENAME$" Exclude="bin\**;obj\**;**\*.xproj;packages\**" />
-    <!-- move this too? -->
     <TrimmerRootDescriptor Include="WasmLinkerDescription.xml" Condition="'$(RunAOTCompilation)' == 'true'" />
+    <!-- Static Web Assets only physically copies wwwroot/ content during publish.
+         Since BDN uses dotnet build, explicitly copy benchmark-main.mjs to the output. -->
+    <None Include="wwwroot\**" CopyToOutputDirectory="PreserveNewest" LinkBase="wwwroot" />
```
maraf: These are already part of Content. We can just CopyToOutputDirectory on Content items

tests/BenchmarkDotNet.IntegrationTests/WasmTests.cs
```diff
@@ -33,7 +34,11 @@ private ManualConfig GetConfig(MonoAotCompilerMode aotCompilerMode)
             var dotnetVersion = "net8.0";
             var logger = new OutputLogger(Output);
             var netCoreAppSettings = new NetCoreAppSettings(dotnetVersion, null, "Wasm", aotCompilerMode: aotCompilerMode);
-            var mainJsPath = Path.Combine(AppContext.BaseDirectory, "AppBundle", "test-main.js");
+            const string mainJsName = "benchmark-main.mjs";
+            var mainJsDir = Path.Combine(Path.GetTempPath(), "BenchmarkDotNet.IntegrationTests");
+            Directory.CreateDirectory(mainJsDir);
+            var mainJsPath = Path.Combine(mainJsDir, mainJsName);
+            File.WriteAllText(mainJsPath, ResourceHelper.LoadTemplate(mainJsName));
```
maraf: Can you please explain why we do it?

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -46,7 +46,9 @@ protected void GenerateProjectFile(BuildPartition buildPartition, ArtifactsPaths
 
             var xmlDoc = new XmlDocument();
             xmlDoc.Load(projectFile.FullName);
-            var (customProperties, sdkName) = GetSettingsThatNeedToBeCopied(xmlDoc, projectFile);
+            var (customProperties, _) = GetSettingsThatNeedToBeCopied(xmlDoc, projectFile);
+            // Always use Microsoft.NET.Sdk.WebAssembly for WASM projects. It auto-defaults UseMonoRuntime=true.
```
pavelsavara: Does CoreCLR also work already (without re-link) ?

tests/BenchmarkDotNet.IntegrationTests/WasmTests.cs
ilonatommy: Right, this is no longer needed because `WasmGenerator` now puts the file directly in `wwwroot`

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
ilonatommy: To have CoreCLR supported we could extend `WasmRuntime` class and allow overwriting `UseMonoRuntime` in the generated project. Do you want me to make it supported already?

pavelsavara: It would be great!

Maybe there would be trouble with `<WasmBuildNative>true</WasmBuildNative>` in `src/benchmarks/micro/MicroBenchmarks.Wasm.props`. I hope that is gracefully ignored for now.

maraf: WasmBuildNative is ignored in CoreCLR

tests/BenchmarkDotNet.IntegrationTests/WasmTests.cs
ilonatommy: It looks like it's more complex. `ForcedNoDependenciesForIntegrationTests` passes `excludeOutput: true` which makes our `/p:OutDir` for integration tests ignored and build output goes to `bin/Release/net8.0/` instead of `bin/Release/net8.0/browser-wasm` and this change was a workaround for it. We can remove the exclusion for wasm.

timcassell: Feel free to update `ForcedNoDependenciesForIntegrationTests` to exclude wasm.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -135,21 +136,25 @@ public DotNetCliCommandResult PublishNoRestore()
...
                 .AppendArgument(GetMsBuildBinLogArgument(buildPartition, binLogSuffix))
-                .MaybeAppendOutputPaths(artifactsPaths, true, excludeOutput)
+                .MaybeAppendOutputPaths(artifactsPaths, isRestore: true, excludeOutput: excludeOutput, excludeArtifactsPath: excludeArtifactsPath)
                 .ToString();
+        }
 
         internal static string GetBuildCommand(ArtifactsPaths artifactsPaths, BuildPartition buildPartition, string filePath, string tfm, string? extraArguments = null, string? binLogSuffix = null, bool excludeOutput = false)
-            => new StringBuilder()
+        {
+            bool excludeArtifactsPath = buildPartition.RepresentativeBenchmarkCase.Job.GetToolchain() is WasmToolchain;
```
timcassell: Better to pass it in as an optional argument.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -216,12 +225,17 @@ internal static class DotNetCliCommandExtensions
...
                 ? stringBuilder
                 : stringBuilder
                     // Use AltDirectorySeparatorChar so it's not interpreted as an escaped quote `\"`.
-                    .AppendArgument($"/p:ArtifactsPath=\"{artifactsPaths.BuildArtifactsDirectoryPath}{Path.AltDirectorySeparatorChar}\"")
+                    // When excludeArtifactsPath is true, we skip /p:ArtifactsPath and set /p:UseArtifactsOutput=false instead.
+                    // This prevents the SDK from adding $(ArtifactsPath)/** to DefaultItemExcludes, which would
+                    // exclude all project files (including wwwroot/) from default Content globs.
+                    .AppendArgument(excludeArtifactsPath
+                        ? "/p:UseArtifactsOutput=false"
+                        : $"/p:ArtifactsPath=\"{artifactsPaths.BuildArtifactsDirectoryPath}{Path.AltDirectorySeparatorChar}\"")
```
timcassell: We use artifacts path so that builds are safe to run in parallel. Without it, the build must be ran sequentially. If this is truly required, you'll need to update https://github.com/dotnet/BenchmarkDotNet/blob/c342c49f1b68d3e5e04bc05c5ad12b0e9140a7b1/src/BenchmarkDotNet/Running/BenchmarkRunnerClean.cs#L90-L97

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmGenerator.cs
```diff
@@ -19,7 +19,7 @@ public WasmGenerator(string targetFrameworkMoniker, string cliPath, string packa
             : base(targetFrameworkMoniker, cliPath, packagesPath, runtimeFrameworkVersion: null)
         {
             CustomRuntimePack = customRuntimePack;
-            MainJS = (targetFrameworkMoniker == "net5.0" || targetFrameworkMoniker == "net6.0") ? "main.js" : "test-main.js";
+            MainJS = (targetFrameworkMoniker == "net5.0" || targetFrameworkMoniker == "net6.0") ? "main.js" : "benchmark-main.mjs";
```
timcassell: It looks like the net5.0 and net6.0 path can be removed now, and the `MainJS` field can be made a constant.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
ilonatommy: I see. Setting `/p:ArtifactsPath` causes the SDK to add `$(ArtifactsPath)/**` to `DefaultItemExcludes`. That excludes all project files (including `wwwroot`) from the WebAssembly SDK's default `Content Include="wwwroot\**"` set, preventing `benchmark-main.mjs` and the `_framework` static web assets from being copied to the build output.

Code for adding `$(ArtifactsPath)/**` to `DefaultItemExcludes comes from SDK, we cannot work it around.

@maraf, the only workaround would be going back to the original remove+add instead of update: https://github.com/dotnet/BenchmarkDotNet/pull/2998#discussion_r2802793524. That is a build time operation, we don't measure it in the benchmarks. Do you insist on keeping update, even if the job would be running longer, sequentially?

src/BenchmarkDotNet/Templates/WasmCsProj.txt
timcassell: If it makes sense for the toolchain, you can also change it to use `DotNetCliPublisher` instead of `DotNetCliBuilder`.

ilonatommy: The `DefaultItemExcludes` containing `ArtifactsPath` is same for build and publish. Using publish solves only the issue we discussed here: https://github.com/dotnet/BenchmarkDotNet/pull/2998#discussion_r2803553850 that was fixed by excluding wasm in `ForcedNoDependenciesForIntegrationTests`. Which is, in fact, a cleaner approach and I would like to apply it.

I figured out something else. If we point `ArtifactsPath` to a `./artifacts` subdirectory then `DefaultItemExcludes` won't contain project-level files and `Content` would preserve `wwwroot` dir. They don't conflict then. I will commit this approach, unless there are any concerns about changing `ArtifactsPath` construction for all the apps.

timcassell: It sounds reasonable. You will need to change all the output paths to include it, though.

timcassell:
Did these changes break `WasmNet50`-`WasmNet70`? We don't have any CI coverage of those, and copilot is saying that the new mjs template is incompatible with the older tfms. If that's the case, I don't think it's worth continuing to support them, and we should probably just remove those monikers.

ilonatommy:
> Did these changes break `WasmNet50`-`WasmNet70`? We don't have any CI coverage of those, and copilot is saying that the new mjs template is incompatible with the older tfms. If that's the case, I don't think it's worth continuing to support them, and we should probably just remove those monikers.

What are the failures, where can I see them? I am not sure about the theory of no support. https://github.com/dotnet/runtime/pull/61313, https://github.com/dotnet/runtime/pull/62292 are in .net 7 and imply modules were working then already.

ilonatommy:
Sorry, I misunderstood. It could have broken `WasmNet50`-`WasmNet60` but not `WasmNet70`. Regardless of that, removing monikers for `WasmNet50`-`WasmNet70` sounds reasonable.

PR #3001, opened 2026-02-12
chore: Add .NET 10 tfm to analyzers tests project

filzrev:
This PR add `net10.0` tfm to analyzer tests projects.

And to resolve following Roslyn version conflict issue.
I've added settings for .NET 10 tfm. (It's same code that is removed and #2913)
```
Diagnostics:
// error CS1705: Assembly 'BenchmarkDotNet' with identity 'BenchmarkDotNet, Version=0.16.0.0, Culture=neutral, PublicKeyToken=aa0ca2f9092cefc4' uses 'System.Runtime, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Runtime' with identity 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```

### DIFF
--- a/build/BenchmarkDotNet.Build/Runners/BuildRunner.cs
+++ b/build/BenchmarkDotNet.Build/Runners/BuildRunner.cs
@@ -128,7 +128,7 @@ public class BuildRunner
     public void BuildAnalyzers()
     {
         context.Information("BuildSystemProvider: " + context.BuildSystem().Provider);
-        string[] mccVersions = ["2.8", "3.8", "4.8", "5.0"];
+        string[] mccVersions = ["2.8", "3.8", "4.8"];
         foreach (string version in mccVersions)
         {
             context.DotNetBuild(context.AnalyzersProjectFile.FullPath, new DotNetBuildSettings


build/BenchmarkDotNet.Build/Runners/BuildRunner.cs
```diff
@@ -128,7 +128,7 @@ public void BuildProjectSilent(FilePath projectFile)
     public void BuildAnalyzers()
     {
         context.Information("BuildSystemProvider: " + context.BuildSystem().Provider);
-        string[] mccVersions = ["2.8", "3.8", "4.8"];
+        string[] mccVersions = ["2.8", "3.8", "4.8", "5.0"];
```
timcassell: These are the versions that get packed into the nuget package. I think this is not necessary since we're not analyzing C# 14 features.

filzrev: I'll try to remove this setting and confirm it pass CI or not later.

filzrev: CI succeeded after reverting above line.
It seems not needed until C#14 specific rules need to be added.

PR #3003, opened 2026-02-12
chore: Enable nullable annotations on BenchmarkDotNet project

filzrev:
This PR fix a part of issue https://github.com/dotnet/BenchmarkDotNet/issues/2815

By enabling #nullable enable and fix nullable errors.

### DIFF
--- a/src/BenchmarkDotNet/Reports/Summary.cs
+++ b/src/BenchmarkDotNet/Reports/Summary.cs
@@ -163,12 +163,7 @@ namespace BenchmarkDotNet.Reports
 
         private static SummaryStyle GetConfiguredSummaryStyleOrDefaultOne(ImmutableArray<BenchmarkCase> benchmarkCases)
             => benchmarkCases
-                   .Where(benchmark => benchmark.Config.SummaryStyle != SummaryStyle.Default
-#nullable disable
-                                       // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract ConditionIsAlwaysTrueOrFalse
-                                       // TODO: remove this check once the nullability migration is finished
-                                       && benchmark.Config.SummaryStyle != null) // Paranoid
-#nullable enable
+                   .Where(benchmark => benchmark.Config.SummaryStyle != SummaryStyle.Default)
                    .Select(benchmark => benchmark.Config.SummaryStyle)
                    .Distinct()
                    .FirstOrDefault()


src/BenchmarkDotNet/Extensions/StringAndTextExtensions.cs
```diff
@@ -151,5 +151,14 @@ internal static StringBuilder AppendArgument(this StringBuilder stringBuilder, o
...
         public static bool IsNotBlank([NotNullWhen(true)] this string? value) => !value.IsBlank();
+
+        /// <summary>
+        /// Helper method to ensure not null value.
+        /// It's expected to be used for public API parameters handling.
+        /// </summary>
+        public static string EnsureNotNull(this string? value)
+        {
+            return value ?? "";
+        }
```
filzrev, line 160: This helper method is used for `[PublicAPI]` related parameter.
When caller is not nullable aware. `string` type parameter can be null.

src/BenchmarkDotNet/Running/BuildPartition.cs
```diff
@@ -35,6 +37,14 @@ public BuildPartition(BenchmarkBuildInfo[] benchmarks, IResolver resolver)
             GenerateMSBuildBinLog = benchmarks[0].Config.Options.IsSet(ConfigOptions.GenerateMSBuildBinLog);
         }
 
+        private BuildPartition()
+        {
+            Resolver = default!;
+            RepresentativeBenchmarkCase = default!;
+            Benchmarks = [];
+            ProgramName = default!;
+        }
```
filzrev, line 44: These properties are not accessed when using `BuildPartition.Empty`

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliPublisher.cs
```diff
@@ -1,26 +1,43 @@
...
+    public string TargetFrameworkMoniker { get; }
+    public string CustomDotNetCliPath { get; }
+    public string ExtraArguments { get; }
+    public IReadOnlyList<EnvironmentVariable>? EnvironmentVariables { get; }
+
+    public DotNetCliPublisher(
+        string tfm,
+        string customDotNetCliPath = "",
+        string extraArguments = "",
+        IReadOnlyList<EnvironmentVariable>? environmentVariables = null)
```
filzrev: It can also remove nullable for collection.
It'll be modified on another PR.

src/BenchmarkDotNet/Running/BenchmarkCase.cs
```diff
@@ -26,7 +26,7 @@ internal BenchmarkCase(Descriptor descriptor, Job job, ParameterInstances parame
         {
             Descriptor = descriptor;
             Job = job;
-            Parameters = parameters;
+            Parameters = parameters ?? ParameterInstances.Empty;
```
filzrev, line 27: Replace null to EmptyParameterInstances.
To support existing non-null aware test projects.

src/BenchmarkDotNet/Detectors/Cpu/Windows/PowershellWmiCpuDetector.cs
```diff
@@ -8,6 +8,8 @@
 using BenchmarkDotNet.Helpers;
 using Perfolizer.Models;
 
+#nullable enable
```
timcassell: I don't think adding this to files is necessary anymore since you added it to the csproj, right?

filzrev: `<Nullable>enable</Nullable>` is expected to be enabled on `common.props` settings.
https://github.com/dotnet/BenchmarkDotNet/blob/master/build/common.props#L26

After nullable annotations are enabled on all projects.
`#nullable enable` and project-level `<Nullable>` setting is expected to be removed.

timcassell: Sure, but all these files are under the BenchmarkDotNet.csproj that you added it to, so you can already remove them from the cs files rather than add more.

filzrev: OK, I'll add commit to remove `#nullable enable` in this PR.

timcassell:
Need to fix merge conflicts.

PR #3007, opened 2026-02-14
1802 - Disabling overhead calculation

### LINKED ISSUE #1802: Inaccurate results reported for small methods
Unless doing instruction level profiling, the highest precision timer on a modern computer is about `25-32 cycles` (which is, under ideal circumstances, about `6ns` on a 5GHz processor at best and `32ns` on a 1GHz processor).

Due to platform specific differences, the maximum reported difference in the high precision timer APIs exposed by the OS is about `100ns`. Additionally, it is well documented that due to the latency between calls and other factors on the OS or hardware, the latency for such a call can be much worse, such as closer to `300ns` when a CPU level timer such as `RDTSC` is not available: https://docs.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps#resolution-precision-accuracy-and-stability.

While Benchmark.NET does try to account for small methods and while it also tries to account for noise due to call overhead and the like, there are many cases where the numbers it reports are of questionable accuracy.

One such example is the following:
image

In particular, if we look at the first entry `GetShortName_opt` is reporting a time of `0.2082 ns`. Even in an "ideal" scenario where the JIT is able to fully optimize the comparison against a constant value and optimize it to simple be `xor rax, rax`, this is still reporting that it takes approximately 1 cycle on a 5GHz CPU.
* It also shouldn't be able to optimize it like this. AFAIR, Benchmark.NET should be passing the value in and preventing the actual benchmark body from being inlined to avoid such issues.

It would be beneficial, IMO, if Benchmark.NET was more proactive about labeling potentially problematic results and had guidance on how to optimally write a test in a way that will provide accurate results.
* I would view a problematic result, at the very least, as anything taking less than 10ns. Most of these methods should be testing more than a single instruction and are running on 2-4GHz computers. So in an "ideal" environment, 10ns represents no more than 20 instructions and likely no memory accesses. Very few instructions take 0 cycles. Several take 1 cycle and can be pipelined for up to 4 to be in simultaneous dispatch, but its rare to actually have this. Many take 2-3 cycles and if you have any kind of memory access they will take about 3-11 cycles in the fastest scenario (potentially longer for uncached results among other things).

ladeak:
- Overhead evaluation is disabled by default, but the corresponding code is not yet removed.
- Obsolete attribute added to WithEvaluateOverhead

Closes #1802

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/EngineTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/EngineTests.cs
@@ -62,10 +62,7 @@ namespace BenchmarkDotNet.IntegrationTests
                 Assert.True(count > 0, $"AllMeasurements don't contain {mode}{stage}");
             }
 
-            Check(IterationMode.Overhead, IterationStage.Jitting);
             Check(IterationMode.Workload, IterationStage.Jitting);
-            Check(IterationMode.Overhead, IterationStage.Warmup);
-            Check(IterationMode.Overhead, IterationStage.Actual);
             Check(IterationMode.Workload, IterationStage.Warmup);
             Check(IterationMode.Workload, IterationStage.Actual);
             Check(IterationMode.Workload, IterationStage.Result);


src/BenchmarkDotNet/Attributes/Mutators/EvaluateOverheadAttribute.cs
```diff
@@ -10,7 +10,9 @@ namespace BenchmarkDotNet.Attributes
     [PublicAPI]
     public class EvaluateOverheadAttribute : JobMutatorConfigBaseAttribute
```
timcassell, line 13: This should also be deprecated (you can remove the pragma when that's done).

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
```diff
@@ -176,10 +176,12 @@ private IEnumerable<Summary> ApplesToApples(ImmutableConfig effectiveConfig, IRe
                 return Array.Empty<Summary>();
             }
 
+#pragma warning disable CS0618 // WithEvaluateOverhead is obsolete
```
timcassell, line 179: Only put the pragmas around the line with the obsolete call.

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
```diff
@@ -204,6 +206,7 @@ private IEnumerable<Summary> ApplesToApples(ImmutableConfig effectiveConfig, IRe
 
             int iterationCount = baselineJob.Run.IterationCount;
             BenchmarkRunInfo[] benchmarksWithoutInvocationCount = TypeFilter.Filter(effectiveConfig, benchmarksToFilter);
+#pragma warning disable CS0618 // WithEvaluateOverhead is obsolete
```
timcassell, line 209: Same

src/BenchmarkDotNet/Jobs/JobExtensions.cs
```diff
@@ -301,8 +301,10 @@ public static Job WithMsBuildArguments(this Job job, params string[] msBuildArgu
 
         /// <summary>
         /// Specifies if the overhead should be evaluated (Idle runs) and it's average value subtracted from every result.
-        /// True by default, very important for nano-benchmarks.
+        /// False by default.
+        /// <remarks>WithEvaluateOverhead is deprecated.</remarks>
         /// </summary>
+        [Obsolete("WithEvaluateOverhead is deprecated by #1802. Use solutions outlined in #654 instead.")]
```
timcassell:
```suggestion
        [Obsolete("Overhead evaluation affects the accuracy of benchmark results. This will be removed soon.")]
```

timcassell (changes requested):

src/BenchmarkDotNet/Attributes/Mutators/EvaluateOverheadAttribute.cs
ladeak: I will do so.

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
ladeak: I don't think I can do that in this fluent syntax (it was my intention as you suggest here):

image

Would you prefer to refactor the syntax? Or is there another trick?

src/BenchmarkDotNet/Attributes/Mutators/EvaluateOverheadAttribute.cs
ladeak: Updated, also added a few more pragma warnings for the tests failing, so it should build correctly.

src/BenchmarkDotNet/Running/BenchmarkSwitcher.cs
timcassell: Oh that's annoying, I thought it worked. It's fine then.

ladeak:
Addressed the failing tests.

PR #3010, opened 2026-02-17
Remove unsupported wasm monikers

timcassell:
Follow-up to #2998


filzrev:
It seems [IntroWasm.cs](https://github.com/dotnet/BenchmarkDotNet/blob/be37238a4930ebe20f7414d4a8d42487495e4c5c/samples/BenchmarkDotNet.Samples/IntroWasm.cs#L32-L45) samples need to be updated.
Because it's targetting net5.0 wasm.

I'm trying to update samples. but it seems requires Linux machine. (not works on Windows)

PR #3013, opened 2026-02-19
chore: Enable nullable on build/common.props

### LINKED ISSUE #2815: [Feature Request] Enable nullable type checks

filzrev:
This PR intended to fix #2815.

**What's changed in this PR**
- Add nullable setting on `build/common.props`
- Remove project-level nullable setting from projects and from `tests/Directory.Build.props`
- Remove `#nullable enable` settings from source code.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -23,7 +23,7 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodingStyle.ruleset</CodeAnalysisRuleSet>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsNotAsErrors>NU1900</WarningsNotAsErrors>
-    <Nullable Condition="'$(Nullable)' == ''">enable</Nullable>
+    <Nullable>enable</Nullable>
     <!-- Suppress warning for nuget package used in old (unsupported) tfm. -->
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
   </PropertyGroup>


build/common.props
```diff
@@ -23,7 +23,7 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodingStyle.ruleset</CodeAnalysisRuleSet>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsNotAsErrors>NU1900</WarningsNotAsErrors>
-    <Nullable Condition="'$(Nullable)' == ''">annotations</Nullable>
+    <Nullable Condition="'$(Nullable)' == ''">enable</Nullable>
```
timcassell: Is the condition still necessary?

timcassell:
Thanks @filzrev

PR #3019, opened 2026-02-22
docs: Update `docs/articles/guides/console-args.md`

filzrev:
This PR update [console-args.md](https://github.com/dotnet/BenchmarkDotNet/blob/f54f764266722c6eba4e6f6ded8c808a4012a34e/docs/articles/guides/console-args.md) file content based on current BenchmarkDotNet `--help` output.

**What's changed in this PR**
1. Add `## Example Usages` section.
2. Update `## More` section .

Note: Following section is preserved.

> Note: the docs that you are currently reading might get outdated, to get the most up-to-date info about supported console arguments run the benchmarks with --help.

It can be removed after this markdown generation is automated. and relating unit tests are  are added in future.

### DIFF
--- a/docs/articles/guides/console-args.md
+++ b/docs/articles/guides/console-args.md
@@ -217,41 +217,41 @@ dotnet run -c Release -- --filter * --runtimes net6.0 net8.0 --statisticalTest 5
 ## Example Usages
 
 * Use Job.ShortRun for running the benchmarks:  
-    `-d -j short`
+    `-j short`
 * Run benchmarks in process:  
-    `-d -i`
+    `-i`
 * Run benchmarks for .NET 4.7.2, .NET 8.0 and Mono. .NET 4.7.2 will be baseline because it was first.:
-    `--disasm --runtimes net472 net8.0 Mono`
-* Run benchmarks for .NET Core 3.1, .NET 6.0 and .NET 8.0. .NET Core 3.1 will be baseline because it was first.:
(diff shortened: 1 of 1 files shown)


docs/articles/guides/console-args.md
```diff
@@ -214,37 +214,114 @@ Example: run Mann–Whitney U test with relative ratio of 5% for all benchmarks
 dotnet run -c Release -- --filter * --runtimes net6.0 net8.0 --statisticalTest 5%
 ```
 
+## Example Usages
+
+* Use Job.ShortRun for running the benchmarks:  
+    `-d -j short`
```
timcassell: Is this what --help says? It's a bit confusing. I think `-d`/`--disasm` should be removed on all of these that aren't disassembly examples.

filzrev:
> Is this what --help says? It's a bit confusing. I think -d/--disasm should be removed on all of these that aren't disassembly examples.

Currently `-d/--disasm` option are shown on help messages.
Because following check returns `true` when initialized with `new CommandLineOptions` constructor.
(It's not occurred when created by CommandLineParser)

https://github.com/dotnet/BenchmarkDotNet/blame/95b9776dccad5ca0e4e5b6767d34e49d5425d7e7/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs#L47 

I've removed unneeded options from example usages.
Additionally modify example target frameworks to currently supported versions.

PR #3025, opened 2026-02-25
chore: Resolve CA1510 analyzer messages

filzrev:
This PR intended to suppress following CA1510 analyzer messages.

> CA1510 'ArgumentNullException.ThrowIfNull' instead of explicitly throwing a new exception instance

**What's changed in this PR**

1. Add following polyfills for `net6.0`/`netstandard2.0` by using C# 14 extension member syntax. 
1.1. ArgumentException.ThrowIfEmpty/ThrowIfWhitespace
1.2. ArgumentNullException.ThrowIfNull
1.3. `Enum.Parse<T>`/ `GetNames<T>()`/`GetValues<T>`
1.4. Add MethodInfo extensions.
3. Cleanup CA1510 messages. And remove redundant null checks.

### DIFF
--- a/src/BenchmarkDotNet/Extensions/Polyfills/EnumExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/Polyfills/EnumExtensions.cs
@@ -1,4 +1,4 @@
-﻿#if NETFRAMEWORK || NETSTANDARD2_0
+﻿#if NETSTANDARD2_0
 namespace System;
 
 internal static partial class EnumExtensions


src/BenchmarkDotNet/Extensions/Polyfills/EnumExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿#if NETFRAMEWORK || NETSTANDARD2_0
```
timcassell:
```suggestion
﻿#if NETSTANDARD2_0
```

Issue #2890, opened 2025-12-02
improve wasm support


FreyLuis:
I found a few issues while setting up WASM benchmarks & would like to fix those. Since those would include breaking changes, I’m asking about them here before starting any work.

- Currently [`--expose_wasm`](https://github.com/dotnet/BenchmarkDotNet/blob/c1dc751ad98588c80b2c8141fab388b1a94bd120/src/BenchmarkDotNet/Environments/Runtimes/WasmRuntime.cs#L31) is the default parameter for `javaScriptEngineArguments`, but the flag was [removed in 2024](https://github.com/v8/v8/commit/47766a75ec7bcd1a0d7fd86432a5b6e26aaf52fe).
- The JS entry point is hardcoded to be `AppBundle/test-main.js` (or `AppBundle/main.js` with .net 5 / 6). The user then has to create the file under the correct path & still needs to set the `WasmMainJSPath` as an msbuild argument. The path shouldn’t be hardcoded & BenchmarkDotNet could autogenerate the file.
- The default parameter for the .net monikers are still .net 5. IMHO, those should either be removed or replaced with factory methods (`CreateWasm10_0(string javaScriptEngine, string javaScriptEngineArguments, bool aot)`).

The docs are also outdated (already tracked by https://github.com/dotnet/BenchmarkDotNet/issues/1818): 

- The current example only describes how to set benchmarks up using a custom dotnet runtime. This is unnecessary and looks scary to anyone not working on dotnet itself.
- The example also recommends using v8 directly via `jsvu`, which does not provide standard JavaScript APIs. Dotnet relies on these APIs (for creating GUID’s, it uses `crypto.getRandomValues()` for example). IMHO, recommending node might be the better choice.

Would you accept such a PR, or is someone already working on something similar?

timcassell:
> * The default parameter for the .net monikers are still .net 5. IMHO, those should either be removed or replaced with factory methods (`CreateWasm10_0(string javaScriptEngine, string javaScriptEngineArguments, bool aot)`).

I have mentioned in the past that the default moniker should probably use the current runtime version, but no one seemed interested in doing it.

We are definitely open to accepting PRs that update the wasm toolchain. Just make sure you validate any changes against dotnet/performance.

timcassell:
[...]

PR #3026, opened 2026-02-25
Fix perf test failures WASM mono & coreclr

ilonatommy:
- Changing from `test-main.mjs` to BDN-based `bechmark-main.mjs` in https://github.com/dotnet/BenchmarkDotNet/pull/2998 did not port the polyfill https://github.com/dotnet/runtime/blob/85b9bc521939ab950171921e021508d5f267f51b/src/mono/browser/test-main.mjs#L50 to `bechmark-main.mjs`, causing
```
[2026/02/24 18:15:05][INFO] MONO_WASM: This engine doesn't support crypto.getRandomValues. Please use a modern version or provide polyfill for 'globalThis.crypto.getRandomValues'.
[2026/02/24 18:15:05][INFO] System.Reflection.TargetInvocationException: Arg_TargetInvocationException
[2026/02/24 18:15:05][INFO]  ---> System.TypeInitializationException: TypeInitialization_Type, System.ComponentModel.TypeDescriptor
[2026/02/24 18:15:05][INFO]  ---> System.Security.Cryptography.CryptographicException: Arg_CryptographyException
[2026/02/24 18:15:05][INFO]    at Interop.GetCryptographicallySecureRandomBytes(Byte* buffer, Int32 length)
[2026/02/24 18:15:05][INFO]    at System.Guid.NewGuid()
[2026/02/24 18:15:05][INFO]    at System.ComponentModel.TypeDescriptor..cctor()
[2026/02/24 18:15:05][INFO]    Exception_EndOfInnerExceptionStack
```

- Mono interpreter and coreCLR tests were hitting timeouts. BDN does not have a way to manipulate test run timeouts, this PR introduces a param for it.

```
[2026/02/24 17:49:36][INFO] // The benchmarking process did not finish within 10 minutes, it's going to get force killed now.
[2026/02/24 17:49:36][INFO] 
[2026/02/24 17:49:36][INFO] 
[2026/02/24 17:49:36][INFO] #
[2026/02/24 17:49:36][INFO] # Fatal error in , line 0
[2026/02/24 17:49:36][INFO] # d8: Received SIGTERM signal (likely due to a TIMEOUT)
```

cc @timcassell

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -706,7 +706,8 @@ namespace BenchmarkDotNet.ConsoleArguments
                 aot: wasmAot,
                 wasmDataDir: options.WasmDataDirectory?.FullName ?? "",
                 moniker: moniker,
-                isMonoRuntime: !options.WasmCoreCLR);
+                isMonoRuntime: !options.WasmCoreCLR,
+                processTimeoutMinutes: options.WasmProcessTimeoutMinutes);
 
             var toolChain = WasmToolchain.From(new NetCoreAppSettings(
                 targetFrameworkMoniker: wasmRuntime.MsBuildMoniker,
--- a/src/BenchmarkDotNet/Templates/benchmark-main.mjs
+++ b/src/BenchmarkDotNet/Templates/benchmark-main.mjs
@@ -1,9 +1,27 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+const ENVIRONMENT_IS_NODE = typeof process == "object" && typeof process.versions == "object" && typeof process.versions.node == "string";
+const ENVIRONMENT_IS_WEB_WORKER = typeof importScripts == "function";
+const ENVIRONMENT_IS_WEB = typeof window == "object" || (ENVIRONMENT_IS_WEB_WORKER && !ENVIRONMENT_IS_NODE);
+
+if (!ENVIRONMENT_IS_NODE && !ENVIRONMENT_IS_WEB && typeof globalThis.crypto === 'undefined') {
+    // **NOTE** this is a simple insecure polyfill for JS shells (like v8/d8) that lack Web Crypto API.
+    // /dev/random doesn't work on js shells, so define our own.
+    globalThis.crypto = {
+        getRandomValues: function (buffer) {
+            for (let i = 0; i < buffer.length; i++)
+                buffer[i] = (Math.random() * 256) | 0;
+        }
+    }
+}
+
 import { dotnet } from './_framework/dotnet.js'
+
+// Get command line arguments: Node.js uses process.argv, v8 uses arguments/scriptArgs
+const args = typeof process !== 'undefined' ? process.argv.slice(2) : (typeof arguments !== 'undefined' ? [...arguments] : (typeof scriptArgs !== 'undefined' ? scriptArgs : []));
 
 await dotnet
     .withDiagnosticTracing(false)
-    .withApplicationArguments(...arguments)
+    .withApplicationArguments(...args)
     .run()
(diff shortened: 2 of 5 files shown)


timcassell:
Is there any reason why we shouldn't just copy the entire file?

ilonatommy:
> Is there any reason why we shouldn't just copy the entire file?

Copy `test-main.mjs`? It's made in mind to serve runtime tests, there's a lot of unnecessary bits inside, not focusing only on v8 tests but also on browser and nodejs tests. I would prefer to keep BDN file minimal. The big downside of `test-main.mjs` is that it's obscure, hard to maintain and that's why we decided to go away from it in perf tests.

The worth considering in the future could be:
- Node.js uncaughtException handler and Node version check (but currently perf runs on v8 so no point right now)
- Firefox/Chromium detection, Error.stackTraceLimit (same as above, it's for browser only)

The rest is irrelevant.

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -706,7 +706,8 @@ private static Job MakeWasmJob(Job baseJob, CommandLineOptions options, string m
                 aot: wasmAot,
                 wasmDataDir: options.WasmDataDirectory?.FullName ?? "",
                 moniker: moniker,
-                isMonoRuntime: !options.WasmCoreCLR);
+                isMonoRuntime: !options.WasmCoreCLR,
```
pavelsavara, line 709: We use `RuntimeFlavor` in runtime repo. Adopting it here would make it easier to have `NAOT` in the future.

src/BenchmarkDotNet/Templates/benchmark-main.mjs
```diff
@@ -1,9 +1,27 @@
...
 // The .NET Foundation licenses this file to you under the MIT license.
 
+const ENVIRONMENT_IS_NODE = typeof process == "object" && typeof process.versions == "object" && typeof process.versions.node == "string";
+const ENVIRONMENT_IS_WEB_WORKER = typeof importScripts == "function";
+const ENVIRONMENT_IS_WEB = typeof window == "object" || (ENVIRONMENT_IS_WEB_WORKER && !ENVIRONMENT_IS_NODE);
+
+if (!ENVIRONMENT_IS_NODE && !ENVIRONMENT_IS_WEB && typeof globalThis.crypto === 'undefined') {
+    // **NOTE** this is a simple insecure polyfill for JS shells (like v8/d8) that lack Web Crypto API.
+    // /dev/random doesn't work on js shells, so define our own.
+    globalThis.crypto = {
```
pavelsavara, line 11: This is already polyfilled in coreCLR. Is it missing for Mono ? I would rather fix it there.

ilonatommy: @timcassell, we'll try to do in in runtime repo

pavelsavara: I see it now, the polyfill is there for NodeJS when it has `node:crypto`
When `node:crypto` is missing we fail. For V8 we fail too.
I guess better be secure than sorry.

Also we will need this polyfill for Net8-Net10 anyway.

pavelsavara: > [@timcassell](https://github.com/timcassell), we'll try to do in in runtime repo

I changed my mind after looking at the details on runtime side, sorry

timcassell:
> The worth considering in the future could be:
> 
> * Node.js uncaughtException handler and Node version check (but currently perf runs on v8 so no point right now)
> * Firefox/Chromium detection, Error.stackTraceLimit (same as above, it's for browser only)
> 
> The rest is irrelevant.

Yeah, handling other JS engines besides v8 was my thought as well. But it can be done separately (#2890).

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
timcassell: Sounds like a good idea. It can be done separately, though.

ilonatommy: I totally agree but I would prefer to put it out of scope, we would like to avoid new out-of-sync changes on the stage we're now. Such change touches public API, it would be probably linked with changing `--wasmCoreCLR` flag to `--wasmRuntimeFlavor coreclr` etc.

timcassell: `--wasmCoreCLR` option hasn't been released to public nuget yet, so it's perfectly fine to change it at this point.

ilonatommy: I made a follow up https://github.com/dotnet/BenchmarkDotNet/pull/3028.

PR #3028, opened 2026-02-25
Adopt `RuntimeFlavor`

ilonatommy:
Follow up for https://github.com/dotnet/BenchmarkDotNet/pull/3026#discussion_r2852886342.

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/RuntimeFlavor.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/RuntimeFlavor.cs
@@ -1,14 +1,13 @@
-namespace BenchmarkDotNet.Environments
-{
-    /// <summary>
-    /// Specifies the .NET runtime flavor to use for WASM benchmarks.
-    /// </summary>
-    public enum RuntimeFlavor
-    {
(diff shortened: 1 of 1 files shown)


src/BenchmarkDotNet/ConsoleArguments/RuntimeFlavor.cs
```diff
@@ -0,0 +1,14 @@
+namespace BenchmarkDotNet.Environments
```
timcassell: Use file-scoped namespace

src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
```diff
@@ -449,7 +449,7 @@ private static Job GetBaseJob(CommandLineOptions options, IConfig? globalConfig)
             {
                 baseJob = baseJob.WithEnvironmentVariables(options.EnvironmentVariables.Select(text =>
                 {
-                    var separated = text.Split([EnvVarKeyValueSeparator], 2);
+                    var separated = text.Split(new[] { EnvVarKeyValueSeparator }, 2);
```
timcassell: I don't think this change is necessary?

Issue #2284, opened 2023-03-09
Integrate RAPL energy usage measurements


ta1H3n:
This could be a great addition to this framework if energy usage measurements could be included in the benchmarks. Energy efficiency is becoming more relevant, and it is known that energy usage does not always correlate with execution time.

Recent research suggests that RAPL provides good enough measurements:

> Our observations reveal that RAPL readings are highly correlated with plug power, promisingly accurate enough, and have negligible performance overhead. Experimental results suggest RAPL can be a very useful tool to measure and monitor the energy consumption of servers without deploying any complex power meters.

https://dl.acm.org/doi/abs/10.1145/3177754

wild0ne:
@AndreyAkinshin Hi!

Frankly, the feature looks quite interesting. Would you be so kind as to share a couple of ideas regarding PoC? What might the minimum results look like? I assume that RAPL counters might be useless if **Frequency and Time** (and certain scalar values, of course) are enough to calculate power consumption.

AndreyAkinshin:
Hey @wild0ne, thanks for the interest! Here are some thoughts on a possible PoC.

**Why RAPL counters matter**

Frequency and time alone are not sufficient to reliably estimate power consumption. Two code paths can run at the same frequency for the same duration but have very different energy profiles — e.g., AVX-heavy workloads vs simple integer arithmetic. Modern CPUs also apply dynamic voltage/frequency scaling (DVFS), selectively power-gate execution units, and transition through various C-states — all of which affect real consumption in ways that a simple formula cannot capture. RAPL provides energy readings via hardware counters that are far more accurate than any frequency/time-based estimation — especially on modern hardware (Haswell+), where RAPL uses integrated voltage regulators for direct measurement. It also covers DRAM energy, which cannot be derived from CPU frequency or timing data.

**What the output could look like**

The whole point is to surface cases where faster ≠ more efficient:

```
| Method | Mean     | Energy/Op |
|--------|----------|-----------|
| Foo    | 1.234 ms | 0.045 J   |
| Bar    | 1.198 ms | 0.071 J   |
```

Bar is slightly faster but consumes ~58% more energy — this kind of insight is invisible with time-only measurements.

**Package structure**

This should live in a separate package (e.g., `BenchmarkDotNet.Diagnostics.Energy`) to keep the core lightweight and avoid pulling in platform-specific dependencies for users who don't need this.

**Minimum PoC scope**

[...]

PR #3032, opened 2026-02-27
Linux RAPL counters #2284

wild0ne:
**DRAFT** PR adding usage of RAPL energy counters on Linux

### DIFF
--- a/src/BenchmarkDotNet.Diagnostics.Energy/LinuxEnergyCounter.cs
+++ b/src/BenchmarkDotNet.Diagnostics.Energy/LinuxEnergyCounter.cs
@@ -22,7 +22,7 @@ namespace BenchmarkDotNet.Diagnosers
             {
                 string fileContents = File.ReadAllText(_energyPath);
                 bool good = long.Parse(fileContents) > 0;
-                return (true, string.Empty);
+                return (good, string.Empty);
             }
             catch (Exception e)
             {


src/BenchmarkDotNet.Diagnostics.Energy/LinuxEnergyCounter.cs
```diff
@@ -0,0 +1,68 @@
...
+            _energyPath = !string.IsNullOrEmpty(energyPath) ? energyPath : throw new ArgumentException(nameof(energyPath));
+        }
+
+        public override (bool, string) TestRead()
+        {
+            try
+            {
+                string fileContents = File.ReadAllText(_energyPath);
+                bool good = long.Parse(fileContents) > 0;
+                return (true, string.Empty);
```
wild0ne: use 'good'

src/BenchmarkDotNet.Diagnostics.Energy/EnergyCountersSetup.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace BenchmarkDotNet.Diagnosers
+{
+    /// <summary>
+    /// Energy counters setup
+    /// </summary>
+    public enum EnergyCountersSetup
+    {
+        /// <summary>
+        /// Default setup (core only)
```
wild0ne, line 15: Probably, we should display the whole package energy consumption by default instead of the core-only

wild0ne: > Probably, we should display the whole package energy consumption by default instead of the core-only

I guess so

PR #3037, opened 2026-03-01
Fix PathTooLongException for benchmarks with long parameters

### LINKED ISSUE #2093: PathTooLongException with benchmark that has long parameter
The following invocation in dotnet/performance:
```bash
dotnet run -c Release -f net7.0 -- --filter "System.Numerics.Tests.Perf_BigInteger.Parse" --corerun /mnt/c/dev/dotnet/core_roots/cf2187_release/corerun -p EP
```

hangs on a benchmark with a very long parameter. When I ctrl-C it, I get the following:
```scala
An exception occurred during reading trace stream: System.IO.PathTooLongException: The path '/home/jakob/dev/dotnet/performance/artifacts/bin/MicroBenchmarks/Release/net7.0/BenchmarkDotNet.Artifacts/System.Numerics.Tests.Perf_BigInteger.Parse(numberString_ 12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890)-20220830-133655.nettrace' is too long, or a component of the specified path is too long.
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirError)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, UnixFileMode openPermissions, Int64& fileLength, UnixFileMode& filePermissions, Func`4 createOpenException)
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access)
   at BenchmarkDotNet.Diagnosers.EventPipeProfiler.CopyEventStreamToFile(EventPipeSession session, String fileName, ILogger logger)
ExitCode != 0 and no results reported
```
Seems like BDN is trying to include all the parameters in the file name of the event pipe trace.

abdulrahmanhossam:
Fixes #2093

### Description
This PR addresses the `PathTooLongException` that occurs on Linux/macOS when a benchmark has very long parameter values. 

**The Root Cause:**
Previously, `ArtifactFileNameHelper` only checked if the *total* path length exceeded the `CommonSenseLimit` (1024 characters). However, on Linux/macOS, while the total path can be long, the individual **file name component** has a strict OS limit of 255 characters. If the parameters are extremely long, the total path might still be under 1024, but the file name itself exceeds 255 characters, causing the OS to throw an exception.

**The Fix:**
I introduced a `MaxFileNameLength` constant (255) and updated the validation in `GetFilePath` to check both the total path length and the individual file name length:
```csharp
string fileNameOnly = Path.GetFileName(nameNoLimit);
if (nameNoLimit.Length <= limit && fileNameOnly.Length <= MaxFileNameLength)

### DIFF
--- a/src/BenchmarkDotNet/Helpers/ArtifactFileNameHelper.cs
+++ b/src/BenchmarkDotNet/Helpers/ArtifactFileNameHelper.cs
@@ -1,13 +1,12 @@
-﻿using System;
-using System.IO;
-using System.Linq;
-using BenchmarkDotNet.Detectors;
+﻿using BenchmarkDotNet.Detectors;
 using BenchmarkDotNet.Diagnosers;
 using BenchmarkDotNet.Exporters;
 using BenchmarkDotNet.Extensions;
-using BenchmarkDotNet.Portability;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains;
+using System;
+using System.IO;
+using System.Linq;
 
 namespace BenchmarkDotNet.Helpers
 {
@@ -15,6 +14,7 @@ namespace BenchmarkDotNet.Helpers
     {
         private const int WindowsOldPathLimit = 260;
         private const int CommonSenseLimit = 1024; // for benchmarks that use args like "new string('a', 200_000)"
+        private const int MaxFileNameLength = 255;
 
         internal static string GetTraceFilePath(DiagnoserActionParameters details, DateTime creationTime, string fileExtension)
         {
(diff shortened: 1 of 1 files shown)


timcassell:
What does this change do? The original code already checked the length of the full path (which includes the filename), how does checking only the filename length do anything?

abdulrahmanhossam:
Hi @timcassell ,

That's a great question! The issue here isn't the total path length, but rather the OS-level limitation on **individual file name components** (`NAME_MAX`).

On Linux/macOS (and specifically file systems like `ext4`), the maximum length of an entire path is usually quite large (e.g., 4096 characters). However, the strict limit for a **single file name** is exactly **255 characters**. 

In the original code, if a benchmark had an extremely long parameter (as described in #2093), the generated filename alone could easily be 300+ characters. 
For example, if the base path is `30` characters and the filename is `320` characters, the total path length is `350`. The old check (`350 <= 1024`) would evaluate to `true` and bypass the truncation logic. Then, when it actually attempts to create the file, the OS throws a `PathTooLongException` (`ENAMETOOLONG`) because the 320-character filename violates the 255-character `NAME_MAX` rule.

By explicitly checking `fileNameOnly.Length <= MaxFileNameLength` (255), we ensure that these edge cases correctly fall back to `GetLimitedFilePath()` to be truncated and hashed, preventing the OS-level crash. 

Hope this clarifies the intent! Let me know if you'd like me to adjust anything.

timcassell:
Thanks (sorry I missed the explanation in the OP).

PR #847, opened 2018-08-08
Added Memory Information - Free and Total memory for different platforms

ChiragRupani:
Added Memory Information - Free and Total memory for different platforms, for #846 Add RAM to summary

PR #3038, opened 2026-03-02
Fix #846: Add RAM to environment summary

### LINKED ISSUE #846: Add RAM to summary
Should include RAM as well as CPU so GC reasons between two different computers are easier to spot
```
BenchmarkDotNet=v0.11.0, OS=Windows 10.0.16299.547 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
Frequency=3515624 Hz, Resolution=284.4445 ns, Timer=TSC
.NET Core SDK=2.1.302
  [Host] : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
  Core   : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT

Job=Core  Runtime=Core  
```

abdulrahmanhossam:
Resolves #846.

This PR adds the total physical memory (RAM) to the environment summary, appending it directly to the CPU information line to keep the summary compact and readable.

**Implementation details:**
- **Windows:** Uses `GlobalMemoryStatusEx` via P/Invoke.
- **Linux:** Reads `MemTotal` from `/proc/meminfo`.
- **macOS:** Uses `sysctl -n hw.memsize` process output.

The total memory is formatted in GB (e.g., `16 GB RAM`). I used `Lazy<T>` to ensure the OS is only queried once and doesn't impact performance.

Let me know if any adjustments are needed!

### DIFF
--- a/src/BenchmarkDotNet/Environments/PhysicalMemoryInfo.cs
+++ b/src/BenchmarkDotNet/Environments/PhysicalMemoryInfo.cs
@@ -156,10 +156,18 @@ namespace BenchmarkDotNet.Environments
                 {
                     string output = process.StandardOutput.ReadToEnd();
                     process.WaitForExit();
+
+                    long pageSize = 4096;
+                    var pageSizeMatch = Regex.Match(output, @"page size of (\d+) bytes");
+                    if (pageSizeMatch.Success && long.TryParse(pageSizeMatch.Groups[1].Value, out long parsedPageSize))
+                    {
+                        pageSize = parsedPageSize;
+                    }
+
                     var match = Regex.Match(output, @"Pages free:\s+(\d+)");
                     if (match.Success && long.TryParse(match.Groups[1].Value, out long pagesFree))
                     {
-                        available = pagesFree * 4096;
+                        available = pagesFree * pageSize;
                     }
                 }
             }


timcassell:
RAM is not part of the CPU technically. Just make it a new line (yield return it separately). Also, if the info is available, it could be useful to include the frequency of the RAM, and type (DDR4/5).

timcassell:
Also free or used memory if possible (see #847 convo regarding it).

abdulrahmanhossam:
thanks for the lightning-fast review!

1. You are absolutely right, it makes much more sense on a separate line. I will update this right away.
2. I will also update the implementation to include the `Available/Free` memory alongside the `Total` memory as discussed in the #847 thread.
3. Regarding the RAM Frequency and Type (DDR4/5): Gathering this accurately cross-platform natively in C# is quite complex without external dependencies (e.g., parsing `dmidecode` on Linux often requires sudo, and `system_profiler` on macOS is quite heavy). Would it be acceptable to scope this PR to `Total/Available` memory first, and perhaps track the hardware details (Frequency/Type) in a separate future issue?

Let me know if this sounds like a good plan, and I'll push the updates shortly!

src/BenchmarkDotNet/Environments/HostEnvironmentInfo.cs
```diff
@@ -96,6 +99,10 @@ public override IEnumerable<string> ToFormattedString()
                 yield return $"{BenchmarkDotNetCaption} v{BenchmarkDotNetVersion}, {Os.Value.ToBrandString()}";
 
             yield return Cpu.Value.ToFullBrandName();
+
+            if (PhysicalMemory.Value != null)
+                yield return $"Memory: {PhysicalMemory.Value.ToFormattedString()}";
```
timcassell: Please move it after the cpu frequency

src/BenchmarkDotNet/Environments/PhysicalMemoryInfo.cs
```diff
@@ -0,0 +1,170 @@
...
+            using (var process = Process.Start(vmStatInfo))
+            {
+                if (process != null)
+                {
+                    string output = process.StandardOutput.ReadToEnd();
+                    process.WaitForExit();
+                    var match = Regex.Match(output, @"Pages free:\s+(\d+)");
+                    if (match.Success && long.TryParse(match.Groups[1].Value, out long pagesFree))
+                    {
+                        available = pagesFree * 4096;
```
filzrev: I thought PageSize need to be extracted from `vm_stat` command output.

As far as I've confirmed.
`macos-latest` GitHub Actions ARM64 image returns following page size.

> Mach Virtual Memory Statistics: (page size of 16384 bytes)

And `macos-15-intel`/`macos-26-intel` image returns 4096 bytes

abdulrahmanhossam: I've updated the logic to parse the page size dynamically from the `vm_stat` output header.

timcassell (approved):
Doing a separate issue/PR for those is ok. Thanks @abdulrahmanhossam

PR #3039, opened 2026-03-02
fix: InProcess command line option behaviors

filzrev:
This PR intended to fix `--inProcess` command line option behaviors.

Currently, when `--inProcess` argument is contained as argument.
ConfigParser create job based on `Job.Default`. and remaining `baseConfig` setting seems silently ignored.

This PR modify code to use `baseJob` settings when creating InProcess toolchain job.


timcassell:
Need to fix merge conflicts

PR #3040, opened 2026-03-02
feat: Wasm MemoryDiagnoser

### LINKED ISSUE #2226: Add MemoryDiagnoser support for WASM

FreyLuis:
This PR enables the `MemoryDiagnoser` to run on wasm.

Resolves #2226

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/MemoryDiagnoserTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/MemoryDiagnoserTests.cs
@@ -49,19 +49,10 @@ namespace BenchmarkDotNet.IntegrationTests
 
         public class AccurateAllocations
         {
-            public class Empty { }
+            [Benchmark] public byte[] EightBytesArray() => new byte[8];
+            [Benchmark] public byte[] SixtyFourBytesArray() => new byte[64];
 
-            [Benchmark]
-            public byte[] EightBytesArray() => new byte[8];
-
-            [Benchmark]
-            public byte[] SixtyFourBytesArray() => new byte[64];
-
-            [Benchmark]
-            public Task<int> AllocateTask() => Task.FromResult<int>(-12345);
-
-            [Benchmark]
-            public Empty AllocateEmptyObject() => new Empty();
+            [Benchmark] public Task<int> AllocateTask() => Task.FromResult<int>(-12345);
         }
 
         [Theory, MemberData(nameof(GetToolchains), DisableDiscoveryEnumeration = true)]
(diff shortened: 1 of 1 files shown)


FreyLuis:
Would the unit test be fine as is? `SizeOfAllFields` calculates the size platform specific so we can't use it here.

tests/BenchmarkDotNet.IntegrationTests/MemoryDiagnoserTests.cs
```diff
@@ -86,6 +99,30 @@ public void MemoryDiagnoserSupportsModernMono()
...
+
+            var runtime = new WasmRuntime(
+                netCoreAppSettings.TargetFrameworkMoniker, RuntimeMoniker.WasmNet80,
+                "Wasm", aotCompilerMode == MonoAotCompilerMode.wasm, "v8");
+
+            AssertAllocations(WasmToolchain.From(netCoreAppSettings), typeof(AccurateAllocations), new Dictionary<string, long>
+            {
+                { nameof(AccurateAllocations.EightBytesArray), 8 + objectAllocationOverhead + arraySizeOverhead },
+                { nameof(AccurateAllocations.SixtyFourBytesArray), 64 + objectAllocationOverhead + arraySizeOverhead },
+                { nameof(AccurateAllocations.AllocateEmptyObject), objectAllocationOverhead },
```
timcassell: Missing `AllocateTask` (and the other test is missing the new `AllocateEmptyObject`).

FreyLuis: AllocateTask is the one that depends on SizeOfAllFields to calculate its expected size.

FreyLuis: Actually, I could also just set the expected value to a constant...

FreyLuis: I removed the `AllocateEmptyObject` because it didn't work with the normal and native toolchains. It allocated 24 bytes instead of 16 (`objectAllocationOverhead` and what `CalculateRequiredSpace` calculated for an empty class). It was the expected 2 * pointer size with Mono and MonoWasm though...

https://github.com/twobrainsgmbh/BenchmarkDotNet/actions/runs/22619464127/job/65540582849

Since this issue occurred with other toolchains then wasm, it seems out of scope for this PR.

timcassell: 3 * ptr size is the minimum object size in CLR, I'm surprised mono is able to eliminate the padding.

Issue #2803, opened 2025-07-04
Consider shipping a test adapter for MTP


Youssef1313:
The current BDN.TestAdapter is specifically for VSTest. It would be good to ship another one for Microsoft.Testing.Platform.

timcassell:
I think we should just update the test adapter to use the new platform, rather than creating a separate package.

adamsitnik:
cc @caaavik-msft

sheddy123:
@timcassell is this up for implementation? I can pick this

timcassell:
@sheddy123 Yes this is available, sorry for the delayed response.

sheddy123:
Thank you @timcassell  can you assign to me

PR #3043, opened 2026-03-06
chore: Separate solution file for sample projects

filzrev:

1. Temporary comment out samples projects from root solution file.
2. Add `samples/BenchmarkDotNet.Samples.slnx` solution file instead.
3. Fix wrong `launchSettings.json` settings.

**Background**
Currently VS TestAdapter and test projects both using VSTest based dependencies.

When replacing TestAdapter to use MTP(Microsoft.Testing.Platform) based implementation (#2803)
Both VSTest/MTP based projects are mixed and it seems Test Explorer unstable state.
And `dotnet test` failed when both VSTest/MTP based project mixed in solution.

So it need to separate solution files until TestAdapter/Tests are migrated to use MTP.

### DIFF
--- a/BenchmarkDotNet.slnx
+++ b/BenchmarkDotNet.slnx
@@ -3,11 +3,14 @@
     <File Path="build/common.props" />
     <File Path="build/common.targets" />
   </Folder>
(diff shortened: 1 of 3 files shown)


timcassell:
I'd rather this be done in the same PR as MTP migration if necessary. It seems pointless by itself.

PR #3047, opened 2026-03-09
Add ReadyToRun flag to restore step

DrewScoggins:
When BDN is run with the `--no-restore` option we need to pass this parameter as part of the restore step or crossgen fails in the publish step.


timcassell:
The title/description no longer matches the changes. I assume commit 3 is the intended fix. Can you update them?

DrewScoggins:
I am going to switch this over to draft, as now I am doubting some of my testing procedures. Once I have this in better shape I will flip it back over.

DrewScoggins:
This PR was never actually needed. Closing.

PR #3050, opened 2026-03-13
feat: Add UseLocalJobOnlyAttribute and modify related codes

filzrev:

1. Add `UseLocalJobOnlyAttribute`
2. Add `ConfigUnionRule.UnionAndUseLocalJob`
3. Modify `BenchmarkConverter.cs` code to handle `UnionAndUseLocalJobAttribute`.
4. Modify samples benchmarks that using local jobs.

**Background**
When using global config that contains jobs and benchmark local config also jobs.
Both jobs that are defined by global/local config are executed.
(e,g, When using `DebugInProcessConfig` as global config. and benchmark has attributes also define jobs)

This PR intended to exclude unnecessary jobs that defined by global config.

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkConverter.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkConverter.cs
@@ -99,9 +99,9 @@ namespace BenchmarkDotNet.Running
             var typeAttributes = type.GetCustomAttributes(true).OfType<IConfigSource>().ToArray();
             if (typeAttributes.Any(x => x is UseLocalJobOnlyAttribute))
             {
-                // UseLocalJobOnlyAttribute.can't be combined with other benchmark attribute that set non-default UnionRule.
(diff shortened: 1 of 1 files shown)


meiranzheng:
Nice addition — I like the intent (let per-type attributes control the jobs, without global/assembly jobs leaking in), and aggregating type attributes into a single config avoids any attribute-ordering surprises.

A couple small suggestions:
- Could we add a unit test to lock in the new `ConfigUnionRule.UnionAndUseLocalJob` behavior (e.g., global config has jobs + other settings; local config has jobs; expect union keeps global non-job settings but uses only the local jobs)?
- Minor: the `InvalidBenchmarkDeclarationException` message has a couple typos and could include *which* attribute set a non-default `UnionRule` to make it more actionable.
- Optional: if this is meant for end users too (not just samples), it might be worth a short doc note about when to use `UseLocalJobOnlyAttribute` / `UnionAndUseLocalJob`.

PR #3051, opened 2026-03-14
feat: Add API to set custom `DefaultConfig.Instance`

filzrev:
This PR add `DefaultConfig.SetDefault(config)` API to customize **default global config**.

**Background**
BenchmarkDotNet use `DefaultConfig.Instance` when **global** config is not passed to `BenchmarkRunner`/`BenchmarkSwitcher`.

On normarl use case, It can specify **global** config by using `Main` entry point.
But when running benchmark via `TestAdaper`, It can't use `Main` entry point (and `launchsettings.json`)
So. it need to customize config via assembly-level ConfigAttribute.
https://benchmarkdotnet.org/articles/features/vstest.html?q=testadapter#setting-a-default-configuration

This PR add `DefaultConfig.SetDefault(config)` API.
By this change, it can use custom config by TestAdapter, (Settings are loaded from `.runsettings`/`testconfig.json`)

### DIFF
--- a/src/BenchmarkDotNet/Configs/DefaultConfig.cs
+++ b/src/BenchmarkDotNet/Configs/DefaultConfig.cs
@@ -35,7 +35,7 @@ namespace BenchmarkDotNet.Configs
         {
         }
 
-        public static void SetDefault(IConfig? config)
+        public static void SetCustomConfig(ImmutableConfig? config)
         {
             customDefaultConfig = config;
         }


timcassell:
How is it expected to be used? The test adapter supports providing your own entry point by setting `<GenerateProgramFile>false</GenerateProgramFile>`.

filzrev:
> How is it expected to be used?

It's needed when running benchmarks from VS TestExplorer.

When running benchmark via TestExplorer.
It always use `DefaultConfig.Instance` because `Program.cs` entry point is not used.
I want to customize this behavior by this PR.

timcassell:
> It's needed when running benchmarks from VS TestExplorer.

Right, you said that already, but it doesn't explain how.

filzrev:
> Right, you said that already, but it doesn't explain how.

The following code is expected API usage on VSTestAdapter (with VsTest).
https://github.com/filzrev/BenchmarkDotNet/commit/c438bd31f23e6e07fdf00c79030aa1c5945eb301

By this changes, It can specify custom arguments when running benchmark on Test Explorer.

src/BenchmarkDotNet/Configs/DefaultConfig.cs
```diff
@@ -23,13 +23,23 @@ namespace BenchmarkDotNet.Configs
...
+
         private readonly static Conclusion[] emptyConclusion = [];
 
+        private static IConfig? customDefaultConfig = null;
+
         private DefaultConfig()
         {
         }
 
+        public static void SetDefault(IConfig? config)
```
timcassell: The name is misleading, it's actually setting `Instance`.

filzrev: I've modified API name to `SetCustomConfig`. and accepted type to `ImmutableConfig` to prevent mutable config  is passed.

timcassell (approved):

PR #3055, opened 2026-03-21
feat(jobs): add MsBuildProperty for escaped MSBuild properties

### LINKED ISSUE #2719: Feature Request: Add MsBuildArgument supports for MSBuild special characters

meiranzheng:
Fixes #2719.

### What
- Add opt-in MSBuild special character escaping via `new MsBuildArgument(value, escapeSpecialCharacters: true)`.
- Add `MsBuildProperty : MsBuildArgument` for `/p:name=value` (joins multiple values with `;` and escapes automatically).
- Keep reports readable by showing the unescaped value via `ToString()` when escaping is enabled.
- Add unit + integration coverage for semicolon-separated values.
- Docs: troubleshooting guide now mentions `MsBuildProperty`.

### Why
MSBuild treats `;` as a property separator in `/p:` so values like `DefineConstants=TEST1;TEST2` must be escaped (eg `%3B`) to work reliably.

### Verification
- `dotnet test tests/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj -c Release -f net8.0 --filter FullyQualifiedName~MsBuildArgumentTests`
- `dotnet test tests/BenchmarkDotNet.IntegrationTests.ManualRunning/BenchmarkDotNet.IntegrationTests.ManualRunning.csproj -c Release -f net8.0 --filter FullyQualifiedName~MsBuildArgumentTests`

### Context
This follows the combined direction discussed in #2730 / #2732 (opt-in escaping + a dedicated property type).

### DIFF
--- a/docs/articles/guides/troubleshooting.md
+++ b/docs/articles/guides/troubleshooting.md
@@ -45,7 +45,7 @@ How to troubleshoot the build process:
 The recommended order of solving build issues:
 
 1. Change the right settings in your project file which defines benchmarks to get it working.
-2. Customize the `Job` settings using available options like `job.WithCustomBuildConfiguration($name)`or `job.With(new Argument[] { new MsBuildArgument("/p:SomeProperty=Value")})`.
+2. Customize the `Job` settings using available options like `job.WithCustomBuildConfiguration($name)` or `job.WithArguments(new Argument[] { new MsBuildArgument("/p:SomeProperty=Value") })`. For list-like MSBuild properties that contain special characters (for example, `DefineConstants=TEST1;TEST2`), use `MsBuildProperty` (it escapes the value for you): `job.WithArguments(new Argument[] { new MsBuildProperty("DefineConstants", "TEST1", "TEST2") })`.
 3. Implement your own `IToolchain` and generate and build all the right things in your way (you can use existing Builders and Generators and just override some methods to change specific behaviour).
 4. Report a bug in BenchmarkDotNet repository.
 
@@ -94,4 +94,3 @@ public void Setup()
 ### One of the above, but with a Debug build
 
 By default, BDN builds everything in Release. But debugging Release builds even with full symbols might be non-trivial. To enforce BDN to build the benchmark in Debug please use `DebugBuildConfig` and then attach the debugger.
-
--- a/src/BenchmarkDotNet/Jobs/Argument.cs
+++ b/src/BenchmarkDotNet/Jobs/Argument.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Text;
 using JetBrains.Annotations;
 
 namespace BenchmarkDotNet.Jobs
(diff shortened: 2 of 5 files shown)


docs/articles/guides/troubleshooting.md
```diff
@@ -45,7 +45,7 @@ How to troubleshoot the build process:
 The recommended order of solving build issues:
 
 1. Change the right settings in your project file which defines benchmarks to get it working.
-2. Customize the `Job` settings using available options like `job.WithCustomBuildConfiguration($name)`or `job.With(new Argument[] { new MsBuildArgument("/p:SomeProperty=Value")})`.
+2. Customize the `Job` settings using available options like `job.WithCustomBuildConfiguration($name)` or `job.WithArguments(new Argument[] { new MsBuildArgument("/p:SomeProperty=Value") })`. For list-like MSBuild properties that contain special characters (for example, `DefineConstants=TEST1;TEST2`), use `MsBuildProperty` (it escapes the value for you): `job.WithArguments(new Argument[] { new MsBuildProperty("DefineConstants", "TEST1", "TEST2") })`.
```
filzrev, line 48: I though pre-defined MSBuild property(`DefineConstants`) should not be set via `WithArgument` API.
It might be better to use something like `ExtraDefineConstants` as examples.

Because it's handled as **global** MSBuild property, and properties that are defined at project level is ignored.

src/BenchmarkDotNet/Jobs/Argument.cs
```diff
@@ -47,6 +48,96 @@ public MonoArgument(string value) : base(value)
...
+        {
+            if (value is null)
+                throw new ArgumentNullException(nameof(value));
+
+            var builder = new StringBuilder(value.Length);
+            foreach (var character in value)
+            {
+                switch (character)
+                {
+                    // See: https://learn.microsoft.com/en-us/visualstudio/msbuild/special-characters-to-escape
```
filzrev, line 73: It's not documented thought, following chars need to be escaped also.  
- Space (` `)
- Comma (`,`)
- Escaped double quote(`"`)

I've tested following tests and it failed when above chars are contained.
https://github.com/dotnet/BenchmarkDotNet/pull/2730/changes#diff-f676d4da73a1a3aef25299c6073c3fc8fc33c925794de73338ba35e9c367719bR70-R82

PR #3078, opened 2026-04-09
Include feature as part of the csproj copying

DrewScoggins:
This PR is from Copilot and is a first attempt to allow us to do performance testing with the new runtime async feature.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
@@ -34,6 +34,7 @@ namespace BenchmarkDotNet.Toolchains.CsProj
             "EnablePreviewFeatures",
             "RuntimeHostConfigurationOption",
             "WarningsAsErrors",
+            "Features",
         }.ToImmutableArray();
 
         public string RuntimeFrameworkVersion { get; }


timcassell (approved):
Nice, you'll get to exercise the new async engine at the same time. 😄

Do note that any historical async benchmark data will be invalidated when you update the BDN version, since we now install a SynchronizationContext and changed to proper awaits instead of synchronous blocking. Results may vary depending on the nature of the benchmark. For example:

```cs
[Benchmark]
public async Task Yield1()
{
    await Task.Yield();
}

[GlobalSetup(Target = nameof(Yield2))]
public void Setup2()
{
    SynchronizationContext.SetSynchronizationContext(null);
}

[Benchmark]
public async Task Yield2()
{
    await Task.Yield();
}

[Benchmark]
public async Task Yield3()
{
    SynchronizationContext.SetSynchronizationContext(null);
    await Task.Yield();
}
```

Results on my machine:

| Method | Mean      | Error    | StdDev   |
|------- |----------:|---------:|---------:|
| Yield1 |  79.40 ns | 0.472 ns | 0.442 ns |
| Yield2 | 304.09 ns | 1.047 ns | 0.928 ns |
| Yield3 | 506.57 ns | 1.170 ns | 1.037 ns |

While all 3 of those were ~455ns with the old engine.

And now you can even do `[Benchmark] public YieldAwaitable Yield4() => Task.Yield();`, which previously would not yield.

Issue #3076, opened 2026-04-05
Benchmark process freeze when running DisassemblyDiagnoserTests with xUnit v3 on macos(x64/arm64)


filzrev:
I'm trying to update BenchmarkDotNet tests to use xUnit v3. (#3065)

But [DisassemblyDiagnoserTests](https://github.com/dotnet/BenchmarkDotNet/blob/d2462300db2247ae73a6a817331482f3b89ad7cb/tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs#L21) cause freeze and cause CI timeout(`6h`) when running tests with xUnit v3 on macos(x64/arm64)

It's not happen on following conditions.
1. Running tests on other OS (Windows/Ubuntu).
2. Run benchmark with `dotnet run` command on `macos`

So I'm suspecting the issue has caused by SynchronizationContext(xUnit/BenchmarkDotNet) related problem,
But I don't have any proof.

**Minimum reproduceable example code**
https://github.com/filzrev/test_ci_on_macos

filzrev:
It seems following `DiagnosticsClient::WriteDump` method causing timeout.

https://github.com/dotnet/BenchmarkDotNet/blob/d2462300db2247ae73a6a817331482f3b89ad7cb/src/BenchmarkDotNet/Disassemblers/ClrMdDisassembler.cs#L59-L78

I don't knows why it cause freeze when benchmark running under xUnit v3.

timcassell:
It sounds like an xUnit or DiagnosticsClient bug. Can you repro with just that code in a test?

timcassell:
Also can you try updating `Microsoft.Diagnostics.NETCore.Client`? It looks like the dependent version from ClrMD is quite old now.

Could also try the new [4.0.0-beta.26203.1](https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/4.0.0-beta.26203.1) version of ClrMD.

filzrev:
It can be reproduced with simple console application without using BenchmarkDotNet/xUnit.v3/`dotnet test` 
- https://github.com/filzrev/test_ci_on_macos/blob/main/Benchmarks/Tests.cs#L13-L42
- https://github.com/filzrev/test_ci_on_macos/actions/runs/24135008739

On macos(x64/arm64).
WriteDump method stop processing after outputting following logs.

> [createdump] 00030bc6 TickFrequency: 1000000 ticks per ms
[...]

PR #3080, opened 2026-04-11
chore: Add AppHost validation to DisassemblyDiagniser  on macos environment

filzrev:
This PR add AppHost validation to DisassemblyDiagniser  on macos environment.

As commented on #3076.
Currently DisassemblyDiagniser  cause freeze when following conditions met.
1. Running on macos(x86/arm64).
2. Using InProcessDiagnoser and using AppHost

By this PR it can avoid unintended benchmark freeze problem.

### DIFF
--- a/src/BenchmarkDotNet/Detectors/AppHostDetector.cs
+++ b/src/BenchmarkDotNet/Detectors/AppHostDetector.cs
@@ -1,0 +1,35 @@
+using BenchmarkDotNet.Extensions;
+using System.Diagnostics;
+using System.Reflection;
+
+namespace BenchmarkDotNet.Detectors;
+
+internal static class AppHostDetector
+{
+    public static bool HasAppHost()
+    {
+        var entryAssembly = Assembly.GetEntryAssembly();
+
+        // Check NativeAOT
+        if (entryAssembly == null)
+            return false;
+
+        using var process = Process.GetCurrentProcess();
+        var processName = process.ProcessName;
+
+        // Check executed with `dotnet run`
+        if (processName.Equals("dotnet", StringComparison.OrdinalIgnoreCase))
+            return false;
+
+        // Check Single-file executable (AppHost is bundled)
+        if (entryAssembly.Location.IsBlank())
+            return true;
+
+        // Check entry assembly extension.
+        if (entryAssembly.Location.EndsWith(".dll", StringComparison.OrdinalIgnoreCase))
+            return true;
+
+        // Return false for unknown environment
+        return false;
+    }
+}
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Detectors/AppHostDetector.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace BenchmarkDotNet.Detectors;
+
+internal static class AppHostDetector
+{
+    public static bool HasAppHost()
+    {
+        var entryAssembly = Assembly.GetEntryAssembly();
+
+        // Check NativeAOT
+        if (entryAssembly == null)
```
timcassell, line 14: GetEntryAssembly also returns null in test frameworks, so this doesn't even catch the issue you found.

filzrev: > GetEntryAssembly also returns null in test frameworks

Is it able to provide code/environment that reproduce issue?

I've testes `AppHostDetector.cs` behavior on Windows with following combinations .
But it can't reproduce issue.  
- Use xUnit.v2/v3.
- Use VSTest/MTP mode.
- Run tests by using TestExplorer/dotnet run/run exe

As far as I've confirmed `testhost.dll` is resolved as entry assembly when using xunit.v2.
And test DLL is resolved when using xunit.v3.

timcassell: In my first prototype of #2682 I tried using it and found it to return null in the test framework. https://github.com/dotnet/BenchmarkDotNet/pull/2682/commits/35b008238277de0d146a5f8e92def723a55432d3#diff-3e3d929255c5bc775e61d756a968c3a994409c578dba06e5630745484c7ce62bR52-R54

timcassell: https://stackoverflow.com/a/53228317

timcassell: That was using the old manual running multiple frameworks test project, which at the time was using an ancient version of xUnit, so maybe it's doesn't repro anymore. But it seems unreliable at least.

filzrev:
Close this PR.
Because `AppHost` existence detection might be unstable.

ClrMD 4.x seems support `CreateSnapshotAndAttach` on macOS.
https://github.com/microsoft/clrmd/blob/main/doc/Migrating4.md#macos-createsnapshotandattach-support

So macOS specific workaround code can be removed after updated to ClrMD 4.x

Issue #3088, opened 2026-04-18
Add settings to skip tests that take times on draft PR


filzrev:
Currently, BenchmarkDotNet CI tests takes about 1.5h - 3h. (Especially slow on `macos(x64)` image)
And it cause following issues.

1. CI tests turnaround time is too long if there are failed tests.
2. It takes GitHub Actions compute resources  
   (e.g. Running CI on `macos` 3h takes about $11, it's not charged on OSS project though)

So, It's needed to reduce CI execution times on PR.

---

**Expected workaround**
Add setting to run **full** tests after `Ready for Review`.

When draft PR is created, Run only limited set of tests. 
1. Skip some tests that are take long times (It's not works inProcessToolchain, and require build/publish operations)
2. Skip Wasm/NativeAOT/R2R related tests and installation of `v8`/`wasm-tool` workload
3. Skip CI on `macos(x64)` image

PR #3090, opened 2026-04-18
chore: Add ThreadingDiagnoserTests with InProcessToolchain

filzrev:
Currently, ThreadingDiagnoserTests with InProcessEmitToolchain is commented out with following commit.
https://github.com/dotnet/BenchmarkDotNet/commit/852bb8cd9c2ac2530866dc53723c5f2ce3d411fa

This PR enable these tests that use InProcessToolin.
Because it seems works without problem on latest codebase (As far as I've tested)

By this changes, it can implement skip long-running tests on **Draft PR** (#3088) (Currently, it takes about 4 minutes)

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
@@ -9,14 +9,10 @@ using BenchmarkDotNet.Reports;
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Tests.Loggers;
 using BenchmarkDotNet.Toolchains;
+using BenchmarkDotNet.Toolchains.InProcess.Emit;
 using BenchmarkDotNet.Toolchains.NativeAot;
-using System;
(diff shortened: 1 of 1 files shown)


timcassell (approved):

filzrev:
On latest master CI.
`CompletedWorkItemCountIsAccurate` test failed on `macos(arm64)` with following error.

**Error message**
```
CompletedWorkItemCountIsAccurate(toolchain: InProcessEmitToolchain) [FAIL]
       Assert.Equal() Failure: Values are not within 2 decimal places
       Expected: 1 (rounded from 1)
       Actual:   1.02 (rounded from 1.015625)
```

The following actions are required to fix this issue.
1. Revert this PR commit
2. Reduce precision when comparing metrics value (Because original CompletedWorkItemCount/LockContentionCount value is integer) https://github.com/dotnet/BenchmarkDotNet/blob/d11792c6c58da1e00bb44d6458ca015d5928bab9/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs#L182.

Issue #2610, opened 2024-08-07
A benchmark job with a Native AOT tool chain fails on package restore when package source mapping is used


akamsteeg:
(BenchmarkDotNet on Windows 11.)

I have a solution that uses [NuGet package source mapping](https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping). There's a benchmark project with a `Program.cs` that looks like this:

```csharp
public static class Program
{
  public static void Main(string[] args)
  {
    var config = GetConfig();
    BenchmarkSwitcher
      .FromAssembly(Assembly.GetExecutingAssembly())
      .Run(args, config);
  }

  private static ManualConfig GetConfig()
  {
    var job = Job.Default;

    var config = ManualConfig.Create(DefaultConfig.Instance)
      .AddJob(
        job.WithToolchain(CsProjCoreToolchain.NetCoreApp80).AsBaseline(),
        Job.Default.WithRuntime(NativeAotRuntime.Net80),
        job.WithToolchain(CsProjCoreToolchain.NetCoreApp60)
        )
      .AddDiagnoser(MemoryDiagnoser.Default);

    if (Environment.OSVersion.Platform == PlatformID.Win32NT)
    {
      config.AddJob(job.WithToolchain(CsProjClassicNetToolchain.Net481));
    }

    config.SummaryStyle = SummaryStyle.Default
      .WithRatioStyle(RatioStyle.Percentage);

    config.AddValidator(JitOptimizationsValidator.FailOnError); // Fail when any of the referenced assemblies are not optimized

    config.WithOrderer(new DefaultOrderer(SummaryOrderPolicy.FastestToSlowest));

    return config;
  }
}
```

So I'm running benchmarks against .NET 8 as my baseline, .NET 8 with Native AOT, .NET 6 and .NET 4.8.1 when on Windows. When running this using `dotnet build -c Release && dotnet run -c Release -f net8.0` the console shows this for the "Native AOT on .NET 8" job:

```
// Build Error: Standard output:

 Standard error:
   Determining projects to restore...
[...]

PR #3097, opened 2026-04-24
chore: modify NativeAotToolchain related settings

filzrev:

**1. Skip creating `NuGet.config` file when custom settings are not contained**
This change is intended to resolve some part of issue #2610.
When using default NuGet feed only, There is no need for packageSourceMapping for `nativeAotNuGetFeed`.

Note: Currently NativeAotToolchain may create following keys on NuGet.config's packageSources.
- `nativeAotNuGetFeed`
- `local`
- `dotnet11`

**2. Modify `NativeAotNuGetFeed` to use default nuget feed**
It's same changes as #2221.

Modify code to use default NuGet feed for `net8.0`/`net9.0`/`net10.0` tfms.
Only preview version tfm need to reference developer feed.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
+++ b/src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
@@ -108,9 +108,7 @@ namespace BenchmarkDotNet.Toolchains.NativeAot
             // Skip creating NuGet.config file if clear tag is not specified, and feeds only contain a default nuget.org feed.
             if (!useNuGetClearTag && Feeds.Count == 1)
             {
-                var feed = Feeds.First();
-                if (feed.Key == NativeAotNuGetFeed
-                 && feed.Value == "https://api.nuget.org/v3/index.json")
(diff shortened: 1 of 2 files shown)


src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -105,6 +105,15 @@ protected override async ValueTask GenerateNuGetConfigAsync(ArtifactsPaths artif
             if (!Feeds.Any())
                 return;
 
+            // Skip creating NuGet.config file if clear tag is not specified, and feeds only contain a default nuget.org feed.
+            if (!useNuGetClearTag && Feeds.Count == 1)
+            {
+                var feed = Feeds.First();
+                if (feed.Key == NativeAotNuGetFeed
+                 && feed.Value == "https://api.nuget.org/v3/index.json")
```
timcassell:
```suggestion
                if (Feeds.TryGetValue(NativeAotNuGetFeed, out var value) && value == "https://api.nuget.org/v3/index.json")
```

timcassell (approved):

PR #3100, opened 2026-04-26
chore: Cleanup run-tests workflow

filzrev:

**1. Modify jsvu `os` parameter to `default`**
Modify `--os` parameter value to use `default` (guess from environment)
As far as I've confirmed, it works without problem with `default` setting . (It can be confirmed by CI logs)

**2. Remove unused tools installation steps**
Remove following tools installation  steps from CI.
- `clang`
- `libkrb5-dev`

These package are listed as Mono prerequisite.
But it seems not be used on CI.
https://benchmarkdotnet.org/articles/configs/toolchains.html?q=libkrb5-dev#wasm

#### TODO:
Following tasks is expected to be on another PR.

**1. Separate `WasmTests` related tool installations to shared actions**
These tool installation can be shared between jobs and `run-tests-selected.yaml`.
And `wasm-tools` workload installation takes about 7 minitus (on `macos-latest`)
So it's better to skip tool installation when it's not needed (On `win-arm64`/`linux-arm64` and wasm related tests are not executed)

**2. Add skip v8 install steps 
Currently,  `jsvu --os=linux64 --engines=v8` command is **silently** failed on `ubuntu-24.04-arm`.
And it seems x64 version is installed on `windows-11-arm`.

On these environment, WasmTests skip tests with following message,
> JSVU does not support ARM on Windows or Linux

So it's better to skip tools installations.

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -32,7 +32,7 @@ jobs:
       - name: Set up v8
         run: |
           npm install jsvu -g
-          jsvu --os=win64 --engines=v8
+          jsvu --os=default --engines=v8
           Add-Content -Path $env:GITHUB_PATH -Value "$env:USERPROFILE\.jsvu\bin"
       - name: Install wasm-tools workload
         run: ./build.cmd install-wasm-tools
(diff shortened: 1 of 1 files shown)


filzrev:
> 2. Remove unused tools installation steps

`clang` & `zlib` dependencies are also required for NativeAOT tests. (used for linker)
And pre-installed version of clang on CI seems be a bit old (`18.1.3`)
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

~~Currently, it seems work without issues.~~
~~But if any problems arise, it need to revert the changes.~~

As far as I've confirmed, there is not clang version restrictions.
https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=linux-ubuntu%2Cnet8#prerequisites

timcassell (approved):

Issue #3098, opened 2026-04-24
ThreadingDiagnoserTests failed randomly when using InProcessTookchain


filzrev:
After PR #3090 is merged.
`CompletedWorkItemCountIsAccurate` test is randomly failed.

Currently, this error is observed on `macos(arm64)` only.

**Error message**
```
CompletedWorkItemCountIsAccurate(toolchain: InProcessEmitToolchain) [FAIL]
       Assert.Equal() Failure: Values are not within 2 decimal places
       Expected: 1 (rounded from 1)
       Actual:   1.02 (rounded from 1.015625)
```

**Logs**
- https://github.com/dotnet/BenchmarkDotNet/actions/runs/24639422508/job/72040644748
- https://github.com/dotnet/BenchmarkDotNet/actions/runs/24885975618/job/72865987114

---
It can be resolved by reduce precision when comparing metrics value. https://github.com/dotnet/BenchmarkDotNet/blob/d11792c6c58da1e00bb44d6458ca015d5928bab9/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs#L182.

Though, it need to determine root cause of this problems.

filzrev:
This issue is suppressed with following temporary workaround code.
https://github.com/dotnet/BenchmarkDotNet/blob/91f71ca99230115452af0e104d53ad3a7258b2ff/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs#L188-L190

It seems there are some background operations running after warmup phase that specific to macOS.
Related issue: #2779

filzrev:
It also occurred on macos(x64)
https://github.com/dotnet/BenchmarkDotNet/actions/runs/26924753402

```
LockContentionCountIsAccurate(toolchain: InProcessEmitToolchain) [FAIL]
  Assert.Equal() Failure: Values are not within 1 decimal place
  Expected: 1 (rounded from 1)
  Actual:   1.1000000000000001 (rounded from 1.125)
```

In that case, `precision = 1` seems not enough.

timcassell:
It seems likely to be caused by xunit v2. See https://github.com/dotnet/BenchmarkDotNet/pull/3065#issuecomment-5013263049

filzrev:
I've removed temporary workaround setting on xunit.v3 PR branch.

Close this issue,
If this flaky test issue persists in xunitv3, I'll reopen the ticket.

PR #3101, opened 2026-04-28
chore: Add temporary workaround for `ThreadingDiagnoserTests::CompletedWorkItemCountIsAccurate` on macos

filzrev:
This PR add temporary workaround for #3098

Currently, on macOS environment.
`ThreadingDiagnoserTests::CompletedWorkItemCountIsAccurate` test failed randomly.
Because some background operation executed during running benchmarks.

At this point, the root cause remains unknown.
So this PR **temporary** suppress CI failure by reducing assert value of floating point precision.

Note: Temporary modify precision to `0` when it's not running with InProcessToolchain.
It'll be reverted if it's not works as expected

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
@@ -175,15 +175,13 @@ namespace BenchmarkDotNet.IntegrationTests
 
                 var metric = selectedReport.Metrics.Single(m => m.Key == assertion.Value.metricName);
 
-                int precision = 0;
-                if (toolchain.IsInProcess)
-                {
-                    // precision is set to 2 because CoreCLR might schedule some work item on it's own and hence affect the results..
-                    // precision = 3 is not enough (e.g., sometimes the actual value may be equal 1.0009765625 while the expected value is 1.0)
-                    precision = !OsDetector.IsMacOS()
-                        ? 2
-                        : 1; // On macos, there are some additional background operations (https://github.com/dotnet/BenchmarkDotNet/issues/3098)
-                }
+                // precision is set to 2 because CoreCLR might schedule some work item on it's own and hence affect the results..
+                // precision = 3 is not enough (e.g., sometimes the actual value may be equal 1.0009765625 while the expected value is 1.0)
+                int precision = 2;
+
+                // On macos, there are some additional background operations (https://github.com/dotnet/BenchmarkDotNet/issues/3098)
+                if (OsDetector.IsMacOS())
+                    precision = 1;
 
                 Assert.Equal(assertion.Value.expectedValue, metric.Value.Value, precision);
             }
--- a/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
@@ -55,7 +55,7 @@ namespace BenchmarkDotNet.IntegrationTests
                 throw;
             }
 
-            AssertStats(summary, new Dictionary<string, (string metricName, double expectedValue)>
+            AssertStats(summary, toolchain, new Dictionary<string, (string metricName, double expectedValue)>
             {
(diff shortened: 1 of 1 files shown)


tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
```diff
@@ -169,16 +170,24 @@ private IConfig CreateConfig(IToolchain toolchain)
...
+                if (toolchain.IsInProcess)
+                {
+                    // precision is set to 2 because CoreCLR might schedule some work item on it's own and hence affect the results..
+                    // precision = 3 is not enough (e.g., sometimes the actual value may be equal 1.0009765625 while the expected value is 1.0)
+                    precision = !RuntimeInformation.IsOSPlatform(OSPlatform.OSX)
+                        ? 2
+                        : 1; // On macos, there are some additional background operations (https://github.com/dotnet/BenchmarkDotNet/issues/3098)
+                }
+
                 Assert.Equal(assertion.Value.expectedValue, metric.Value.Value, precision: 2);
```
timcassell: You're not actually passing the calculated precision.

tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
```diff
@@ -169,16 +170,24 @@ private IConfig CreateConfig(IToolchain toolchain)
...
                 var metric = selectedReport.Metrics.Single(m => m.Key == assertion.Value.metricName);
 
-                // precision is set to 2 because CoreCLR might schedule some work item on it's own and hence affect the results..
-                // precision = 3 is not enough (e.g., sometimes the actual value may be equal 1.0009765625 while the expected value is 1.0)
+                int precision = 0;
+                if (toolchain.IsInProcess)
+                {
+                    // precision is set to 2 because CoreCLR might schedule some work item on it's own and hence affect the results..
+                    // precision = 3 is not enough (e.g., sometimes the actual value may be equal 1.0009765625 while the expected value is 1.0)
+                    precision = !RuntimeInformation.IsOSPlatform(OSPlatform.OSX)
```
timcassell: nit
```suggestion
                    precision = !OsDetector.IsMacOS()
```

timcassell (changes requested):

tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
filzrev: I've fixed wrong parameter on latest commit

tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
filzrev: I've applied reviewed changed on latest commit.

tests/BenchmarkDotNet.IntegrationTests/ThreadingDiagnoserTests.cs
```diff
@@ -167,17 +167,25 @@ private IConfig CreateConfig(IToolchain toolchain)
...
             foreach (var assertion in assertions)
             {
                 var selectedReport = summary.Reports.Single(report => report.BenchmarkCase.DisplayInfo.Contains(assertion.Key));
 
                 var metric = selectedReport.Metrics.Single(m => m.Key == assertion.Value.metricName);
 
-                // precision is set to 2 because CoreCLR might schedule some work item on it's own and hence affect the results..
-                // precision = 3 is not enough (e.g., sometimes the actual value may be equal 1.0009765625 while the expected value is 1.0)
-                Assert.Equal(assertion.Value.expectedValue, metric.Value.Value, precision: 2);
+                int precision = 0;
```
timcassell: Default precision 0 for out-of-proc toolchains? I think it should still be 2.

filzrev: I've modified logics to use precision `2` as default value on latest commit.

As far as I've tested,
Test always returns exactly matched value when using  out-of-proc toolchain.
Though, to exclude long-running benchmark side effects (e.g. Garbage Collection).
It should use precision `2`.

timcassell (approved):

PR #3103, opened 2026-04-29
Fix async benchmarks with MemoryRandomization

DrewScoggins:
When `MemoryRandomization=true`, `Engine.RunCore` calls `RandomizeManagedHeapMemory` between iterations, which invokes `GlobalCleanupAction` followed by `GlobalSetupAction`. The generated `__GlobalCleanup` previously called
`workloadContinuerAndValueTaskSource?.Complete()` to signal the async workload loop to exit, but left the field non-null in a terminal state. The next iteration's `WorkloadActionNoUnroll` then called `Continue()` on the same source, which calls `SetResult(false)` on a `ManualResetValueTaskSourceCore` that had already been signaled, throwing `InvalidOperationException` from `SignalCompletion`. This broke every async benchmark that opted into memory randomization.

Fix:
- Add `WorkloadValueTaskSource.IsContinuerPending` so the generated cleanup can detect whether the workload loop is still awaiting the next iteration. Calling `Complete()` when not pending would mask the original exception (e.g. when the workload itself threw and the loop has already exited via the catch block) with a double-signal `InvalidOperationException`.
- Update `PrependExtraGlobalCleanupImpl` to:
  1. Only call `Complete()` when the continuer source is pending.
  2. Set `workloadContinuerAndValueTaskSource = null` after signalling.

The existing `WorkloadActionNoUnroll` codegen already handles the null case by allocating a fresh `WorkloadValueTaskSource` and restarting `__StartWorkload()`, so iterations after a between-iteration `GlobalCleanup` now run against a fresh, pending source.


DrewScoggins:
This is output from copilot, so not sure if this is the correct approach.

timcassell:
It would be good to add a test for this case.

> The existing `WorkloadActionNoUnroll` codegen already handles the null case by allocating a fresh `WorkloadValueTaskSource` and restarting `__StartWorkload()`, so iterations after a between-iteration `GlobalCleanup` now run against a fresh, pending source.

This will affect memory allocation measurements. I'm not sure the best way to handle it. My initial thought was just implement a custom `IClock` that does the diagnostic actions before starting the true clock, but it doesn't really work. Maybe we could change the API to use a new interface to make it work. I'm also open to any other ideas. [Edit] Actually looking at the `IClock` interface and `Start` extension, they just call `IClock.GetTimestamp()`, so we could theoretically implement it with that, though it does feel a bit hacky.

timcassell:
Maybe a simpler way would be to create the workload continuer in `__IterationSetup`, which is already called before the diagnosers, and then the `WorkloadActionNoUnroll` can simply call `Continue()` branchlessly. And then we can even also complete it in `__IterationCleanup` instead of `__GlobalCleanup`.

[Edit] Or more simply, we can create it in `__GlobalSetup` unconditionally, complete it in `__GlobalCleanup` unconditionally. That way for non-MemoryRandomization jobs it won't allocate between iterations.

DrewScoggins:
Decided to go with the second suggestion.

src/BenchmarkDotNet/Code/DeclarationsProvider.cs
```diff
@@ -182,10 +186,35 @@ public override string[] GetExtraFields() =>
             "public long invokeCount;"
         ];
 
+        protected override string PrependExtraGlobalSetupImpl(string impl)
```
timcassell: It should be Append to setup, Prepend to cleanup.

src/BenchmarkDotNet/Code/DeclarationsProvider.cs
```diff
@@ -182,10 +186,35 @@ public override string[] GetExtraFields() =>
...
+                        {{impl}}
+            """;
+
         protected override string PrependExtraGlobalCleanupImpl(string impl)
-            => $"""
-            this.__fieldsContainer.workloadContinuerAndValueTaskSource?.Complete();
-                        {impl}
+            => $$"""
+            if (this.__fieldsContainer.workloadContinuerAndValueTaskSource != null
+                            && this.__fieldsContainer.workloadContinuerAndValueTaskSource.IsContinuerPending)
```
timcassell: The guard is unnecessary.

src/BenchmarkDotNet/Engines/WorkloadValueTaskSource.cs
```diff
@@ -26,6 +26,12 @@ public ValueTask<bool> GetIsComplete()
     public void Complete()
         => continuerSource.SetResult(true);
 
+    // Used by __GlobalCleanup to safely call Complete only when the async workload loop
+    // is awaiting the next iteration. When the workload threw or the loop has already exited,
+    // calling SetResult(true) again on an already-signaled source would throw.
+    public bool IsContinuerPending
+        => continuerSource.GetStatus(continuerSource.Version) == ValueTaskSourceStatus.Pending;
```
timcassell: Unnecessary.

timcassell (changes requested):
Also need to update both InProcess toolchains.

timcassell (approved):

PR #3109, opened 2026-05-03
chore: Update macos(x64) CI image to `macos-26-intel` and increase build timeout

filzrev:

#### 1. Migrate `macos(x64)` CI image from `macos-15-intel` to `macos-26-intel`
`macos-26-intel` is expected to be last version that support `Intel Mac`

#### 2. Increase build timeout value that randomly failed on macos(x64) image
Sometimes `macos(x64)`  takes over 3h and randomly failed with build timeout error.
So I've increased build timeout for failed tests.

#### Other Notes
CI on `macos(x64)` is about 2x slower than other images and takes over 2h-3h to complete.
It might be better to skip some integration tests that takes long times to execute.

### DIFF
--- a/.github/workflows/run-tests-selected.yaml
+++ b/.github/workflows/run-tests-selected.yaml
@@ -15,7 +15,7 @@ on:
           - macos-latest
           - windows-11-arm
           - ubuntu-24.04-arm
-          - macos-15-intel
+          - macos-26-intel
       project:
         type: string
         description: Specify test project path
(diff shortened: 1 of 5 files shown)


timcassell (approved):

filzrev:
After this PR is merged.
`test-macos (x64)` CI randomly failed and logs are recorded incomplete state.
https://github.com/dotnet/BenchmarkDotNet/actions/runs/25769366932/job/75688976383

It looks like caused by `macos-26-intel` OS image issue.
https://github.com/actions/runner-images/issues/13882

PR #3116, opened 2026-05-07
Ignore hidden directories in CsProjGenerator (fixes #3110)

### LINKED ISSUE #3110: Make the IgnoredDirectoryNames from CsProjGenerator.Helpers configurable
AI-tools are everywhere. I use e.g. Claude and the **worktree** feature. So I have inside my project folder a `.claude\worktrees` folder. When I use BenchmarkDotNet. This worktree folder will be picked up, and I get an Exception / Error. The project name needs to be unique.

> // Generate Exception: Found more than one matching project file for XXX.Benchmark in C:\ws\my-project and its subfolders: 'C:\ws\my-project\XXX.Benchmark\XXX.Benchmark.csproj','C:\ws\my-project\.claude\worktrees\Download-Benmark\XXX.Benchmark\XXX.Benchmark.csproj'. Benchmark project names needs to be unique.

Thanks to open source, I found here, that there is a value of "excluded" folders. 

In the class `CsProjGenerator` -> `Helpers` there is a private field [IgnoredDirectoryNames](https://github.com/dotnet/BenchmarkDotNet/blob/7e583dc9dd1cd84040ea2fb1f6170dfe27d2c50f/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs#L394-L400).

**My ideas**

* It would be nice to make that filed configurable, or create an option to add "additional" folders via config or Benchmark-Setup.

* Or just add some more default-folders to exclude, like ".vscode" and ".claude" ... ".github", ".gitlab".

 I am on Win 11 Pro, DotNet 8, BenchmarkDotNet v0.15.8.

Many Thanks for your great work

**Update** what I have tried:

* AddJob (AddJob(Job.Default.WithToolchain(InProcessEmitToolchain.Instance))) was not unblocking me
* Create a directory junctions/symlinks, move the .claude\worktree outside the real tree,  do not solve the problem

NehaGitHubAcc:
This PR addresses [Issue #3110](https://github.com/dotnet/BenchmarkDotNet/issues/3110) by updating the CsProjGenerator to automatically ignore hidden directories and directories starting with a dot (e.g., .claude, .git).

Changes
Automatic Filtering: Instead of manually expanding a hardcoded list of directory names, I implemented a ShouldIgnoreDirectory helper that checks for the FileAttributes.Hidden flag and the . prefix.

Case Insensitivity: Updated the IgnoredDirectoryNames and ProjectExtensions sets to use StringComparer.OrdinalIgnoreCase to prevent issues on different file systems.

Unit Tests: Added a new test case FindProjectFileIgnoresDotAndHiddenDirectories to ensure that project files inside hidden folders (like .claude/worktrees) are correctly ignored, preventing duplicate project file errors.

Verification
[x] Verified that the new test case passes.

[x] Confirmed the logic works for both Windows (hidden attribute) and Unix-like (dot-prefix) conventions.

Fixes #3110

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
@@ -389,22 +389,25 @@ namespace BenchmarkDotNet.Toolchains.CsProj
             => HashCode.Combine(TargetFrameworkMoniker, RuntimeFrameworkVersion, CliPath, PackagesPath);
     }
 
-    file static class Helpers
+    internal static class Helpers
     {
-        private static readonly HashSet<string> IgnoredDirectoryNames = new(StringComparer.Ordinal)
+        private static readonly HashSet<string> IgnoredDirectoryNames = new(StringComparer.OrdinalIgnoreCase)
         {
-            ".git",
-            ".vs",
             "bin",
             "obj",
         };
 
-        private static readonly HashSet<string> ProjectExtensions = new(StringComparer.Ordinal)
+        private static readonly HashSet<string> ProjectExtensions = new(StringComparer.OrdinalIgnoreCase)
         {
             ".csproj",
             ".fsproj",
             ".vbproj"
         };
+
+        private static bool ShouldIgnoreDirectory(DirectoryInfo directory)
+            => IgnoredDirectoryNames.Contains(directory.Name)
+               || directory.Name.StartsWith(".", StringComparison.Ordinal)
+               || directory.Attributes.HasFlag(FileAttributes.Hidden);
 
         public static FileInfo FindProjectFile(DirectoryInfo rootDirectory, string projectName)
         {
@@ -448,7 +451,7 @@ namespace BenchmarkDotNet.Toolchains.CsProj
                 // 2. Handle sub directories.
                 foreach (var dir in subDirectories)
                 {
-                    if (IgnoredDirectoryNames.Contains(dir.Name))
+                    if (ShouldIgnoreDirectory(dir))
                         continue;
 #if NETSTANDARD2_0
                     // Ignore reparse point / symlink to avoid infinite loops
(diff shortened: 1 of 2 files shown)


NehaGitHubAcc:
@dotnet-policy-service agree

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -389,23 +389,26 @@ public override int GetHashCode()
...
         {
             ".csproj",
             ".fsproj",
             ".vbproj"
         };
 
+        private static bool ShouldIgnoreDirectory(DirectoryInfo directory)
+            => IgnoredDirectoryNames.Contains(directory.Name)
+               || directory.Name.StartsWith(".", StringComparison.Ordinal)
+               || directory.Attributes.HasFlag(FileAttributes.Hidden);
```
timcassell: It looks like you can add that to the `DirectoryEnumerationOptions` below, and wrap this with `#if NETSTANDARD2_0`, and move the `ReparsePoint` check here also.

NehaGitHubAcc:
Hi @timcassell, I've pushed the updates to address your feedback.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -481,7 +488,11 @@ private static IEnumerable<DirectoryInfo> GetSubDirectories(DirectoryInfo curren
         {
             try
             {
+#if NETSTANDARD2_0
                 return currentDir.EnumerateDirectories();
+#else
+                return currentDir.EnumerateDirectories("*", DirectoryEnumerationOptions);
+#endif
```
timcassell: This is already within `#if NETSTANDARD2_0` so this change is superfluous.

src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -448,7 +455,7 @@ private static IEnumerable<FileInfo> EnumerateProjectFiles(DirectoryInfo rootDir
                 // 2. Handle sub directories.
                 foreach (var dir in subDirectories)
                 {
-                    if (IgnoredDirectoryNames.Contains(dir.Name))
+                    if (ShouldIgnoreDirectory(dir))
                         continue;
 #if NETSTANDARD2_0
```
timcassell: This can be removed now.

timcassell (approved):
Thanks @NehaGitHubAcc

PR #3117, opened 2026-05-07
Defer BenchmarkDotNetSynchronizationContext install past discovery

DrewScoggins:
Below is Copilot's analysis of another issue that we are hitting with moving forward to the BDN that contains the async refactor. If this is us using `ParamsSource` and `ArgumentsSource` wrong, I am happy to make the changes in our code, but I was not sure what the right solution was.

The sync entrypoints BenchmarkSwitcher.Run/RunAll/RunAllJoined and BenchmarkRunner.Run<T> overloads were installing
BenchmarkDotNetSynchronizationContext (a single-threaded post/pump context) before benchmark discovery. Discovery evaluates user [ParamsSource]/[ArgumentsSource] members. If any of them perform sync-over-async work (e.g. someTask.GetAwaiter().GetResult()), the async continuation captures the BDN ctx and is queued; the pump thread is blocked inside .GetResult(); classic single-threaded deadlock.

Repro in dotnet/performance:
    SslStreamTests.GetTls13Support() does HandshakeAsync(...).GetAwaiter().GetResult()
    inside a [ArgumentsSource(nameof(TlsProtocols))] code path. With
    BDN 0.16.0 (post Async Refactor 4229d4fc), `--list flat` hangs
    forever. Pre-Refactor sync Run() did not install the SyncCtx and
    the same code worked for years.

Fix: defer SyncCtx install. Sync entrypoints now route through a tiny ExecuteSynchronously<T>(ValueTask<T>) helper that:
  - returns immediately if the task is already completed (covers --list/--info/--help/no-benchmarks early exits)
  - otherwise installs the BDN ctx and pumps until completion (preserves SyncCtx semantics for execution awaits)

Internal awaits use ConfigureAwait(false) so nothing is actually posted to the BDN ctx in modern code; the SyncCtx is essentially decorative during the pump.

### DIFF
--- a/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
+++ b/src/BenchmarkDotNet/Running/BenchmarkRunnerDirty.cs
@@ -16,51 +16,42 @@ namespace BenchmarkDotNet.Running
     {
         [PublicAPI]
         public static Summary Run<T>(IConfig? config = null, string[]? args = null)
-        {
-            using var context = BenchmarkSynchronizationContext.CreateAndSetCurrent();
-            return context.ExecuteUntilComplete(RunAsync<T>(config, args));
-        }
+            => ExecuteSynchronously(RunAsync<T>(config, args));
 
         [PublicAPI]
         public static Summary Run(Type type, IConfig? config = null, string[]? args = null)
-        {
-            using var context = BenchmarkSynchronizationContext.CreateAndSetCurrent();
-            return context.ExecuteUntilComplete(RunAsync(type, config, args));
-        }
+            => ExecuteSynchronously(RunAsync(type, config, args));
 
         [PublicAPI]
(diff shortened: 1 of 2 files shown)


timcassell:
This change causes the async chain to lose the context, causing `ConfigureAwait(true)` to be incorrect. Sync-over-async is well-known to be prone to deadlock and is a code smell, I don't think we should be encouraging its use. Rather the correct fix is to update `ParamsSource` and `ArgumentsSource` to support `IAsyncEnumerable`.

timcassell:
To unblock you now, you can switch to the `Run*Async` APIs that do not install the sync context.

DrewScoggins:
Sounds good. I appreciate all your time looking at these PRs, I know there have been a lot lately, and I don't have the deepest understanding of BDN. So appreciate your help.

DrewScoggins:
Going to go ahead and close this.

PR #3126, opened 2026-05-12
chore: Reduce CI execution time by changing some tests to use InProcessToolchain

filzrev:
This PR intended to reduce CI execution time by using `InProcessToolchain` when possible.

### What's changed in this PR

**1. Add shared configs to `TestConfigs.cs`**
Add following test configs that used for faster test execution.
- SingleRunInProcessConfig
- SingleRunOutOfProcessConfig

**2. Add `Helpers/CaptureConsoleHelper.cs`**
This helper class is used to capture console logs that are outputted by benchmark.

It's required because `StandardOutput` is not recorded to Summary/Report instance. when running benchmark with `InProcessToolchain`.

As a side effect, captured log contains benchmark execution logs also.
Though, It's not affects test results when assert contents line by line.

**3. Modify some test code to use `InProcessToolchain`**

Modify following tests to use `SingleRunInProcessConfig` when possible
- AsyncBenchmarksTests
- AttributesTests
- BaselineRatioColumnTest
- BenchmarkSwitcherTest (Partially)
- ParamsTests

**4. Modify some test code to use `SingleRunOutOfProcessConfig`**
Some `BenchmarkSwitcherTest` tests can't convert to use `InProcessToolchain` by following reason.
- Test that using `[DryJob]` attribute. 
- Test that pass `--job` command line argument.
- Test that using `--resume` option (It seems not works because InProcessToolchain don't output logs for resume) 

So these test need to use `SingleRunOutOfProcessConfig` instead. (It's basically same behavior as original code)

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/BenchmarkSwitcherTest.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/BenchmarkSwitcherTest.cs
@@ -20,13 +20,19 @@ namespace BenchmarkDotNet.IntegrationTests
 
         public BenchmarkSwitcherTest(ITestOutputHelper output) => Output = output;
 
+        private IConfig GetRunInProcessConfig()
+            => new SingleRunInProcessConfig(Output);
+
+        private IConfig GetRunOutOfProcessConfig()
+           => new SingleRunOutOfProcessConfig(Output);
+
         [FactEnvSpecific(
             "When CommandLineParser wants to display help, it tries to get the Title of the Entry Assembly which is an xunit runner, which has no Title and fails..",
             EnvRequirement.DotNetCoreOnly)]
         public void WhenInvalidCommandLineArgumentIsPassedAnErrorMessageIsDisplayedAndNoBenchmarksAreExecuted()
         {
-            var logger = new OutputLogger(Output);
-            var config = ManualConfig.CreateEmpty().AddLogger(logger);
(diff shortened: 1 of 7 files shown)


tests/BenchmarkDotNet.IntegrationTests/BenchmarkSwitcherTest.cs
```diff
@@ -262,7 +262,7 @@ public void WhenJobIsDefinedInTheConfigAndArgumentsDontContainJobArgumentOnlySin
             Assert.Single(results);
             Assert.Single(results.SelectMany(r => r.BenchmarksCases));
             Assert.Single(results.SelectMany(r => r.BenchmarksCases.Select(bc => bc.Job)));
-            Assert.True(results.All(r => r.BenchmarksCases.All(bc => bc.Job == Job.Dry)));
+            Assert.True(results.All(r => r.BenchmarksCases.All(bc => bc.Job.Id == "Dry")));
```
filzrev, line 265: It need to compare job by `Id`.
Because `SingleRunInProcessConfig` using `Job.Dry` based config.

timcassell:
I'm not sure we should stop testing in the default toolchain. If anything, we should be testing both toolchains.

filzrev:
> I'm not sure we should stop testing in the default toolchain. If anything, we should be testing both toolchains.

Understood. 
I’d like to update tests run on both toolchains on another PR.

PR #3133, opened 2026-05-14
Catch TimeoutException in WasmExecutor to restore pre-async-refactor …

DrewScoggins:
…behavior

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
+++ b/src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
@@ -209,7 +209,7 @@ namespace BenchmarkDotNet.Toolchains.MonoWasm
                     process.TrySetAffinity(benchmarkCase.Job.Environment.Affinity, logger);
                 }
 
-try
+                try
                 {
 #pragma warning disable CA2016 // Forward the 'CancellationToken' parameter to methods
                     await broker.ProcessData(cancellationToken)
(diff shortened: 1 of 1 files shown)


DrewScoggins:
PTAL ~~@alinasmirnova~~ Apologies, that was the wrong person @ilonatommy

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
```diff
@@ -209,11 +209,21 @@ private static async ValueTask<ExecuteResult> Execute(Process process, Benchmark
...
+                        .AsTask()
+                        .WaitAsync(TimeSpan.FromMinutes(wasmRuntime.ProcessTimeoutMinutes)).ConfigureAwait(false);
 #pragma warning restore CA2016 // Forward the 'CancellationToken' parameter to methods
+                }
+                catch (TimeoutException)
+                {
+                    // Preserve pre-async-refactor behavior: log a message and force-kill the
+                    // process tree rather than propagating the timeout to the caller.
+                    logger.WriteLineInfo($"// The benchmarking process did not finish within {wasmRuntime.ProcessTimeoutMinutes} minutes, it's going to get force killed now.");
+                    process.KillTree();
```
timcassell: `ProcessCleanupHelper` already does it. I think you can extend the try block around the `!process.WaitForExit`, and move the broker read to after the catch.

src/BenchmarkDotNet/Toolchains/MonoWasm/WasmExecutor.cs
```diff
@@ -209,19 +209,28 @@ private static async ValueTask<ExecuteResult> Execute(Process process, Benchmark
                     process.TrySetAffinity(benchmarkCase.Job.Environment.Affinity, logger);
                 }
 
+try
```
timcassell:
```suggestion
                try
```

timcassell (approved):

PR #3138, opened 2026-05-16
Extend OpenMetrics exporter to include allocations

martincostello:
- Extend the OpenMetrics exporter to include a gauge for allocated bytes if the memory diagnoser is enabled.
- Only emit GC metrics if the memory diagnoser is enabled.


timcassell:
`gc_genX_collections` are collected unconditionally, I think they should also be conditional.

martincostello:
What should they be conditional on? Would that be a breaking change given that's something that's already shipped?

Or did you mean for `allocated_bytes`?

timcassell:
> What should they be conditional on?

The same thing. Both allocated bytes and genX collections are gathered with the MemoryDiagnoser.

> Would that be a breaking change given that's something that's already shipped?

Yes, but it's likely reporting incorrect data if the MemoryDiagnoser is not attached. And next version already has breaking changes, it's fine to add another.

PR #3141, opened 2026-05-23
chore: Update package dependencies and .NET SDK version

filzrev:

1. ~~Update `AsmResolver.DotNet` version to `6.0.0`~~
2. Update .NET SDK version to `10.0.300`
3. Update `Cake.Frosting` package to 6.2.0 (to suppress vulnerable warnings on build)
4. Remove `Microsoft.Bcl.AsyncInterfaces` dependency from `Samples.FSharp` project.
5. Update other outdated packages. (As far as I've confirmed, there is no breaking changes on release notes)

### DIFF
--- a/build/BenchmarkDotNet.Build/BenchmarkDotNet.Build.csproj
+++ b/build/BenchmarkDotNet.Build/BenchmarkDotNet.Build.csproj
@@ -7,8 +7,8 @@
     <NoWarn>$(NoWarn);CA2007</NoWarn>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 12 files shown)


timcassell:
> 1. Update `AsmResolver.DotNet` version to `6.0.0`

I believe the new version added a `ReferencePathAssemblyResolver` that we can use instead of rolling our own.

filzrev:
> I believe the new version added a ReferencePathAssemblyResolver that we can use instead of rolling our own.

I've reverted `AsmResolver.DotNet` update on latest commit.
It's expected to be updated on separated PR.

timcassell (approved):

PR #3143, opened 2026-05-23
Update JIT stage timing heuristic

### LINKED ISSUE #3114: WorkloadJitting executed more than once
I am trying to benchmark some of the new `Process` APIs using latest preview from the MyGet feed (what we offer at nuget.org does not support .NET 11 yet). When running following benchmark:

```csharp
public class GlobalLock
{
    private ProcessStartInfo info;

    [Params(true, false)]
    public bool SetInheritedHandles { get; set; }

    [GlobalSetup]
    public void Setup()
    {
        info = OperatingSystem.IsWindows()
            ? new("cmd.exe", ["/c", "exit 42"])
            : new("sh", ["-c", "exit 42"]);

        info.InheritedHandles = SetInheritedHandles ? [] : null;
    }

    [Benchmark]
    public ParallelLoopResult Run() => Parallel.For(0, 1_000, (_, _) => _ = Process.Run(info));
}
```

With explicitly asking for running the benchmark once per iteration, for 10 iterations, with a single warmup:

```cmd
dotnet run -c Release -f net11.0 --filter "*Lock*" --launchCount 1 --iterationCount 10 --invocationCount 1 --unrollFactor 1 --warmupCount 1
```

I can see `WorkloadJitting` being printed more than once:

```log
// BenchmarkDotNet v0.16.0-nightly.20260505.517
// Runtime=.NET 11.0.0 (11.0.0-preview.5.26255.101, 11.0.26.25601), X64 RyuJIT x86-64-v3
// GC=Concurrent Workstation
// HardwareIntrinsics=AVX2+BMI1+BMI2+F16C+FMA+LZCNT+MOVBE,AVX,SSE3+SSSE3+SSE4.1+SSE4.2+POPCNT,X86Base+SSE+SSE2,AES+PCLMUL VectorSize=256
// Job: Job-HFOVVH(InvocationCount=1, IterationCount=10, LaunchCount=1, UnrollFactor=1, WarmupCount=1)

WorkloadJitting  1: 1 op, 4193300600.00 ns, 4.1933 s/op
WorkloadJitting  2: 1 op, 4045230800.00 ns, 4.0452 s/op
WorkloadJitting  3: 1 op, 4034639500.00 ns, 4.0346 s/op
WorkloadJitting  4: 1 op, 4043227900.00 ns, 4.0432 s/op

WorkloadWarmup   1: 1 op, 4118601300.00 ns, 4.1186 s/op

// BeforeActualRun
WorkloadActual   1: 1 op, 3993833200.00 ns, 3.9938 s/op
WorkloadActual   2: 1 op, 4038450400.00 ns, 4.0385 s/op
WorkloadActual   3: 1 op, 4028608700.00 ns, 4.0286 s/op
WorkloadActual   4: 1 op, 3990243600.00 ns, 3.9902 s/op
WorkloadActual   5: 1 op, 4001137100.00 ns, 4.0011 s/op
WorkloadActual   6: 1 op, 4014296800.00 ns, 4.0143 s/op
WorkloadActual   7: 1 op, 4003703800.00 ns, 4.0037 s/op
WorkloadActual   8: 1 op, 4004910800.00 ns, 4.0049 s/op
WorkloadActual   9: 1 op, 4008993200.00 ns, 4.0090 s/op
WorkloadActual  10: 1 op, 4058972600.00 ns, 4.0590 s/op

// AfterActualRun
WorkloadResult   1: 1 op, 3993833200.00 ns, 3.9938 s/op
WorkloadResult   2: 1 op, 4038450400.00 ns, 4.0385 s/op
WorkloadResult   3: 1 op, 4028608700.00 ns, 4.0286 s/op
WorkloadResult   4: 1 op, 3990243600.00 ns, 3.9902 s/op
WorkloadResult   5: 1 op, 4001137100.00 ns, 4.0011 s/op
WorkloadResult   6: 1 op, 4014296800.00 ns, 4.0143 s/op
WorkloadResult   7: 1 op, 4003703800.00 ns, 4.0037 s/op
WorkloadResult   8: 1 op, 4004910800.00 ns, 4.0049 s/op
WorkloadResult   9: 1 op, 4008993200.00 ns, 4.0090 s/op
WorkloadResult  10: 1 op, 4058972600.00 ns, 4.0590 s/op
```

```log
WorkloadJitting  1: 1 op, 1868394200.00 ns, 1.8684 s/op
WorkloadJitting  2: 1 op, 1809145900.00 ns, 1.8091 s/op
WorkloadJitting  3: 1 op, 1818846700.00 ns, 1.8188 s/op
WorkloadJitting  4: 1 op, 1873981600.00 ns, 1.8740 s/op
WorkloadJitting  5: 1 op, 1889514800.00 ns, 1.8895 s/op
WorkloadJitting  6: 1 op, 1836730200.00 ns, 1.8367 s/op
WorkloadJitting  7: 1 op, 2030291800.00 ns, 2.0303 s/op

WorkloadWarmup   1: 1 op, 1935149300.00 ns, 1.9351 s/op

// BeforeActualRun
WorkloadActual   1: 1 op, 1890218000.00 ns, 1.8902 s/op
WorkloadActual   2: 1 op, 1879179200.00 ns, 1.8792 s/op
WorkloadActual   3: 1 op, 2079051200.00 ns, 2.0791 s/op
WorkloadActual   4: 1 op, 2031099700.00 ns, 2.0311 s/op
WorkloadActual   5: 1 op, 2016355900.00 ns, 2.0164 s/op
WorkloadActual   6: 1 op, 1997384600.00 ns, 1.9974 s/op
WorkloadActual   7: 1 op, 1966403300.00 ns, 1.9664 s/op
WorkloadActual   8: 1 op, 1988965700.00 ns, 1.9890 s/op
WorkloadActual   9: 1 op, 1885926700.00 ns, 1.8859 s/op
WorkloadActual  10: 1 op, 1848450100.00 ns, 1.8485 s/op

// AfterActualRun
WorkloadResult   1: 1 op, 1890218000.00 ns, 1.8902 s/op
WorkloadResult   2: 1 op, 1879179200.00 ns, 1.8792 s/op
WorkloadResult   3: 1 op, 2079051200.00 ns, 2.0791 s/op
WorkloadResult   4: 1 op, 2031099700.00 ns, 2.0311 s/op
WorkloadResult   5: 1 op, 2016355900.00 ns, 2.0164 s/op
WorkloadResult   6: 1 op, 1997384600.00 ns, 1.9974 s/op
WorkloadResult   7: 1 op, 1966403300.00 ns, 1.9664 s/op
WorkloadResult   8: 1 op, 1988965700.00 ns, 1.9890 s/op
WorkloadResult   9: 1 op, 1885926700.00 ns, 1.8859 s/op
WorkloadResult  10: 1 op, 1848450100.00 ns, 1.8485 s/op
```

@timcassell @AndreyAkinshin have we recently changed something around this particular heuristic? I would expect JITting to be executed just once.

timcassell:
Bails out after the second invocation if it detects a long-running benchmark using the same calculation as the pilot stage, instead of continuing invokes for 10 seconds.

Fixes #3114

### DIFF
--- a/src/BenchmarkDotNet/Attributes/Mutators/JitTieringAttribute.cs
+++ b/src/BenchmarkDotNet/Attributes/Mutators/JitTieringAttribute.cs
@@ -1,0 +1,9 @@
+using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Attributes;
+
+/// <inheritdoc cref="RunMode.JitTieringMode"/>
+public class JitTieringAttribute(JitTieringMode mode) : JobMutatorConfigBaseAttribute(Job.Default.WithJitTieringMode(mode))
+{
+}
--- a/docs/articles/guides/console-args.md
+++ b/docs/articles/guides/console-args.md
@@ -313,6 +313,7 @@ dotnet run -c Release -- --filter * --runtimes net6.0 net8.0 --statisticalTest 5
 * `--maxWidth`                Max parameter column width, the default is 20.
 * `--envVars`                 Colon separated environment variables (key:value)
 * `--memoryRandomization`     Specifies whether Engine should allocate some random-sized memory between iterations. It makes [GlobalCleanup] and [GlobalSetup] methods to be executed after every iteration.
+* `--jitTieringMode`          (Default: Auto) Controls the behavior of the JIT stage when tiering is enabled. Auto/Force/Skip.
 * `--wasmEngine`              (Default: v8) Specifies the executable (in PATH) or full path to a java script engine used to run the benchmarks, used by Wasm toolchain.
 * `--wasmArgs`                (Default: --expose_wasm) Arguments for the javascript engine used by Wasm toolchain.
 * `--customRuntimePack`       Path to a custom runtime pack. Only used for wasm/MonoAotLLVM currently.
--- a/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs
@@ -198,6 +198,9 @@ namespace BenchmarkDotNet.ConsoleArguments
         [Option("memoryRandomization", Required = false, HelpText = "Specifies whether Engine should allocate some random-sized memory between iterations. It makes [GlobalCleanup] and [GlobalSetup] methods to be executed after every iteration.")]
         public bool MemoryRandomization { get; set; }
 
+        [Option("jitTieringMode", Required = false, Default = JitTieringMode.Auto, HelpText = "Controls the behavior of the JIT stage when tiering is enabled. Auto/Force/Skip.")]
+        public JitTieringMode JitTieringMode { get; set; }
+
         [Option("wasmEngine", Required = false, HelpText = "Specifies the executable (in PATH) or full path to a java script engine used to run the benchmarks, used by Wasm toolchain.", Default = "v8")]
         public string? WasmJavaScriptEngine { get; set; } = "v8";
 
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -455,6 +455,8 @@ namespace BenchmarkDotNet.ConsoleArguments
                 baseJob = baseJob.RunOncePerIteration();
             if (options.MemoryRandomization)
                 baseJob = baseJob.WithMemoryRandomization();
+            if (options.JitTieringMode != Engines.JitTieringMode.Auto)
+                baseJob = baseJob.WithJitTieringMode(options.JitTieringMode);
             if (options.NoForcedGCs)
                 baseJob = baseJob.WithGcForce(false);
             if (options.EvaluateOverhead is bool evaluateOverhead)
--- a/src/BenchmarkDotNet/Engines/EngineJitStage.cs
+++ b/src/BenchmarkDotNet/Engines/EngineJitStage.cs
@@ -74,8 +74,9 @@ internal sealed class EngineJitStage : EngineStage
         }
         yield return GetWorkloadIterationData(userInvokeCount);
 
-        // If the jit is not tiered, we're done.
-        if (!JitInfo.IsTiered)
+        JitTieringMode jitTieringMode = parameters.TargetJob.ResolveValue(RunMode.JitTieringModeCharacteristic, parameters.Resolver);
+        // If the jit is not tiered, or the user wants to skip, we're done.
+        if (!JitInfo.IsTiered || jitTieringMode == JitTieringMode.Skip)
         {
             yield break;
         }
@@ -90,9 +91,11 @@ internal sealed class EngineJitStage : EngineStage
         // it agrees. Same cutoff value that pilot stage uses.
         // We do not bail out immediately if the first iteration is long-running because it could
         // be due to cctors or other lazy initialization that won't be hit in steady-state. #2004
+        // JitTieringMode.Force opts out of this heuristic and always promotes through every tier.
         TimeInterval iterationTime = parameters.TargetJob.ResolveValue(RunMode.IterationTimeCharacteristic, parameters.Resolver);
         long remainingCalls = JitInfo.TieredCallCountThreshold;
-        if (iterationTime.Nanoseconds / (lastMeasurement.Nanoseconds / (double)userInvokeCount) < 1.5)
+        if (jitTieringMode == JitTieringMode.Auto
+            && iterationTime.Nanoseconds / (lastMeasurement.Nanoseconds / (double)userInvokeCount) < 1.5)
         {
             ++iterationIndex;
             yield return GetWorkloadIterationData(userInvokeCount);
--- a/src/BenchmarkDotNet/Engines/EngineJitStage.cs
+++ b/src/BenchmarkDotNet/Engines/EngineJitStage.cs
@@ -11,9 +11,6 @@ namespace BenchmarkDotNet.Engines;
 // the following stages (Pilot and Warmup) will likely take it the rest of the way. Long-running benchmarks may never fully reach tier1.
 internal sealed class EngineJitStage : EngineStage
 {
-    // It is not worth spending a long time in jit stage for macro-benchmarks.
-    private static readonly TimeInterval MaxTieringTime = TimeInterval.FromSeconds(10);
-
     // Jit call counting delay is only for when the app starts up. We don't need to wait for every benchmark if multiple benchmarks are ran in-process.
     private static TimeSpan s_tieredDelay = JitInfo.TieredDelay;
 
@@ -88,49 +85,40 @@ internal sealed class EngineJitStage : EngineStage
         // Don't make the next jit stage wait if it's ran in the same process.
         s_tieredDelay = TimeSpan.Zero;
 
-        // Attempt to promote methods to tier1, but don't spend too much time in jit stage.
-        StartedClock startedClock = parameters.TargetJob.ResolveValue(InfrastructureMode.ClockCharacteristic, parameters.Resolver)!.Start();
-
-        int remainingTiers = JitInfo.MaxTierPromotions;
-        long lastInvokeCount = userInvokeCount;
-        while (remainingTiers > 0)
+        // If the first iteration suggests a long-running benchmark (a single invocation already
+        // takes ~2/3 of IterationTime or more), run one confirmation iteration and bail out if
+        // it agrees. Same cutoff value that pilot stage uses.
+        // We do not bail out immediately if the first iteration is long-running because it could
+        // be due to cctors or other lazy initialization that won't be hit in steady-state. #2004
+        TimeInterval iterationTime = parameters.TargetJob.ResolveValue(RunMode.IterationTimeCharacteristic, parameters.Resolver);
+        long remainingCalls = JitInfo.TieredCallCountThreshold;
+        if (iterationTime.Nanoseconds / (lastMeasurement.Nanoseconds / (double)userInvokeCount) < 1.5)
         {
-            --remainingTiers;
-            long remainingCalls = JitInfo.TieredCallCountThreshold;
+            ++iterationIndex;
+            yield return GetWorkloadIterationData(userInvokeCount);
+            if (iterationTime.Nanoseconds / (lastMeasurement.Nanoseconds / (double)userInvokeCount) < 1.5)
+            {
+                didStopEarly = true;
+                yield break;
+            }
+            remainingCalls -= userInvokeCount;
+        }
+
+        // Promote methods to tier1.
+        for (int remainingTiers = JitInfo.MaxTierPromotions; remainingTiers > 0; --remainingTiers)
+        {
             while (remainingCalls > 0)
             {
-                long invokeCount;
-                if (hasUserInvocationCount)
-                {
-                    invokeCount = userInvokeCount;
-                }
-                else
-                {
-                    // If we can run a batch of calls within the time limit (based on the last measurement), do that instead of multiple single-invocation iterations.
-                    // For long-running benchmarks where even a small batch wouldn't fit, fall back to one invocation per iteration.
-                    var remainingTimeLimit = MaxTieringTime.ToNanoseconds() - startedClock.GetElapsed().GetNanoseconds();
-                    var lastMeasurementSingleInvocationTime = lastMeasurement.Nanoseconds / lastInvokeCount;
-                    long allowedCallsWithinTimeLimit = (long)Math.Floor(remainingTimeLimit / lastMeasurementSingleInvocationTime);
-                    invokeCount = allowedCallsWithinTimeLimit > 0
-                        ? Math.Min(remainingCalls, allowedCallsWithinTimeLimit)
-                        : 1;
-                }
-                lastInvokeCount = invokeCount;
-
+                // Run the whole tier's call budget in a single iteration unless the user pinned InvocationCount.
+                long invokeCount = hasUserInvocationCount ? userInvokeCount : remainingCalls;
                 remainingCalls -= invokeCount;
                 ++iterationIndex;
                 // The generated __Overhead method is aggressively optimized, so we don't need to run it again.
                 yield return GetWorkloadIterationData(invokeCount);
-
-                if ((remainingTiers > 0 || remainingCalls > 0)
-                    && startedClock.GetElapsed().GetTimeValue() >= MaxTieringTime)
-                {
-                    didStopEarly = true;
-                    yield break;
-                }
             }
 
             Engine.SleepIfPositive(JitInfo.BackgroundCompilationDelay);
+            remainingCalls = JitInfo.TieredCallCountThreshold;
         }
 
         // Empirical evidence shows that the first call after the method is tiered up may take longer,
--- a/src/BenchmarkDotNet/Engines/EngineResolver.cs
+++ b/src/BenchmarkDotNet/Engines/EngineResolver.cs
@@ -33,6 +33,7 @@ namespace BenchmarkDotNet.Engines
             Register(AccuracyMode.MinInvokeCountCharacteristic, () => 4);
             Register(AccuracyMode.EvaluateOverheadCharacteristic, () => false);
             Register(RunMode.MemoryRandomizationCharacteristic, () => false);
+            Register(RunMode.JitTieringModeCharacteristic, () => JitTieringMode.Auto);
             Register(AccuracyMode.OutlierModeCharacteristic, job =>
             {
                 // if Memory Randomization was enabled and the benchmark is truly multimodal
--- a/src/BenchmarkDotNet/Engines/JitTieringMode.cs
+++ b/src/BenchmarkDotNet/Engines/JitTieringMode.cs
@@ -1,0 +1,25 @@
+namespace BenchmarkDotNet.Engines;
+
+/// <summary>
+/// Controls the behavior of the JIT stage when tiering is enabled.
+/// </summary>
+public enum JitTieringMode
+{
+    /// <summary>
+    /// Default.
+    /// If the benchmark is long-running, tiering is skipped and the benchmark method is left at the tier following the initial invoke (usually tier0).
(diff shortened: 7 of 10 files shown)


adamsitnik (commented):
Hello @timcassell !

Thank you for addressing my feedback. I haven’t reviewed the original changes in #2806, so I’d like to ask a few questions first to better understand the current design and behavior.

- Is the new tiering heuristic based on deterministic signals such as JIT events emitted by the runtime or disassembly data, or is it intended to approximate the CLR’s current heuristic?
- How is this handled across different runtimes? For example, is the new heuristic enabled for AOT and .NET Framework, and how does it behave on older versus newer Mono? Also, do we respect all environment variables that allow users to disable tiered JIT, such as `DOTNET_TieredCompilation` and `COMPlus_TieredCompilation`?
- How does it affect the total time it takes to run a typical micro benchmark?
- How does this interact with other stages, especially Warmup? For example, if the JITting phase runs long enough to satisfy the warmup heuristic, do we then skip the warmup phase?
- How can users configure the JITting phase? Do we expose new Job APIs, attributes, and command-line arguments?

Thanks,
Adam

timcassell:
> * Is the new tiering heuristic based on deterministic signals such as JIT events emitted by the runtime or disassembly data, or is it intended to approximate the CLR’s current heuristic?

It does not listen to JIT events, it simply queries the values used by the JIT for number of invokes for each stage, with an approximate wait time between JIT stages. You can see how those values are queried in JitInfo.cs.

> * How is this handled across different runtimes? For example, is the new heuristic enabled for AOT and .NET Framework, and how does it behave on older versus newer Mono? Also, do we respect all environment variables that allow users to disable tiered JIT, such as `DOTNET_TieredCompilation` and `COMPlus_TieredCompilation`?

It only handles JIT stages in CoreCLR, and when tiering is enabled. In all other runtimes `JitInfo.IsTiered` returns false. I'm not sure if new Mono has a tiered JIT, I didn't look into it or test it. And yes we respect all env vars and knobs that allow users to control JIT behavior.

> * How does it affect the total time it takes to run a typical micro benchmark?

It depends on how fast the benchmark is. By default, it invokes it 30x per jit stage promotion, with 250ms wait time between stages. On the latest runtime there are 3 calculated stage promotions (it should be 2, but there is a runtime bug that we account for).

> * How does this interact with other stages, especially Warmup? For example, if the JITting phase runs long enough to satisfy the warmup heuristic, do we then skip the warmup phase?

Other stages were left unchanged. Warmup stage has existed since even before tiered JIT was a thing and I didn't want to mess with it. We could definitely improve things there, though it's tracked by #1993.

> * How can users configure the JITting phase? Do we expose new Job APIs, attributes, and command-line arguments?

Via environment variables or MSBuild properties that the runtime exposes. No new APIs were added to BDN for it.

timcassell:
> > how does it behave on older versus newer Mono?
> 
> It only handles JIT stages in CoreCLR

The CoreCLR detection was broken, fixed in #3150.

adamsitnik:
> The CoreCLR detection was broken, fixed in #3150.

Great, thanks for fixing it!

> It depends on how fast the benchmark is. By default, it invokes it 30x per jit stage promotion, with 250ms wait time between stages. On the latest runtime there are 3 calculated stage promotions (it should be 2, but there is a runtime bug that we account for).

So for benchmarks that are neither very short or long running, like:

```csharp
[Benchmark]
public void Sleep() => Thread.Sleep(100);
```

We are going to spend a lot of time trying to promote them. In this case it was 9+ seconds while the whole benchmarking took 25s. 

```log
WorkloadJitting  1: 1 op, 103811800.00 ns, 103.8118 ms/op
WorkloadJitting  2: 30 op, 3247958900.00 ns, 108.2653 ms/op
WorkloadJitting  3: 30 op, 3259087200.00 ns, 108.6362 ms/op
WorkloadJitting  4: 30 op, 3247723900.00 ns, 108.2575 ms/op
WorkloadJitting  5: 1 op, 108950800.00 ns, 108.9508 ms/op
```

16s -> 25s is quite a lot. Do you see any way to reduce the time it takes to run it?

> Via environment variables or MSBuild properties that the runtime exposes. No new APIs were added to BDN for it.

I think it would be nice to offer an ability to disable (or hardcode) it. So for example if I am having plenty of benchmarks similar to the `100ms` sleep described above, I can avoid regressing the time it takes to run all of them. Or just have a config that I can use during a conference talk. FWIW the dotnet/performance repo has been using [these](https://github.com/dotnet/performance/blob/96cbaa1351d0efc47ad6550906e2be41720e22f7/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L36-L41) settings for years.

```csharp
config.WithJittingCount(1);
[JittingCount(1)]
```

timcassell:
The only way to reduce the time is to reduce the CallCountThreshold. Originally I had set it to 1 via TC_AggressiveTiering, but Egor explained in that PR why that was a bad idea, so I went with a timeout instead. We already have config options to skip the JIT stage (`RunStrategy.Monitoring/ColdStart`).

Generally users are most interested in measuring tier1 performance, which is the default `Throughput` strategy, so it doesn't make sense to not push it all the way there unless the user selects a different run strategy.

> ```cs
> config.WithJittingCount(1);
> [JittingCount(1)]
> ```

What is that supposed to do? Just run a single JIT iteration? I guess it would kinda make sense to have it get to tier0 before the pilot stage runs. But probably it should just be a bool value for whether to go through all the tiers or not, because I don't know what `JittingCount(2)` would even mean.

timcassell:
> FWIW the dotnet/performance repo has been using [these](https://github.com/dotnet/performance/blob/96cbaa1351d0efc47ad6550906e2be41720e22f7/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L36-L41) settings for years.

It seems that with those settings you were actually measuring tier0 performance in the first iterations (unless the pilot stage invoked it enough times to bump it up), and possibly higher tiers in later iterations depending how fast the benchmark is. So each iteration was not measuring the same thing.

But the perf repo is on a version of BDN that has the new JIT stage, so it should always be measuring tier1 now. I don't have access to the data, so I don't know what effect that had on the total runtime and measurement stability.

timcassell:
> Do you see any way to reduce the time it takes to run it?

Also if the runtime would fix the duplicated tier0 with OSR enabled we could reduce the time by 1/3, but that's out of our control.

adamsitnik:
> Generally users are most interested in measuring tier1 performance, which is the default `Throughput` strategy, so it doesn't make sense to not push it all the way there unless the user selects a different run strategy.

They are, but Tiered JIT was introduced 7(?) years ago and they have figured various ways of dealing with this problem. Most of them were lucky enough to just run micro-benchmarks that were getting promoted to Tier 1 with the default Pilot/Warmup settings. Others increased warmup count or just disabled Tiered JIT.

> It seems that with those settings you were actually measuring tier0 performance in the first iterations (unless the pilot stage invoked it enough times to bump it up), and possibly higher tiers in later iterations depending how fast the benchmark is. So each iteration was not measuring the same thing.
> 
> But the perf repo is on a version of BDN that has the new JIT stage, so it should always be measuring tier1 now.

In case of dotnet/performance we have went back and forth about Tiered JIT when the feature got introduced (https://github.com/dotnet/performance/pull/247, https://github.com/dotnet/performance/pull/320). What we actually ended up doing (https://github.com/dotnet/performance/pull/1536, https://github.com/dotnet/performance/pull/666) was setting explicit min/max warmup count for just [a couple of benchmarks](https://github.com/search?q=repo%3Adotnet%2Fperformance+%22WarmupCount%28%22&type=code) that were not getting promoted to Tier 1 with our custom settings.

And what is very important in case of dotnet/performance is the time it takes to run all the benchmarks. Assuming that we run only nano-benchmarks (which is not true) and the new logic takes 1s per benchmark, our 5k benchmarks will need additional 83 minutes to run. And we are aiming at running the benchmarks multiple time per day, for multiple architectures. So in our (.NET Team) case, it really matters to not prolong the time it takes to run all the benchmarks.

> But probably it should just be a bool value for whether to go through all the tiers or not, because I don't know what `JittingCount(2)` would even mean.

Bool is enough for our needs. 

> Also if the runtime would fix the duplicated tier0 with OSR enabled we could reduce the time by 1/3, but that's out of our control.

@egorbo @AndyAyersMS Is this something that can be easily fixed? Do we have any plans to do it?

> Other stages were left unchanged. Warmup stage has existed since even before tiered JIT was a thing and I didn't want to mess with it. We could definitely improve things there, though it's tracked by #1993.

Thank you for providing the link. Ideally all of this should be combined and the total time it takes to get the perfect invocation count, code promoted to Tier 1 and warmed up should be minimal.

Also another thing to consider are the "worst case" benchmarks here. Benchmarks that take less than one iteration time to execute (so we are not going to skip the extra warmup), but running them 3x30 times takes a lot of time. Most of them just delegate the work to smaller methods, and these small, hot path methods are usually executed multiple times even for a single invocation. And they usually get promoted to Tier 1 with the default settings. If my memory serves me well, this was the case for all the [C# Computer Game benchmarks](https://github.com/dotnet/performance/tree/main/src/benchmarks/micro/runtime/BenchmarksGame). What I am trying to say is that based on the evidence we had in the past, very few benchmarks required the extra warmup.

timcassell:
> They are, but Tiered JIT was introduced 7(?) years ago and they have figured various ways of dealing with this problem. Most of them were lucky enough to just run micro-benchmarks that were getting promoted to Tier 1 with the default Pilot/Warmup settings. Others increased warmup count or just disabled Tiered JIT.

Right, but that involved extra knowledge of the JIT tiering strategy and how the benchmarks were being invoked. New users run into the same problem. Now that's no longer a concern.

> In case of dotnet/performance we have went back and forth about Tiered JIT when the feature got introduced ([dotnet/performance#247](https://github.com/dotnet/performance/pull/247), [dotnet/performance#320](https://github.com/dotnet/performance/pull/320)). What we actually ended up doing ([dotnet/performance#1536](https://github.com/dotnet/performance/pull/1536), [dotnet/performance#666](https://github.com/dotnet/performance/pull/666)) was setting explicit min/max warmup count for just [a couple of benchmarks](https://github.com/search?q=repo%3Adotnet%2Fperformance+%22WarmupCount%28%22&type=code) that were not getting promoted to Tier 1 with our custom settings.

I think now the recommended config can be updated to set warmup count to 0, and remove all the specific benchmark warmup settings since tiering up is automatically handled now.

> And what is very important in case of dotnet/performance is the time it takes to run all the benchmarks. Assuming that we run only nano-benchmarks (which is not true) and the new logic takes 1s per benchmark, our 5k benchmarks will need additional 83 minutes to run. And we are aiming at running the benchmarks multiple time per day, for multiple architectures. So in our (.NET Team) case, it really matters to not prolong the time it takes to run all the benchmarks.

Dotnet/performance updated to the new JIT stage in https://github.com/dotnet/performance/pull/5073, do you have the numbers from before/after?

> Bool is enough for our needs.

I think that option will probably not be needed. With the JIT stage already promoting to tier1, the pilot stage should be able to reach stable more quickly, and with setting warmup count to 0 that's less time spent "guessing" at tiering up. So total time I would think would be not much more than before (except for the duplicated tier0 of course).

> Thank you for providing the link. Ideally all of this should be combined and the total time it takes to get the perfect invocation count, code promoted to Tier 1 and warmed up should be minimal.

I agree. @AndreyAkinshin mentioned combining the pilot and warmup stages in #2787 (also #1210), which I think would also help here, but I'm not sure about concrete plans for how to go about it.

> Also another thing to consider are the "worst case" benchmarks here. Benchmarks that take less than one iteration time to execute (so we are not going to skip the extra warmup), but running them 3x30 times takes a lot of time.

I mean that's what this PR is for, to skip the tiering-up for long-running benchmarks. We can adjust the heuristic for what we consider to be the optimal cutoff value. I don't know what that cutoff value should be, though, so I just started with copying the pilot stage's heuristic.

EgorBo:
>> Also if the runtime would fix the duplicated tier0 with OSR enabled we could reduce the time by 1/3, but that's out of our control.

> @EgorBo @AndyAyersMS Is this something that can be easily fixed? Do we have any plans to do it?

If I understand it correctly, it's https://github.com/dotnet/runtime/issues/76402 we probably should indeed address it for .NET 11.0 (am not promising, but will try)

timcassell:
> If I understand it correctly, it's [dotnet/runtime#76402](https://github.com/dotnet/runtime/issues/76402) we probably should indeed address it for .NET 11.0 (am not promising, but will try)

I don't think it's the same issue. It was described in https://github.com/dotnet/runtime/issues/117787#issuecomment-3090771091, I'm not sure if there's a dedicated issue for it. [Edit] Also later comment in that issue https://github.com/dotnet/runtime/issues/76402#issuecomment-2511471547.

timcassell:
Also working with Claude it found that we can listen to JIT events in-process, so I'm going to try building a prototype to do that for deterministic tiering-up instead of waiting 250ms.

adamsitnik:
> Right, but that involved extra knowledge of the JIT tiering strategy and how the benchmarks were being invoked. New users run into the same problem. Now that's no longer a concern.

It's not a concern for the new users, but it can be for the existing users that have used different solution to get their code promoted to Tier 1, have plenty of benchmarks and don't want to prolong the time it takes to run the benchmarks.

> I think now the recommended config can be updated to set warmup count to 0, and remove all the specific benchmark warmup settings since tiering up is automatically handled now.

In case of this particular repo (which sets `WarmupCount=1`) it may be true, but it should require running all the benchmarks before and after and using tools like [ResultsComparer](https://github.com/dotnet/performance/tree/main/src/tools/ResultsComparer) to find all the differences. Then go over all of them, study and make the call. It takes plenty of time to do that. And it may require having an attribute to disable it for certain benchmarks.

> Dotnet/performance updated to the new JIT stage in [dotnet/performance#5073](https://github.com/dotnet/performance/pull/5073), do you have the numbers from before/after?

@DrewScoggins do we have such numbers?

> I mean that's what this PR is for, to skip the tiering-up for long-running benchmarks. 

You are right, it's skipped for the long-running benchmarks. 2/3 of the iteration time as of now:

https://github.com/dotnet/BenchmarkDotNet/blob/bb136afd70f8974a6a930c064cb6f2f5384f199b/src/BenchmarkDotNet/Engines/EngineJitStage.cs#L88-L92

What I meant  by the "worst case" benchmarks are benchmarks that are not long-running, but running them 30x3 takes a lot of time. Like 100ms for a single invocation with 250ms for iteration time. Running it 90 times takes 9s just to warmup the code. While in the performance repo we always aimed at getting the result in 5s.

> We can adjust the heuristic for what we consider to be the optimal cutoff value. I don't know what that cutoff value should be, though, so I just started with copying the pilot stage's heuristic.

TBH I am not sure if it's even possible to find the optimal cutoff value. That is why I would prefer to provide some good defaults and an ability to customize it (either by disabling the new warmup or providing explicit cutoff value for the heuristic). 

> Also working with Claude it found that we can listen to JIT events in-process, so I'm going to try building a prototype to do that for deterministic tiering-up instead of waiting 250ms.

This is great, but please make sure we know what are the side effects of listening for these events on every platform (on Windows I would expect whatever it is to use ultra-fast ETW, while on Unix it could be EventPipe with non zero overhead).

timcassell:
I added `JitTieringMode` option to force the tiering either way.

> This is great, but please make sure we know what are the side effects of listening for these events on every platform (on Windows I would expect whatever it is to use ultra-fast ETW, while on Unix it could be EventPipe with non zero overhead).

The prototype was a success. I went through several iterations to resolve all the edge cases until I got to what I think is the best it can be with existing APIs in #3169 (it's still draft because it was built on top of this branch). The overhead shouldn't matter since the listener is torn down at the end of the JIT stage.

timcassell:
@adamsitnik Are your concerns addressed?

DrewScoggins:
Apologies @adamsitnik, I missed this ping. These are the four issues that got filed across Windows and Linux x64/x86 that correspond with that change. This is out of about 6000 tests, so it was certainly not widespread, but it did show up.

On another note, Tim. You can find all of our perf data here, https://pvscmdupload.z22.web.core.windows.net/reports/allTestHistory/TestHistoryIndexIndex.html

* https://github.com/dotnet/perf-autofiling-issues/issues/68044
* https://github.com/dotnet/perf-autofiling-issues/issues/68041
* https://github.com/dotnet/perf-autofiling-issues/issues/68008
* https://github.com/dotnet/perf-autofiling-issues/issues/67961

timcassell:
> You can find all of our perf data here, https://pvscmdupload.z22.web.core.windows.net/reports/allTestHistory/TestHistoryIndexIndex.html

We were talking about the total time it takes to run the benchmark suite. I don't see that data there.

> These are the four issues that got filed across Windows and Linux x64/x86 that correspond with that change. This is out of about 6000 tests, so it was certainly not widespread, but it did show up.
> 
> * [[Perf] Linux/x64: 50 Regressions on 1/23/2026 10:32:11 PM +00:00 perf-autofiling-issues#68044](https://github.com/dotnet/perf-autofiling-issues/issues/68044)
> * [[Perf] Linux/x64: 36 Regressions on 1/22/2026 3:57:05 AM +00:00 perf-autofiling-issues#68041](https://github.com/dotnet/perf-autofiling-issues/issues/68041)
> * [[Perf] Linux/x64: 23 Regressions on 1/22/2026 3:57:05 AM +00:00 perf-autofiling-issues#68008](https://github.com/dotnet/perf-autofiling-issues/issues/68008)
> * [[Perf] Windows/x86: 87 Regressions on 1/23/2026 10:32:11 PM +00:00 perf-autofiling-issues#67961](https://github.com/dotnet/perf-autofiling-issues/issues/67961)

Those regressions look unrelated to the BDN update. The other major change that was included there was the direct calls to improve accuracy, but that would've only changed results by a few clock cycles at max (<1ns).

src/BenchmarkDotNet/Attributes/Mutators/JitTieringAttribute.cs
```diff
@@ -0,0 +1,9 @@
+using BenchmarkDotNet.Engines;
+using BenchmarkDotNet.Jobs;
+
+namespace BenchmarkDotNet.Attributes;
+
+/// <inheritdoc cref="RunMode.JitTieringMode"/>
+public class JitTieringAttribute(JitTieringMode mode) : JobMutatorConfigBaseAttribute(Job.Default.WithJitTieringMode(mode))
```
adamsitnik, line 7: Thank you for providing the attribute @timcassell !

adamsitnik (commented):
@timcassell thank you for addressing my feedback!

PR #3144, opened 2026-05-24
chore: Remove `System.Management` dependency

filzrev:
This PR remove `System.Management` (and transitive `System.CodeDom`) package dependency from BenchmarkDotNet.
By replacing WMI query logics to use external `powershell`/`pwsh` command instead.

1. Remove `System.Management` package dependency
2. Remove `MosCpuDetector` that is used on .NET Framework
3. Add `ProcessHelper::RunPowerShellCommandAndReadOutput` helper method to invoke powershell/pwsh command
  3.1. Add `-NoProfile` argument to faster launch
  3.2. Add `ErrorActionPreference='Stop'` statement to throw exception on error.
  3.3. Set `TERM=dumb` environment variable by default to suppress ANSI escape sequence
  3.4. Add throw logics **when running on IntegrationTests** to avoid exception is silently ignored.
4. Modify following methods to call powershell/pwsh command instead of using `ManagementObjectSearcher`.
  4.1.  `RuntimeInformation::GetAntivirusProducts`
  4.2.  `RuntimeInformation::GetVirtualMachineHypervisor`
5. Add `HostEnvironmentTests.cs` to validate modified logics

---

Calling external  powershell/pwsh command required some extra overhead.
Though, these value are cached and called once per `CPU`/`VirtualMachineHypervisor`.
Therefore, it is expected no observable impact on execution time.
(Note: `AntivirusProducts` is used only when benchmark returns non-0 exit code)

1. Using `ManagementObjectSearcher` : 55 [ms]
2. Using `powershell.exe`: 300 [ms]
3. Using `pwsh.exe` : 600 [ms]

---
On PowerShell execution is restricted environment (Retricted by AppLocker, WDAC, or GPO)
`Cpu`/`VirtualMachineHypervisor`/`AntivirusProducts` information might not be available.

In this case, CPU is displayed as `Unknown Processor` and `VirtualMachineHypervisor`/`AntivirusProducts` information will be empty.

Close this PR.  If it need to support on PowerShell restricted environment.
In this case, I'll rewrite related logics to use `Microsoft.Management.Infrastructure` package instead.


filzrev:
Close this PR.

It seems using the PowerShell/pwsh command introduces extra complexity and only a few benefits by removing the `System.Management` (and transitive `System.CodeDom`) dependency.

And it seems migrating to `Microsoft.Management.Infrastructure` has license issue
`https://github.com/PowerShell/PowerShell/issues/20798`

Issue #3086, opened 2026-04-15
DisassemblyDiagnoserTests randomly failed on CI


filzrev:
On CI environment,
DisassemblyDiagnoserTests randomly failed with following exception.

**StackTrace**

```
CanDisassembleAllMethodCallsUsingFilters(jit: RyuJit, platform: X64, toolchain: .NET 8.0) [FAIL]
System.InvalidOperationException : Sequence contains no elements
Stack Trace:
     at System.Linq.ThrowHelper.ThrowNoElementsException()
     at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source)
  /Users/runner/work/BenchmarkDotNet/BenchmarkDotNet/tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs(114,0): at BenchmarkDotNet.IntegrationTests.DisassemblyDiagnoserTests.CanDisassembleAllMethodCallsUsingFilters(Jit jit, Platform platform, IToolchain toolchain)
     at InvokeStub_DisassemblyDiagnoserTests.CanDisassembleAllMethodCallsUsingFilters(Object, Span`1)
     at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

It seems disassembler silently failed on some conditions https://github.com/dotnet/BenchmarkDotNet/issues/2637

Currently this issue is observed on `test-macos (x64)` only.
- https://github.com/dotnet/BenchmarkDotNet/actions/runs/24426573098
- https://github.com/dotnet/BenchmarkDotNet/actions/runs/24440083195
- https://github.com/dotnet/BenchmarkDotNet/actions/runs/24927496657/job/72999787674

filzrev:
When error occurred.
Following logs are outputted additionally.

It seems benchmark process is crashed with SIGSEGV signal (ExitCode: 139)

```
[createdump] Gathering state for process 21852 
[createdump] Writing full dump to file /var/folders/kk/699g20m50k56v2wnfvqxt3mm0000gn/T/tmpTzTOoa.tmp
[createdump] Written 5661995304 bytes (1382323 pages) to core file
[createdump] Target process is alive
[createdump] Dump successfully written in 17504ms
ProcessData operation is interrupted by EndOfStream.
 // Benchmark Process 21852 has exited with code 139.
```

filzrev:
[...]

PR #3153, opened 2026-05-27
chore: Add setting to skip disassembly tests on macos(x64)

filzrev:
This PR add temporary workaround code for #3086
By adding setting to skip disassembly tests on `macos(x64)`

Currently this SIGSEGV  issue is observed on `macos(x64)`/CsProjectCoreToolchain only.
So just skip related tests on this environment.

Note: As far as I've confirmed, there is no related issue reported on ClrMD repository.

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs
@@ -1,6 +1,7 @@
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Columns;
 using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Detectors;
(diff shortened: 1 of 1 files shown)


filzrev:
Currently following command is **silently** failed on `windows-latest` environment.
>  jsvu --os=default --engines=v8

It looks like downloaded v8 binary is something wrong.
~~And it seems be resolved on latest release.~~

~~I've added `$PSNativeCommandUseErrorActionPreference = $true` setting to thrown when native command exit non-0 code.~~

-> This setting seems not works. jsvu returns exit code 0. And It also confirmed on `ubuntu-latest`

`jsvu` installation issue seems occurred when running on `20260525.56.1` runner image.
Currently, It seems no related issue reported on `actions/runner-images` or `jsvu` repository.

timcassell (approved):

Issue #719, opened 2018-04-14
If fail to build in Parallel due to file access issues, try to build sequentially


adamsitnik:

martincostello:
Is there an ETA on a version with this fix being published to NuGet? I'm having this problem with `0.10.14` with a branch of a project using preview 2 of .NET Core 2.1. It's fine with 2.0.

adamsitnik:
hello @martincostello 

I don't have ETA for `0.11.0`, but you can use our preview package from our CI feed today:

```xml
<packageSources>
  <add key="bdn-nightly" value="https://ci.appveyor.com/nuget/benchmarkdotnet" />
</packageSources>
```

martincostello:
Thanks, I’ll use that for now. My 2.1 branch of my project is going to stay as something experimental until it reaches final non-preview/rc release anyway.

martincostello:
I've tried using `0.10.14.524` from the AppVeyor feed, but I'm still getting errors about files being in use when trying to run my normal benchmarks:

```
// Build Exception: Microsoft (R) Build Engine version 15.7.145.53551 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  You are working with a preview version of the .NET Core SDK. You can define the SDK version via a global.json file in the current project. More at https://go.microsoft.com/fwlink/?linkid=869452

C:\Program Files\dotnet\sdk\2.1.300-preview2008530\Microsoft.Common.CurrentVersion.targets(4343,5): warning MSB3026: Could not copy "C:\Coding\Thing\src\Thing\bin\Release\netcoreapp2.1\Thing.dll" to "bin\Release\netcoreapp2.1\Thing.dll". Beginning retry 1 in 1000ms. The process cannot access the file 'C:\Coding\Thing\tests\Thing.Benchmarks\bin\Release\netcoreapp2.1\Thing.dll' because it is being used by another process  [C:\Coding\Thing\tests\Thing.Benchmarks\Thing.Benchmarks.csproj]
```

adamsitnik:
@martincostello have you tried killing all `dotnet` processes before running the benchmarks?

martincostello:
@adamsitnik I didn't explicitly try that, no. I'll give that a try and get back to you.

martincostello:
I checked no instances of `dotnet` were running and then did `dotnet run` to run my benchmark project. It does the initial compilation for the run fine, but then the internal compilations for the benchmarks themselves fail.

[...]

PR #3166, opened 2026-06-02
chore: Re-enable shared compilation feature

filzrev:
This PR re-enable `UseSharedCompilation` setting(Use shared `VBCSCompiler.exe` process) for faster **local** build.

This setting is introduced by #719.
Though,  currently it seems not be required because there are no shared `obj`/`bin` content between BenchmarkDotNet projects.

And when building AutoGenerated project. following settings are passed explicitly.
So these build is not expected be affected. 
https://github.com/dotnet/BenchmarkDotNet/blob/bc899f3c8c218f7c829a69e3643ef51ab2682e2a/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs#L213

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -16,7 +16,6 @@
     <CLSCompliant>true</CLSCompliant>
     <ComVisible>false</ComVisible>
 
-    <UseSharedCompilation>false</UseSharedCompilation>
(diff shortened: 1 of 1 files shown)


adamsitnik (approved):
Let's give it a try!

PR #3169, opened 2026-06-03
Listen to JIT events

timcassell:
Improves the JIT stage by listening to events instead of waiting 250ms per tier. It's still not 100% deterministic, but it's much better than before.

Master: ~750ms + benchmark time
PR: ~90ms + benchmark time + compilation time


timcassell:
cc @EgorBo @AndyAyersMS

AndyAyersMS:
Dismayed that it is so complicated. I don't believe we have any better solution to offer.

timcassell:
> Dismayed that it is so complicated. I don't believe we have any better solution to offer.

Right, it's the consequence of having to compensate for racing with the JIT and event dispatcher threads. Dedicated APIs would vastly simplify it (https://github.com/dotnet/runtime/issues/129020).

PR #3171, opened 2026-06-06
chore: Remove unneeded project reference from auto-generated project

filzrev:
This PR remove `ProjectReference` of benchmark project from `BenchmarkDotNet.Autogenerated.csproj`

CsProjGenerator gather referenced dlls of benchmark project.
Then add referenced dlls as `<Reference Include="..." />` node.

So original `<ProjectReference Include="..." />` node is not expected to be used when building `BenchmarkDotNet.Autogenerated.csproj`

**Test**
This code path is not used when running IntegrationTest project or when using non Release custom configuration.
https://github.com/dotnet/BenchmarkDotNet/blob/bcdcc98407faf392179fb76ee96a187ca7b1e591/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs#L57

So I've confirmed behavior by using generated `binlog`.
After this PR content is merged. Reference project evaluation stage is skipped on local build.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
+++ b/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
@@ -191,6 +191,9 @@ namespace BenchmarkDotNet.Toolchains.CsProj
                 ));
             }
 
+            // Remove ProjectReference node of benchmark project after gathered dll references are added.
+            doc.Root.Descendants("ProjectReference").Single().Remove();
+
             using var projectStream = File.Create(artifactsPaths.ProjectFilePath);
 #if NETSTANDARD2_0
             doc.Save(projectStream, SaveOptions.None);


timcassell:
When I tried it originally the disassemble tests failed. Maybe that was fixed with clrmd v4. [Edit] Actually, I see the tests are still failing on arm.

filzrev:
Disassembler tests on `linux(arm64)`/`macos(arm64)` failed with following error.

```
System.DllNotFoundException: Unable to load shared library 'capstone' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: 
dlopen(/Users/runner/work/BenchmarkDotNet/BenchmarkDotNet/.dotnet/shared/Microsoft.NETCore.App/8.0.27/capstone.dylib, 0x0001):
(...Fallback to available system paths...)
```

It's occurred because DllGatherer add `<Reference>` node for DLLs.
And if `ProjectReference` is removed.  Generated `deps.json` file don't contains `runtimeTargets`.
So it failed to resolve native dependencies.

And as far as I've confirmed.
It require hacky MSBuild operations to add these dependencies with `GenerateBuildDependencyFile` hook. 
e.g. https://www.nuget.org/packages/Asmichi.InjectNativeFileDepsHack

@timcassell
May I ask what's the purpose of adding `<Reference>` tag on AutoGenerated project? (#2508/#2921)
If it's introduced for DLL copy related issue. It might be resolved by another approach.

timcassell:
It's to resolve issues with dll references that weren't picked up by the auto-generated project previously (see the linked issues that #2508 closed). If you have a better solution for it, please send a PR. 😄

timcassell:
I see the reference gatherer only gets the top directory files, not nested directories. Maybe we simply need to pass `SearchOption.AllDirectories` to fix it.

https://github.com/dotnet/BenchmarkDotNet/blob/bcdcc98407faf392179fb76ee96a187ca7b1e591/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs#L185

filzrev:
> I see the reference gatherer only gets the top directory files, not nested directories. Maybe we simply need to pass SearchOption.AllDirectories to fix it.

When `ProjectReference` is removed,
Satellite assemblies and native DLL are not included in the build tree.
Though, Dlls are already placed to `bin` directory.
Because `DllGatherer` and `Autogenerated` projects build are sharing same binary output directory.

---
I though it's desired to refactor DllGatherer/Autogenerated projects build process with following paths.

1. Build **original** benchmark projects.
2. Run DllGatherer's **custom** MSBuild task to collect `ReferenceCopyLocalPaths` information.   
   (Don't build DllGatherer project in this step)
4. Modify `Autogenerated` project based on collected information (Add Reference and `Aliases` information)
5. Build `Autogenerated` project with `--no-dependencies`

Close this PR, and I'll create another issue when trying to testing above refactoring.

PR #3175, opened 2026-06-17
feat: Add cleanup logic for workdirectory before build projects

### LINKED ISSUE #3174: Some benchmark failed on second execution when using `--keepFiles` option
When running benchmark **multiple** times with `--keepFiles` options.
It works the first time, but fails on the second and subsequent execution on some benchmarks.

It's caused because `DllGatherer` collect unintended assembly references that are produced by `Autogenerated` project build/publish.

I thought, It need to add directory cleanup logics before building `BuildPartition`.

**How to reproduce problem**
1. Add benchmark project with `net8.0` tfm.
2. Add **nightly** build package reference. and configure NuGet.config.
3. Add following code
    ```csharp
    internal class Program
    {
        {
            var config = ManualConfig.CreateEmpty()
                .AddLogger(ConsoleLogger.Default)
                .AddJob(Job.Dry.WithRuntime(MonoRuntime.Mono80).WithStrategy(RunStrategy.Monitoring))
                .WithBuildTimeout(TimeSpan.FromSeconds(240))
                .WithOptions(ConfigOptions.KeepBenchmarkFiles);

            var _ = BenchmarkRunner.Run(typeof(Program).Assembly, config);
        }
    }

    public class Benchmarks
    {
        [Benchmark]
        public void Mono()
        {
            if (Type.GetType("Mono.RuntimeStructs") == null)
            {
                throw new Exception("This is not Mono runtime");
            }
        }
    }
    ```
4. Run benchmark and confirm it's succeed on first execution
5. Run benchmark and confirm failed on `dotnet publish`

filzrev:
This PR intended to fix #3174

1. Add `FileCleanupHelper` helper class
2. Modify `GeneratorBase.cs` and add cleanup logics for files/directories that returned by `GetArtifactsToCleanup`
3. Modify `BenchmarkRunnerClean.cs` logics to support following scenario.
3.1  Delete file that have `ReadOnly` attribute (on Windows)
3.2. Continue processing when exception occurred.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/GeneratorBase.cs
+++ b/src/BenchmarkDotNet/Toolchains/GeneratorBase.cs
@@ -27,6 +27,10 @@ namespace BenchmarkDotNet.Toolchains
             {
                 artifactsPaths = GetArtifactsPaths(buildPartition, rootArtifactsFolderPath);
 
+                // Cleanup artifact files/directories before generating project files.
+                foreach (var path in GetArtifactsToCleanup(artifactsPaths))
+                    FileCleanupHelper.Cleanup(path);
+
                 // There is no async file copy API, so we just do it synchronously. We are likely on a ThreadPool thread here anyway if this generator is ran in parallel.
                 CopyAllRequiredFiles(artifactsPaths);
 
--- a/src/BenchmarkDotNet/Helpers/FileCleanupHelper.cs
+++ b/src/BenchmarkDotNet/Helpers/FileCleanupHelper.cs
@@ -1,0 +1,91 @@
+namespace BenchmarkDotNet.Helpers;
+
+internal static class FileCleanupHelper
+{
+    public static void Cleanup(string path)
+    {
+        if (File.Exists(path))
+        {
+            CleanupFile(new FileInfo(path));
+            return;
+        }
+
+        if (Directory.Exists(path))
+        {
+            CleanupDirectory(new DirectoryInfo(path));
+        }
+    }
+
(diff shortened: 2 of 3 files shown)


src/BenchmarkDotNet/Toolchains/GeneratorBase.cs
```diff
@@ -160,5 +162,17 @@ private ArtifactsPaths GetArtifactsPaths(BuildPartition buildPartition, string r
                 programName: programName,
                 packagesDirectoryName: GetPackagesDirectoryPath(buildArtifactsDirectoryPath));
         }
+
+        private static void CleanupBuildArtifactDirectory(BuildPartition buildPartition, ArtifactsPaths artifactsPaths)
+        {
+            var buildArtifactsDirectoryName = new DirectoryInfo(artifactsPaths.BuildArtifactsDirectoryPath).Name;
+
+            // Check directory name to skip BenchmarkDotNet.Toolchains.Roslyn.Generator
```
timcassell: Instead of checking this way, you could make this method virtual and override it in DotNetCliGenerator.

filzrev: I've simplified code to  cleanup files/directories that are returned by existing `GetArtifactsToCleanup` method.

timcassell:
Any changes you want to do based on https://github.com/dotnet/BenchmarkDotNet/issues/3174#issuecomment-4745457427?

timcassell (approved):

Issue #3154, opened 2026-05-28
Failed to install v8 with jsvu on CI environment


filzrev:
Currently, following jsvu command *silently* failed on CI.
and Wasm related tests are failed.

> jsvu --os=default --engines=v8

This issue is observed on following `20260525` runner images. (It's not occurred on `20260518` image)
- `windows-latest`
- `windows-11-arm`
- `ubuntu-latest`
- `macos-intel25`

### Failed log on runner image `20260525.149.1`
```
📦 jsvu v3.0.3 — the JavaScript engine Version Updater 📦
✔ Read OS from config: win64
✔ Read engines from config: v8
❯ Finding the latest V8 version…
✔ Found latest V8 version: v15.0.232.
❯ Predicting URL…
✔ URL: https://storage.googleapis.com/chromium-v8/official/canary/v8-win64-rel-15.0.232.zip
❯ Downloading…


✔ Download completed.
❯ Extracting…
```

It looks like jsvu command **silently** exited with `0`. after `Extracting…` log.
https://github.com/GoogleChromeLabs/jsvu/blob/55f97f394f07f222f23d4a27e6cecc4a70b59370/shared/engine.js#L48-L55

### Successful log on runner image `20260518.141.1`

```
📦 jsvu v3.0.3 — the JavaScript engine Version Updater 📦
✔ Read OS from config: win64
✔ Read engines from config: v8
❯ Finding the latest V8 version…
✔ Found latest V8 version: v15.0.207.
❯ Predicting URL…
✔ URL: https://storage.googleapis.com/chromium-v8/official/canary/v8-win64-rel-15.0.207.zip
❯ Downloading…


✔ Download completed.
❯ Extracting…
Installing library to ~\.jsvu\engines\v8\icudtl.dat…
Installing library to ~\.jsvu\engines\v8\snapshot_blob.bin…
Installing binary to ~\.jsvu\engines\v8\v8.exe…
Installing wrapper script to ~\.jsvu\bin\v8.cmd…
✔ Extraction completed.
❯ Testing…
✔ Testing completed.
✔ V8 v15.0.207 has been installed! 🎉
```

### What's I've confirmed
- Disable Windows Defender
- Run command without using `pwsh`
- Fix v8 install version from latest canary to `v15.0.207` (or stable version `v14.8.x`)
- Confirm runner image diffs on https://github.com/actions/runner-images

If this problem is continued,
it need to report issue to appropriate repository.

[...]

PR #3176, opened 2026-06-18
chore: Downgrade node.js version to 22

filzrev:
On recent GitHub Actions runner images.
`Setup Additional Tools` failed.

It seems cached Node.js version `24.16.0` is used and explicitly specified `24.15.0` version seems ignored in some cases.

This PR downgrade Node.js version to `22.x` temporary.
Because Node 22 is not affected by #3154. and it can receive security updates.

### DIFF
--- a/.github/actions/setup-additional-tools/action.yaml
+++ b/.github/actions/setup-additional-tools/action.yaml
@@ -11,7 +11,7 @@ runs:
     - name: Set up node
       uses: actions/setup-node@v6
       with:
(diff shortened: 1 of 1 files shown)


filzrev:
I'm not sure why cached Node.js version is used on some CI executions.
https://github.com/dotnet/BenchmarkDotNet/actions/runs/27738890485

On latest CI executions. It seems `20260615` runner images are used.
And properly installed explicitly specified Node.js version (`24.15.0`)

Though, It's better to use Node.js `22` to receive security update until  #3154 issue is resolved.

timcassell (approved):

PR #2953, opened 2026-01-04
Add support for custom hardware counter names

ramartinez7:
Fixes #1520

This PR adds the ability to specify custom hardware counter names by their ETW profile source names, giving developers flexibility to use CPU-specific counters that aren't covered by the HardwareCounter enum.

Changes:
- Add CustomCounter class to represent custom PMC counters with configurable profile source name, short display name, interval, and higher-is-better flag
- Add CustomCountersAttribute for declarative counter specification
- Add IConfig.GetCustomCounters() and ManualConfig.AddCustomCounters() methods
- Update ImmutableConfig/ImmutableConfigBuilder to handle custom counters
- Update PreciseMachineCounter to support both enum and custom counters
- Update PmcStats to track custom counters separately
- Update PmcMetricDescriptor to use ShortName for column display
- Update EtwProfiler to collect both hardware and custom counters
- Update HardwareCounters.Validate() and Sessions to handle custom counters
- Fix ManualConfig.Add() to propagate custom counters during config merge

Usage example:
// Using attribute
[CustomCounters("DCMiss", "ICMiss", "BranchMispredictions")] public class MyBenchmark { }

// Using ManualConfig
var config = ManualConfig.CreateEmpty()
    .AddCustomCounters(
        new CustomCounter("DCMiss", shortName: "L1D$Miss"),
        new CustomCounter("ICMiss", shortName: "L1I$Miss")
    );

Developers can discover available counters on their system using: TraceEventProfileSources.GetInfo().Keys

**Note:** This is an initial implementation that has not been
thoroughly tested across different CPU architectures or edge cases. Further review
and testing is recommended before merging to confirm correctness of the changes.

If you spot issues or have improvements in mind,
please feel free to push commits directly to this branch or open your own PR with the changes.

Issue #388, opened 2017-03-12
Precise Machine Counter Diagnoser


adamsitnik:
Few days ago I saw a discussion at corefxlab regarding Spans and Benchmarking. It was clear to me that the developers struggle with a problems that BenchmarkDotNet solves in a very good way.

However when I mentioned BenchmarkDotNet as the cure, @KrzysztofCwalina [replied](https://github.com/dotnet/corefxlab/pull/1278#issuecomment-285082017):

> Yeah, we have. In general I love BenchmarkDotNet and use it in many cases. It does have better front end reporting capabilities and is more mature from reliability and statistical analysis perspective. But it seems like **BenchmarkDotNet does not support measuring instructions retired, cache misses, and branch mispredictions** that we often find very useful. Would be great to combine the efforts/features :-)

So I decided to implement this feature. I have analyzed the docs and mainly the code of xunit.performance, perfview and Microsoft.Diagnostics.Tracing.TraceEvent and with some luck found the way to get these PMC counters working for BenchmarkDotNet.

It would be great if some of our users could give me more suggestions and use cases. What do I mean by that? Following counters are available:

```
Timer
TotalIssues
BranchInstructions
CacheMisses
BranchMispredictions
TotalCycles
UnhaltedCoreCycles
InstructionRetired
UnhaltedReferenceCycles
LLCReference
LLCMisses
BranchInstructionRetired
BranchMispredictsRetired
```

Currently, I am using only:
```
BranchInstructions
CacheMisses
BranchMispredictions
InstructionRetired
```

and calculate Misspredict rate as `BranchMispredictions / BranchInstructions`. 

Should I include more counters? Is there anything more that we could calculate to make the tool more usefull? Reminder: the number of counters that can be used is limited and dependent on hardware. I can not simply record all of them (usually only 3 or 4).

/cc @goldshtn @benaadams @mattwarren @AndreyAkinshin @redknightlois 

Sample results:

```
 |             Method |        Mean |    StdDev | Misspredict rate |
 |------------------- |------------ |---------- |----------------- |
 |       SortedBranch |  22.9457 us | 1.4673 us |           0.03 % |
 |     UnsortedBranch | 147.0984 us | 2.8415 us |          23.32 % |
 |   SortedBranchless |  31.4378 us | 1.0527 us |           0.04 % |
 | UnsortedBranchless |  30.3965 us | 0.7525 us |           0.05 % |

redknightlois:
[...]

Issue #1520, opened 2020-08-21
Unable to read some hardware counters on ZEN2 CPU


alexcovington:
Having trouble reading 'CacheMisses' and 'InstructionsRetired' whenever I run the [dotnet/performance](https://github.com/dotnet/performance) benchmark suite. If I run the command from within the `src\benchmarks\micro` directory:

```powershell
dotnet run -c Release -f netcoreapp5.0 --filter '*Bilinear*' --counters CacheMisses InstructionsRetired
```

I get the following error message:

```
// Validating benchmarks:
The counter CacheMisses is not available. Please make sure you are Windows 8+ without Hyper-V
The counter InstructionRetired is not available. Please make sure you are Windows 8+ without Hyper-V
```

I am able to get some counters to read by enabling IBS in my BIOS (technically wasn't called IBS, I had to disable SVM to get it to work). So the following command does work for me:

```powershell
dotnet run -c Release -f netcoreapp5.0 --filter '*Bilinear*' --counters BranchMispredictions
```

If I profile a C# application outside of BenchmarkDotNet using AMD uProf, I can get cache miss and instructions retired statistics. I am also able to read all of the counters on my Skylake machine using the BenchmarkDotNet CLI without any issues.

Would appreciate any help that can be provided!

adamsitnik:
Hi @alexcovington 

BenchmarkDotNet is using `TraceEvent` which internally uses ETW to get hardware counters information. I am afraid that there is an AMD-specific bug somewhere.

Could you please run the following command and share your output here?

```cmd
tracelog.exe -profilesources Help
```

[tracelog.exe](https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/tracelog) might not be present in your $PATH so you can use Visual Studio Command Prompt to run this command:

![obraz](https://user-images.githubusercontent.com/6011991/96728961-ec093700-13b4-11eb-8f03-81663a99d267.png)

alexcovington:
Hi @adamsitnik, sorry for the wait on this. With COVID, I don't have physical access to this machine all the time, so I wasn't able to work on this till now.

Here's the output from tracelog running in admin Developer Command Prompt for VS 2019:

```
C:\Windows\System32>tracelog -profilesources Help
Id  Name                        Interval  Min      Max
--------------------------------------------------------------
  0 Timer                          10000  1221    1000000
  2 TotalIssues                    65536  4096 2147483647
[...]

PR #3184, opened 2026-06-24
Add hardware counter profile

bugRanger:
**SUMMARY**

It's great that version [1](https://github.com/dotnet/BenchmarkDotNet/issues/388) added counter support, but the expected counter names don't always match what the system provides.

Based on previous attempts ([1](https://github.com/dotnet/BenchmarkDotNet/pull/1438), [2](https://github.com/dotnet/BenchmarkDotNet/pull/2953)), providing free input strings provides versatility, but there are also drawbacks:
- Inconvenient for teams working in different configurations;
- Strings in attributes lead to a lot of duplicate code;

Therefore, an option with a profile description for extracting the required value sets could solve these problems.

*UPD:* Added simple test for ETW with custom profile, output result:
```
BenchmarkDotNet v0.16.0-develop (2026-06-25), Windows 10 (10.0.19045.6456/22H2/2022Update)
AMD Ryzen 7 5700G with Radeon Graphics 3.80GHz, 1 CPU, 16 logical and 8 physical cores
Memory: 27,9 GB Total, 13,77 GB Available
.NET SDK 10.0.103
  [Host] : .NET 8.0.24 (8.0.24, 8.0.2426.7010), X64 RyuJIT x86-64-v3
  Dry    : .NET 8.0.24 (8.0.24, 8.0.2426.7010), X64 RyuJIT x86-64-v3

Job=Dry  IterationCount=1  LaunchCount=1  
RunStrategy=ColdStart  UnrollFactor=1  

| Method         | Mean     | Error | FakeCacheMisses1/Op | FakeCacheMisses3/Op | FakeCacheMisses2/Op |
|--------------- |---------:|------:|--------------------:|--------------------:|--------------------:|
| SortedBranch   | 143.1 μs |    NA |             581,632 |             581,632 |              45,056 |
| UnsortedBranch | 180.3 μs |    NA |             577,536 |             577,536 |              45,056 |
```


bugRanger:
@dotnet-policy-service agree

timcassell:
It's an interesting approach. But it looks like you can only remap existing `HardwareCounter` enum members to machine-specific hardware counter names. I'm by no means an expert when it comes to hardware counters, but do we need to handle the case of a hardware counter that is not mapped to an existing enum member? Or in those cases we will just need to expand the enum to cover new cases?

tests/BenchmarkDotNet.IntegrationTests.ETW/BenchmarkDotNet.IntegrationTests.ETW.csproj
```diff
@@ -0,0 +1,50 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
filzrev, line 1: If this project is not expected to be executed on CI.

Is it able to define these tests on `BenchmarkDotNet.IntegrationTests.ManualRunning` project side?

tests/BenchmarkDotNet.IntegrationTests.ETW/HardwareCounterTests.cs
```diff
@@ -0,0 +1,130 @@
...
+using BenchmarkDotNet.Diagnosers;
+using BenchmarkDotNet.Diagnostics.Windows;
+using BenchmarkDotNet.Jobs;
+using Microsoft.Diagnostics.Tracing.Session;
+
+namespace BenchmarkDotNet.IntegrationTests.ETW;
+
+public class HardwareCounterTests(ITestOutputHelper output) : BenchmarkTestExecutor(output)
+{
+    [Fact]
```
filzrev, line 13: This test has `ETW` namespace.
Is this test works on non-Windows environment?

timcassell: I would not expect it to. https://github.com/dotnet/BenchmarkDotNet/issues/388#issuecomment-3461545178

filzrev: > I would not expect it to.

In that case, I think it need to replace `[Fact]` to use `[EnvSpecificFact(EnvRequirement.WindowsOnly)]`.

bugRanger:
> It's an interesting approach. But it looks like you can only remap existing `HardwareCounter` enum members to machine-specific hardware counter names. I'm by no means an expert when it comes to hardware counters, but do we need to handle the case of a hardware counter that is not mapped to an existing enum member? Or in those cases we will just need to expand the enum to cover new cases?

This is a good question! :+1:

I welcome any criticism, as even now I feel like my solution doesn't fully address or reflect the problem.

*A little context*
Apparently, performance counters across different manufacturers aren't coordinated in any way and don't adhere to any clear standard. This is evident in how the Windows operating system [HAL](https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/windows-kernel-mode-hal-library) itself attempts to solve the same problem using [ProfileSources](https://www.geoffchappell.com/studies/windows/km/hal/api/hwperfcnt/index.htm) (I don't claim this information is up-to-date; perhaps someone has more recent data than I do).

As far as I understand, ProfileSources includes a set of counters familiar to Windows and excludes many vendor-specific counters. But, apparently, these counters aren't ignored, but rather somehow registered by the system and made available in ETW.

However, it is worth noting that all the approaches described above understand the configuration on which they are running, while our performance testing script is described by a set of attributes or a launch configuration and does not know what system it will run on.

*Results*
I don't think we can't fully cover the entire set of counters from different manufacturers or attempt to fully map them as is done in HAL.

In my solution, the old version of HardwareCounter (which mapped directly to the counter name) is used as an abstraction when connecting the counter. This provides an advantage when setting up a benchmark scenario, as you don't need to specify which processor models the code should run on, but it does require the executor to load a profile that will implement the necessary performance counter mapping.

However, I wouldn't remove the old version, which attempts to directly map HardwareCounter to the required counters. I think it could be improved by allowing it to determine which counters it can use from the available enumeration for a given machine at the time the benchmark is launched (use table [ProfileSources](https://www.geoffchappell.com/studies/windows/km/hal/api/hwperfcnt/index.htm)).

P.S.
After all that, I took another look at my proposed solution and it seems the API may be unintuitive and confusing for developers.

bugRanger:
https://github.com/dotnet/BenchmarkDotNet/issues/1520

bugRanger:
Planned list of changes:
- Add predefined counter sets (see [Hardware Performance Counters - Emon/AMD/Default](https://www.geoffchappell.com/studies/windows/km/hal/api/hwperfcnt/index.htm)).
  * **Selection criteria:** only counters that are present in the majority and are frequently used.
- Add a log of available counters when running a benchmark (if there is at least one unknown counter).
- Allow the ability to create a custom counter set (named on the machine, named in the report).

Issue #3181, opened 2026-06-23
Benchmark process freeze when `Ctrl+C` is pressed multiple time


filzrev:
When using latest version of nightly build (`0.16.0-nightly.20260623.567`)
Benchmark process freeze when cancelling benchmark by pressing `Ctrl+C` **multiple** times.

Benchmark freeze after following log output.
> Cancelling benchmark run... press ctrl+c again to force abort.

 It seems there is problem on `force abort` logics.

#### How to reproduce problem
1. Create benchmark dotnet project with `net8.0` tfm.
2. Reference nightly build version of BenchmarkDotNet
3. Start benchmark and press `Ctrl+C` multiple times when executing `start dotnet restore`.
4. Confirm there is no progress after `Cancelling benchmark run... press ctrl+c again to force abort.` log.

#### Other issue
Currently, when benchmark is cancel cancelled with `Ctrl+C`.
Full stack trace is printed.
It might be better to show custom message and return custom exit code.  
(e.g. [ExitCode of Microsoft.Testing.Platform](https://learn.microsoft.com/en-us/dotnet/core/testing/microsoft-testing-platform-troubleshooting#exit-codes))

timcassell:
Pressing ctrl+c once cancels the CancellationToken, pressing it twice simply lets the normal dotnet exit flow run. The only thing I can think that would cause it to get stuck is if another `Console.CancelKeyPress` listener sets `ConsoleCancelEventArgs.Cancel = true`, but I don't see anything like that in the codebase. So I'm not sure what's happening. I also have not repro'd so far, it always exits when I press ctrl+c twice. [Edit] Actually when I press ctrl+c once during `start dotnet restore`, it cancels too fast for me to press it a second time, so I can't repro anyway.

> #### Other issue
> Currently, when benchmark is cancel cancelled with `Ctrl+C`. Full stack trace is printed. It might be better to show custom message and return custom exit code.

We link the user-provided CancellationToken and ctrl+c, so either one can cause the run to be canceled. I suppose we could distinguish them by catching `OperationCanceledException` and check if the user-provided token was actually canceled.

filzrev:
> I also have not repro'd so far, it always exits when I press ctrl+c twice. [Edit] Actually when I press ctrl+c once during start dotnet restore, it cancels too fast for me to press it a second time, so I can't repro anyway.

On my environment (Windows). It can be reproduced 100% of the time.
It’s tested with `dotnet run -c Release` command.
[...]

PR #3189, opened 2026-06-30
Handle `OperationCanceledException` from ctrl+c

timcassell:
Secondary part of #3181.


timcassell:
cc @filzrev

src/BenchmarkDotNet/Helpers/ExceptionHelper.cs
```diff
@@ -7,14 +7,12 @@ namespace BenchmarkDotNet.Helpers
     public static class ExceptionHelper
     {
         public static bool IsProperCancelation(Exception ex, CancellationToken cancellationToken)
```
filzrev, line 9: It's not relating this PR though,
I thought this method should be renamed to `IsProperCancellation`.
And also need to rename `IsCancelation` to `IsCancellation`

PR #3190, opened 2026-07-03
Remove dependency to `Microsoft.DotNet.PlatformAbstractions`.

teo-tsirpanis:
`Microsoft.DotNet.PlatformAbstractions` is a package that [no longer receives updates since .NET 5](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/platformabstractions-package-removed). This PR removes this dependency from BenchmarkDotNet, and replaces its usage with built-in APIs.

### DIFF
--- a/src/BenchmarkDotNet/Detectors/OsDetector.cs
+++ b/src/BenchmarkDotNet/Detectors/OsDetector.cs
@@ -5,7 +5,6 @@ using Perfolizer.Models;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using static System.Runtime.InteropServices.RuntimeInformation;
-using RuntimeEnvironment = Microsoft.DotNet.PlatformAbstractions.RuntimeEnvironment;
 
 namespace BenchmarkDotNet.Detectors;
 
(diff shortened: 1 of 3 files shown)


timcassell (approved):
Thanks

src/BenchmarkDotNet/Detectors/OsDetector.cs
```diff
@@ -53,8 +52,8 @@ private static OsInfo ResolveOs()
             }
         }
 
-        string operatingSystem = RuntimeEnvironment.OperatingSystem;
-        string operatingSystemVersion = RuntimeEnvironment.OperatingSystemVersion;
+        string operatingSystem = OSDescription;
+        string operatingSystemVersion = Environment.OSVersion.ToString();
```
filzrev, line 56: These APIs do not seem to be compatible.

`Microsoft.DotNet.PlatformAbstractions` APIs returns following outputs.
- Windows
- 10.0.26200

In contrast, replaced API returns following outputs.
- Microsoft Windows 10.0.26200
- Microsoft Windows NT 6.2.9200.0

Issue #3191, opened 2026-07-07
DoubleWrites message is recorded when using `--output` parameter on dotnet command


filzrev:
When benchmark project and dependent projects supports multiple TFMs.
**DoubleWrites** messages are recorded unexpectedly.

It's occurred when using `--output` parameter on dotnet command without `--no-dependencies`.

#### How to reproduce problem

1. Run `samples\BenchmarkDotNet.Samples` benchmark with following parameter.
    > dotnet run -c Release -f net10.0 --no-launch-profile -- --filter IntroBasic --keepFiles --generateBinLog
2. Open generated binlog file that is generated in `bin\Release\net8.0\BenchmarkDotNet.Samples-DefaultJob-1`
3. Confirm **DoubleWrites** logs exists. 
    > <img width="1296" height="100" alt="Image" src="https://github.com/user-attachments/assets/6bde3892-3f1e-4308-ab94-7c122a3aa5f4" />

Additionally, following warning is also recorded.   
(It should not be occurred. Because there is no System.Threading.Channels dependency on net8.0 tfm build)

> warning MSB3277:
>     Found conflicts between different versions of "System.Threading.Channels" that could not be resolved.

#### Background

Currently, BenchmarkDotNet using `--output` parameter when buildinding **Autogenerated** project.
Though, it's not works as expected when dependent projects tfm is resolved to multiple TFMs.

Example:  
- `BenchmarkDotNet.csproj` (net8.0) -> BenchmarkDotNet.dll (net8.0) 
- `BenchmarkDotNet.Diagnostics.Windows` (netstandard2.0)  ->`BenchmarkDotNet.csproj` (netstandard2.0) -> `BenchmarkDotNet.dll` (netstandard2.0) 

`Double Writes` issue is occurred because both project output are copied to single directory (Specified with `--output`)

It also might affects benchmark results. (It's not defined which version of DLL is used)

#### Expected solution
- Use .NET 8 SDK  as minimum requirements to use ArtifactsPath
- Add .NET SDK validator to check `dotnet --version` return .NET 8 SDK or greater.
- Build  **Original** benchmark projects with `ArtifactsPath` option (Don't use `--output`)
- Modify DllGatherer related logics
- Build **Autogenerated** project with `--no-restore --no-dependencies`  and  `ArtifactsPath` and `--output` option.

filzrev:
Following part of changes are handled by #3194
[...]

PR #3194, opened 2026-07-10
chore: Add validator to check  .NET SDK as minimum requirements

filzrev:
This PR include following changes that are commented on #3191

- Use .NET 8 SDK as minimum requirements to use `ArtifactsPath`
- Add .NET SDK validator to check dotnet --version return .NET 8 SDK or greater.

#### 1. `DotNetSdkValidator.cs`

Add logics to validate `dotnet --version` command results.
When SDK major version is lower that `8`. It raise validation error.

This might be **Breaking Changes** who running benchmarks on environment that install only .NET 7 SDK or earlier.

#### 2. `BenchmarkRunnerClean.cs`
Remove unneeded sequencial build logics.
When using .NET 8 SDK or later, `ArtifactsPath` feature can be used for older TFMs.

#### 3. `DotNetCliCommand.cs`

**1. Set initial capacity for StringBuilder.**

**2. Add `ToRelativePath` helper method. And shorten path argument.**

This helper method returns relative path when specified path starts with `BuildArtifactsDirectoryPath`
(It's used as `WorkingDirectory` on  DotNetCliCommandExecutor)

Note: MSBuild parameter(`/p:`) is leaved as is. Because it's propagated to dependent project and path is evaluated based on project file directory.

**3. Modify `GetMsBuildBinLogArgument` quotes location**

**4. Rename `MaybeAppendOutputPaths` to `MaybeAppendExtraArguments` and modify logics**
- Remove unneeded `OutDir`/`OutputPath` parameters.
- Remove unneeded trailing `Path.AltDirectorySeparatorChar` addition.
- Simplify dotnet parameters (Use relative path)
- Modify code to add arguments based on specified parameters.

### DIFF
--- a/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
+++ b/src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
@@ -148,22 +148,22 @@ namespace BenchmarkDotNet.Toolchains.DotNetCli
                 cancellationToken);
 
         internal static string GetRestoreCommand(ArtifactsPaths artifactsPaths, BuildPartition buildPartition, string filePath, string? extraArguments = null, string? binLogSuffix = null, bool excludeOutput = false)
-            => new StringBuilder()
+            => new StringBuilder(256)
                 .AppendArgument("restore")
-                .AppendArgument($"\"{filePath}\"")
+                .AppendArgument($"\"{filePath.ToRelativePath(artifactsPaths)}\"")
                 // restore doesn't support -f argument.
                 .AppendArgument(artifactsPaths.PackagesDirectoryName.IsBlank() ? string.Empty : $"--packages \"{artifactsPaths.PackagesDirectoryName}\"")
                 .AppendArgument(GetCustomMsBuildArguments(buildPartition.RepresentativeBenchmarkCase, buildPartition.Resolver))
                 .AppendArgument(extraArguments)
                 .AppendArgument(GetMandatoryMsBuildSettings(buildPartition.BuildConfiguration))
                 .AppendArgument(GetMsBuildBinLogArgument(buildPartition, binLogSuffix))
-                .MaybeAppendOutputPaths(artifactsPaths, true, excludeOutput)
+                .MaybeAppendExtraArguments(artifactsPaths, isRestore: true, excludeOutput: excludeOutput)
                 .ToString();
 
         internal static string GetBuildCommand(ArtifactsPaths artifactsPaths, BuildPartition buildPartition, string filePath, string tfm, string? extraArguments = null, string? binLogSuffix = null, bool excludeOutput = false)
-            => new StringBuilder()
+            => new StringBuilder(256)
                 .AppendArgument("build")
-                .AppendArgument($"\"{filePath}\"")
+                .AppendArgument($"\"{filePath.ToRelativePath(artifactsPaths)}\"")
                 .AppendArgument($"-f {tfm}")
                 .AppendArgument($"-c {buildPartition.BuildConfiguration}")
                 .AppendArgument(GetCustomMsBuildArguments(buildPartition.RepresentativeBenchmarkCase, buildPartition.Resolver))
@@ -171,13 +171,13 @@ namespace BenchmarkDotNet.Toolchains.DotNetCli
                 .AppendArgument(GetMandatoryMsBuildSettings(buildPartition.BuildConfiguration))
                 .AppendArgument(artifactsPaths.PackagesDirectoryName.IsBlank() ? string.Empty : $"/p:NuGetPackageRoot=\"{artifactsPaths.PackagesDirectoryName}\"")
                 .AppendArgument(GetMsBuildBinLogArgument(buildPartition, binLogSuffix))
-                .MaybeAppendOutputPaths(artifactsPaths, excludeOutput: excludeOutput)
+                .MaybeAppendExtraArguments(artifactsPaths, isBuild: true, excludeOutput: excludeOutput)
                 .ToString();
 
         internal static string GetPublishCommand(ArtifactsPaths artifactsPaths, BuildPartition buildPartition, string filePath, string tfm, string? extraArguments = null, string? binLogSuffix = null)
-            => new StringBuilder()
+            => new StringBuilder(256)
                 .AppendArgument("publish")
-                .AppendArgument($"\"{filePath}\"")
+                .AppendArgument($"\"{filePath.ToRelativePath(artifactsPaths)}\"")
                 .AppendArgument($"-f {tfm}")
                 .AppendArgument($"-c {buildPartition.BuildConfiguration}")
                 .AppendArgument(GetCustomMsBuildArguments(buildPartition.RepresentativeBenchmarkCase, buildPartition.Resolver))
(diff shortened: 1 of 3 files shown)


src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -228,18 +228,37 @@ internal static class DotNetCliCommandExtensions
...
+            bool excludeOutput = false)
+        {
+            if (!excludeOutput)
+            {
+                // Use a subdirectory for ArtifactsPath so that DefaultItemExcludes (which the SDK
+                // sets to $(ArtifactsPath)/**) doesn't cover project-level files like wwwroot/.
+                stringBuilder.AppendArgument($"--artifacts-path .artifacts");
+
+                if (!isRestore)
+                    stringBuilder.AppendArgument($"--output \"{artifactsPaths.BinariesDirectoryPath.ToRelativePath(artifactsPaths)}\"");
```
timcassell, line 246: I thought we don't need to pass `--output` with artifacts path?

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -228,18 +228,37 @@ internal static class DotNetCliCommandExtensions
...
+            }
+
+            if (isPublish)
+                stringBuilder.AppendArgument($"/p:PublishDir=\"{artifactsPaths.PublishDirectoryPath}\"");
```
timcassell, line 250: I think we can also omit `PublishDir` since we're always using artifacts path, the toolchains that use it can update their location (`.artifacts/publish` instead of `bin/publish`).

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
```diff
@@ -228,18 +228,37 @@ internal static class DotNetCliCommandExtensions
...
-                    .AppendArgument($"/p:OutDir=\"{artifactsPaths.BinariesDirectoryPath}{Path.AltDirectorySeparatorChar}\"")
-                    // OutputPath is legacy, per-project version of OutDir. We set both just in case. https://github.com/dotnet/msbuild/issues/87
-                    .AppendArgument($"/p:OutputPath=\"{artifactsPaths.BinariesDirectoryPath}{Path.AltDirectorySeparatorChar}\"")
-                    .AppendArgument($"/p:PublishDir=\"{artifactsPaths.PublishDirectoryPath}{Path.AltDirectorySeparatorChar}\"")
-                    .AppendArgument(isRestore ? string.Empty : $"--output \"{artifactsPaths.BinariesDirectoryPath}{Path.AltDirectorySeparatorChar}\"");
+        internal static StringBuilder MaybeAppendExtraArguments(
+            this StringBuilder stringBuilder,
+            ArtifactsPaths artifactsPaths,
+            bool isRestore = false,
+            bool isBuild = false,
```
timcassell, line 235: Unused param

src/BenchmarkDotNet/Validators/DotNetSdkValidator.cs
```diff
@@ -182,5 +192,25 @@ private static string CheckFor45PlusVersion(int releaseKey)
...
+            [NotNullWhen(true)] out Version? sdkVersion)
+        {
+            string exePath = customDotNetCliPath.IsBlank()
+                ? "dotnet"
+                : customDotNetCliPath!;
+
+            rawSdkVersion = ProcessHelper.RunAndReadOutput(exePath, "--version") ?? "";
+
+            // Trim `-preview`/`-rc` part.
+            var versionText = rawSdkVersion.Split('-')[0];
```
timcassell: You can use `CoreRuntime.GetParsableVersionPart` helper method.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
filzrev: I also agree `--output` argument should be removed from command line.   
(This setting output **all** project dependencies output to single directory and cause confilecs)

Though, it will be handled on another PR.

---

`artifactsPaths.BinariesDirectoryPath` value is set on early stage of `GenerateProjectAsync`.
and this path also used on `GenerateAppConfigAsync`
So BinariesDirectoryPath setting itself will not be changed.

Instead, following steps are expected.
1. Gets `TargetDir` property(`.artifacts\bin\BenchmarkDotNet.Autogenerated\release_net8.0`) on project build phase.
2. Copy output files from `TargetDir` to `BinariesDirectoryPath` when build succeeded.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
filzrev: `Publish` target directory also needs MSBuild project evaluation to determine actual value.
(e.g. it's resolved to `artifacts\publish\ConsoleApp7\release_net10.0`)

So it seems better to use **fixed** path that relative to `BuildArtifactsDirectoryPath`.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
filzrev: This parameter is being added for API consistency between commands.

It'll be removed on another PR if it's not expected to be used.

src/BenchmarkDotNet/Validators/DotNetSdkValidator.cs
filzrev: I've modified code to use `CoreRuntime.GetParsableVersionPart` on latest commit.

src/BenchmarkDotNet/Toolchains/DotNetCli/DotNetCliCommand.cs
timcassell: It seems the same issue with non-publishing toolchains for the bin directory. If we're removing `--output`, we should unify the logic for both. It seems pretty straight forward for us to construct that path since we're generating the project name, and we have the tfm and configuration.

timcassell: It looks like we could also add `<ArtifactsPivots></ArtifactsPivots>` property to the csproj to strip the `release_net10.0` part without affecting dependencies, so we can just build the path from the generated project name (we could also set `ArtifactsProjectName`, but it'd be redundant).

filzrev: > It seems the same issue with non-publishing toolchains for the bin directory. If we're removing --output, we should unify the logic for both.

It'll be expected to be handled on another PR.

> It looks like we could also add <ArtifactsPivots></ArtifactsPivots> property to the csproj

I'll create another PR to supports custom `props`/`targets` on benchmark project build.
`/p:PublishDir` setting will be removed by using this feature.

Additionally, `ErrorOnDuplicatePublishOutputFiles` setting also should be removed.

timcassell: > It'll be expected to be handled on another PR.

Let's just include the full fix in a single PR, please.

PR #3195, opened 2026-07-11
feat: add custom props/targets to customize project build

filzrev:
This PR add infrastructure to customize build with props/targets on CsProj based toolchain.

#### Background
Existing implementation have following limitations.
- `ImportDirectoryBuildProps`/`ImportDirectoryBuildTargets` are disabled.  
   So there is no way to use **custom** setting when build **original** benchmark project.
- When using `/p:` command line parameter. it's handled as **global property** and it's applied to all dependent projects.  
   It need a way to set MsBuild property to specific project.
- Ther is no generic way to add custom MSBuild target to all CsProj based toolchain.

**1. Add custom props/targets template files**
Following template files are added. (Currently, no settings are contained)
- BenchmarkDotNet.Build.props.txt
- BenchmarkDotNet.Build.targets.txt

These files expected to be used for following purposes.
- Set **project specific** MSBuild settings.
- Add custom MSBuild target. (e.g. DllGather targets without project build)

**2. `GeneratorBase.cs`**
Add `GenerateCustomBuildHooksAsync` **virtual** API and invoke on `GenerateProjectAsync`.
This API is used for generate custom props/targets files that are used on project build.

**3. `CsProjGenerator.cs.cs`**
Add `GenerateCustomBuildHooksAsync` API implementation. 

Additionally, some fixed project names are refactored to constant field.

**4. Other `Templates/*.txt` project templates**
Add import for props/targets and apply code formatter.


timcassell:
> * `ImportDirectoryBuildProps`/`ImportDirectoryBuildTargets` are disabled.
>   So there is no way to use **custom** setting when build **original** benchmark project.

Removed in #3196.

> * Set **project specific** MSBuild settings.

What do you mean? Each toolchain already sets project-specific settings.

> * Add custom MSBuild target. (e.g. DllGather targets without project build)

Can you make that change in the same PR?

I'm uncomfortable merging this without a use.

Also you missed the NativeAot generator that has its template in source.

PR #3197, opened 2026-07-16
Handle null values in joined summary columns

### LINKED ISSUE #2656: Bug: NRE possible on joining categories
If `--join` is omitted, no error is thrown.

Repro:

Run this command:

```ps1
 dotnet run -c Release -f net9.0 --join --anyCategories Linq
```

Over the [micro benchmarks in dotnet/performance](https://github.com/dotnet/performance/tree/main/src/benchmarks/micro)

Error message:

```txt
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Linq.Enumerable.MaxInteger[TSource,TResult](IEnumerable`1 source, Func`2 selector)
   at BenchmarkDotNet.Reports.SummaryTable.SummaryTableColumn..ctor(SummaryTable table, Int32 index, IColumn column, Boolean hide)
   at BenchmarkDotNet.Reports.SummaryTable..ctor(Summary summary, SummaryStyle style)
   at BenchmarkDotNet.Reports.Summary.GetTable(SummaryStyle style)
   at BenchmarkDotNet.Reports.Summary..ctor(String title, ImmutableArray`1 reports, HostEnvironmentInfo hostEnvironmentInfo, String resultsDirectoryPath, String logFilePath, TimeSpan totalTime, CultureInfo cultureInfo, ImmutableArray`1 validationErrors, ImmutableArray`1 columnHidingRules, SummaryStyle summaryStyle)
   at BenchmarkDotNet.Reports.Summary.Join(List`1 summaries, ClockSpan clockSpan)
   at BenchmarkDotNet.Running.BenchmarkSwitcher.RunWithDirtyAssemblyResolveHelper(String[] args, IConfig config, Boolean askUserForInput)
   at MicroBenchmarks.Program.Main(String[] args) in D:\forks\performance\src\benchmarks\micro\Program.cs:line 43
```

infisble:
Fixes #2656.

When summaries with different column providers are joined, a column may not have a value for benchmarks originating from another summary. `SummaryTable` stored that `null` value and later dereferenced it while calculating the column width, which caused the reported `NullReferenceException`.

This change renders missing column values as `NA` and adds a regression test that joins two summaries where a custom column only applies to one of them.

Tests:

- `dotnet test tests/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj -c Release -f net10.0 --no-restore --filter FullyQualifiedName~JoinedSummaryHandlesMissingColumnValues` (1 passed)
- `dotnet test tests/BenchmarkDotNet.Tests/BenchmarkDotNet.Tests.csproj -c Release -f net10.0 --no-restore --no-build --filter FullyQualifiedName~BenchmarkDotNet.Tests.Reports.SummaryTableTests` (23 passed)
- Full `BenchmarkDotNet.Tests` run: 1022 passed, 4 skipped, 2 failed. The two Linux CPU parser fixture failures reproduce unchanged on a clean `upstream/master` worktree.

### DIFF
--- a/src/BenchmarkDotNet/Reports/SummaryTable.cs
+++ b/src/BenchmarkDotNet/Reports/SummaryTable.cs
@@ -61,7 +61,7 @@ namespace BenchmarkDotNet.Reports
             ColumnCount = columns.Length;
             FullHeader = columns.Select(c => c.GetColumnTitle(style)).ToArray();
 
-            FullContent = summary.Reports.Select(r => columns.Select(c => c.GetValue(summary, r.BenchmarkCase, style) ?? string.Empty).ToArray()).ToArray();
+            FullContent = summary.Reports.Select(r => columns.Select(c => c.GetValue(summary, r.BenchmarkCase, style) ?? "-").ToArray()).ToArray();
             IsDefault = columns.Select(c => summary.Reports.All(r => c.IsDefault(summary, r.BenchmarkCase))).ToArray();
 
             var highlightGroupKeys = summary.BenchmarksCases.Select(b => b.Config.Orderer.GetHighlightGroupKey(b)).ToArray();
--- a/src/BenchmarkDotNet/Reports/SummaryTable.cs
+++ b/src/BenchmarkDotNet/Reports/SummaryTable.cs
@@ -61,7 +61,7 @@ namespace BenchmarkDotNet.Reports
             ColumnCount = columns.Length;
             FullHeader = columns.Select(c => c.GetColumnTitle(style)).ToArray();
 
-            FullContent = summary.Reports.Select(r => columns.Select(c => c.GetValue(summary, r.BenchmarkCase, style)).ToArray()).ToArray();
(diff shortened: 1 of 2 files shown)


timcassell:
I thought usually the non-existing values are displayed as `-` rather than empty. It would be good to unify the behavior.

infisble:
Good point. I updated the fallback to use `-` and adjusted the regression test in 6c3b992. The focused test passes, along with all 23 `SummaryTableTests`.

timcassell:
Actually looking at the code it seems the common display values for non-existant are `NA` or `?`. `-` is used for 0 values in metric columns. I think `NA` fits better here.

infisble:
Thanks for double-checking. I switched the fallback to `NA` and updated the regression test in 626fef8. The focused test passes, along with all 23 `SummaryTableTests`.

timcassell (approved):
Thanks @infisble

PR #3198, opened 2026-07-18
chore: Fix error when running benchmark with file-based app

filzrev:
This PR intended following validation error when running benchmark with file-based apps.

> //    * Assembly run, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null is located in temp. If you are running benchmarks from xUnit you need to disable shadow copy. It's not supported by design.

**1. `Portability/RuntimeInformation.cs`**
Add `IsFileBasedApp` API to detect benchmark is running on [File-based apps](https://learn.microsoft.com/en-us/dotnet/core/sdk/file-based-apps) or not.

**2. `ShadowCopyValidator.cs`**
Add logics to skip validatation when running file-based apps.

**3. `MosCpuDetector.cs`**
Add logics to avoid following errors when running file-based apps without `PublishAot=false` setting.  
(Note: As far as I knows, Currently **Benchmark Host** is not support NativeAot, Because it use reflections)

```
System.TypeInitializationException: The type initializer for 'System.Management.ManagementPath' threw an exception.
 ---> System.NotSupportedException: Built-in COM has been disabled via a feature switch. See https://aka.ms/dotnet-illink/com for more information.
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean wrapExceptions)
   at System.Management.MTAHelper.CreateInMTA(Type type)
   at System.Management.ManagementPath.CreateWbemPath(String path)
   at System.Management.ManagementPath..ctor(String path)
   at System.Management.ManagementPath..cctor()
   at System.Runtime.CompilerServices.InitHelpers.CallClassConstructor(Void* cctor, Void* instantiatingArg, Exception* pException)
```

#### TestCode

> dotnet main.cs -c Release

```csharp
#!/usr/bin/env dotnet

// Use latest preview version
#:package BenchmarkDotNet@*-preview*
#:property PublishAot=false

using System.Collections;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Toolchains;
using BenchmarkDotNet.Toolchains.InProcess.Emit;

var config = DefaultConfig.Instance
    .AddJob(Job.ShortRun.WithToolchain(InProcessEmitToolchain.Default));

if(args.Length == 0)
    args = ["--filter", "*"];
    
var summaries = BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly)
                                 .Run(args, config)
                                 .ToArray();

if (summaries.Any(x => x.Reports.Any(r => !r.Success)))
    return 1;
return 0;

namespace Benchmarks
{
    public class SampleBenchmarks
    {
        [Benchmark]
        public bool Execute() 
        {
            return true;
        }
    }
}

```

### DIFF
--- a/src/BenchmarkDotNet/Validators/ShadowCopyValidator.cs
+++ b/src/BenchmarkDotNet/Validators/ShadowCopyValidator.cs
@@ -1,3 +1,4 @@
+using BenchmarkDotNet.Portability;
 using System.Reflection;
 
 namespace BenchmarkDotNet.Validators
@@ -11,7 +12,12 @@ namespace BenchmarkDotNet.Validators
         public bool TreatsWarningsAsErrors => false;
 
         public IAsyncEnumerable<ValidationError> ValidateAsync(ValidationParameters validationParameters)
-            => validationParameters
+        {
+            // Skip validation when running on file-based app. because it's executed on temp directory.
+            if (RuntimeInformation.IsFileBasedApp)
+                return AsyncEnumerable.Empty<ValidationError>();
+
+            return validationParameters
                 .Benchmarks
                 .Select(benchmark => benchmark.Descriptor.Type.GetTypeInfo().Assembly)
                 .Distinct()
@@ -22,5 +28,6 @@ namespace BenchmarkDotNet.Validators
                         $"Assembly {assembly} is located in temp. If you are running benchmarks from xUnit you need to disable shadow copy. It's not supported by design.")
                 )
                 .ToAsyncEnumerable();
+        }
     }
 }
--- a/src/BenchmarkDotNet/Detectors/Cpu/Windows/MosCpuDetector.cs
+++ b/src/BenchmarkDotNet/Detectors/Cpu/Windows/MosCpuDetector.cs
@@ -10,7 +10,7 @@ namespace BenchmarkDotNet.Detectors.Cpu.Windows;
 internal class MosCpuDetector : ICpuDetector
 {
     [SupportedOSPlatform("windows")]
-    public bool IsApplicable() => OsDetector.IsWindows() && !RuntimeInformation.IsMono;
+    public bool IsApplicable() => OsDetector.IsWindows() && !RuntimeInformation.IsNativeAOT && !RuntimeInformation.IsMono;
 
     [SupportedOSPlatform("windows")]
     public CpuInfo? Detect()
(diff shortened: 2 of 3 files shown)


timcassell:
> **3. `MosCpuDetector.cs`**
> Add logics to avoid following errors when running file-based apps without `PublishAot=false` setting.
> (Note: As far as I knows, Currently **Benchmark Host** is not support NativeAot, Because it use reflections)

We should rather catch the exception like we do in other parts of the code that use COM. Built-in COM can be disabled without AOT, and I believe it can be enabled with AOT.

src/BenchmarkDotNet/Validators/ShadowCopyValidator.cs
```diff
@@ -11,7 +12,12 @@ private ShadowCopyValidator() { }
         public bool TreatsWarningsAsErrors => false;
 
         public IAsyncEnumerable<ValidationError> ValidateAsync(ValidationParameters validationParameters)
-            => validationParameters
+        {
+            // Skip validation when running on file-based app.
+            if (RuntimeInformation.IsFileBasedApp)
+                AsyncEnumerable.Empty<ValidationError>();
```
timcassell:
```suggestion
                return AsyncEnumerable.Empty<ValidationError>();
```

src/BenchmarkDotNet/Validators/ShadowCopyValidator.cs
```diff
@@ -11,7 +12,12 @@ private ShadowCopyValidator() { }
         public bool TreatsWarningsAsErrors => false;
 
         public IAsyncEnumerable<ValidationError> ValidateAsync(ValidationParameters validationParameters)
-            => validationParameters
+        {
+            // Skip validation when running on file-based app.
```
timcassell: An explanation for _why_ it's being skipped would help.

timcassell (changes requested):

filzrev:
> We should rather catch the exception like we do in other parts of the code that use COM. Built-in COM can be disabled without AOT, and I believe it can be enabled with AOT.

I've updated code based on following try-catch code.
https://github.com/dotnet/BenchmarkDotNet/blob/2365829b82d95843e561f9ef666f4e9e86761d38/src/BenchmarkDotNet/Helpers/TaskbarProgress.cs#L125-L130

timcassell (approved):

PR #3159, opened 2026-05-29
deps: Update Perfolizer package version to `0.7.1`

filzrev:
This PR update `Perfolizer` package dependency to latest version.

Currently, this package version is specified explicitly.
So if there are any known issues on updating `Perfolizer` version, Please close this PR.

As far as I've confirmed, there is following comment.
https://github.com/dotnet/BenchmarkDotNet/issues/3056#issuecomment-4113773740

Issue #3056, opened 2026-03-21
--statisticalTest silently accepts bare numbers (e.g. 0.02) as absolute nanosecond threshold instead of percentage


danmoseley:
Ran benchmarks with `--statisticalTest 0.02` thinking this meant "2% threshold". It appears that without units supplied, it assumes nanoseconds (0.02ns here). That is far too low for most benchmarks. So every row is shown as Faster or Slower, not Same. And the emitted results table does not show the units so it isn't clear why this happened.

## Root cause

Perfolizer `MeasurementFormatter.TryParse` (in `Perfolizer.Metrology`):
- Splits input into numeric part + unit suffix
- If no unit suffix: creates `Measurement(value)` with default unit (`MeasurementUnit.Number`)
- `MeasurementUnit.Number` applies as `x + value` (absolute additive shift)

Valid formats require a unit suffix: `5%`, `10ms`, `100ns`, `1s`, `3ms`.
Bare numbers like `0.02` parse successfully but produce meaningless results.

BDN `ConfigParser.cs` calls `Threshold.TryParse` and only checks the bool return —
no validation that the parsed threshold makes sense for the context.

## What should happen

Include units in column header — instead of `MannWhitney(0.02)` show `MannWhitney(0.02ns)`.

Other options: warn if no unit supplied; or reject without units (breaking change)

## Relevant files

Used BDN v0.16.0-nightly, looked at Perfolizer from AndreyAkinshin/perfolizer main branch.

- `AndreyAkinshin/perfolizer`: `src/Perfolizer/Perfolizer/Metrology/MeasurementFormatter.cs`
  (the TryParse that accepts bare numbers)
- `dotnet/BenchmarkDotNet`: `src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs`
  (could add validation before creating StatisticalTestColumn)
- `dotnet/BenchmarkDotNet`: `src/BenchmarkDotNet/Columns/StatisticalTestColumn.cs`
  (column header format, `ColumnName => $"MannWhitney({Threshold})"`)

meiranzheng:
Opened PR #3059 to address this: normalize unitless --statisticalTest values to 'ns' (so headers show units) and emit a warning pointing to '%' for relative thresholds. Also updated docs/help text + added a regression test. https://github.com/dotnet/BenchmarkDotNet/pull/3059

AndreyAkinshin:
[...]

PR #3209, opened 2026-07-26
chore(deps): upgrade Perfolizer from 0.6.6 to 0.7.5

jn8128:
Perfolizer 0.7.x follows Pragmastat 11-14, which removed the configurable presentation layer from metrology, so unit formatting moves into BenchmarkDotNet. Reported output stays byte-for-byte the same: microseconds still print as `μs`, and the statistics are untouched, since the deprecated significance testing API is kept on purpose while [that subsystem awaits its own rework](https://github.com/dotnet/BenchmarkDotNet/issues/3056#issuecomment-4113773740). This picks up where #3159 stopped, adapting the incompatible code it ran into.

## Appendix

**What upstream removed.** `UnitPresentation`, `MeasurementUnitFormatter`, `SampleFormatter`, and the settings-aware `MeasurementFormatter` are gone, along with `AbbreviationAscii` and the `NumberValue`/`RatioValue`/`DisparityValue` wrappers. How a measurement looks is now the consumer's decision, so `UnitHelper` becomes that layer: it formats value, gap, and abbreviation, using the invariant culture unless one is supplied, and it owns the abbreviation shown to users. Cells, column headers, and the time-unit legend all resolve abbreviations through it, so they cannot drift apart.

**Microseconds.** Perfolizer reports them as ASCII `us` since 0.7.0. BenchmarkDotNet keeps printing `μs` and lets `ConsoleLogger` downgrade it for terminals without Unicode support, exactly as before. Since perfolizer parsers now recognize only the ASCII spelling, thresholds arriving from the command line are normalized on the way in, so `--statisticalTest 5μs` keeps working.

**Statistics left alone.** `SimpleEquivalenceTest` and `MannWhitneyTest` are obsolete in favor of `Pragmastat.Toolkit.Compare2`, which reports per-metric verdicts against typed thresholds instead of a single equivalence result. Adopting it would change every Same/Faster/Slower verdict and every rank, so the three call sites suppress the warning and keep today's behavior; the switch belongs in a dedicated change with validation against real runs.

**Smaller adaptations.** `Threshold` becomes ambiguous between `Perfolizer.Metrology` and `Pragmastat` once both namespaces are in scope, resolved with an alias. Perfonar table snapshots are updated rather than preserved: those exporters are internal and unreachable from any public config, and the layout shift comes from perfolizer dropping the gap in its own table renderer.

**Why 0.7.5 and not 0.7.4.** In 0.7.4 a missing format provider falls through to the ambient culture, so on a non-English machine a processor brand renders as `(Max: 3,10GHz)` and a column name as `MannWhitney(2,5%)`. Fixed in AndreyAkinshin/perfolizer#33, released as 0.7.5, which also pins the formatting contract with round-trip and culture tests.

Summary tables were checked against a live benchmark run, not just unit tests: `12.63 μs` in markdown and Unicode console output, `Mean [μs]` when units go into the header, `us` in the plain console.


jn8128:
@dotnet-policy-service agree

AndreyAkinshin (approved):

PR #3211, opened 2026-07-28
Disambiguate type display names in benchmark selection

### LINKED ISSUE #3202: [Feature Request]: Customize/Disambiguate Benchmark Fixture Names for BenchmarkSwitcher.Run()
I frequently have to deal with the fact that benchmarks with the same class name, but in different namespaces, are displayed ambigously, by `BenchmarkSwitcher.Run()`.

E.G.

```cs
namespace BenchmarkTest.Foo
{
    public class Fixture
    {
        [Benchmark]
        public void Baz() { }
    }
}

namespace BenchmarkTest.Bar
{
    public class Fixture
    {
        [Benchmark]
        public void Baz() { }
    }
}
```

When running this project via `BenchmarkSwitcher.Run()`, we get...

```
Available Benchmarks:
  #0 Fixture
  #1 Fixture
```

What I've taken to doing to work around this is duplicate each fixture class's namespace, within its name, E.G. `Foo_Fixture` and `Bar_Fixture`. Which gets the job done, but is annoying.

### Proposal

A. Adjust the logic within [`UserInteraction.PrintAvailable()`](https://github.com/dotnet/BenchmarkDotNet/blob/2365829b82d95843e561f9ef666f4e9e86761d38/src/BenchmarkDotNet/Running/UserInteraction.cs#L108) to detect when there's ambiguous options in the list, and add namespace information to disambiguate. I'd probablydo this by using the fully-qualified class name, by default, and then trimming out redundant chunks, or something along those lines.

B. Add a new [FixtureName] or similarly-named attribute, which [`ReflectionExtensions.GetDisplayName()`](https://github.com/dotnet/BenchmarkDotNet/blob/2365829b82d95843e561f9ef666f4e9e86761d38/src/BenchmarkDotNet/Extensions/ReflectionExtensions.cs#L118) would look for and use, when present. Alternatively, maybe and existing attribute could be reused, such as [DisplayName] from `System.ComponentModel.DataAnnotations` or `[Display(Name = "")] from `BenchmarkDotNet.Attributes`.

I'd be happy to submit a PR for either of these options.

sheddy123:
#3202
Disambiguate type display names in benchmark selection

Added GetDisambiguatedDisplayNames to qualify ambiguous type names with namespaces. Updated UserInteraction to use these names for clearer benchmark selection and filtering. Added unit tests and supporting fixture classes to verify correct behavior.

image

### DIFF
--- a/src/BenchmarkDotNet/Extensions/ReflectionExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ReflectionExtensions.cs
@@ -131,9 +131,9 @@ namespace BenchmarkDotNet.Extensions
             var displayNames = new string[types.Count];
             for (int i = 0; i < types.Count; i++)
             {
-                string? @namespace = types[i].Namespace;
-                displayNames[i] = ambiguousNames.Contains(simpleNames[i]) && !string.IsNullOrEmpty(@namespace)
-                    ? $"{@namespace}.{simpleNames[i]}"
+                string? fullName = types[i].FullName;
+                displayNames[i] = ambiguousNames.Contains(simpleNames[i]) && !string.IsNullOrEmpty(fullName)
+                    ? fullName
                     : simpleNames[i];
             }
 
--- a/src/BenchmarkDotNet/Extensions/ReflectionExtensions.cs
+++ b/src/BenchmarkDotNet/Extensions/ReflectionExtensions.cs
@@ -113,6 +113,34 @@ namespace BenchmarkDotNet.Extensions
         internal static string GetDisplayName(this Type type) => GetDisplayName(type.GetTypeInfo());
 
         /// <summary>
+        /// Returns a display name per type, aligned with <paramref name="types"/>. When more than one
+        /// type shares the same simple <see cref="GetDisplayName(Type)"/> (e.g. same class name in
+        /// different namespaces), the colliding names are qualified with their namespace so they can be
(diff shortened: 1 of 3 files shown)


sheddy123:
@dotnet-policy-service agree

src/BenchmarkDotNet/Extensions/ReflectionExtensions.cs
```diff
@@ -112,6 +112,34 @@ private static IEnumerable<string> GetNestedTypeNames(Type type, bool includeGen
...
+            var ambiguousNames = new HashSet<string>(
+                simpleNames.GroupBy(name => name).Where(group => group.Count() > 1).Select(group => group.Key),
+                StringComparer.Ordinal);
+
+            var displayNames = new string[types.Count];
+            for (int i = 0; i < types.Count; i++)
+            {
+                string? @namespace = types[i].Namespace;
+                displayNames[i] = ambiguousNames.Contains(simpleNames[i]) && !string.IsNullOrEmpty(@namespace)
+                    ? $"{@namespace}.{simpleNames[i]}"
```
timcassell: You can use `Type.FullName` instead.

sheddy123: This has been resolved @timcassell

timcassell (approved):
Thanks @sheddy123

PR #3214, opened 2026-07-29
Feature/2621 add export name for joined exports

### LINKED ISSUE #2621: Feature Request: Add ability to set export name for joined exports.
In Summary.cs line 92 the title and therefore the file name for exports are set to `$"BenchmarkRun-joined-{DateTime.Now:yyyy-MM-dd-HH-mm-ss}"`
If you run many sets of joined benchmarks it is hard do find the right one.

There should be an option to at least replace the DateTime part or the whole title with some custom string.

sheddy123:
#2621 
These implementations were carried out:
Added a configurable Title on the config, used as the joined summary's title. This drives the exported file names (via ExporterBase.GetFileName → summary.
Title for multi-type/joined summaries). When unset, the existing BenchmarkRun-joined-{timestamp} default is used, so behavior is unchanged for everyone else..

You can now set it two ways:

// Fluent API
ManualConfig.Create(DefaultConfig.Instance).WithTitle("MyBenchmarks");
// CLI
--join --title=MyBenchmarks

| Scenario | How title is set | Exported files | HTML `<title>` |
|----------|-------------------|----------------|----------------|
| **Code API** | `.WithTitle("MyCustomBenchmarks")` | `MyCustomBenchmarks-report.csv` / `-report-github.md` / `-report.html` | `MyCustomBenchmarks` |
| **CLI** | `--join --title CliCustomTitle` | `CliCustomTitle-report.csv` / `...` | `CliCustomTitle` |
| **Default (unchanged)** | *(nothing)* | `BenchmarkRun-joined-2026-07-29-15-25-19-report.csv` / `...` | `BenchmarkRun-joined-2026-07-29-15-25-19` |

### DIFF
--- a/docs/articles/configs/exporters.md
+++ b/docs/articles/configs/exporters.md
@@ -10,6 +10,16 @@ By default, files with results will be located in
 `.\BenchmarkDotNet.Artifacts\results` directory, but this can be changed via the `ArtifactsPath` property in the `IConfig`. 
 Default exporters are: csv, html and markdown.
 
+The exported files are named after the title of the summary they belong to, which by default is the full name of
+the benchmarked type (or `BenchmarkRun-joined` when the results of many types are joined into a single summary).
+You can override it via the `Title` property in the `IConfig` (`--title` on the command line); when the run reports
+more than one summary, the name of the benchmarked type is appended to the title, so that the results of one type
+don't overwrite the results of another:
+
+```cs
+ManualConfig.Create(DefaultConfig.Instance).WithTitle("MyBenchmarks"); // MyBenchmarks-report-github.md
+```
+
 ---
 
 [!include[IntroExport](../samples/IntroExport.md)]
--- a/docs/articles/guides/console-args.md
+++ b/docs/articles/guides/console-args.md
@@ -274,7 +274,7 @@ dotnet run -c Release -- --filter * --runtimes net6.0 net8.0 --statisticalTest 5
 * `--anyCategories`           Any Categories to run
 * `--attribute`               Run all methods with given attribute (applied to class or method)
 * `--join`                    (Default: false) Prints single table with results for all benchmarks
-* `--title`                   Custom title for the joined summary and the base name of its exported result files
+* `--title`                   Custom title for the produced summaries and the base name of the result files exported for them
 * `--keepFiles`               (Default: false) Determines if all auto-generated files should be kept or removed after running the benchmarks.
 * `--noOverwrite`             (Default: false) Determines if the exported result files should not be overwritten (be default they are overwritten).
 * `--counters`                Hardware Counters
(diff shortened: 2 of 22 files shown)


sheddy123:
@timcassell this is up for approval/merge

timcassell:
Does it not affect the title of non-joined summaries? Should it?

sheddy123:
@timcassell  
It doesn't config.Title is only read for Summary.Join; non-joined summaries still get their title from GetTitle(...).

timcassell:
I think it should. The simplest fix looks like would be to prepend the title in the single type case in https://github.com/dotnet/BenchmarkDotNet/blob/95d469d181edca15f9ac7d634ef3ca5d63a7eb59/src/BenchmarkDotNet/Exporters/ExporterBase.cs#L57-L58
```cs
$"{summary.Title}-{FolderNameHelper.ToFolderName(targets.Single())}"
```
With an added benefit of preventing overwriting multiple run results (the same way the joined title behaves).

Also, this option should override the log file name. Currently it looks like the log file uses different logic and results in a different name than the exported files, it might be good to unify them.

sheddy123:
@timcassell 
Based on the literal suggestion, for a non-joined run, summary.Title is already {FullTypeName}-{timestamp} 
so $"{summary.Title}-{ToFolderName(targets.Single())}" produces:

BenchmarkDotNet.Samples.IntroBasic-20260730-101112-BenchmarkDotNet.Samples.IntroBasic-report.csv

The type name lands twice whenever no custom title is set. The prepend only adds information when the title is custom.

timcassell:
We should redesign how it works, then. The implementation for `ExporterBase.GetFileName` looks pretty hacky as it is. It's fine if it requires breaking changes to get it right.

sheddy123:
@timcassell 

This was the redesign I did

# The design
`Summary.Title` is now the single source of truth; the runner guarantees it is unique within a run, and `GetArtifactFullName` is one expression over it; the heuristic is gone. All the rules live in the new `Helpers/TitleHelper.cs`:
- custom title set - used verbatim; `-{TypeName}` appended only when the run reports more than one summary, so their files can't collide
- no custom title - `{FullTypeName}` for a single type, `BenchmarkRun-joined` for a joined summary
- no timestamp in the base name; `ToFolderName` already includes the namespace so #529 is covered, and `DontOverwriteResults` (timestamped folder) already owns run retention
- the `.log` file stays `{title}-{timestamp}.log`, since `--resume` has to find previous runs; `GetIdToResume` now matches `{title}-*.log`
- titles are escaped for path validity and truncated against a budget that accounts for the results folder, so export paths are protected too

I have included a screenshot of testing this locally. Let me know if this implementation suffices

image

sheddy123:
Hi, @timcassell any update on this

timcassell (approved):
Thanks @sheddy123

Issue #3182, opened 2026-06-23
Reorganize command line interface options/parameters


filzrev:
As commented at https://github.com/dotnet/BenchmarkDotNet/pull/3162#issuecomment-4592633337

> I wonder if we should use this as an opportunity to consider redesigning the existing command line arguments.
> 
> What we have as of today is literally two commands (run benchmarks or display info) and the benchmarking command has way too many options/arguments (I am the one to blame for that). Moreover, it certain things are impossible to configure with command line.

BenchmarkDotNet command line interface are complicated and need to be simplified.

Currently, there are following options/arguments categories are exists. and these options are specified on flat hierarchy.
So it can't specify multiple config/job level options via command line interface. 

### `Global` level options
- `--help`
- `--version`
- `--info`
- `--list`

### `Config` level options
- `--exporters`
- `--counters`
- `--memory`
- `--threading`
- `--exceptions`
- `--disasm`
- `--profiler`
- `--allStats`
- `--statisticalTest`
- `--artifacts`
- `--filter`
- `--hide`
- Options that are defined at [ConfigOptions](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/Configs/ConfigOptions.cs)
- `--maxWidth`
- `--buildTimeout`
- `--wakeLock`

### `Job` level options
Other that are not listed above.

filzrev:
Personally, I think `configuration via code`  is sufficient for general purpose benchmark.
So `Config level options`/`Job level options` should be removed.   
(Attritionary, attribute-based options that adding `Job` should be removed also. Because it can't be merged with code based config)

And I want functionality to `specifying a Benchmark profile key`  from the Command Line feature.
(e.g. Run benchmark InProcess for debugging / Compare performance between .NET versions)

sheddy123:
@timcassell does this implementation suffice?

timcassell:
I will defer to @adamsitnik.

sheddy123:
@adamsitnik does this implementation suffice?

PR #3216, opened 2026-07-30
Reorganize command line interface options/parameters #3182

sheddy123:
Issue #3182's 
@filzrev based on your suggestions, "I have specified a benchmark profile key from the command line".
A profile is now a job Id, matched with globs, case-insensitively:

dotnet run -c Release -- --filter '*JsonSerialization*' --profile debug
dotnet run -c Release -- --filter '*' --profile net8 net9

Ids come from Job.Default.WithId("net8") in code or [SimpleJob(RuntimeMoniker.Net80, id: "net8")]. 
Job.Id already existed but was only ever used for display, folder names, and ordering, never selection.


filzrev:
This PR appears to provide a **filtering** feature based on JobId.

It's a bit different feature that I've commented at #3182.
And adding `--profile` in addition to `--profiler` seems to be confusing as commented on documentation. 

I personally thought it's better to provide `--filter-jobid` like syntax that similar to xUnit.v3 MTP-mode command line .
https://xunit.net/docs/getting-started/v3/microsoft-testing-platform#enabling-the-command-line-experience

By using `--filter-*` argument syntax,
It can be easily extend/group filter related command line argument.

sheddy123:
@filzrev does this suffice

filzrev:
It looks good to me about the PR content that adding `--filterJobId` cli argument.

If there are no other issues, please go ahead and merge it.
I'll expected to add another `--filterUid` option on another PR.

**Note:**
Personally,  I don't sure whether it's better to add new filter options to CLI options or not.

Original issue's intent is to reorganize command line arguments and reduce options.
So it might be desirable to introduce another **integrated filter syntax** for `--filter`

**Example**
- VSTest filter  
(It' can be integrate via [Microsoft.TestPlatform.Filter.Source](https://www.nuget.org/packages/Microsoft.TestPlatform.Filter.Source) package
- MTP's `--treenode-filter`
- [XUnit's Query Filter Language](https://xunit.net/docs/query-filter-language)

sheddy123:
Hi @filzrev thanks for the review

Adding the advantage of adding --filterJobId option is  makes the CLI semantics unambiguous, enables distinct error messages, and reflects the real distinction between "which benchmarks?" and "in which jobs?"
In addition, I cannot merge from my end, can you please help with that

timcassell:
I'm hesitant to merge this. Adding a bunch of different --filter* arguments is smelly. I also am not happy using job ID (I mentioned a while back that it's a bad API). We should probably do something more like job categories, and expand the --filter query instead of new filter arguments like @filzrev suggested.

sheddy123:
@timcassell Thanks for the update. I was wondering if this is what you meant:
- The selection key should be based on job categories rather than the job ID. 
- Adding a Categories property to Job, most likely under MetaMode, since this is metadata and should not affect the job’s execution. 
- Provide WithCategory/WithCategories methods and support an attribute path, mirroring the existing [BenchmarkCategory] approach.

timcassell:
Looks like this is superseded by #3239.

PR #3223, opened 2026-08-05
chore: Modify NativeAotToolchain and suppress some warnings

filzrev:
This PR contains following NativeAotToolchain related changes.

#### 1. Modify NuGet feed URL for `NativeAot90`/`NativeAot100`/`NativeAotToolchain.Net10_0`
Replace to normal NuGet feed URL.

This URL seems not be actually used.
Because `PackageReference` to `Microsoft.DotNet.ILCompiler` is added when ilCompilerVersion is empty. and it's handled by `<PublishAot>true</PublishAot>` setting.

It might be better to remove these URLs on another PR.  
(It need to verify .NET 6 NativeAot behavior because it don't support `PublishAot`)

#### 2. Add `RuntimeInformation.GetCoreLibDllLocation()` helper method

Add helper method to get `System.Private.CoreLib.dll` location and suppress IL Linker related warning.
And replace existing code that using `typeof(object).Assembly.Location`.

#### 3. Modify `NativeAot/Generator.cs`

`<TrimMode>link</TrimMode>` and `<TrimmerDefaultAction>link</TrimmerDefaultAction>` seems not supported by .NET 7 or later.

> Microsoft.NET.ILLink.targets(227,5): warning : Property 'TrimmerDefaultAction' is deprecated in .NET 7 and higher and will be ignored. Use TrimMode instead. 

So I've modified setting to use `<TrimMode>full</TrimMode>`


src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs
```diff
@@ -104,8 +104,8 @@ internal static bool TryGetVersion([NotNullWhen(true)] out Version? version)
                 return true;
             }
 
-            string coreclrLocation = typeof(object).Assembly.Location;
-            // Single-file publish has empty assembly location.
+            // Single-file publish and NativeAot aoo has empty assembly location.
```
timcassell:
```suggestion
            // Single-file publish and NativeAot app have empty assembly location.
```

src/BenchmarkDotNet/Portability/RuntimeInformation.cs
```diff
@@ -160,8 +161,8 @@ private static string GetNetCoreVersion()
 
             string GetDetailedVersion()
             {
-                string coreclrLocation = typeof(object).GetTypeInfo().Assembly.Location;
-                // Single-file publish has empty assembly location.
+                // Single-file publish and NativeAot app has empty assembly location.
```
timcassell:
```suggestion
                // Single-file publish and NativeAot app have empty assembly location.
```

timcassell:
> (It need to verify .NET 6 NativeAot behavior because it don't support `PublishAot`)

We should simply remove NativeAot 6.0. It was experimental until .Net 7.

src/BenchmarkDotNet/Toolchains/NativeAot/NativeAotToolchain.cs
```diff
@@ -8,50 +8,37 @@ namespace BenchmarkDotNet.Toolchains.NativeAot
...
-            .UseNuGet("", "https://api.nuget.org/v3/index.json")
             .TargetFrameworkMoniker("net10.0")
             .ToToolchain();
 
         /// <summary>
-        /// compiled as net11.0, targets latest NativeAOT build from the .NET 11 feed: https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet11/nuget/v3/index.json
+        /// compiled as net11.0.
         /// </summary>
         public static readonly IToolchain Net11_0 = CreateBuilder()
-            .UseNuGet("", "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet11/nuget/v3/index.json")
```
timcassell, line 54: Does NativeAot11.0 work without that feed?

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
```diff
@@ -174,9 +174,6 @@ private static IToolchain GetToolchain(RuntimeMoniker runtimeMoniker)
                 case RuntimeMoniker.Net11_0:
                     return CsProjCoreToolchain.NetCoreApp11_0;
 
-                case RuntimeMoniker.NativeAot60:
```
timcassell, line 177: Remove the moniker itself, and the `NativeAotRuntime` also.

filzrev:
I've added following changes on latest commit.

1. Remove `NativeAot60` toolchain support
Because it's **experimental** feature and  looks like  not working [after migrated to use `PublishAot`](https://github.com/dotnet/BenchmarkDotNet/pull/2095))

2. Remove NuGet feeds for nativeAot
It's not used when `ilCompilerVersion`  is not specified.
And it's not expected specified URL is used after `NativeAotToolchain` instance is generated.

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
```diff
@@ -190,16 +190,20 @@ private string GetCustomProperties(BuildPartition buildPartition, ILogger logger
...
-                // Use the defaults
-                ? ""
-                // TrimMode is set in explicit way as for older versions it might have different default value
-                : "<TrimMode>link</TrimMode><TrimmerDefaultAction>link</TrimmerDefaultAction>";
+        {
+            if (rootAllApplicationAssemblies)
+                return ""; // Use the defaults
+
+            // TrimMode is set in explicit way as for older versions it might have different default value
+            if (TargetFrameworkMoniker.StartsWith("net6.0"))
```
timcassell: Can drop this branch since 6.0 was removed.

src/BenchmarkDotNet/Toolchains/NativeAot/NativeAotToolchain.cs
timcassell: I guess it should since .Net 11 preview is already public. We should at least leave a comment to include the .Net 12 feed when it's needed in the future.

filzrev: ILCompiler package is continuously published to NuGet.org from .NET 11 preview.2.
https://www.nuget.org/packages/Microsoft.DotNet.ILCompiler

So it's expected to be works. 

As far as I've confirmed, specified NuGet feed seems not used after https://github.com/dotnet/BenchmarkDotNet/pull/2095 is merged

Though, It's not tested on my local environment,
Because NativeAot related tests failed on my local environment with following error.

> LINK : fatal error LNK1104: cannot open file 'C:\Projects\GitHub\thirdparty\BenchmarkDotNet\tests\BenchmarkDotNet.IntegrationTests\bin\Release\net10.0\BenchmarkDotNet.IntegrationTests-1/.artifacts/\obj\BenchmarkDotNet.Autogenerated\release_net10.0_win-x64\native\BenchmarkDotNet.IntegrationTests-1.sourcelink' .0]

src/BenchmarkDotNet/Toolchains/ToolchainExtensions.cs
filzrev: I've removed NativeAot60 on latest commit that cause build errors.

timcassell: I don't see `RuntimeMoniker.NativeAot60` and `NativeAotRuntime.Net60` removed.

timcassell:
Test failure:

> System.InvalidOperationException : You need to use UseNuGet or UseLocalBuild methods to tell us which ILCompiler to use.

Error seems outdated.

src/BenchmarkDotNet/Toolchains/NativeAot/Generator.cs
timcassell, line 1: Line 153, `<PublishAot Condition="$([MSBuild]::VersionGreaterThan('$(NETCoreSdkVersion)', '6.0'))">true</PublishAot>` can drop the condition.

timcassell (approved):
Thanks

PR #3225, opened 2026-08-07
chore: Remove System.Management dependency from Portability.RuntimeInformation

filzrev:

#### 1. Modify `RuntimeInformation.GetAntivirusProducts`

Remove dependency to `System.Management` and migrate code to used PowerShell/Pwsh based implementation. 

This code is called only when `EnvironmentAnalyser` raise warnings.
So  external process invocation overhead can be ignored (it takes about 500-700ms)

#### 2. Modift `RuntimeInformation.GetVirtualMachineHypervisor`
Remove dependency to `System.Management` and modify code to get values from equivalent registry settings.

#### 3. Add test on `BenchmarkDotNet.Tests/Environments/HostEnvironmentInfoTests.cs`

Add tests to verify GetAntivirusProducts lazy initialized value.

### DIFF
--- a/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
+++ b/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
@@ -2,12 +2,13 @@ using BenchmarkDotNet.Detectors;
 using BenchmarkDotNet.Environments;
 using BenchmarkDotNet.Extensions;
 using BenchmarkDotNet.Helpers;
+using Microsoft.Win32;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
-using System.Management;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
+using System.Text.Json;
 using static System.Runtime.InteropServices.RuntimeInformation;
 
 namespace BenchmarkDotNet.Portability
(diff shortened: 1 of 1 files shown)


tests/BenchmarkDotNet.Tests/Environments/HostEnvironmentInfoTests.cs
```diff
@@ -47,5 +50,24 @@ public void DoesntReturnHypervisorNameWhenItsNotDetected()
                               $"{info.Os.Value.ToBrandString()}";
             Assert.Equal(expected, line);
         }
+
+        [FactEnvSpecific(EnvRequirement.WindowsOnly)]
+        public void GetAntivirusProducts()
```
timcassell: This would be better as an integration test (tests in this project should be fast).

filzrev: I've removed this test.
Because it seems already tested on [ValidateLazyInitializedValues integration test](https://github.com/dotnet/BenchmarkDotNet/blob/7f04030181a66789cfdd46abe5b3c8be6957386d/tests/BenchmarkDotNet.IntegrationTests/Environments/HostEnvironmentTests.cs#L18)

tests/BenchmarkDotNet.Tests/Environments/HostEnvironmentInfoTests.cs
timcassell, line 1: Revert this file

timcassell (commented):
What's the gain here? It doesn't look like the dependency is removed from BDN entirely.

tests/BenchmarkDotNet.Tests/Environments/HostEnvironmentInfoTests.cs
filzrev: I've reverted test code in latest commit.

filzrev:
> What's the gain here? It doesn't look like the dependency is removed from BDN entirely.

When running NativeAot benchmarks following `System.Management` related warning are raised.
> warning IL2104: Assembly 'System.Management' produced trim warnings. For more information see https://aka.ms/il2104
> warning IL3053: Assembly 'System.Management' produced AOT analysis warnings. 

As far as I've confirmed,
Existing WMI/CIM queries works under NativeAot when `BuiltInComInteropSupport=true` is enabled.

Though, it's better to remove `System.Management` dependency completely. Because `System.Management` is not expected to support trimming.

Remaining `System.Management` dependency is `MosCpuDetector`.
And it'll be handled on another PR. (It needs to add another Win32 API dependencies)

timcassell (approved):

PR #3131, opened 2026-05-14
chore: Add unit test for CpuDetector

filzrev:
This PR add unit test for CpuDetector and output raw values to `ITestOutputHelper`.

By this unit test. It can verify CPU info mismatch issue that occurred on Windows environment.
https://github.com/dotnet/BenchmarkDotNet/issues/859#issuecomment-4414842406

PR #3230, opened 2026-08-14
chore: Modify Windows CpuDetector and remove System.Management dependency

filzrev:

#### 2. Remove `MosCpuDetector` and add `DefaultCpuDetector`
Instead of using WMI/CIM query.
new implementation get CPU information from registry and Win32 API (`GetLogicalProcessorInformationEx`)

#### 3. Add/Modify CpuDetector test codes
CPU frequency information is compared by `BeApproximately` with accepting 1％ difference.
It's required because value that taken from registry, Win32API and WMI CIM query returns slightly different value randomly.
And cause flaky test.

### Other notes
I've also tested [CPUID x86 instruction](https://en.wikipedia.org/wiki/CPUID) to get **correct** CPU frequency information.
Though it's not works as expected on virtualized environment. and when running on AMD CPUs.


timcassell:
> I've also tested [CPUID x86 instruction](https://en.wikipedia.org/wiki/CPUID) to get **correct** CPU frequency information.
> Though it's not works as expected on virtualized environment. and when running on AMD CPUs.

Is this a regression, or same behavior on master? Or am I misunderstanding the comment, and you just meant you tried that approach and didn't go with it?

tests/BenchmarkDotNet.IntegrationTests/Detectors/CpuDetectorTests.cs
```diff
@@ -24,8 +26,49 @@ public void DetectCpuInfo()
...
+        {
+            // On Windows, CPU frequency that are returned by CIM/registory value has slightly different.
+            // https://github.com/dotnet/BenchmarkDotNet/issues/859#issuecomment-4414842406
+        }
+        else if (OsDetector.IsLinux())
+        {
+            // On Linux, There is issue wrong CPU frequency is returned on some CPU.
+            // https://github.com/dotnet/BenchmarkDotNet/pull/3131#issuecomment-4455965694
+        }
+        else
```
timcassell:
```suggestion
        // On Windows, CPU frequency that are returned by CIM/registory value has slightly different.
        // https://github.com/dotnet/BenchmarkDotNet/issues/859#issuecomment-4414842406
        // On Linux, There is issue wrong CPU frequency is returned on some CPU.
        // https://github.com/dotnet/BenchmarkDotNet/pull/3131#issuecomment-4455965694
        else if (!OsDetector.IsWindows() && !OsDetector.IsLinux())
```

src/BenchmarkDotNet/Detectors/Cpu/Windows/DefaultCpuDetector.cs
```diff
@@ -0,0 +1,199 @@
...
+                    case LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorPackage:
+                        ++processorPackageCount;
+                        break;
+
+                    case LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore:
+                        ++physicalProcessorCoreCount;
+                        break;
+
+                    case LOGICAL_PROCESSOR_RELATIONSHIP.RelationGroup:
+                        ref readonly GROUP_RELATIONSHIP group = ref info->Group;
```
timcassell:
```suggestion
                        ref GROUP_RELATIONSHIP group = ref info->Group;
```

**Correctness bug on machines with >64 logical processors (≥2 processor groups).**

`group` is a `ref readonly`, and CsWin32 generates `VariableLengthInlineArray<T>` as a non-`readonly` struct whose indexer getter is also non-`readonly`:

```csharp
internal ref T this[int index]
{
    [UnscopedRef] get => ref Unsafe.Add(ref this.e0, index);   // note: not 'readonly'
}
```

Calling that indexer on a `ref readonly` receiver forces the compiler to insert a **defensive copy** of the `GroupInfo` inline array. Its only field is the first element `e0` — elements `[1..]` physically live in the trailing bytes of the OS buffer and are *not* part of the struct's size, so they aren't copied. `Unsafe.Add(ref copy.e0, i)` for `i >= 1` then reads adjacent stack memory instead of the real `PROCESSOR_GROUP_INFO`, producing a garbage `LogicalCoreCount`. No exception is thrown, so the `try/catch` never fires and the wrong value is reported silently. `[0]` is copied correctly, which is why ≤64-core machines are unaffected.

Dropping `readonly` (the buffer is genuinely writable memory) removes the defensive copy so the indexer reads straight from the buffer.

Verified against generated CsWin32 0.3.298 with a 3-group layout (64+32+16, expected 112):

| read pattern | Debug | Release |
|---|---|---|
| `ref readonly` + indexer (current) | 229 ❌ | 486 ❌ |
| `ref` + indexer (this suggestion) | 112 ✅ | 112 ✅ |
| `ref readonly` + `.AsSpan(...)` | 71 ❌ | 258 ❌ |

Note `.AsSpan(...)` is **not** a safe alternative here — it's also a non-`readonly` member, so it hits the same defensive-copy trap.

<sub>🤖 Automated review by Claude (Claude Code, Opus 4.8), verified empirically against generated CsWin32 code. Please double-check before merging.</sub>

timcassell:
## Review of #3230 — Windows CPU detector rewrite

Overall this is a nice cleanup: dropping the `System.Management` dependency in favour of the registry + `GetLogicalProcessorInformationEx` via CsWin32 is the right direction, and the csproj / `NativeMethods.txt` / PolySharp-exclude changes are all correct and consistent.

One blocking correctness bug (posted as an inline suggestion on the `RelationGroup` case) plus a few lower-priority items below.

### 🟡 Re-enabled frequency assertion may reintroduce flaky macOS CI
`tests/.../CpuDetectorTests.cs` — the `MaxFrequencyHz >= NominalFrequencyHz.Value` assertion (previously commented out because it failed on *some environment*, per [#3131](https://github.com/dotnet/BenchmarkDotNet/pull/3131#issuecomment-4455965694)) now executes in the final `else`, which is reached only on macOS. It also risks an NRE via `.Value` if `NominalFrequencyHz` is null there. Worth confirming this holds on macOS CI before relying on it.

### 🟡 Unreachable Linux branch (copy-paste)
`tests/.../CpuDetectorTests.cs`:
```csharp
if (OsDetector.IsWindows() || OsDetector.IsLinux()) { /* Windows comment */ }
else if (OsDetector.IsLinux()) { /* Linux comment */ }   // dead: Linux already matched above
else { ...assert... }
```
The first branch was presumably meant to be `if (OsDetector.IsWindows())`. Net *behavior* is still correct (the assert only runs on macOS), so this is dead/misleading code rather than a functional bug — but the Linux-specific intent is silently lost.

### 🟡 Cross-detector equality test is prone to string flakiness
`DetectCpuInfoOnWindowsAndCompareValues` uses `BeEquivalentTo`, which compares `ProcessorName` for exact string equality. `DefaultCpuDetector` builds the name from the trimmed registry `ProcessorNameString`; `PowershellWmiCpuDetector` uses WMI `Win32_Processor.Name`. Any whitespace/formatting difference (or a per-group core-count difference) will fail the test. Consider normalizing, or asserting the individual numeric fields rather than full equivalence.

### Minor / optional
- **`ParseData` reads `info->Size` / `info->Relationship` before the boundary check** — one read too early on truncated data, but the OS returns well-formed, correctly-sized buffers, so this is effectively theoretical.
- **`ExtensionMethods.ToNullableInteger` (`== 0 ? null`) duplicates the inline `> 0 ? … : null` frequency idiom** with slightly different semantics — a negative count would slip through `ToNullableInteger` as a valid value, whereas the `> 0` guard rejects it. Cosmetic/consistency.

<sub>🤖 Automated review by Claude (Claude Code, Opus 4.8). Findings verified against the diff; please sanity-check before acting.</sub>

filzrev:
> Is this a regression, or same behavior on master?
> Or am I misunderstanding the comment, and you just meant you tried that approach and didn't go with it?

I've tried that approach and didn't go with it.
And it's commented for future reference..

---
Getting CPU frequency from Registry/Win32 API/CIM query has value drift.
So I've tried  to get `Nominal` frequency from CPUID x86 instruction.

Though It turned out to be unreliable for the intended purpose.
(It's expected to works only non-virtualized and intel CPU machine)

src/BenchmarkDotNet/Detectors/Cpu/Windows/DefaultCpuDetector.cs
filzrev: I've modified code that causing struct defensive copy.
and remove struct temporary variable.

filzrev:
#### 1. Re-enabled frequency assertion may reintroduce flaky macOS CI

Currently CPU frequency issue is not observed error on macos.
So I've re-enable this test for macos.


#### 2. Unreachable Linux branch (copy-paste)

Code is fixed based on review comment.

#### 3. Cross-detector equality test is prone to string flakiness

The CPU name value that are retrieved from registry contains extra trailing spaces.
So I've trimmed ProcessorNameString value.
Name normalization will be added when there are CPUs that cause test failurer.

timcassell (approved):

Issue #2344, opened 2023-06-27
Build fails when returning type from PackageReference


timcassell:
I added a nuget package.

```xml
<PackageReference Include="ProtoPromise" Version="2.5.4" />
```

Created a benchmark to return a type from that package.

```cs
[Benchmark] public CancelationToken CanceledToken() => CancelationToken.Canceled();
```

And the benchmark project build failed.

> C:\Users\Tim\Documents\git\ProtoPromiseBenchmarks\bin\Release\net7.0\a07bed7c-cb6d-4d10-b849-74756c9d53ac\a07bed7c-cb6d-4d10-b849-74756c9d53ac.notcs(342,55): error CS0246: The type or namespace name 'Proto' could not be found (are you missing a using directive or an assembly reference?) [C:\Users\Tim\Documents\git\ProtoPromiseBenchmarks\bin\Release\net7.0\a07bed7c-cb6d-4d10-b849-74756c9d53ac\BenchmarkDotNet.Autogenerated.csproj]

Is there any way to get it to work?

[Edit] Note: it works when I use the types in my benchmarks, but when I return a type, it breaks.

adamsitnik:
Is the generated code using a fully qualified type name (#1009)?

timcassell:
> Is the generated code using a fully qualified type name (#1009) when referring to `CancelationToken`?

Yes, I checked the .notcs file, it's `Proto.Promises.CancelationToken`.

AndreyAkinshin:
This bug is specific to `ProtoPromise`. Let's create a new project with the following package references:
```xml
    <ItemGroup>
        <PackageReference Include="Newtonsoft.Json" Version="13.0.3"/>
        <PackageReference Include="ProtoPromise" Version="2.5.4"/>
    </ItemGroup>
```
Here is the relevant fragment of `obj/project.assets.json`:
```json
      "Newtonsoft.Json/13.0.3": {
        "type": "package",
        "compile": {
          "lib/net6.0/Newtonsoft.Json.dll": {
            "related": ".xml"
          }
        },
        "runtime": {
          "lib/net6.0/Newtonsoft.Json.dll": {
            "related": ".xml"
          }
        }
      },
      "ProtoPromise/2.5.4": {
        "type": "package",
        "build": {
          "build/net6.0/ProtoPromise.targets": {}
        }
      }
```

[...]

Issue #1197, opened 2019-07-11
Enable including references in autogenerated csproject


billwert:
Today, if a benchmark needs a reference to a Nuget package it goes into the original project (microbenchmarks.csproj, say) and then does not need to be in the auto generated csproj.

I’d like to enable us to remove the requirement that we run .NET Framework tests on VMs / machines that have Visual Studio installed: https://github.com/dotnet/performance/pull/627. The new Microsoft.NetFramework.ReferenceAssemblies package is precisely the answer.

But it fails because the generated project doesn’t include the package reference I added:

```
[2019/07/11 19:29:39][INFO] $ dotnet run --project C:\dotnetbuild\work\274ec05f-7f6a-45af-b6aa-aa3f40fa2e5d\Payload\src\benchmarks\micro\MicroBenchmarks.csproj --configuration Release --framework net461 --no-restore --no-build -- --anyCategories coreclr corefx --iterationCount 1 --warmupCount 0 --invocationCount 1 --unrollFactor 1 --strategy ColdStart --stopOnFirstError true --partition-count 5 --partition-index 0 --packages C:\dotnetbuild\work\274ec05f-7f6a-45af-b6aa-aa3f40fa2e5d\Payload\artifacts\packages --runtimes net461
[2019/07/11 19:29:42][INFO] // Validating benchmarks:
[2019/07/11 19:29:48][INFO] // ***** BenchmarkRunner: Start   *****
[2019/07/11 19:29:48][INFO] // ***** Found 437 benchmark(s) in total *****
[2019/07/11 19:29:48][INFO] // ***** Building 1 exe(s) in Parallel: Start   *****
[2019/07/11 19:29:50][INFO] // start dotnet restore --packages "C:\dotnetbuild\work\274ec05f-7f6a-45af-b6aa-aa3f40fa2e5d\Payload\artifacts\packages"  /p:UseSharedCompilation=false /p:BuildInParallel=false /m:1 in C:\dotnetbuild\work\274ec05f-7f6a-45af-b6aa-aa3f40fa2e5d\Payload\artifacts\bin\MicroBenchmarks\Release\net461\2397be0f-2eba-4afc-b8e3-34ec43808e6d
[2019/07/11 19:29:52][INFO] // command took 2.7s and exited with 0
[...]

PR #3235, opened 2026-08-18
chore: Modify csproj-based project build logics

### LINKED ISSUE #3191: DoubleWrites message is recorded when using `--output` parameter on dotnet command

filzrev:
This PR intended to fix #3191.

#### 1. Cherry picks other PR commits
Following PR contents are included
- #3194 
- #3195 

#### 2. Remove DllGatherer related logics

After this PR changes, DllGatherer logics seems not be needed.
So I've removed DllGatherer related logics.

#### 3. Add settings for Autogenerated project.
Modify `BenchmarkDotNet.Build.props.txt` template.
And enable following settings for Autogenerated project.
- `OutputPath`
- `PublishDir`

#### 4. Cleanup `DotNetCliCommand.cs` logics.
Previously, `--output` and `-p:PublishDir` command line parameters are used.
These settings are handled as MSBuild Global property and that causing DoubleWrite Issues.

So this PR move these settings to Autogenerated.csproj level.
other projects build outputs are placed under `.artifacts` directory.

#### 4. Other Changes
1. Remove `ErrorOnDuplicatePublishOutputFiles` settings.
2. Add `XmlDocument.SaveAsync` helper method for netstandard2.0

### Test
Setting `ForcedNoDependenciesForIntegrationTests: false` version of CI. (InProgress)
- https://github.com/filzrev/BenchmarkDotNet/actions/runs/32147625697


src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs
```diff
@@ -124,91 +146,6 @@ private async ValueTask<string> GenerateBuildProject(BuildPartition buildPartiti
                 .ToString();
         }
 
-        private static string GetDllGathererPath(string filePath)
-            => Path.Combine(Path.GetDirectoryName(filePath)!, $"DllGatherer{Path.GetExtension(filePath)}");
-
-        protected async ValueTask GatherReferencesAsync(BuildPartition buildPartition, ArtifactsPaths artifactsPaths, ILogger logger, CancellationToken cancellationToken)
```
timcassell, line 130: What's the replacement for this?

filzrev: On this PR, following `DllGatherer` steps are removed.
1. Create `DllGatherer.csproj` that reference original benchmark project
2. Build DllGatherer project
3. Collect generated DLLs from **bin** directory
4. Add DLL `<Reference>` nodes to `Autogenerated` project.

By this PR changes, Csproj-based Toolchain build **Antogenerated** project directly .
And currently no behavior problem is observed.

1. `NativeAssets` are successfully copied to bin directory.
2. Satellite assemblies are successfully copied to bin directory.   
    (That commended  on https://github.com/dotnet/BenchmarkDotNet/pull/3171#issuecomment-4646214743)
3. DoubleWrite issues are resolved.

---

If it need to add steps that modify **Autogenerated** project in future. (e.g. DLL aliaes support)
It should be implemented by using custom MSBuild task that collect `ReferenceCopyLocalPaths` output without building actual project build. 

**Example of custom MSBuild task usage I've tested before**
https://github.com/filzrev/BenchmarkDotNet/commit/9821463375e68ca5812120ed816f707b703936bc

timcassell: This is a regression from master. Dll references are not transitive, which is why this exists. None of our tests exercise that case.

filzrev: >  Dll references are not transitive, which is why this exists.

I'd like to test how it works, Is it able to provide a specific example that  reproduce problem?
(When benchmark project using DLL `Reference` case?)

timcassell: See #1197 and #2344. The code gen has changed since then, so it might only repro if a type from the reference is used in a param/argument or returned from an async benchmark (sync benchmarks drop the returned value without caring what the return type is).

filzrev: I've reverted DllGatherer steps and modify related build logics.
By this changes, build steps works as following.

1. Build `DllGatherer` project **with** dependencies
2. Gather Dlls from **bin** directory that generated under `.artifacts` directory.
3. Add gathered DLL references to `Autogenerated` project
4. Build `Autogenerated` project **without** dependencies.

DllGather's satellite assemblies copy issue is expected be handled on another PR.
(Because It need to exclude NativeAssets)

**Test CI that disable `ForcedNoDependenciesForIntegrationTests`**
https://github.com/filzrev/BenchmarkDotNet/actions/runs/32337919575

filzrev:
I've re-enable `<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles>` setting for WASM project.

Because it cause following errors for `xunit.runner.visualstudio` package related DLLs.
> NETSDK1152: Found multiple publish output files with the same relative path

It seems occurred because `xunit.runner.visualstudio` package reference has `<PrivateAssets>all</PrivateAssets>` setting.

filzrev:
Wasm related CI tests are still failed when using `ForcedNoDependenciesForIntegrationTests: false` setting.

timcassell:
It seems that xUnit v2 is incompatible with wasm. It's probably fixed by #3232.

filzrev:
> Wasm related CI tests are still failed when using ForcedNoDependenciesForIntegrationTests: false setting.

This issue is caused commit 7f18102 is not pushed to target CI branch.

And I've added following commits to avoid issue when using `ForcedNoDependenciesForIntegrationTests=false`
- 75973bf (Increase timeout from 1 min to 5 min to avoid timeout)
- ~~20fc0fe (Increase job timeout to 720 minutes to avoid job timeout(Default: 360 minutes))~~  
  (This setting is not works on GitHub Actions hosted runner. and timeout at 360 min)

filzrev:
By changing some settings.
It pass CI with `ForcedNoDependenciesForIntegrationTests=false` setting.

CI: https://github.com/filzrev/BenchmarkDotNet/actions/runs/32571372114/workflow 
(`windows-latest` CI is cancelled because it execution time overs 360 minutes (GitHub Actions hosted-runner limitation))

timcassell:
@filzrev I put up a PR against this branch. https://github.com/filzrev/BenchmarkDotNet/pull/16

filzrev:
@timcassell 
Please feel free to use the content of this PR.
There is no need to keep a commit histories.

timcassell (approved):

PR #3237, opened 2026-08-22
feat: Add filter by UID feature by using existing --filter syntax

filzrev:
This PR extend existing `--filter` syntax to support filter by UID of BenchmarkCase.
By modyfing existing filter logics to use `UidFilter` when filter text can be parsed as GUID format.

#### Other changes
- Move filter related test under `Filters` directory/namespace.

### DIFF
--- a/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
+++ b/src/BenchmarkDotNet/ConsoleArguments/ConfigParser.cs
@@ -792,11 +792,24 @@ namespace BenchmarkDotNet.ConsoleArguments
         private static IEnumerable<IFilter> GetFilters(CommandLineOptions options)
         {
             if (options.Filters.Any())
-                yield return new GlobFilter(options.Filters.ToArray());
+            {
+                // Filter by UID when specified filter can be parsed as GUID.
+                var uids = options.Filters.Where(x => Guid.TryParse(x, out _)).ToArray();
+                if (uids.Length > 0)
+                    yield return new UidFilter(uids);
+
+                // Use GlobFilter for remaining filters.
+                var globFilters = options.Filters.Except(uids).ToArray();
+                if (globFilters.Length > 0)
+                    yield return new GlobFilter(globFilters);
+            }
+
             if (options.AllCategories.Any())
                 yield return new AllCategoriesFilter(options.AllCategories.ToArray());
+
             if (options.AnyCategories.Any())
                 yield return new AnyCategoriesFilter(options.AnyCategories.ToArray());
+
             if (options.AttributeNames.Any())
                 yield return new AttributesFilter(options.AttributeNames.ToArray());
         }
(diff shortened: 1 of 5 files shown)


adamsitnik (commented):
@filzrev I am trying to understand when filtering by guid is going to provide value. 

Let's say that I have the `BenchmarkDotNet.Tests.Filters.TypeWithBenchmarksAndParams.TheBenchmark` benchmark and I want to run it.

If I use `--filter BenchmarkDotNet.Tests.Filters.TypeWithBenchmarksAndParams.TheBenchmark` it just works.

When I use UID I need to:

- write some code that is going to generate the unique ID using `GetUniqueId` (not mentioned in samples or XML docs)
- somehow store this value
- provide the value when filtering.

Now as soon as the namespace, type name, method name or any other part of the benchmark ID changes, the UID is no longer valid. So what is the advantage of using it over the typical `--filter`?

filzrev:
It's intended to be used for debugging purpose. (`launchSetting.json` or CI input parameter)

Existing **filter** syntax match `filterName` or `fullName`.
So it can't distinguish BenchmarkCases that have different job.

**Example**

```csharp
[InProcess]
[Config(typeof(DebugInProcessConfig))]
public class Benchmarks
{
    [Benchmark]
    public void Benchmark01() {}
}
```

And when filtering benchmarkCase with `fullName`.
It needs double-quote for filter value (When Params exists empty space is included on fullName)

To run **specific** benchmark for debugging purpose,
It's easier to specify BenchmarkCase by uid that are retrieved by `--list json` command.

adamsitnik:
> It's intended to be used for debugging purpose. (`launchSetting.json` or CI input parameter)
> 
> Existing **filter** syntax match `filterName` or `fullName`. So it can't distinguish BenchmarkCases that have different job.

Thanks. In such case I would expect the user to modify the config and just comment out the Job they don't want to use.

filzrev:
> Thanks. In such case I would expect the user to modify the config and just comment out the Job they don't want to use.

I want to avoid modify/rebuild benchmark if possible.
A more specific example would be testing across multiple .NET versions, as shown below:

I want to run specific benchmark case(Job) by editing `launchSettings.json`.

```
[SimpleJob(RuntimeMoniker.Net80)]
[SimpleJob(RuntimeMoniker.Net90)]
[SimpleJob(RuntimeMoniker.Net10_0)]
public class MyBenchmark
{
    [Benchmark]
    public void Test()
    {
        // benchmark
    }
}
```

If reusing the `--filter` option is not desirable,
I want other options to execute **specific** benchmark case.

adamsitnik:
> I want to avoid modify/rebuild benchmark if possible.

This is a great goal and that is why we should encourage users to specify all the jobs arguments via command line arguments.

My personal approach is to always have a config with the default settings I usually want (like 1 warmup iteration), with a single job marked `AsDefault` and all the exporters and then everything else (diagnosers, filters, tfms) specified via command like arg.

But in this particular example we talk about Debugging, which requires and IDE (at least for most of us). So I would expect the users to simply go and modify the code by commenting out the job they don't like just a moment before they hit F5.

PR #3238, opened 2026-08-22
chore: Modify powercfg output parse logics

filzrev:
This PR intended to fix `BenchmarkRunnerAcquiresWakeLock` tests are failed with timeout error when running on non-english Windows environment.

`powercfg /requests` command output is localized based on UI language setting of Windows.
So it need to handle localized message.

So I've added `IsNoneToken` helper method to check `None.` equivalent string.

---
The added language mapping are generated by LLM.
And it's confirmed `powercfg /requests` output message data source per language.

There are some languages that can't find reliable message sources.
So test still failed on these language environment.

### DIFF
--- a/tests/BenchmarkDotNet.IntegrationTests/WakeLockTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/WakeLockTests.cs
@@ -99,8 +99,14 @@ public class WakeLockTests : BenchmarkTestExecutor
         async Task WaitForBenchmarkRunningAndGetPowerRequests()
         {
             await AsTask(ping, testTimeout);
-            pwrRequests = GetPowerRequests("BenchmarkDotNet Running Benchmarks");
-            pong.Set();
+            try
+            {
+                pwrRequests = GetPowerRequests("BenchmarkDotNet Running Benchmarks");
+            }
+            finally
+            {
+                pong.Set();
+            }
         }
     }
 
--- a/tests/BenchmarkDotNet.IntegrationTests/WakeLockTests.cs
+++ b/tests/BenchmarkDotNet.IntegrationTests/WakeLockTests.cs
@@ -98,9 +98,9 @@ public class WakeLockTests : BenchmarkTestExecutor
 
         async Task WaitForBenchmarkRunningAndGetPowerRequests()
         {
-            await AsTask(ping, testTimeout);
             try
             {
+                await AsTask(ping, testTimeout);
                 pwrRequests = GetPowerRequests("BenchmarkDotNet Running Benchmarks");
             }
             finally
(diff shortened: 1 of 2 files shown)


timcassell:
Good diagnosis, but the language table isn't needed — that line is positionally determined. After a `RequestType` token the next line can only be a `[...]` requester or the "none" marker; a `Reason` only ever follows a `RequesterName`. `ParseRequestType` already relies on this (it unconditionally `Take`s `None` there), so matching the word is a redundant extra constraint that happens to be the locale-dependent one.

Drop `IsNoneToken` and track the previous token in `Tokens()` instead:

```csharp
else if (line[0] == '[')
{
    int pos = line.IndexOf(']');
    yield return new Token(TokenType.RequesterType, line.Substring(1, pos - 1));
    previous = TokenType.RequesterName;
    yield return new Token(TokenType.RequesterName, line.Substring(pos + 2));
}
else if (previous == TokenType.RequestType)
{
    // Any single line directly after a request type header is the localized "None."
    previous = TokenType.None;
    yield return new Token(TokenType.None, line);
}
```

I ran both versions over the same input:

```
English:              PR #3238  OK    positional  OK
German (in table):    PR #3238  OK    positional  OK
Hungarian (absent):   PR #3238  FAIL  positional  OK
Czech (absent):       PR #3238  FAIL  positional  OK
```

Identical where this PR works, plus every locale it doesn't — so the caveat in the description goes away. It's also a smaller diff, and avoids a table that CI can't verify (English-only) and that has real ambiguity in it (`Ninguna.` vs `Ninguno.`).

Two adjacent things, either PR or follow-up:

- `pong.Set()` in `WakeLockTests.WaitForBenchmarkRunningAndGetPowerRequests` should be in a `finally`. A parse failure currently costs a full `testTimeout` and reports `TimeoutException` instead of the real `InvalidCastException` — that's why this presented as a timeout.
- `PowerRequestsParser` has no unit tests. A couple of canned-string cases would run everywhere, whereas `WakeLockTests` needs Windows *and* elevation.

*Reviewed by Claude (Opus 5), posted by @timcassell.*

filzrev:
I've modified skip logics based on review comment. 
And confirmed `WakeLockTests` tests are passed with elevated privilege.

tests/BenchmarkDotNet.IntegrationTests/WakeLockTests.cs
```diff
@@ -99,8 +99,14 @@ public async Task BenchmarkRunnerAcquiresWakeLock(Type type, string expected)
         async Task WaitForBenchmarkRunningAndGetPowerRequests()
         {
             await AsTask(ping, testTimeout);
```
timcassell: `await AsTask(ping, testTimeout)` sits outside the new `try`, so `finally { pong.Set(); }` doesn't cover it. If the out-of-process build/start exceeds the 1-minute `testTimeout` (plausible on a cold machine), `AsTask` faults, `pong` is never set, and the child blocks another full minute in `pong.WaitOne(testTimeout)` — the same hang symptom this PR fixes, via the other arm. Fix: wrap the whole method body including the `await`. `pong` is an `AutoResetEvent`, so signalling early is harmless.

filzrev: I've modified try-finally block code on latest commit.

timcassell (approved):
Thanks

PR #3239, opened 2026-08-22
Add job categories

sheddy123:
Issue #3182
Follow-up to the discussion in #3216, and a step towards #3182.
## What this adds
This PR introduces job categories to provide a way to group and filter jobs, similar to [BenchmarkCategory].

- Added MetaMode.Categories as hidden job metadata, so categories can be used for selection without affecting job IDs, folder names, summaries, or generated source.
- Added WithCategory and WithCategories for assigning categories in code. Categories are case-insensitive and automatically de-duplicated.
- Added [JobCategory] for applying categories to all jobs at the class or assembly level.
- Ensured categories from attributes and code are merged rather than overwritten.
- Added JobCategoryFilter to select jobs based on their assigned categories.

cc @filzrev @adamsitnik @timcassell


docs/articles/configs/jobs.md
```diff
@@ -15,6 +15,36 @@ There are several categories of characteristics which you can specify. Let's con
...
+    .AddJob(Job.Default.WithRuntime(CoreRuntime.Core90).WithCategories("runtimes", "net9"))
+    .AddJob(Job.Dry.WithCategory("debug"));
+```
+
+`WithCategories` overrides the categories of the job, `WithCategory` adds to them.
+
+You can also use the `[JobCategory]` attribute, which adds its categories to every job defined for the given class or assembly:
+
+```cs
+[JobCategory("runtimes")]
```
timcassell: I don't think we should do this. Rather, add a categories property to `SimpleJob`, etc.

sheddy123:
@timcassell 
One thing worth knowing since it isn't obvious from the diff: the old attribute's "adds to categories already defined in code" behavior is gone along with it. Categories now come from exactly one place per job whichever attribute or fluent call defined that job.

sheddy123:
Hi @timcassell, please any updates on this PR?

timcassell:
Thanks for the PR! I went through the diff (merge-base `b6d2177` .. `1fc3f7f`), built the branch, and ran the unit suite — 1082 passed / 0 failed, including all 25 new `JobCategoryTests`. The findings below were reproduced against that build.

### Categories affect job identity but not the job name

`src/BenchmarkDotNet/Jobs/MetaMode.cs:20`

`CategoriesCharacteristic` is created with `CreateHiddenCharacteristic`, which only hides it from the presenters (`IsPresentableCharacteristic`). `JobComparer.Compare` walks `GetAllCharacteristics()`, so it *does* compare categories.

The result is that two jobs differing only by category survive `Distinct(JobComparer.Default)` but produce byte-identical benchmark names:

```
ManualConfig.CreateEmpty()
    .AddJob(Job.Dry.WithCategory("a"))
    .AddJob(Job.Dry.WithCategory("b"))

// two cases, both named:
// B.M: Dry(IterationCount=1, LaunchCount=1, RunStrategy=ColdStart, UnrollFactor=1)
```

Without a `JobCategoryFilter` in play, the user silently runs the same job twice, gets two indistinguishable summary rows, and any artifact keyed on the benchmark name (EventPipe traces, per-benchmark exports) collides.

`JobsWhichDifferOnlyByCategoriesAreNotConsideredDuplicates` currently locks this behavior in. Two ways out: exclude the characteristic from `JobComparer`, or merge categories when deduplicating.

### `WithCategories()` with an empty array silently doubles the run

`src/BenchmarkDotNet/Jobs/JobExtensions.cs:343`

The assignment is unconditional, so an empty array still *sets* the characteristic:

```csharp
Job.Default.WithCategories().Categories.Count           // 0
Job.Default.WithCategories().HasValue(CategoriesCharacteristic)  // true
```

So `AddJob(Job.Default).AddJob(Job.Default.WithCategories())` yields two `DefaultJob`s and runs the benchmark twice. That makes `job.WithCategories(userSelection.ToArray())` a hazard whenever the selection happens to be empty. Suggest skipping the assignment (or clearing the characteristic) when the deduplicated list is empty.

Same line: `WithCategories(null)` is legal via `params`, and currently surfaces as `ArgumentNullException: ... (Parameter 'source')` thrown from `Distinct` inside `MetaMode.Unique`. A guard naming `categories` would be friendlier.

### `NullReferenceException` on `Categories = null`

`src/BenchmarkDotNet/Attributes/Jobs/JobConfigbaseAttribute.cs:32`

`Categories.Length` is dereferenced without a null check. `[DryJob(Categories = null)]` is valid C# for a named attribute argument of array type, and throws a bare `NullReferenceException` out of `BenchmarkConverter.TypeToBenchmarks` during discovery. `Categories is not { Length: > 0 }` handles both cases.

### Memoized `Config` silently drops later `Categories` assignments

`src/BenchmarkDotNet/Attributes/Jobs/JobConfigbaseAttribute.cs:30`

`Config` memoizes into `config` while `Categories` remains publicly settable, so any assignment after the first `Config` read is dropped (read `Config`, set `Categories = ["late"]`, re-read → 0 categories on the same instance). Relatedly, `job.WithCategories(Categories)` *replaces*, so a user-derived attribute whose base job already carries categories loses them when a consumer sets `Categories`. Making `Categories` init-only, or using append (`WithCategory`) semantics, would avoid both.

### Unrelated whitespace churn

`src/BenchmarkDotNet/Configs/ImmutableConfigBuilder.cs:242`

After 92f6955b, the only change left in this file is two deleted blank lines around `copy.Apply(mutatorJob)`. Worth reverting so the PR touches nothing it doesn't need.

(The underlying behavior change here is fine — `Apply` skips characteristics the mutator has no value for, and `[DryJob(Categories = ["a"])]` + `[GcServer(true)]` correctly yields `gcServer=True cats=[a]`, so removing the explicit preservation code was safe.)

### Checked and correct

Categories are properly excluded from `JobIdGenerator`, `FolderPresenter`, `SourceCodePresenter` (nothing leaks to the child process) and from `JobCharacteristicColumn.AllColumns`, so the "no effect on id/folder/summary" intent holds. `WithCore` copies rather than mutating the shared static jobs; `AddCategories`/`Unique` dedupe case-insensitively keeping the first spelling; `JobCategoryFilter`'s empty-array behavior matches the existing `AnyCategoriesFilter`; and collection expressions in attribute arguments (`Categories = ["a", "b"]`) do compile.

---

*This review was generated by Claude (Opus 5) via Claude Code and posted by @timcassell. The findings were reproduced against a local build, but please treat the analysis as a starting point rather than a verdict.*

sheddy123:
Hi @timcassell the finds you outlined have now been fixed. 

Addressed the findings around job categories, deduplication, null handling, and configuration mutability.

* **Job identity & category merging:** `MetaMode.Categories` was incorrectly affecting job identity because hidden characteristics are still considered by `JobComparer`. The comparer now excludes the categories characteristic, while `GetRunnableJobs` groups equivalent jobs and merges their categories. For example, jobs with categories `"a"` and `"b"` now collapse into a single job containing both categories instead of silently losing one.
* **Empty categories:** `WithCategories()` no longer creates a separate job. Empty categories are stored as `null`, allowing deduplication to work correctly while preserving the documented override behaviour.
* **Null handling:** `Categories` now safely handles `null`. The nullable-array definition also prevents the previous `NullReferenceException` scenario. The test uses `null!` to reproduce the same IL where nullable reference types are disabled.
* **Configuration mutability:** `Categories` is now `init`-only, preventing late assignments from being silently ignored by the memoized configuration.
* **Argument validation:** `WithCategories(null)` now throws `ArgumentNullException` with `categories` as the parameter name.
* **Cleanup:** Reverted unrelated whitespace changes. The only remaining `ImmutableConfigBuilder` change is the intentional grouping and category-merging logic.

timcassell:
Re-reviewed at `c5478c88`. All five points from the earlier comment are addressed — four verified fixed by execution, one partly. Build is clean across all TFMs (netstandard2.0 picks up the repo's `ArgumentNullException.ThrowIfNull` polyfill), unit suite 1087 passed / 4 skipped / 0 failed, `JobCategoryTests` 30 passed. The explanatory comments you added at each fix are genuinely good.

| | verified |
|---|---|
| jobs differing only by category | 1 job, `categories=[a,b]`; filters match `a`, `b`, `A` (case-insensitive), not `zzz` |
| equality/hash contract | same `Id`, `GetHashCode` equal, `Equals`/`Compare` consistent — also with an explicit `WithId` |
| empty categories | `HasValue == False`, `ResolvedId` unchanged, two jobs collapse to one |
| `WithCategories(null)` | throws naming `categories` |
| `[DryJob(Categories = null)]` | discovery succeeds, no NRE; `init` doesn't block a derived attribute; `Config` still cached |

Taking both routes for the identity problem — skipping the characteristic in `JobComparer` *and* merging in `ImmutableConfigBuilder` — was the right call; either alone would have left a hole. Worth noting why the pairing is safe: `JobComparer.GetHashCode` is `obj.Id.GetHashCode()`, and categories are excluded from `JobIdGenerator`, so collapsed jobs share an Id and land in the same `GroupBy` bucket. Had categories leaked into the Id, the merge would have silently never fired. Whitespace churn is gone (`git diff -w` is byte-identical to the plain diff). `MergeCategories`'s short-circuit can't be wrong, only conservatively false — a needless copy when categories overlap.

Four things left, none blocking:

**1. A mutator carrying categories wipes the target's categories.** `AddJob(Dry.WithCategory("keep"))` + `AddJob(Default.WithGcServer(true).WithCategory("mut").AsMutator())` → `Dry[mut]`; `"keep"` is gone, because `copy.Apply(mutatorJob)` treats the characteristic as an ordinary overwriting one. That's the same "selecting by a category matches nothing" failure `MergeCategories` was written to prevent, reached via the mutator path. Pre-existing to this round rather than introduced by it, and it only bites users who put categories on a mutator. Marking the characteristic `IgnoreOnApply` (as `IsMutatorCharacteristic` is) and merging explicitly would close it.

**2. `Meta.Categories = null` still throws `ArgumentNullException("source")`** — the unhelpful message the new guard was added to remove. The guard is only in the `WithCategories` extension; the property it delegates to is public. `AddCategories(null)` has the same shape.

**3. `WithCategory(null)` silently stores a null category** — `items=[<null>]`, and it survives merging. The array is guarded, the elements are not; a null in the list is a latent NRE for anything that formats or compares categories.

**4. `JobCategoryFilter` has no attribute and no CLI option.** The predicate itself is consistent with the rest of the codebase — every filter is inclusion-only, and this one is a near-verbatim `AnyCategoriesFilter` with `Descriptor` swapped for `Job.Meta`. What is inconsistent is the surface: `AnyCategoriesFilter` has both `AnyCategoriesFilterAttribute` and `--anyCategories`, `AllCategoriesFilter` likewise, but `JobCategoryFilter` has neither, so the only way to use it is constructing it by hand in a config. Given the headline use case is "run only the net8 job", the missing CLI option is the notable one, and it is much cheaper to add before the API ships.

One related design question worth settling explicitly, since the docs are the place users will look: applying `JobCategoryFilter` drops every case whose job has **no** categories. That is fine in the documented example where all three jobs are categorized, but in a config mixing categorized and uncategorized jobs, selecting one category silently removes the uncategorized baseline. Either document that, or adopt the convention that a job with no categories always passes.

*Reviewed by Claude (Opus 5), posted by @timcassell.*

sheddy123:
@timcassell this is up for approval. The issues have been addressed

timcassell:
**`ImmutableConfigBuilder.MergeCategories` strips `IsMutator`, turning a mutator into an extra standalone job.**

```csharp
return merged.Length == first.Meta.Categories.Count ? first : first.WithCategories(merged).Freeze();
```

`WithCategories` goes through `JobExtensions.WithCore` → `new Job(job.Id, job)` → `Apply`, and `MetaMode.IsMutatorCharacteristic` is created with `CreateIgnoreOnApplyCharacteristic`, so `ApplyCore` never copies it. Any group of ≥2 comparer-equal jobs whose summed category count differs from the first job's count takes the merge branch — which includes two *identical* categorized jobs, not just jobs that differ by category. If they're mutators, the survivor comes back with `IsMutator == false`, so `unique.Where(job => !job.Meta.IsMutator && !job.Meta.IsDefault)` adds it to `result` as a normal job and the mutator loop never applies it.

Repro against this branch:

```csharp
ManualConfig.CreateEmpty()
    .AddJob(Job.Dry.WithId("base"))
    .AddJob(Job.Default.WithWarmupCount(2).WithCategory("m").AsMutator())
    .AddJob(Job.Default.WithWarmupCount(2).WithCategory("m").AsMutator());
```

→ 2 jobs: `id=Job-PGIYFE warmup=2 isMutator=False cats=[m]` and `id=base warmup=0 cats=[]`. `base` was never mutated, and a spurious second summary row appears. The same config without the categories correctly yields 1 job (`id=base warmup=2`). `IsDefault` and `Baseline` survive the merge; only `IsMutator` is lost.

Fix: re-apply the flag, e.g. `var merged = first.WithCategories(...); if (first.Meta.IsMutator) merged = merged.AsMutator();` — or skip the merge entirely for mutator groups.

---

**`docs/articles/configs/jobs.md:31`** — "Every job attribute has a `Categories` property" isn't accurate. `Categories` was added to `JobConfigBaseAttribute` only; the 24 mutator attributes (`[WarmupCount]`, `[IterationCount]`, `[RunOncePerIteration]`, `[MemoryRandomization]`, `[Outliers]`, …) derive from `JobMutatorConfigBaseAttribute`, which is a separate `Attribute`/`IConfigSource` and got no `Categories`. Someone following the doc and writing `[WarmupCount(3, Categories = ["ci"])]` gets a compile error.

**`ImmutableConfigBuilder.cs:247`** — the merge comment cites "a mutator that carries categories (eg `[SimpleJob(Categories = ...)]` on a method)", but `JobConfigBaseAttribute` is `[AttributeUsage(AttributeTargets.Class | AttributeTargets.Assembly)]`, so `[SimpleJob]` can't go on a method, and job attributes never produce mutator jobs. The only way to reach that branch is the fluent `.WithCategory(...).AsMutator()` — which is the path that hits the bug above. Worth correcting so the next reader doesn't assume the branch is attribute-driven and covered.

<sub>Reviewed with Claude Code.</sub>

sheddy123:
@timcassell 
Updated `MergeCategories` to correctly preserve `IsMutator` and deduplicate merged categories before the count check. Added two regression tests covering the original repro and a differing-category scenario.

Also updated the documentation and code comments to clarify that categorized mutators must be created via the fluent API using `.WithCategory(...).AsMutator()`, as mutator attributes do not support `Categories`.

timcassell:
`Categories` in `JobConfigBaseAttribute` is `{ get; init; }`, which makes the new attribute API uncompilable for consumers on C# 7.3 — the default `LangVersion` for `net472`/`netstandard2.0`. `[DryJob(Categories = new[] { "a" })]` from such a project fails with `CS8370: Feature 'init-only setters' is not available in C# 7.3`. The rationale for `init` (a later assignment being silently ignored) no longer applies now that `Config` is built lazily — all named arguments are assigned before anything reads `Config`, so plain `{ get; set; }` behaves identically and compiles on every language version.

`MetaMode.Unique(IEnumerable<string> categories)` does the validation, so every throw reports `ParamName = "categories"` regardless of caller. `WithCategory(null)` (parameter `category`) points the user at a parameter that doesn't exist on the method they called; `Meta.Categories = null` reports `"categories"` instead of `"value"`. Validate in the callers, or rethrow with the right name. `ANullCategoryIsRejected` currently asserts the misleading name.

`--jobCategories` has OR semantics (`JobCategoryFilter.Predicate` uses `Any`), but the name doesn't say so while the siblings do (`--anyCategories`/`--allCategories`). `--jobCategories a b` reads as AND. This is public CLI surface that can't be renamed later — `--anyJobCategories` matches the convention and leaves room for an `--all` counterpart.

`JobCategoryFilter` is missing from the filter table in `docs/articles/configs/filters.md`, which lists every other filter with its console argument.

Reviewed with Claude Code.

timcassell (approved):
Thanks @sheddy123

PR #3243, opened 2026-09-02
chore: Drop net9.0 from target frameworks

filzrev:
This PR remove `net9.0` from TargetFrameworks.
By dropping `net9.0` tfm, **build solution** time is reduced.

Currently `net9.0` tfm is only used on preview/nightly versions.
So dropping support for net9.0 will have little impact.

### What's affected when `net9.0` tfm is dropped.
When running benchmarks on .NET 9.0. `net8.0` version of BenchmarkDotnet Dlls are referenced.
So it might affects benchmark results slightly because ` `allows ref struct` support is dropped.

---
Close this PR if this PR changes are not desirable.

### DIFF
--- a/src/BenchmarkDotNet/BenchmarkDotNet.csproj
+++ b/src/BenchmarkDotNet/BenchmarkDotNet.csproj
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>BenchmarkDotNet</AssemblyTitle>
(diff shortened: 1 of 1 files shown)


timcassell:
Unless we're going to drop `netstandard2.0` at some point, we should continue to support tfms that are newer than it which have features that users may be interested in, even if they are officially out of support. Slightly longer build times for us is a small price to pay. Also, `net9.0` is technically still supported for 2 more months (I'm hoping to release v0.16 before then). Thanks, though.

PR #3241, opened 2026-08-25
chore: Modify ClrMD related code to use interface

filzrev:
This PR modify ClrMD related code to use interface instead of sealed classes (`ClrRuntime`/`ClrMethod`)

These changes are required to write unit tests for `Arm64Disassembler::Decode`.

PR #3240, opened 2026-08-25
chore: Modify arm64 instruction formatter and cleanup code for arm64 disassembler

filzrev:
This  PR contains following changes.

### `Arm64InstructionFormatter.cs`
Modify existing `PadLeft` logics to insert spaces at least 1.

On existing implementation, When specifying `FirstOperandCharIndex` value is lower than Mnemonic length.
First operand is outputted without space between mnemonic. 

### `Arm64Disassembler.cs`
Move `RegisterValueAccumulator` struct to separated file.
And rename to `Arm64RegisterValueAccumulator`.

### `Arm64RegisterValueAccumulator.cs`
Modify code that using `ClrRuntime` to `IClrRuntime`.
This changes are needed to write unit tests for ValueAccumulator.

Issue #16, opened 2015-07-30
Reporting: better number display


ppanyukov:
Probably minor.

This threw me off a little:

```
 Method |  AvrTime |    StdDev |        op/s |
------- |--------- |---------- |------------ |
 Sqrt13 | 63.43 ns |  0.768 ns | 15764394.26 |
 Sqrt14 | 68.17 ns | 0.0794 ns |  14669862.3 |
```

and then I realised the numbers are not aligned.

To make it easier to inspect the results, we can do a couple of things:

1. **Ops/sec**: thousand separators and fixed decimals (say 2).

    Effect of this:

	```
	 Method |  AvrTime |    StdDev |           op/s |
	------- |--------- |---------- |--------------- |
	 Sqrt13 | 63.43 ns |  0.768 ns |  15,764,394.26 |
	 Sqrt14 | 68.17 ns | 0.0794 ns |  14,669,862.30 |
	```

2. **AvrTime** and **StdDev** is trickier, because I can see you use variable decimal precision depending on how small the value is. Nevertheless, several options here too:

  - **Same precision**. Just use the same precision for everything (say 3 or 4). The simplest but may not work if the numbers are really small.

    So the effect of this would be:

    ```
     Method |  AvrTime   |    StdDev |           op/s |
    ------- |----------- |---------- |--------------- |
     Sqrt13 | 63.4300 ns | 2.1000 ns |  15,764,394.26 |
     Sqrt13 | 63.4300 ns | 0.7680 ns |  15,764,394.26 |
     Sqrt14 | 68.1700 ns | 0.0794 ns |  14,669,862.30 |

    ```

  - **Decimal point alignment**. More tricky but will work in all cases. The result may look unusual and may throw off people :)

    This may look like this (with minor variations):

    ```
              0.753  ns
              0.0335 ns
          23535.1     s
      614119772.34     
    614,119,772.34     
       16496828.35     
     16,496,828.35     
              0.0      
    ```

------------

I have a pull request for ops/sec format (hopefully not a controversial one), and a bit of code to do decimal point alignment which I can plug in.

Let me know what you think.

AndreyAkinshin:
@ppanyukov, I totally agree, it is a good point.

AndreyAkinshin:
@ppanyukov, Can you make a PR with all formating improvements?

ppanyukov:
[...]

PR #3245, opened 2026-09-04
chore: Add unit tests for arm64 disassembler

filzrev:

#### 1. Cleanup arm64 disassembler related code to preparing to add unit tests
See following PR comment for details.
- #3240
- #3241

#### 2. Add AsmArm64 package reference
[AsmArm64](https://github.com/xoofx/AsmArm64) package to unit test project.
Currently it's used for test purpose.
It's expected existing arm64 disassembler is replaced to `AsmArm64` based implementation. (#3246)

#### 3. Add arm64 disassembler related unit tests.
To ensure existing arm64 disassembler behavior.
Unit test codes are added for major code paths. (It can confirm code coverage results with `Analyze Code Coverage` on VS)

Note:
Almost of unit tests on .NET Framework are excluded by `#if NET` directive.
- private methods are tested with `UnsafeAccessor` (It requires .NET 8 or later)
- Capstone's native dependencies seems not copied to bin directory when using xUnit v2 (Because it's `Library` project)

### DIFF
--- a/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
+++ b/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
@@ -2,12 +2,13 @@ using BenchmarkDotNet.Diagnosers;
 using Gee.External.Capstone;
 using Gee.External.Capstone.Arm64;
 using Microsoft.Diagnostics.Runtime;
+using Microsoft.Diagnostics.Runtime.Interfaces;
 
 namespace BenchmarkDotNet.Disassemblers
 {
     internal class Arm64Disassembler : ClrMdDisassembler
     {
-        protected override IEnumerable<Asm> Decode(byte[] code, ulong startAddress, State state, int depth, ClrMethod currentMethod, DisassemblySyntax syntax)
+        protected override IEnumerable<Asm> Decode(byte[] code, ulong startAddress, State state, int depth, IClrMethod currentMethod, DisassemblySyntax syntax)
         {
             const Arm64DisassembleMode disassembleMode = Arm64DisassembleMode.Arm;
             using (CapstoneArm64Disassembler disassembler = CapstoneDisassembler.CreateArm64Disassembler(disassembleMode))
--- a/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
+++ b/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
@@ -5,137 +5,6 @@ using Microsoft.Diagnostics.Runtime;
 
 namespace BenchmarkDotNet.Disassemblers
 {
-    internal struct RegisterValueAccumulator
-    {
-        private enum State
-        {
-            LookingForPattern,
-            ExpectingMovk,
-            ExpectingAdd,
-            LookingForPossibleLdr
-        }
-
-        private State _state;
-        private long _value;
-        private int _expectedMovkShift;
-        private Arm64RegisterId _registerId;
-        private ClrRuntime _runtime;
-
-        public void Init(ClrRuntime runtime)
-        {
-            _state = State.LookingForPattern;
-            _expectedMovkShift = 0;
-            _value = 0;
-            _registerId = Arm64RegisterId.Invalid;
-            _runtime = runtime;
-        }
-
-        public void Feed(Arm64Instruction instruction)
-        {
-            Arm64InstructionDetail details = instruction.Details;
-
-            switch (_state)
-            {
-                case State.LookingForPattern:
-                    if (instruction.Id == Arm64InstructionId.ARM64_INS_MOVZ)
-                    {
-                        _registerId = details.Operands[0].Register.Id;
-                        _value = details.Operands[1].Immediate;
-                        _state = State.ExpectingMovk;
-                        _expectedMovkShift = 16;
-                    }
-                    else if (instruction.Id == Arm64InstructionId.ARM64_INS_ADRP)
-                    {
-                        _registerId = details.Operands[0].Register.Id;
-                        _value = details.Operands[1].Immediate;
-                        _state = State.ExpectingAdd;
-                    }
-                    break;
-                case State.ExpectingMovk:
-                    if (instruction.Id == Arm64InstructionId.ARM64_INS_MOVK &&
-                        details.Operands[0].Register.Id == _registerId &&
-                        details.Operands[1].ShiftOperation == Arm64ShiftOperation.ARM64_SFT_LSL &&
-                        details.Operands[1].ShiftValue == _expectedMovkShift)
-                    {
-                        _value = _value | (instruction.Details.Operands[1].Immediate << details.Operands[1].ShiftValue);
-                        _expectedMovkShift += 16;
-                        break;
-                    }
-                    _state = State.LookingForPossibleLdr;
-                    goto case State.LookingForPossibleLdr;
-                case State.ExpectingAdd:
-                    if (instruction.Id == Arm64InstructionId.ARM64_INS_ADD &&
-                        details.Operands[0].Register.Id == _registerId &&
-                        details.Operands[1].Register.Id == _registerId &&
-                        details.Operands[2].Type == Arm64OperandType.Immediate)
-                    {
-                        _value = _value | instruction.Details.Operands[2].Immediate;
-                        _state = State.LookingForPossibleLdr;
-                    }
-                    break;
-                case State.LookingForPossibleLdr:
-                    if (instruction.Id == Arm64InstructionId.ARM64_INS_LDR &&
-                        details.Operands[1].Type == Arm64OperandType.Memory &&
-                        details.Operands[1].Memory.Base.Id == _registerId && // The source address is in the register we are tracking
-                        details.Operands[1].Memory.Displacement == 0 && // There is no displacement
-                        details.Operands[1].Memory.Index == null) // And there is no extra index register
-                    {
-                        // Simulate the LDR instruction.
-                        long newValue = (long)_runtime.DataTarget.DataReader.ReadPointer((ulong)_value);
-                        _value = newValue;
-                        if (_value == 0)
-                        {
-                            _state = State.LookingForPattern;
-                        }
-                        else
-                        {
-                            // The LDR might have loaded the result in another register
-                            _registerId = details.Operands[0].Register.Id;
-                        }
-                    }
-                    else if (instruction.Id == Arm64InstructionId.ARM64_INS_CBZ ||
-                            instruction.Id == Arm64InstructionId.ARM64_INS_CBNZ ||
-                            instruction.Id == Arm64InstructionId.ARM64_INS_B && details.ConditionCode != Arm64ConditionCode.Invalid)
-                    {
-                        // ignore conditional branches
-                    }
-                    else if (details.BelongsToGroup(Arm64InstructionGroupId.ARM64_GRP_BRANCH_RELATIVE) ||
-                             details.BelongsToGroup(Arm64InstructionGroupId.ARM64_GRP_CALL) ||
-                             details.BelongsToGroup(Arm64InstructionGroupId.ARM64_GRP_JUMP))
-                    {
-                        // We've encountered an unconditional jump or call, the accumulated registers value is not valid anymore
-                        _state = State.LookingForPattern;
-                    }
-                    else if (instruction.Id == Arm64InstructionId.ARM64_INS_MOVZ)
-                    {
-                        // Another constant loading is starting
-                        _state = State.LookingForPattern;
-                        goto case State.LookingForPattern;
-                    }
-                    else
-                    {
-                        // Finally check if the current instruction modified the register that was accumulating the constant
-                        // and reset the state machine in case it did.
-                        foreach (Arm64Register reg in details.AllWrittenRegisters)
-                        {
-                            // Some unexpected instruction overwriting the accumulated register
-                            if (reg.Id == _registerId)
-                            {
-                                _state = State.LookingForPattern;
-                            }
-                        }
-                    }
-                    break;
-            }
-        }
-
-        public bool HasValue => _state == State.ExpectingMovk || _state == State.LookingForPossibleLdr;
-
-        public long Value { get { return _value; } }
-
-        public Arm64RegisterId RegisterId { get { return _registerId; } }
-    }
-
     internal class Arm64Disassembler : ClrMdDisassembler
     {
         protected override IEnumerable<Asm> Decode(byte[] code, ulong startAddress, State state, int depth, ClrMethod currentMethod, DisassemblySyntax syntax)
@@ -147,7 +16,8 @@ namespace BenchmarkDotNet.Disassemblers
                 // disassembled binary code.
                 disassembler.EnableInstructionDetails = true;
                 disassembler.DisassembleSyntax = Map(syntax);
-                RegisterValueAccumulator accumulator = new RegisterValueAccumulator();
+
+                Arm64RegisterValueAccumulator accumulator = new();
                 accumulator.Init(state.Runtime);
 
                 Arm64Instruction[] instructions = disassembler.Disassemble(code, (long)startAddress);
@@ -324,7 +194,7 @@ namespace BenchmarkDotNet.Disassemblers
             return true;
         }
 
-        private static bool TryGetReferencedAddress(Arm64Instruction instruction, RegisterValueAccumulator accumulator, uint pointerSize, out ulong referencedAddress, out bool isReferencedAddressIndirect)
+        private static bool TryGetReferencedAddress(Arm64Instruction instruction, Arm64RegisterValueAccumulator accumulator, uint pointerSize, out ulong referencedAddress, out bool isReferencedAddressIndirect)
         {
             if ((instruction.Id == Arm64InstructionId.ARM64_INS_BR || instruction.Id == Arm64InstructionId.ARM64_INS_BLR) && instruction.Details.Operands[0].Register.Id == accumulator.RegisterId && accumulator.HasValue)
             {
--- a/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
+++ b/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
@@ -138,11 +138,11 @@ namespace BenchmarkDotNet.Disassemblers
             return false;
         }
 
-        // .NET 10 prefixes some AArch64 precode/stub shapes with `DMB ISH` for concurrent
+        // .NET 10 prefixes some AArch64 precode/stub shapes with `DMB ISHLD` for concurrent
         // stub-patching safety. Detect and skip the barrier so the existing 3-instruction pattern
         // match still works, and return the effective PC of the first real stub instruction so
         // LDR-literal offsets are calculated relative to it.
-        // Encoding: DMB ISH = 0xD50339BF.
+        // Encoding: DMB ISHLD = 0xD50339BF.
         private const uint DmbIshInstr = 0xD50339BFu;
 
         private static bool TryReadStubHead(IDataReader reader, ulong address, out ulong parseBase, out uint instr0, out uint instr1, out uint instr2)
--- a/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
+++ b/src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
@@ -143,7 +143,7 @@ namespace BenchmarkDotNet.Disassemblers
         // match still works, and return the effective PC of the first real stub instruction so
         // LDR-literal offsets are calculated relative to it.
         // Encoding: DMB ISHLD = 0xD50339BF.
-        private const uint DmbIshInstr = 0xD50339BFu;
+        private const uint DmbIshLdInstr = 0xD50339BFu;
 
         private static bool TryReadStubHead(IDataReader reader, ulong address, out ulong parseBase, out uint instr0, out uint instr1, out uint instr2)
         {
@@ -157,7 +157,7 @@ namespace BenchmarkDotNet.Disassemblers
 
             int offset = 0;
             uint first = ReadInstr(buffer, 0);
-            if (first == DmbIshInstr)
+            if (first == DmbIshLdInstr)
             {
                 if (read < 16)
                     return false;
--- a/src/BenchmarkDotNet/Disassemblers/Arm64InstructionFormatter.cs
+++ b/src/BenchmarkDotNet/Disassemblers/Arm64InstructionFormatter.cs
@@ -22,7 +22,8 @@ namespace BenchmarkDotNet.Disassemblers
                 FormatInstructionPointer(instruction, formatterOptions, pointerSize, output);
             }
 
-            output.Append(instruction.Mnemonic.ToString().PadRight(formatterOptions.FirstOperandCharIndex));
+            var padRight = Math.Max(formatterOptions.FirstOperandCharIndex, instruction.Mnemonic.Length + 1);
+            output.Append(instruction.Mnemonic.PadRight(padRight));
 
             if (asm.ReferencedAddress.HasValue && !asm.IsReferencedAddressIndirect && symbols.TryGetValue(asm.ReferencedAddress.Value, out var name))
             {
--- a/src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
+++ b/src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
@@ -1,0 +1,41 @@
+#if !NET8_0_OR_GREATER
+namespace System.Diagnostics;
+
+/// <summary>
+/// Exception thrown when the program executes an instruction that was thought to be unreachable.
+/// </summary>
+public sealed class UnreachableException : Exception
+{
+    private const string DefaultMessage = "The program executed an instruction that was thought to be unreachable.";
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="UnreachableException"/> class with the default error message.
+    /// </summary>
+    public UnreachableException()
+        : base(DefaultMessage)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="UnreachableException"/>
+    /// class with a specified error message.
+    /// </summary>
+    /// <param name="message">The error message that explains the reason for the exception.</param>
+    public UnreachableException(string? message)
+        : base(message ?? DefaultMessage)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="UnreachableException"/>
+    /// class with a specified error message and a reference to the inner exception that is the cause of
+    /// this exception.
+    /// </summary>
+    /// <param name="message">The error message that explains the reason for the exception.</param>
+    /// <param name="innerException">The exception that is the cause of the current exception.</param>
+    public UnreachableException(string? message, Exception? innerException)
+        : base(message ?? DefaultMessage, innerException)
+    {
+    }
+}
+#endif
--- a/tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
+++ b/tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
@@ -1,0 +1,203 @@
+// TODO: Remove #if directive when migrated to xunit.v3 or migrated to AsmArm64 based implementation.
+#if NET
+using AsmArm64;
+using AwesomeAssertions;
+using BenchmarkDotNet.Disassemblers;
+using static AsmArm64.Arm64RegisterX;
+using static AsmArm64.Arm64RegisterW;
+
+namespace BenchmarkDotNet.Tests.Disassemblers.Arm64;
+
+public partial class Arm64DisassemblerTests
+{
+    [Fact]
+    public void TryFollowJumpTrampoline_B()
+    {
+        // Arrange
+        var rawInstructions = new[]
+        {
+            Arm64InstructionFactory.B(0x10000), // b 0x10000
+        };
+        PrintInstructions(rawInstructions);
+
+        var clrRuntime = CreateMockClrRuntime(rawInstructions);
+        var state = new State(clrRuntime, DummyTargetFrameworkVersion);
+        ulong baseAddress = DummyBaseAddress;
+
+        // Act
+        var helper = new Arm64DisassemblerHelper();
+        var result = helper.TryFollowJumpTrampoline(state, baseAddress, out var target);
+
+        // Assert
+        result.Should().BeTrue();
+        target.Should().Be(baseAddress + 0x10000);
+    }
+
+    [Fact]
+    public void TryFollowJumpTrampoline_B_MinusOffset()
+    {
+        // Arrange
+        var rawInstructions = new[]
+        {
+            Arm64InstructionFactory.B(-0x10000), // b 0x10000
+        };
+
+        PrintInstructions(rawInstructions);
+
+        var clrRuntime = CreateMockClrRuntime(rawInstructions);
+        var state = new State(clrRuntime, DummyTargetFrameworkVersion);
+        ulong baseAddress = DummyBaseAddress;
+
+        // Act
+        var helper = new Arm64DisassemblerHelper();
+        var result = helper.TryFollowJumpTrampoline(state, baseAddress, out var target);
+
+        // Assert
+        result.Should().BeTrue();
+        target.Should().Be(baseAddress - 0x10000);
+    }
+
+    [Fact]
+    public void TryFollowJumpTrampoline_StubPrecode()
+    {
+        // Arrange
+        var rawInstructions = new[]
+        {
+            Arm64InstructionFactory.DMB(Arm64BarrierOperationLimitKind.ISHLD), // dmb ishld
+            Arm64InstructionFactory.LDR(X10, 0x10000), // ldr x11, #0x10000
+            Arm64InstructionFactory.LDR(X12, 0x20000), // ldr x12, #0x20000
+            Arm64InstructionFactory.BR(X10),           // br x11
+        };
+        PrintInstructions(rawInstructions);
+
+        var clrRuntime = CreateMockClrRuntime(rawInstructions, address =>
+        {
+            return 0x30000;
+        });
+        var state = new State(clrRuntime, DummyTargetFrameworkVersion);
+        ulong baseAddress = DummyBaseAddress;
+
+        // Act
+        var helper = new Arm64DisassemblerHelper();
+        var result = helper.TryFollowJumpTrampoline(state, baseAddress, out var target);
+
+        // Assert
+        result.Should().BeTrue();
+        target.Should().Be(0x30000);
+    }
+
+    [Fact]
+    public void TryFollowJumpTrampoline_FixupPrecode()
+    {
+        // Arrange
+        var rawInstructions = new uint[]
+        {
+            Arm64InstructionFactory.DMB(Arm64BarrierOperationLimitKind.ISHLD), // dmb ishld
+            Arm64InstructionFactory.LDR(X11, 0x10000), // ldr x11, #0x10000
+            Arm64InstructionFactory.BR(X11),         // br x11
+            Arm64InstructionFactory.LDR(X12, 0x20000), // ldr x12, #0x20000
+        };
+        PrintInstructions(rawInstructions);
+
+        var clrRuntime = CreateMockClrRuntime(rawInstructions, address =>
+        {
+            return 0x20000;
+        });
+        var state = new State(clrRuntime, DummyTargetFrameworkVersion);
+        ulong baseAddress = DummyBaseAddress;
+
+        // Act
+        var helper = new Arm64DisassemblerHelper();
+        var result = helper.TryFollowJumpTrampoline(state, baseAddress, out var target);
+
+        // Assert
+        result.Should().BeTrue();
+        target.Should().Be(0x20000);
+    }
+
+    [Fact]
+    public void TryFollowJumpTrampoline_CallCountingStub()
+    {
+        // Arrange
+        var rawInstructions = new uint[]
+        {
+            Arm64InstructionFactory.DMB(Arm64BarrierOperationLimitKind.ISHLD),          // dmb ishld
+            Arm64InstructionFactory.LDR(X9, 0x10000),                                   // ldr x11, #0x10000
+            Arm64InstructionFactory.LDRH(W10, new Arm64ImmediateMemoryAccessor(X9, 0)), // ldrh w10, [x9]
+            Arm64InstructionFactory.SUBS(W10, W10, 1),                                  // subs w10, w10, #1
+        };
+        PrintInstructions(rawInstructions);
+
+        var clrRuntime = CreateMockClrRuntime(rawInstructions, address =>
+        {
+            return 0x30000;
+        });
+        var state = new State(clrRuntime, DummyTargetFrameworkVersion);
+        ulong baseAddress = DummyBaseAddress;
+
+        // Act
+        var helper = new Arm64DisassemblerHelper();
+        var result = helper.TryFollowJumpTrampoline(state, baseAddress, out var target);
+
+        // Assert
+        result.Should().BeTrue();
+        target.Should().Be(0x30000);
+    }
+
+    // TryFollowJumTrampoline seems not support FixupPrecode with pre-backpatch form.
+    [Fact]
+    public void TryFollowJumpTrampoline_FixupPrecodeCode_Fixup_ShouldReturnFalse()
+    {
+        // Arrange
+        var rawInstructions = new uint[]
+        {
+            Arm64InstructionFactory.DMB(Arm64BarrierOperationLimitKind.ISHLD), // dmb ishld
+            Arm64InstructionFactory.LDR(X12, 0x10000), // ldr x12, #0x10000
+            Arm64InstructionFactory.LDR(X11, 0x20000), // ldr x11, #0x20000
+            Arm64InstructionFactory.BR(X11),           // br x11
+        };
+        PrintInstructions(rawInstructions);
+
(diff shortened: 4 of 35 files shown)


timcassell:
<details><summary>AI review</summary>
<p>

  ### Accumulator bugs

  **`Arm64RegisterValueAccumulator.cs:64` — `ExpectingAdd` never resets.** Unlike `ExpectingMovk`, this case has no fall-through to `LookingForPossibleLdr`, so any instruction that isn't the expected `ADD`
  leaves the state machine parked with the stale ADRP value. `adrp x0,#0x1000 ; movz x0,#0x100 ; add x0,x0,#0x100` gives `HasValue == true, Value == 0x1100` after x0 was clobbered, so a following `BR/BLR x0`
  resolves to a bogus address and the disassembly prints the wrong symbol. The PR's own skipped `AdrpThenOther_ThenAdd_ShouldResetValue` fails on this today.

  **`Arm64RegisterValueAccumulator.cs:68` — `ADD` ignores `lsl #12`.** The match checks only `Operands[2].Type == Immediate`, then does `_value | Immediate`. `adrp x0,#0x1000000 ; add x0,x0,#0xfff, lsl #12`
  yields `page | 0xFFF` instead of `page + 0xFFF000`, with `HasValue` still true — a silently wrong address rather than a bail-out. At minimum, reject a shifted immediate. (Skipped
  `AdrpThenAdd_WithShiftedImm_ShouldCalculateAddress` covers it.)

  **`Arm64RegisterValueAccumulator.cs:41` — `MOVZ` discards the shift.** Same class: `_value = details.Operands[1].Immediate` drops the `lsl` amount, so `movz x0,#0x1234, lsl #16` seeds `0x1234` instead of
  `0x12340000`.

  **`Arm64RegisterValueAccumulatorTests.Movz.cs:31` — skipped test asserts MOVK semantics.** `Movz_WithShiftedImmediateValue_ShouldStartNewValue` expects `0x3333_2222_1111` from three `MOVZ`s. A real `MOVZ`
  zeroes the rest of the register, so the architectural result is `0x3333_0000_0000`. Whoever un-skips this and "fixes" the accumulator to satisfy it will encode a decoding bug.

  ### Tests that can't fail

  **`Arm64DisassemblerTests.TryFollowJumpTrampoline.cs:73` — slot displacement untested.** In all four stub tests the `getPointer` callback returns `ExpectedResultAddress` for *any* address, unlike the
  `TryResolvePrecode` tests which assert the requested slot. Changing `parseBase + (ulong)(long)off0` to `parseBase + 4 + (ulong)(long)off0` in the `StubPrecode` branch leaves the whole suite green. Assert the
  address inside `getPointer`, as `TryResolvePrecode_*` does.

  **`Arm64DisassemblerTests.TryFollowJumpTrampoline.cs:178` — `NonStubHead` bails on length, not shape.** The test supplies one instruction, so the reader returns 4 bytes and `TryReadStubHead` bails at `read <
  12`; the stub-shape rejection it names is never reached, and the test passes with all stub matching deleted. Pad to >= 4 non-stub instructions.

  **`Arm64InstructionFormatterTests.cs:19` — padding disabled.** The comment says "Use DisassemblyDiagnoserConfig default config value" but `FirstOperandCharIndex = 10` is commented out, so the theory runs at
  Iced's default of 0 and asserts strings BDN never emits (`"b #8"` vs. production `"b         #8"` — `DisassemblyDiagnoserConfig.cs:86`). The shipped column width is only incidentally covered by the one
  `FirstOperandCharIndex = 6` case.

  **`Arm64InstructionFormatterTests.cs:61` — empty `symbols` map.** `FormatInstruction_B_WithReferencedAddress` passes no symbols, so `TryGetValue` always misses and the one thing gated on `ReferencedAddress`
  — the `Operand.Replace($"#0x{addr:x}", name)` substitution — is never run. Add an entry mapping `0x10000` to a name.

  **`Arm64DisassemblerTests.TryGetReferencedAddress.cs:11` — helper skips `Init(runtime)`.** `_runtime` stays null; it works only because no test here feeds an `LDR`. The first one that does will NRE inside
  `Feed`. The sibling helper (`Arm64RegisterValueAccumulatorTests.cs:15`) does call `Init`.

  **`Arm64DisassemblerTests.TryGetReferencedAddress.cs:23` — `_With_BL` builds `BR X0`.** Duplicate of `_With_BLR`; `BL` is never covered.

  ### Minor

  **`Arm64Disassembler.cs:145` — constant is `ISHLD`, not `ISH`.** `0xD50339BF` has `CRm = 0b1001` (`DMB ISHLD`); `DMB ISH` is `0xD5033BBF`. The new tests construct `Arm64BarrierOperationLimitKind.ISHLD`,
  confirming it. As written a stub prefixed with a plain `DMB ISH` isn't recognised and precode resolution silently fails — fix the comment/name or widen the match.

  **`ClrMdDisassembler.cs:114` — `IClrRuntime` switch introduced unchecked downcasts.** `foreach (ClrModule module in state.Runtime.EnumerateModules())` and the nested `foreach (ClrType type in ...)` went from
  statically-typed iteration to runtime casts, since both interface members are explicit implementations returning `IClrModule`/`IClrType`. Nothing breaks against ClrMD 4.0.732401 (the concrete instances
  still come back), but `FilterAndEnqueue` now throws `InvalidCastException` for any other `IClrRuntime` — including the `MockClrRuntime` this PR adds, which makes that path untestable by the harness being
  introduced.

  **`Helpers/Arm64TestInstructions.cs:180`** — `ValidateMultipleOf8` throws `"...must be multiple of 4"`.

</p>
</details> 

It looks like it found some possible bugs in the accumulator (I did not verify myself). Fine if you want to fix them here, or defer for out-of-scope.

filzrev:
> It looks like it found some possible bugs in the accumulator (I did not verify myself). Fine if you want to fix them here, or defer for out-of-scope.

This PR intended to add tests to verify existing arm64 disassembler behaviors.
So following tasks are handled on another PRs.
- Refactor Arm64ValueAccumulator/Arm64Disassebler codes. And fix some existing issues. (and reviewed content)
- Migrate code to use AsmArm64.

timcassell:
Findings are all in the new test code; the production changes look good (the formatter padding is a real fix — a mnemonic reaching `FirstOperandCharIndex` previously ran straight into its operand with no separator).

**`tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs:71`** — the `read` delegate ignores its `address` parameter and always serves `rawInstructions` from index 0, so the address the disassembler reads from is never asserted for `TryReadStubHead`, `TryFollowJumpTrampoline`, or the multi-hop loop. Mutating the trampoline read to `dataReader.Read(address + 0x1000, head)` leaves all 79 tests green. Suggest keying the delegate off `address`: serve bytes at a registered base, return 0 elsewhere.

**`tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryTranslateAddressToName.cs:97`** — `NonAlignedAddress = 0x10004` is below `MinValidAddress` on macOS-arm64 (`ClrMdDisassembler.GetMinValidAddress` returns `0x100000000` there), so the test returns on the first line and asserts three empty collections without reaching the `GetMethodByHandle`/`GetTypeByMethodTable` path it documents. `macos-latest` is arm64 in the CI matrix. Forcing `GetMinValidAddress` to `0x100000000` locally keeps all 79 tests passing. `Address1 + 4` would fix it — same class of issue as 0ca5d3f4e/019acb64e, just missed for this one address.

**`tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64InstructionFormatterTests.cs:97`** — `BeEquivalentTo` on a `List<string>` is order-insensitive, so the instruction ordering `Decode` is responsible for isn't asserted (swapping two expected lines still passes). `Should().Equal(...)` instead.

**`tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64InstructionFormatterTests.cs:142`** — `{ MOVZ(X0, 0x100), "movz x0, #0x100" }` duplicates line 140 verbatim; xUnit silently drops it (`Skipping test case with duplicate ID` in the run output). Presumably one row was meant to be a different value.

**`tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs:13`** — `DummyMethodNotUsed = default!` is a null `MockClrMethod`. Safe only because the current tests return before `TryTranslateAddressToName` reaches `method.NativeCode == currentMethod.NativeCode`; a future test that lets `GetMethodByInstructionPointer` resolve will NRE inside production code instead of failing usefully.

**`tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.Decode.cs:28,39,63,74`** (and the same copy-paste in the `TryFollowJumpTrampoline`/`TryResolvePrecode` files) — trailing comments name `x11`/`x0` where the code uses `x10`/`x1`. The register number is exactly what distinguishes StubPrecode (x10/x12) from FixupPrecode (x11/x12), so someone "fixing" code to match a comment here would break stub recognition.

**`src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs:146`** — nit: now that the comment correctly reads `DMB ISHLD`, the constant name `DmbIshInstr` is the remaining misnomer (`DmbIshLdInstr`). Separately, only ISHLD is matched — a build emitting `dmb ish` (`0xD5033BBF`) would silently fail stub-head detection. Out of scope for this PR, just noting it.

<sub>Reviewed with Claude Code.</sub>

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
```diff
@@ -0,0 +1,90 @@
...
+    protected static IDataReader CreateMockDataReader(uint[] rawInstructions)
+    {
+        Func<ulong, ulong> dummyFunc = _ => throw new InvalidOperationException("This func is not expected to be called.");
+        return CreateMockDataReader(rawInstructions, dummyFunc);
+    }
+
+    protected static IDataReader CreateMockDataReader(uint[] rawInstructions, Func<ulong, ulong> getPointer)
+    {
+        return new MockDataReader(
+            read: (address, buffer) =>
```
timcassell: Still open from the last round: the `read` delegate ignores its `address` parameter and always serves `rawInstructions` from index 0, so the address the disassembler reads from is never asserted for `TryReadStubHead`, `TryFollowJumpTrampoline`, or the multi-hop loop. Mutating the trampoline read to `dataReader.Read(address + 0x1000, head)` leaves the whole suite green.

Keying the delegate off `address` — serve the bytes at a registered base, return 0 elsewhere — would close it.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
```diff
@@ -0,0 +1,90 @@
...
+
+namespace BenchmarkDotNet.Tests.Disassemblers.Arm64;
+
+public abstract class Arm64DisassemblerTestBase
+{
+    // On macos(arm64), available virtual address space is 48-bit (or 52-bit) and minimum address must be 4GB (0x1_0000_0000)
+    internal const ulong DummyBaseAddress = 0x0000_F000_0000_0000UL;
+
+    internal static readonly Version DummyTargetFrameworkVersion = new Version(10, 0);
+    internal static readonly MockClrMethod DummyMethodNotUsed = default!;
```
timcassell: Still open: `DummyMethodNotUsed = default!` is a null `IClrMethod`, passed as the non-nullable `currentMethod` argument in `TryTranslateAddressToName_GetJitHelperFunctionName_ReturnsNonEmptyValue` and `_NoAlignedAddress`. Those pass only because production returns before reaching `method.NativeCode == currentMethod.NativeCode` (`ClrMdDisassembler.cs:343`). If that early-return ordering ever changes, they fail with a `NullReferenceException` instead of a readable assertion. A real `MockClrMethod` instance would be safer.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
```diff
@@ -0,0 +1,203 @@
...
+        var rawInstructions = new[]
+        {
+            Arm64InstructionFactory.DMB(Arm64BarrierOperationLimitKind.ISHLD), // dmb ishld
+            Arm64InstructionFactory.LDR(X10, 0x10000), // ldr x11, #0x10000
+            Arm64InstructionFactory.LDR(X12, 0x20000), // ldr x12, #0x20000
+            Arm64InstructionFactory.BR(X10),           // br x11
+        };
+        PrintInstructions(rawInstructions);
+
+        var clrRuntime = CreateMockClrRuntime(rawInstructions, address =>
```
timcassell: Raised in the earlier review and still present (also :102 and :131): the `getPointer` callback returns `ExpectedResultAddress` for any address, so the slot arithmetic these paths exist for is never asserted. Changing `dataReader.ReadPointer(countSlot + 8, ...)` to `countSlot` in the CallCountingStub branch, or following `off1`/`offB` instead of `off0`/`offA`, keeps all three tests green.

The sibling `TryResolvePrecode_*` tests already assert `address.Should().Be(mdSlot)` inside the callback — same assertion belongs here.

timcassell (commented):
Follow-up on the updated head. The fixes from the last round all look right — the macOS-arm64 address, `WithStrictOrdering`, the duplicate `MOVZ` row, `DmbIshLdInstr`, and the `Decode.cs` comments. A few items are still open, plus one new one and one follow-up note.

<sub>Reviewed with Claude Code.</sub>

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockDataReader.cs
```diff
@@ -0,0 +1,91 @@
...
+public delegate int ReadBytesDelegate(ulong address, Span<byte> buffer);
+public delegate bool TryReadPointerDelegate(ulong address, out ulong value);
+
+internal class MockDataReader : IDataReader
+{
+    private readonly ReadBytesDelegate _read = (_, _) => throw new InvalidOperationException($"{nameof(_read)} field is not set.");
+    private readonly TryReadPointerDelegate _tryReadPointer = (_, out _) => throw new InvalidOperationException($"{nameof(_tryReadPointer)} field is not set.");
+
+
+    public MockDataReader(ulong dummyValue = 0)
```
timcassell, line 15: This overload sets only `_tryReadPointer`, leaving `_read` as the throwing default. It backs `DummyClrRuntime` and `Arm64InstructionFormatterTests.GetArm64Asms` via `CreateMockClrRuntime(ulong)`, and `GetArm64Asms` runs the full `Decode` loop — so the first test case added there containing an indirect branch (e.g. `movz`/`br`) will hit `FlushCachedDataIfNeeded` -> `Read` and die with `InvalidOperationException: _read field is not set.` rather than an assertion failure.

Delegating to `CreateMockDataReader([], _ => dummyValue)` would give it a working `Read`.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    [Fact]
+    public void TryFollowJumpTrampoline_StubPrecode()
+    {
+        // Arrange
+        var rawInstructions = new[]
+        {
+            Arm64InstructionFactory.DMB(Arm64BarrierOperationLimitKind.ISHLD), // dmb ishld
+            Arm64InstructionFactory.LDR(X10, 0x10000), // ldr x11, #0x10000
```
timcassell: Register comments still mismatch the code here and on :69 (`BR(X10)` commented `br x11`) and :125 (`LDR(X9, ...)` commented `ldr x11`), plus `Arm64DisassemblerTests.TryResolvePreCode.cs:20,22`. `Decode.cs` was fixed in the last push; these were missed.

Worth getting right because the register number is exactly what distinguishes StubPrecode (x10/x12) from FixupPrecode (x11/x12) — someone "fixing" code to match a comment here would break stub recognition.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
```diff
@@ -0,0 +1,203 @@
...
+        // Act
+        var helper = new Arm64DisassemblerHelper();
+        var result = helper.TryFollowJumpTrampoline(state, baseAddress, out var target);
+
+        // Assert
+        result.Should().BeTrue();
+        target.Should().Be(ExpectedResultAddress);
+    }
+
+    // TryFollowJumTrampoline seems not support FixupPrecode with pre-backpatch form.
```
timcassell: Follow-up rather than something to fix here, in the same bucket as the other pre-existing issues you deferred: this test locks in a real asymmetry — `TryResolvePrecode` handles the pre-backpatch `FixupPrecodeCode_Fixup` shape (`Arm64Disassembler.cs:233`) but `TryFollowJumpTrampoline` does not. A direct `BL` onto a fixup precode that has never been backpatched therefore fails the trampoline chase, `GetMethodByHandle(stubAddress)` then fails because the address is the stub rather than the MethodDesc, and the call target is left untranslated in the disassembly.

Pre-existing (from #3208) and out of scope, but a TODO or issue link would read better than the bare "seems not support" comment, which makes it look intentional. (Also a typo there: `TryFollowJumTrampoline`.)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
filzrev: On latest commits, I've added MockMemory abstraction layer to create MockClrRuntime.
So it throw exception when address mapping data is not registered.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
filzrev: I've modified code on latest commits to throw `UnreachableException` when `DummyMethodNotUsed` is used..

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
filzrev: On latest commits changes.
Only registered address data is allowed and throw exception when unexpected address is passed.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
filzrev: Wrong instruction explanations are fixed by 11dfa26

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
filzrev: I've added comment/link and fix typo

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockDataReader.cs
filzrev: I've modified code to return 0 when using fixed dummy value.

timcassell (commented):
Another pass on `7124b6c3a`. `AddFalsePointer` and the absolute-address assertion in `TryGetReferencedAddress_With_BranchRelative` are both good additions. Remaining items below — the `UnreachableException` visibility one is the only one that affects shipped code.

<sub>Reviewed with Claude Code.</sub>

src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
```diff
@@ -0,0 +1,41 @@
+#if !NET8_0_OR_GREATER
+namespace System.Diagnostics;
+
+/// <summary>
+/// Exception thrown when the program executes an instruction that was thought to be unreachable.
+/// </summary>
+public sealed class UnreachableException : Exception
```
timcassell, line 7: This polyfill is `public`, so BenchmarkDotNet's netstandard2.0 and net6.0 assets now export a public `System.Diagnostics.UnreachableException`. A consumer on those TFMs who references another library exporting the same polyfill gets CS0433 ("exists in both"), and one who declares their own gets CS0436 — purely from referencing BenchmarkDotNet.

Its only consumer is `MockClrMethod` in the test assembly, and `Properties/AssemblyInfo.cs:9` already grants `InternalsVisibleTo("BenchmarkDotNet.Tests")`, so `internal` works. The sibling polyfills that declare types (`AsyncEnumerable`, `FileExtensions`) are both `internal`.

src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
```diff
@@ -0,0 +1,41 @@
+#if !NET8_0_OR_GREATER
```
timcassell, line 1: `System.Diagnostics.UnreachableException` shipped in .NET 7, not .NET 8. This guard is correct today only because the TFM list (`netstandard2.0;net6.0;net8.0;net9.0;net10.0`) happens to skip net7.0 — add net7.0 and the polyfill duplicates the BCL type (CS0436), which `TreatWarningsAsErrors` in `build/common.props` makes fatal.

Should be `#if !NET7_0_OR_GREATER`.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryGetReferencedAddress.cs
```diff
@@ -0,0 +1,127 @@
...
+using BenchmarkDotNet.Disassemblers;
+using static AsmArm64.Arm64RegisterX;
+
+namespace BenchmarkDotNet.Tests.Disassemblers.Arm64;
+
+public partial class Arm64DisassemblerTests
+{
+    private Arm64RegisterValueAccumulator CreateValueAccumulator(Arm64RegisterX register, ushort initialValue)
+    {
+        using var clrRuntime = CreateMockClrRuntime();
```
timcassell, line 13: `using` disposes the runtime when this helper returns, but the accumulator it returns keeps it in `_runtime` (`Init(clrRuntime)` on the next line). Every caller therefore gets an accumulator holding a disposed `IClrRuntime`.

Harmless only while the BR/BLR paths never dereference it — the first MOVZ->LDR case added here reads through the disposed object, since `Arm64RegisterValueAccumulator.Feed` calls `_runtime.DataTarget.DataReader.ReadPointer`. Drop the `using`, or return the runtime alongside the accumulator so the caller owns its lifetime.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockClrRuntime.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        DataTarget.Dispose();
+    }
+
+    #region APIs that are used by TryTranslateAddressToName
+    public string? GetJitHelperFunctionName(ulong address)
+    {
+        if (JitHelperFunctionNames.TryGetValue(address, out var value))
+            return value;
+        throw new ArgumentException($"Specified address(0x{address:X}) is not registered.");
```
timcassell, line 34: These four lookups (also :41, :48, :55) throw for an unregistered address, where ClrMD returns `null` and the production code treats `null` as an ordinary outcome. So the `null` branches in `TryTranslateAddressToName` can't be reached through this mock, and a fixture that forgets one registration fails with an opaque exception from inside the mock instead of a readable assertion.

`Arm64InstructionFormatterTests.FormatInstruction_IntroDisassembly_SumLocal` only survives because every branch target happens to be pre-seeded into `AddressToNameMapping`; adding one more branch instruction to that fixture without registering its symbol would throw.

`AddFalsePointer` handled the equivalent problem on the `MockMemory` side nicely — same idea would fit here. (One gap remains there too: `ClrMdDisassembler.FlushCachedDataIfNeeded`'s `Read(...) <= 0` branch is still unreachable, since a 1-byte region returns 1 rather than 0.)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64RegisterValueAccumulatorTests.Movz.cs
```diff
@@ -0,0 +1,33 @@
+using AwesomeAssertions;
```
timcassell, line 1: TFM guards are inconsistent across the new files: `#if NET` in seven, `#if NET8_0_OR_GREATER` in five, and none in eight — including this one, which calls the Capstone-backed `Arm64TestInstructions.Movz`. It survives on net472 only because the single test here is `Skip`ped.

The two guards are equivalent for today's TFMs (net10.0 and net472), but `DmbIshldOffset` is defined under `#if NET` and consumed under `#if NET8_0_OR_GREATER`, so adding a net6.0/net7.0 test TFM — or un-skipping this test — breaks the build. Worth settling on one guard and applying it consistently.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
```diff
@@ -0,0 +1,79 @@
...
+    protected static MockClrRuntime CreateMockClrRuntime()
+    {
+        var dataReader = new MockDataReader();
+        return new MockClrRuntime(new MockDataTarget(dataReader));
+    }
+}
+
+file static class ExtensionMethods
+{
+    public static ReadBytesDelegate ToGetReadBytesDelegate(this uint[] rawInstructions)
```
timcassell, line 55: `ToGetReadBytesDelegate` and `ToTryReadPointerDelegate` (:71) are no longer referenced anywhere — `MockMemory` replaced them, and the single-delegate `MockDataReader` constructors they fed were removed in `2304b834`.

Worth deleting: they're the old address-ignoring helpers, so leaving them around invites reintroducing exactly the gap `MockMemory` closed.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
```diff
@@ -0,0 +1,79 @@
...
+        foreach (var rawInstruction in rawInstructions)
+        {
+            var instruction = rawInstruction.ToCapstoneArm64Instruction();
+            Output.WriteLine(instruction.ToString());
+        }
+    }
+
+    /// <summary>
+    /// Create MockClrRuntime with MockDataReader that returns following data.
+    ///   Read: Throw InvalidOperationException.
```
timcassell, line 43: Stale since `2304b834`: `MockDataReader(ulong)` now sets `_read = (_, _) => 0`, so `Read` returns 0 rather than throwing. (Returning 0 is the right call — it matches the `IDataReader.Read` contract; it's just the doc that needs updating.)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockMemory.cs
```diff
@@ -0,0 +1,183 @@
...
+    private readonly List<MemoryRegion> _regions = new();
+
+    private readonly Dictionary<ulong, string?> JitHelperFunctionNames = new();
+    private Dictionary<ulong, IClrMethod?> MethodByHandle = new();
+    private Dictionary<ulong, IClrMethod?> MethodByInstructionPointer = new();
+    private Dictionary<ulong, IClrType?> TypeByMethodTable = new();
+
+    private readonly ulong BaseAddress;
+
+    public MockMemory(ulong baseAddress = 0)
```
timcassell, line 17: The `baseAddress` parameter is never passed — every call site is `new MockMemory()` — and it's applied only in `AddBytes`, not in `AddJitHelperFunctionName`/`AddMethodByHandle`/`AddMethodByInstructionPointer`/`AddTypeByMethodTable`. If anyone does pass a base, instruction addresses would rebase while the lookup mappings silently would not. Either drop the parameter or apply it uniformly.

src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
filzrev: It's fixed by commit (89efdaa)

src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
filzrev: It's also fixed by commit (89efdaa)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryGetReferencedAddress.cs
filzrev: I've modified code by commit (924add6)
By creating/disposing clrRuntime instance on test code side.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
filzrev: doc comments are modified by commit(5249d5c)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockMemory.cs
filzrev: Unused baseAddress parameter is removed by commit (2d6b4e8)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/Arm64DisassemblerTestBase.cs
filzrev: Unused extension methods are removed by commit (edcdf9e)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64RegisterValueAccumulatorTests.Movz.cs
filzrev: TFM guards issues are changed to use `#if NET` by commit(5aa4678)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockClrRuntime.cs
filzrev: I've added commit (edc10f5)
And change behavior to not throw exception by default.

`ThrowExceptionMode` is expected to be used when writing new unit tests.

FlushCachedDataIfNeeded issue is remaining.
But currently there is no unit tests. So it's not handled on this PR.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
```diff
@@ -0,0 +1,247 @@
...
+        var helper = new Arm64DisassemblerHelper();
+        var result = helper.TryFollowJumpTrampoline(state, BaseAddress, out var target);
+
+        // Assert
+        result.Should().BeFalse();
+    }
+
+
+    [Fact]
+    public void TryFollowJumpTrampoline_NonStubHead_ShouldReturnFalse()
```
timcassell, line 221: **Medium.** This test is vacuous. It maps 2 instructions = 8 bytes, so `TryReadStubHead` bails at `if (read < 12)` (`Arm64Disassembler.cs:155`) and the `BL` opcode is never examined — the test would still pass if `BL` *were* wrongly accepted as a stub head, which is what the comment on the `BL` line says it verifies. The unused `.AddPointer(BaseAddress + 0x10000, ...)` suggests the pattern-match path was the intent.

Pad the region to 4 instructions (e.g. `BL` + 3 `NOP`s) so the length guard passes and the opcode match is what returns false.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockMemory.cs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private int Read(ulong address, Span<byte> buffer)
+    {
+        if (buffer.Length == 0)
+            return 0;
+
+        var region = FindRegion(address);
+        if (region is null)
+            throw new ArgumentException($"Failed to read address 0x{address:X}. Because address data is not registered.");
```
timcassell: **Medium.** Still open from the last round, and here is a concrete consequence: `Read` throws for an unmapped address where the real `IDataReader.Read` returns 0, and production depends on that contract — `TryReadStubHead` and `TryFollowJumpTrampoline` treat a short/zero read as "not a stub", and `FlushCachedDataIfNeeded` treats `<= 0` as "flush and retry".

`TryTranslateAddressToName`'s 8-hop trampoline chase calls `TryFollowJumpTrampoline` on each hop target, so the moment a chain walks one hop past the registered regions the mock throws out of production code instead of letting it `return false`. The graceful-degradation path can't be tested, and any future test that trips it fails with a confusing exception. Return `0`.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockClrRuntime.cs
```diff
@@ -0,0 +1,118 @@
+using Microsoft.Diagnostics.Runtime;
+using Microsoft.Diagnostics.Runtime.Interfaces;
+using System.Collections.Immutable;
+
+namespace BenchmarkDotNet.Tests.Disassemblers.Arm64;
+
+public class MockClrRuntime : IClrRuntime
+{
+    // It is expected that this property will be used for validating registered addresses.
+    private bool ThrowExceptionMode => false;
```
timcassell: **Low.** `ThrowExceptionMode` is hardcoded to `false` and is `private`, so all four `if (ThrowExceptionMode) throw` branches are dead and nothing can turn them on. The stated intent — validating registered addresses — is therefore not in force: a test that forgets an `AddMethodByHandle`/`AddTypeByMethodTable` mapping silently gets `null` and passes for the wrong reason instead of failing loudly.

Either make it settable per-test (`internal bool ThrowExceptionMode { get; set; }`) or drop the dead branches.

src/BenchmarkDotNet/Disassemblers/Arm64InstructionFormatter.cs
```diff
@@ -22,7 +22,8 @@ internal static string Format(Arm64Asm asm, FormatterOptions formatterOptions,
                 FormatInstructionPointer(instruction, formatterOptions, pointerSize, output);
             }
 
-            output.Append(instruction.Mnemonic.ToString().PadRight(formatterOptions.FirstOperandCharIndex));
+            var padRight = Math.Max(formatterOptions.FirstOperandCharIndex, instruction.Mnemonic.Length + 1);
```
timcassell: **Low.** `Math.Max(..., Mnemonic.Length + 1)` guarantees at least one pad character even for operand-less instructions. With a user-supplied `FormatterOptions` where `FirstOperandCharIndex == 0` (the Iced default, reachable because `DisassemblyDiagnoserConfig` accepts a custom formatter), `nop`/`ret` now render as `"nop "`/`"ret "` where they previously rendered `"nop"`.

It also makes the new `FormatInstruction` theory fragile: adding any operand-less instruction to `TestData01` fails the AsmArm64 cross-check purely on the trailing space. Padding only when the operand is non-empty would avoid both.

src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
```diff
@@ -286,7 +157,7 @@ private static bool TryReadStubHead(IDataReader reader, ulong address, out ulong
 
             int offset = 0;
             uint first = ReadInstr(buffer, 0);
-            if (first == DmbIshInstr)
+            if (first == DmbIshLdInstr)
             {
                 if (read < 16)
```
timcassell, line 163: **Low, pre-existing** (master behavior, not introduced here — raising it because the new `TryReadStubHead_InsufficientInstructions_ShouldReturnFalse` locks it in as intended).

Both this guard and `read < 12` above treat a short read as "not a stub", but `IDataReader.Read` may legitimately return fewer bytes than requested at a page boundary or from a chunked reader — `ClrMdDisassembler.Decode` loops for exactly that reason, this call does not. A `DMB ISHLD`-prefixed precode straddling a page boundary such that only 12 of 16 bytes come back would never resolve, and the disassembly prints a raw address instead of the method name. A short-read retry loop would be the safer contract.

src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
```diff
@@ -0,0 +1,41 @@
+#if !NET7_0_OR_GREATER
+namespace System.Diagnostics;
+
+/// <summary>
+/// Exception thrown when the program executes an instruction that was thought to be unreachable.
+/// </summary>
+internal sealed class UnreachableException : Exception
```
timcassell, line 7: **Low.** Thanks for the `internal` + `!NET7_0_OR_GREATER` fix. Remaining point: this polyfill still has zero production callers — its only consumer is `MockClrMethod` in the test project, which is also where the net472 need originates. It would sit better in `BenchmarkDotNet.Tests` than compiled into the shipped netstandard2.0/net6.0 assemblies.

Same theme as `Arm64RegisterValueAccumulator.State` going `private` -> `internal` for the new `Reset` seam: worth a look at whether the test need can be met without widening the product.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryResolvePreCode.cs
```diff
@@ -0,0 +1,323 @@
...
+namespace BenchmarkDotNet.Tests.Disassemblers.Arm64;
+
+public partial class Arm64DisassemblerTests
+{
+    [Fact]
+    public void TryResolvePrecode_StubPrecode()
+    {
+        // Arrange
+        const int JumpAddress = 0x10000;
+        const int MdOffset = 0x20000; // This value is not used for test.
```
timcassell: **Low.** The comment is inverted. In `TryResolvePrecode_StubPrecode` it is `JumpAddress` that production ignores (`out int _`), while `MdOffset` is load-bearing — it determines `MdSlotAddress` below. A future edit trusting this comment would silently desynchronise the mapped pointer from the slot the code actually reads.

timcassell (commented):
Pass on `edc10f573` — all eight items from the last review are addressed, and the `UnreachableException` visibility/guard fix is exactly right. Seven remaining, severity-tagged; the two mediums are both tests that pass for the wrong reason.

<sub>Reviewed with Claude Code.</sub>

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryFollowJumpTrampoline.cs
filzrev: I’ve modified code by commit(abcc4ea)
By adding NOP instructions to pass `return false` code path that don't match StubHead condition.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockMemory.cs
filzrev: I've modified code to return 0 when specified address is not registered at commit(5ea2d39)

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Helpers/MockClrRuntime.cs
filzrev: I've modified code ThrowExceptionMode  can be set be test.
Existing tests are expecting ThrowExceptionMode. So default value is  set to `true`

src/BenchmarkDotNet/Disassemblers/Arm64InstructionFormatter.cs
filzrev: I've modified code to skip setting padding when instruction don't have operands.
at commit (8ddcd31)

src/BenchmarkDotNet/Disassemblers/Arm64Disassembler.cs
filzrev: This PR intended to add unit tests for arm64 disassembler.

It should be handled on next PR that refactor existing arm64 disassembler code.
I've added TODO comment at commit (a3ca150)
(Resetting value accumulator code is handled on this PR though)

src/BenchmarkDotNet/Extensions/Polyfills/UnreachableException.cs
filzrev: `UnreachableException` is useful to identifying code path is not expected to be called. (By code coverage)
So it's intentionally  added on BenchmarkDotNet polyifills.

tests/BenchmarkDotNet.Tests/Disassemblers/Arm64/Arm64DisassemblerTests.TryResolvePreCode.cs
filzrev: I've modified wrong comment on commit (2d97e70)

timcassell (approved):

PR #3253, opened 2026-09-10
Add .NET 12 runtime and toolchain support

LoopedBard3:
Add .NET 12 monikers, cached runtimes and toolchains including composite ReadyToRun and Mono WebAssembly. Update development defaults and templates, and cover runtime parsing and CLI resolution. Based on the .NET 11 enablement PR (#2911) and takes into account any newly added touchpoints like (#2967).

### DIFF
--- a/src/BenchmarkDotNet.Annotations/Jobs/RuntimeMoniker.cs
+++ b/src/BenchmarkDotNet.Annotations/Jobs/RuntimeMoniker.cs
@@ -53,6 +53,8 @@ namespace BenchmarkDotNet.Jobs
         public const string Net10_0 = "net10.0";
         /// <summary>.NET 11.0</summary>
         public const string Net11_0 = "net11.0";
+        /// <summary>.NET 12.0</summary>
+        public const string Net12_0 = "net12.0";
 
         /// <summary>NativeAOT compiled as net7.0</summary>
         public const string NativeAot70 = "nativeaot7.0";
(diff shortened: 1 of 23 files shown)


LoopedBard3:
FYI @DrewScoggins

timcassell (approved):

src/BenchmarkDotNet.Annotations/Jobs/RuntimeMoniker.cs
```diff
@@ -77,6 +81,8 @@ public static class RuntimeMoniker
         public const string Mono10_0 = "mono10.0";
         /// <summary>.NET 11 using MonoVM (not CLR which is the default)</summary>
         public const string Mono11_0 = "mono11.0";
+        /// <summary>.NET 12 using MonoVM (not CLR which is the default)</summary>
+        public const string Mono12_0 = "mono12.0";
```
timcassell, line 85: It looks like Mono is being dropped in .Net 12. Should we remove these?

LoopedBard3: @DrewScoggins Any thoughts? I think we should remove the new .NET 12 Mono entries, including `MonoWasm12_0` and `MonoWasmAot12_0` and their corresponding runtime/toolchain presets. However, I think we should keep up to .NET 11 and older for now as the full removal is still a WIP [dotnet/runtime#134337](https://github.com/dotnet/runtime/issues/134337) and [dotnet/runtime#134397](https://github.com/dotnet/runtime/issues/134397).

